PR #69, opened 2016-11-24
Move and update Azure Documentation

carlos-salinas:
- Move the right category, howto6 (on-premise)
- Update content and attachments


jandevriesmendix:
This is on hold until I've met with Carlos.

PR #82, opened 2016-12-09
Incidents

katamaks:
This article is regarding incidents only


Adam-Dupaski:
If the title changes, all the terminology in the doc must be evaluated and changed if necessary. Please have the meeting with Jan about Support terminology before submitting this request.

PR #97, opened 2016-12-21
Alert thresholds

knorrie:
Add detailed information about the different alerts that exist now, including thresholds.

Note for editors:

I used the :---|:---| tables to get the page render locally. Feel free to adjust everything to decrease the total amount of space usage on the page, since with all separate tables and rows it's rather huge now.

The "5.  If one of the status items is not listed" shows up as 1. now instead of 5... Also, the whole {% markdown %} thing does not render locally or on regular github markdown.


jandevriesmendix:
Pull request was reviewed and merged in separate branch.

PR #125, opened 2017-02-02
[Not to be published before 2017-02-20] Update screenshots of the feedback widget.

reinouts:
- Add a section on whitelisting the mendix.com domain for hybrid app use


JelteMX:
If this is not to be published before 2017-02-20 we should keep this one open until then?

jandevriesmendix:
I will pick this up as part of the documentation project for the Developer Portal.

PR #130, opened 2017-02-08
Document hybrid profiles

johanflint:
- Offline device will convert to hybrid phone
- New: hybrid tablet
- New flag: offline support enabled

### DIFF
--- a/refguide6/hybrid-phone-profile.md
+++ b/refguide6/hybrid-phone-profile.md
@@ -1,0 +1,14 @@
+---
+title: "Hybrid phone profile"
+space: "Reference Guide 6"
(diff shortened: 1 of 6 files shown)


johanflint:
> Navigation.md
> 
> *  Shouldn't the Offline Device and Hybrid Profiles section not explain the new profiles? They currently only list Phone, Tablet, Desktop.
> 
Do you mean the fallback of profiles? Not sure what you are refering to.

jandevriesmendix:
> Do you mean the fallback of profiles? Not sure what you are refering to.
I checked again, but I was mistaken. The doc looks good 👍

PR #134, opened 2017-02-10
HTML was badly converted

RCatersels:
Changed all the HTML (&#123 is {).

Regards,
Ronald


jandevriesmendix:
Thanks for the catch Ronald!

PR #138, opened 2017-02-14
Added best practice document on performance in mendix 7

BartLuijtenMx:
This is related to Mx7, so do not release before release of Mx7. Also, I wanted to do a "more info here" section at the end, but since there is no reference doc for Mx7 live yet, it's hard to get the link URLs right.


jandevriesmendix:
Publishing the doc. Please update the screenshot a.s.a.p.

PR #142, opened 2017-02-15
APM 1.10 an additional feature

TolenE:
Trigger action notify and separate triggers.md


TolenE:
merged mendix:development and solved conflicts

PR #144, opened 2017-02-15
WIP: Add releasenotes for 5.21.6

hwdegroot:


hwdegroot:
@jandevriesmendix Is this already released? Because it was still kinda work in progress, because we will only release tomorrow

PR #158, opened 2017-02-22
Feature/rest timeout

joostverhoog:
Mendix 7.1 will allow the user to specify how long the REST call action should wait for the REST endpoint to respond.


Adam-Dupaski:
Merge for 7.1

PR #159, opened 2017-02-22
Feature/urlencode urldecode

joostverhoog:
Mendix 7.1 will have two extra string functions to be used in microflow expression.


Adam-Dupaski:
Merge for 7.1

PR #166, opened 2017-02-23
Release 1.10.1 with Mx7 support and bug fixes

TolenE:
I'm still testing the Mx7 compatibility, so please wait with putting in production until the release.

Can be tomorrow afternoon, can also be early next week.

### DIFF
--- a/releasenotes/APM/1.10.1.md
+++ b/releasenotes/APM/1.10.1.md
@@ -6,7 +6,16 @@ category: "APM"
 
 ## Mendix 7 Compatibility
 
-This is the first version that is also available for Mendix 7. Due to some incompatibilities between 6 and 7, there are now two packages available in the Mendix App Store. The standard package to download is the Mendix 7 package. Under older versions, you can find the same version packaged for Mendix 6.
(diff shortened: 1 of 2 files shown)


Adam-Dupaski:
Also, please clarify what should be linked in the release notes. The download is available in the App Store, correct? However, the release notes mention two packages available for download, whereas the App Store has only one download available. Please clarify for this set of release notes.

Adam-Dupaski:
Regarding the link, I want to link these release notes to the Web App Store. But only the latest version is available for download there. The older versions are only available for download in the Modeler App Store. This will affect how the release notes are written.

TolenE:
Hi Adam,

How can I see your changes?

As for the versions. Both are 1.10.1. The one mpk runs on Mx7, the other on Mx6.2. I will ask Andrej to place the Mx7 one as the current version in the AppStore and make the Mx6.2 version available under ‘older versions’.

Best regards,

Bart

Adam-Dupaski:
Hi Bart, you can see my changes in the most recent commit here. As for the version naming, let's wait until the packages are in the App Store to see how they are named, and then we can update the release notes here.

PR #171, opened 2017-02-28
Rest proxy

joostverhoog:
New feature in Mendix 7.1: Using a proxy in REST.
This is similar to using a proxy in webservices, but not the same, so I created a new page in the refguide.


Adam-Dupaski:
Ready for merge for 7.1.

PR #185, opened 2017-03-09
Reference the Mendix BuildPack (cf-mendix-buildpack) Documentation

djvdorp:
It seems that some customers were unable to find their way to our cf-mendix-buildpack, so let's help them out a bit more. This is mostly useful for advanced users of our platform, and people that host on-premise on Cloud Foundry, but it might be good for our users in general to be able to find the inner workings of our BuildPack.


djvdorp:
:+1:

PR #208, opened 2017-03-14
Removed iframe Modelshare

Criccle:
Iframes aren't supported in markdown:

http://stackoverflow.com/questions/14951321/how-to-display-html-content-in-github-readme-md

### DIFF
--- a/community/documentation/content-writing-and-formatting-guidelines.md
+++ b/community/documentation/content-writing-and-formatting-guidelines.md
@@ -67,25 +67,17 @@ Use **bold** when identifying buttons, windows, fields, and other entities that 
 
 Use *italics* when the user needs to enter text. For example: "In the **Name** field, enter *Customer_NewEdit_Commit*."
 
-## 4.3 Embedding a Modelshare
-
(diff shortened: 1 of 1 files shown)


JelteMX:
Iframes are normally not supported in Markdown, but HTML can be used in Markdown. So in that sense, iframes do work in Markdown. I changed the example in the contribution guideline, that one works. With the earlier example that you want to delete, it is missing a quote, which does not make it render

JelteMX:
See [this change](https://github.com/mendix/docs/commit/3b4a487e9c957832440a79bf52ec0dec5fe836c9#diff-27a585c5d0ea1abf67cfd25578694743L75). I think we can close this PR, @jandevriesmendix

Criccle:
Does not seem to work though.

image

JelteMX:
True, not on Github. It does work on the website though

JelteMX:
See the result [here](https://docs.mendix.com/community/documentation/content-writing-and-formatting-guidelines)

Criccle:
Thanks Jelte! I was convinced it didn't work, my bad!

PR #236, opened 2017-03-21
no need for a link

isamu-tanaka:


isamu-tanaka:
no review is required.

PR #253, opened 2017-03-23
platform portal release notes - fix title anchors

jtwaleson:
The anchors now have two <a> tags, one for the anchor itself,
and the other so that you have a clickable link which you can
copy and share.

Also removed the duplicate "Release date:" in every title because that is a given here.


jtwaleson:
Never mind, all the h2 in the document have clickable id's already. We can remove the manual anchors completely.

PR #275, opened 2017-03-31
Mappings can now be based on entities

joostverhoog:
As of Mendix 7.2, you can select an entity as the source of import and export mappings


Adam-Dupaski:
Ready for merge for 7.2.0.

PR #276, opened 2017-04-03
Update release notes + backups api documentation

jtwaleson:


RCatersels:
Jouke, while you at it. I think that the domain model is incomplete. I miss the reference between Environment and Package, because the action https://deploy.mendix.com/api/1/apps/<AppId>/environments/<Mode>/package gives the package for that environment. That seems like a one on one relation between those.
Regards,
Ronald

Adam-Dupaski:
@RCatersels - @jtwaleson is going to look into this. Merging this PR.

PR #278, opened 2017-04-04
rewrote trends doc

katamaks:
added introduction and description where to find trends and premissions. Needs additional description from Jouke.


jandevriesmendix:
Still missing several introductions. Will merge when they've been added.

PR #282, opened 2017-04-04
Document navigation profiles in Mendix 7.2

johanflint:
Due to the differences between profiles in 7.1 and 7.2 the old navigation page is renamed and links to the newly written navigation page. The old, hard coded profile pages stay underneath the old navigation page and a new navigation profile page is written to document all fields and options. I've made a copy of all the fields and options to avoid that new features pop-up in the old documentation as those pages use snippets.


Adam-Dupaski:
Ready to be merged for 7.2.0.

johanflint:
Thanks Adam!

PR #307, opened 2017-04-07
delete deploying to a free app

katamaks:
the same description is in getting starten -> create and deploy your first app


jandevriesmendix:
Can't merge this until you remove all the reference to this file from other docs.

PR #322, opened 2017-04-12
Update client documentation due to 7.2.0

mijay:
- Add documentation for `isSynced` expression function.
- Update documentation for `string-function-call` as they are now supported in the client.
- Mark "Use schema" as deprecated.
- Remove wrong mention of the ticket number from releasenotes of 7.1.1
- Add missing documentation for expression-based conditional formatting (one page missed during 7.1.0)
- Remove mentions of client features added/deleted before 7 (missed that during 7)
- Remove mentions of context mechanism removed in 7 (missed that during 7)


refguide7/conditions.md
```diff
@@ -24,70 +24,76 @@ There are four kinds of conditions:
...
+Module role conditions can only be used for the visibility of table rows and tab pages. They cannot be used to make an input widget read-only or to hide buttons. To make an input widget conditionally editable based on roles, use security to give roles read and write access to attributes. For example, by not giving a role write access to an attribute, the corresponding input widget will automatically turn read-only (as in, not editable). To hide a button based on roles, configure the security of the underlying page or action (for example, by not giving a role create access, **New** buttons will be hidden).
 
 ## 2 Context
 
-### 2.1 Show/enable element based on attribute value
```
mijay: This names were aligned with labels in the modeler, so maybe it is better to keep them?

Adam-Dupaski: Okay, will change them back.

PR #341, opened 2017-04-21
remove mx could from doc

katamaks:


jandevriesmendix:
Please check the Travis build to find the errors and update the content.

PR #360, opened 2017-04-26
<br> is better than two spaces '  '

isamu-tanaka:
A white space could be removed during a process like a translation.


JelteMX:
A `<br>` might be better, but the correct way to write that would be `<br />`

isamu-tanaka:
I see.

PR #362, opened 2017-04-27
Some strings of the top page are cut off.

isamu-tanaka:
While translating it into Japanese, I found a solution for this cut-off.
The &nbsp; exists for layout adjustment.


isamu-tanaka:
My solution is not good for expanding Sub-menu by javascript. 
So, I would close this ticket, and please let you know about the issue.

PR #369, opened 2017-05-04
Update and rename data-migration.md to attributes-type-migration.md

nerush:
Title is also updated.


jandevriesmendix:
Please update the PR with updated references to the document you changed, so the documentation will build. See the errors here: https://travis-ci.org/mendix/docs/builds/228694560.

PR #387, opened 2017-05-09
How-to-Mx7-Create-a-to-app

erikkg:
I am not sure if the way I uploaded and direct to the image is in the right way. As it is, I will add more images in a second version. Please check the content for now and publish if it is allright. iframe and notes are workign as far as I know..


jandevriesmendix:
Hi Erik,
Please create a new and up to date fork of the Mendix docs and submit a new pull request, with the new how-to in the Mendix 7 How-To section.

This PR was done from an out of date fork, and therefore wants to update several thousand files.

Thanks!
Jan

PR #389, opened 2017-05-09
Move docs kasja

katamaks:
moved, added documents, added/deleted attachments


katamaks:
@jandevriesmendix ready for review

PR #417, opened 2017-05-22
add release notes for cloud portal release of May 19th

jtwaleson:


Adam-Dupaski:
"The 'view current log' functionality for Mendix Cloud v4 applications was empty by default" - What is the fix? Should this be "is now empty by default" or "is no longer empty by default"?

jtwaleson:
Good point. Elaborated a bit. Let me know if you think it's clear enough now.

PR #478, opened 2017-06-13
Manipulating objects after synchronization is allowed now

hnejra:
This is part of the Mendix 7.4.0 release, so this can be merged when Mendix 7.4.0 is released.


hnejra:
@Adam-Dupaski : I've updated a bit more on the documentation page about offline, as it was quite outdated. It's currently being reviewed by my team member. When that's done, I'll let you know. Can you take a look at the additional changes at that time?

hnejra:
@Adam-Dupaski My team member already finished reviewing, you can take a look when you have time.

PR #517, opened 2017-06-29
move 'provide feedback on mendix' to how-to

katamaks:
move 'provide feedback on mendix' to how-to category.


Adam-Dupaski:
Also updated link on Collaborate index.

PR #547, opened 2017-07-06
Path Based Access Restrictions new doc

katamaks:


knorrie:
Please move this over to #551 as this PR contains only half of the changes. Thanks.

PR #558, opened 2017-07-11
leave app types doc

katamaks:


Adam-Dupaski:
Do you have some introduction text for "Leave App Types"?

PR #567, opened 2017-07-13
Add details and intro to backups in Mendix Cloud

jtwaleson:


Adam-Dupaski:
@jtwaleson and @katamaks - I am merging this for the new content. But the doc does not appear to accurately reflect the Backups page. Some of the options are not labelled correctly, and it is confusing as to whether section 3 is organized according to action or according to the UI button/section available. I improved what I could at this point. Please make a plan to improve this doc in the future.

PR #578, opened 2017-07-17
update webmodeler event app doc

katamaks:


Adam-Dupaski:
Changes added manually during proofreading

PR #586, opened 2017-07-18
update develop index

katamaks:


Adam-Dupaski:
Please make sure that the build will not fail.

PR #612, opened 2017-07-25
Improvements in the SQL Server database documentation

arjenw:
Some improvements to these SQL server documentations based on support ticket request. Also some texts are restructured as they where unclear.


nerush:
Hi @Adam-Dupaski , do you possibly know when these updates will be available in production ?

Adam-Dupaski:
Hi @nerush, the changes should already be published.

PR #623, opened 2017-07-26
Remove fixed known issue

hnejra:
This was fixed in 7.3.0 (Ticket 50650).


Adam-Dupaski:
We should not delete fixed known issues. We need to write that they are fixed and then link to the release notes that describes the fix via the Support ticket number

hnejra:
Ah, yes, that makes sense. Thanks for updating!

PR #677, opened 2017-08-23
Show page update regarding Location part

dkho87:
Added information about which version the Location setting is still relevant for.


Adam-Dupaski:
Thanks for your pull request, Dennis. But after internal verification, we have found that the legacy layout is not supported at all in Mendix 7, so the Location section needs to be removed from this doc.

PR #680, opened 2017-08-23
Update Model SDK 4.2.0 releasenotes

jhuiting:
We missed one story in the releasenotes.


jhuiting:
Thanks @Adam-Dupaski!

PR #691, opened 2017-08-29
Documenting default login behaviour

lhohan:
This page adds documentation on login behaviour in case of login failures.

I am not sure this is the best place to put this information. Ideally I think we could create a page called 'Login behaviour' under runtime section. Would this be possible?


Adam-Dupaski:
Yes, that's possible. It'll be a short doc, but we can add it. Please change this PR or send a new one with the correct location.

Adam-Dupaski:
Merging this for the document title name. Will create new "Login Behavior" doc for the Runtime category manually.

PR #736, opened 2017-09-21
Ats1.8 new bestpractice structure

zwartd:
Fixed all the links.


jandevriesmendix:
Hi,
There are still some build errors. Please check them.

zwartd:
Hi i checked the build error. It says that the images are not found. I changed the image links to make them more robust but it fails again. However the images and all the links are correct.

zwartd:
Another idea is to remove the existing docs and replace them with exact copies.

jandevriesmendix:
@JelteMX can you please check the build error? As far as I can see the links are correct. What is the problem?

zwartd:
@jandevriesmendix I fixed it by removing all the old files and replacing them with the renamed files.

jandevriesmendix:
Ok, I will review tomorrow.

PR #744, opened 2017-09-25
Update migrating-to-v4.md

katamaks:


jtwaleson:
@jandevriesmendix Can we revert this? @katamaks and I miscommunicated in what needed to be updated, this was not the right thing.

jandevriesmendix:
Reverted.

PR #764, opened 2017-09-28
Ats1.8 add uploading a file howto

zwartd:


zwartd:
I have added a folder that contains a document that is shown when you click a link in the uploading-a-file-howto. It should not be visible.

jandevriesmendix:
What should not be visible, exactly?

zwartd:
The Build said that the selenium-files folder is rendered but is not visible. But that is exactly what I wanted, only the build failed. I tried to create the same mechanic as Mendix uses for downloading the how-to template.

PR #798, opened 2017-10-20
Link howto video

Mcpoowl:
Link how to (same as in refguide7)


Mcpoowl:
This will post the howto video @Criccle and I created to the Mendix 6 refguide aswell, since it is now only shown on the Mendix 7 refguide

PR #815, opened 2017-11-06
Added example video

Criccle:


jandevriesmendix:
Will review when back from vacation.

PR #824, opened 2017-11-10
[WEBUI-399] Add known issue on enumeration display in text widget

mlaponder:


Adam-Dupaski:
Added ticket number based on email.

PR #838, opened 2017-11-16
Remove  max_idle_time_s_low_resources from server statistics

akhilless:
RUN-503 - Remove "max_idle_time_s_low_resources" from server statistics because it is no longer provided by Jetty


Adam-Dupaski:
@akhilless This should be published on the release of 7.9, correct?

akhilless:
@Adam-Dupaski yes, we should publish it once we publish 7.9

PR #842, opened 2017-11-17
Remove max_idle_time_s_low_resources from server statistics Mendix 6

akhilless:
RUN-503 - Remove "max_idle_time_s_low_resources" from server statistics because it is no longer provided by Jetty.


akhilless:
Please publish when Mendix 6.10.11 is released.

PR #852, opened 2017-11-27
Add known issue: object does exist on commit

lhohan:
See https://mendix.atlassian.net/browse/RUN-635


Adam-Dupaski:
@lhohan Because this is a known issue, we need a Zendesk ticket number. Can you please add that?

PR #865, opened 2017-12-04
Removed public beta alert for Message Definitions and Published REST

grbinho:
Hi guys,

These features will be released in version 7.10 so from then, these messages are no longer going to be necessary.


content/refguide/export-mapping-action.md
```diff
@@ -39,7 +39,7 @@ If the [export mapping](export-mappings) requires an input, you can select a var
 
 {{% alert type="info" %}}
 
-Introduced in version 7.8.0. This feature is still in private beta, and is subject to change before the final release.
+This was introduced in version 7.8.0.
```
fazzdev: This version should be 7.10.0 instead of 7.8.0 because the feature was behind feature flag in version 7.8.0 and normal user does not have access to it.

content/refguide/gfm-syntax.md
```diff
@@ -5,7 +5,7 @@ parent: "published-rest-services"
 
 {{% alert type="info" %}}
 
-The **Published REST Service** feature was introduced in version 7.8.0. This feature is still in private beta, and it is subject to change before final release.
+The **Published REST Service** feature was introduced in version 7.8.0.
```
fazzdev: This version should be 7.10.0 instead of 7.8.0 because the feature was behind feature flag in version 7.8.0 and normal user does not have access to it.

content/refguide/export-mapping-action.md
fazzdev: This change applies to all changes in this commit.

PR #884, opened 2017-12-21
 Note limitation in search by special characters in Mendix 5 and 6

nerush:
Search by special characters is not supported in Mendix 5 and 6.

### DIFF
--- a/content/refguide5/xpath-contains.md
+++ b/content/refguide5/xpath-contains.md
@@ -12,3 +12,5 @@ The contains() function tests whether a String attribute contains a specific str
 ```
 
 This query returns all customers from which the name contains the string 'an'. Customers with the name 'Jansen' or 'Jans' will be returned, because 'an' is part of the names 'Jansen' and 'Jans'.
+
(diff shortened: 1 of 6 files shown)


content/refguide5/xpath-contains.md
```diff
@@ -12,3 +12,5 @@ The contains() function tests whether a String attribute contains a specific str
 ```
 
 This query returns all customers from which the name contains the string 'an'. Customers with the name 'Jansen' or 'Jans' will be returned, because 'an' is part of the names 'Jansen' and 'Jans'.
+
+Search by special characters, like `%`, `_` and `\` is not supported in Mendix 5. Please upgrade to Mendix 7.11 or higher if you want to search by special characters.
```
roland-mendix: Generic remark [applies to many changes]: a comma is missing in
  Search by special characters, like `%`, `_` and `\` is not supported in Mendix
Comma that delimits the subclause:
  Search by special characters, like `%`, `_` and `\`, is not supported in Mendix

nerush: Fixed in all files and PRs with this kind of statement.

PR #886, opened 2017-12-22
Needed completion

TimvanSteenbergen:
"Use a client certificate" is not restricted to password-protected client certificates but does also use public (not sure if that is the correct term) certificates. Not having this mentioned in this document sent me off the tracks. Therefor this text addition.

Might want to add a reference to https://docs.mendix.com/refguide/project-settings#6-certificates at line 14.

### DIFF
--- a/content/howto/integration/use-a-client-certificate.md
+++ b/content/howto/integration/use-a-client-certificate.md
@@ -6,7 +6,15 @@ tags: ["integration", "web service"]
 
 ## 1 Introduction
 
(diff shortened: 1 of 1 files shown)


joostverhoog:
The text that gets added here talks about a 'public client certificate', and then talks about configuring a server certificate. That is confusing, so I suggest we don't include this.

Adam-Dupaski:
@TimvanSteenbergen Based on the request for verification from @joostverhoog, we are going to have to decline this submission. Thanks.

PR #937, opened 2018-01-13
Minor

TimvanSteenbergen:
Minor


Adam-Dupaski:
Hi, there is no "version 6.0.0" of this module, so I changed it to searching for a version that will work with a version 6 of the DM. Thanks.

PR #938, opened 2018-01-15
Ats add browserstack test files

zwartd:
The idea is that when you click the link in the how to upload a file document that the Browserstack test files appear. It should work the same as the how-to template.


Adam-Dupaski:
Hi @zwartd, I have merged this PR. But looking at it, is there a reason we don't want to publish ats-browserstack-test-files.md as a proper document that will appear in the browser? Publishing it would allow the Markdown formatting to make more sense and appear more readable. Or, if you are saying that "For the latest version please contact BrowserStack", is there a URL we can include instead that lists the latest files?

zwartd:
Hi @Adam-Dupaski,

I never thought of that, that would indeed look a lot better. We can make it a separate document. 
No, unfortunately they do not have this information on their website. I had to contact Support for the list.

PR #977, opened 2018-02-05
More clear and correct texts.

TimvanSteenbergen:
The documentation informed me a lot but also confused and lacked details. So I decided to improve it. First of all, there is no justification for having published-rest-service, published-rest-services and published-rest-operation. That is one page too many.  I propose to remove published-rest-service and have incorporated the information on that page in published-rest-services. Also the texts itself are now more clear and correct and will guide the developers better.

PR #979, opened 2018-02-05
Removal because of duplicate content

TimvanSteenbergen:
See https://github.com/mendix/docs/pull/977 where this info is more appropriate.


Adam-Dupaski:
Hi, thanks for the PR. I have asked for verification of your suggested changes, and they have been assessed as either incorrect or not bringing value. In addition, the content on [https://docs.mendix.com/refguide/published-rest-service](https://docs.mendix.com/refguide/published-rest-service) has been updated with 7.12, so rearranging content via this PR is no longer applicable.. I am going to have to close this PR. Thanks.

PR #987, opened 2018-02-12
Update 7.11.md

tuhinmitra888:
Added a known issue regarding domain model slowness


Adam-Dupaski:
@tuhinmitra888 We need a Zendesk ticket number for each known issue. What is the Zendesk ticket number for this?

PR #996, opened 2018-02-21
A dead link

isamu-tanaka:


Adam-Dupaski:
Thanks for finding this! I replaced the link with one to another PDF.

PR #1052, opened 2018-03-08
Ats update ci/cd with new example

zwartd:


Adam-Dupaski:
@zwartd Please refork development so you are not sending old commits with future PRs.

PR #1082, opened 2018-03-20
Add release notes for model-sdk 4.9

jhuiting:


Adam-Dupaski:
What does "set an edit lock" mean? Should it be "edit a lock"?

PR #1105, opened 2018-03-27
Add warning about discarded changes in unreadable objects

mijay:


mijay:
This is not related to a specific version, so can be merged asap

/cc @Adam-Dupaski

PR #1118, opened 2018-03-30
Anchor fixes

JelteMX:


JelteMX:
This will give you the possibility to set your own anchors (only works in titles!)

```
## 3.6 Start Environment {#start-environment}
```

Will be rendered in the HTML with ID="start-environment"

PR #1145, opened 2018-04-17
Update sap-odata-connector.md

lutraki-yavan:
Initial updates - only made it by section 2.3.2 - so to be continued...


MarkvanMents:
Commits applied in the wrong order - some changes here need to be rejected and the development version used.

PR #1162, opened 2018-04-26
Update and rename database-maintenance-size-reduction.md to database-…

jpulleman:
…maintenance-in-mx-cloud-size-reduction.md

Major simplification of the contents as it contained errors and caused more confusion than clarity.


MarkvanMents:
Hi Jaap (@jpulleman ),

There are two build problems with this. Can you look at them?

/developerportal/operate/metrics.html has the following errors:


   + Has link to /howtogeneral/support/database-maintenance-size-reduction which would resolve to /home/travis/build/mendix/docs/_site/howtogeneral/support/database-maintenance-size-reduction (.html | index.html), but it does not exist


/developerportal/operate/trends.html has the following errors:


   + Has link to /howtogeneral/support/database-maintenance-size-reduction which would resolve to /home/travis/build/mendix/docs/_site/howtogeneral/support/database-maintenance-size-reduction (.html | index.html), but it does not exist

PR #1165, opened 2018-04-30
Update custom-settings.md

gnmarcos:
Changed persistentSessions default value from false to true - this changed for Mendix 7


gnmarcos:
Hi Mark,

I would be happy to, but I can find what you're describing. I see a few details buttons, but can't find "Contributor has signed...". Can you say what I can do to get there please?

Thanks
Gonçalo

JelteMX:
@gnmarcos The CLA can be found here: https://www.clahub.com/agreements/mendix/docs

At the bottom you will need to login with Github and the you can sign it.

gnmarcos:
Done! Although I just realised I was logged in with my gmail account instead of my Mendix account (which I used to say "I agree". I hope that's ok.

MarkvanMents:
Great. I'll release it - just looking at one more PR first.

PR #1188, opened 2018-05-18
Temporarily remove IBM Cloud deployments instructions

MarkvanMents:
IBM Cloud deployment no longer works as we are waiting on IBM to fix some things.
Removing documentation until these issues are resolved.


MarkvanMents:
Hi @Adam-Dupaski ,

Can you approve this? I can do it myself if you would rather, but perhaps we should follow the process in case I've missed something?

We can discuss tomorrow if you want.

Mark

PR #1220, opened 2018-06-14
Page Editor documentation for new 7.17 property 'Documentation URL'

fvanderveen:
We added a property to building blocks for 7.17. While writing this, I also noticed that the paragraph above it (about the image) contained an untrue statement about what happens when you don't select an image.


Adam-Dupaski:
@fvanderveen Will be published upon release of 7.17.

Adam-Dupaski:
@fvanderveen Made change about image to current doc, which will be published shortly.

PR #1236, opened 2018-06-22
Sorted out strange section numbering

MarkvanMents:
Also made a few changes for consistency.


MarkvanMents:
Really wanted to change nillable to nullable, but apparently "Nillable" is the correct term in XML schemas.

Adam-Dupaski:
@MarkvanMents  I put those 2 cases in bullets, think that is clearer than numbers. Thanks for the PR, cheers.

PR #1237, opened 2018-06-25
Feature/charts

Andries-Smit:
Add app store charts documentation:
- Reference guide
- How to


Adam-Dupaski:
@Andries-Smit Hi Andries, there were a ton of problems with this PR. First of all, there were way too many docs in it. It would have been better to send at least 2 PRs - 1 for the how-to's and 1 for the ref guide material. It took me a long time to fix all the build/link errors in here, which should not be happening. And that's without proofreading, and proofreading attention is necessary. Please send smaller PRs next time. Thanks.

PR #1239, opened 2018-06-25
Documentation for delete action (to be merged for 7.17)

ekrem-aktas:
Delete action will be release in 7.17, with this it will be possible to delete object directly for simple cases, without the need to have a microflow for it.


ekrem-aktas:
@Adam-Dupaski added a new commit for another story which also will be released in 7.17, and the changes in the documentation are on the same page, so to avoid conflicts, I committed here. 

Someone from my team will review the last commit first, then I'll notify you whenever it is ready for your review

ekrem-aktas:
Hi @Adam-Dupaski, I've added one more commit about a similar change in grid edit buttons, which also should be in 7.17

PR #1251, opened 2018-07-03
SAP SSO Release Notes PR

MarkvanMents:
For review by DES. Then needs a review by Adam to agree style.

### DIFF
--- a/content/releasenotes/developer-portal/index.md
+++ b/content/releasenotes/developer-portal/index.md
@@ -4,6 +4,13 @@ title: "Developer Portal"
 
 ## 2018
 
+### July 3rd, 2018
+
+#### Improvements (SAP - Cloud Deployment Portal)
+
+* To improve integration between Mendix and SAP we now enforce using the same username (email address) for Mendix and SAP. You will need to use the same username the next time you need to provide your credentials. 
+  *  This is currently implemented only for SAP regions **eu10** and **us10**.
+  
 ### July 2nd, 2018
 
 #### Improvements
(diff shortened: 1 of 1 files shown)


content/releasenotes/developer-portal/index.md
```diff
@@ -4,6 +4,13 @@ title: "Developer Portal"
 
 ## 2018
 
+### July 3rd, 2018
+
+#### Improvements (SAP - Cloud Deployment Portal)
```
Adam-Dupaski: I don't think "SAP - Cloud Deployment Portal" is necessary to identify in heading. It's integrated into the Dev Portal, so we should just identify it as a general improvement.

content/releasenotes/developer-portal/index.md
```diff
@@ -4,6 +4,13 @@ title: "Developer Portal"
...
+
+* To improve integration between Mendix and SAP we now enforce using the same username (email address) for Mendix and SAP. You will need to use the same username the next time you need to provide your credentials. 
+  *  This is currently implemented only for SAP regions **eu10** and **us10**.
```
Adam-Dupaski: Second bullet probably not necessary, just put the last sentence in the main bullet/paragraph.

content/releasenotes/developer-portal/index.md
MarkvanMents: It only affects SAP users. but I'll make sure that is clear in the text. Thanks for the comments.

Adam-Dupaski: Sounds good. Anything like that should be made clear in the text, that way we can keep the headings consistent and generic.

MarkvanMents:
Changes from Adam and review with Mehran.
Changes referred to in release notes have been released.
Releasing Document.

PR #1300, opened 2018-08-01
Integration changes for 7.17

joostverhoog:


MariaShaposhnikova:
@joostverhoog @MarcoBedulli @grbinho @fazzdev  I can see that the clahub check has failed, this means not all contributors have signed CLA. Can you please sign it, so that we can proceed with this pull request.

MariaShaposhnikova:
@fazzdev We checked, please re-sign CLA, because it says you haven't sign it. Thanks!

PR #1327, opened 2018-08-14
Add note that as of 7.19.0 more additional parameters are supported

hnejra:


Adam-Dupaski:
Note: verify footnote formatting on build to dev, document.

gajduk-mansystems:
So exiting, can't wait for Mx 7.19. Can you confirm if this also applies to custom widgets?

PR #1334, opened 2018-08-21
Application Monitoring formatting fix

cenkiravul:
Application Monitoring formatting fix


MarkvanMents:
@Adam-Dupaski can you look at this. The original PR was against Mendix/Master so I changed it to Mendix/Development but it added a whole load of other commits. I don't think this will cause an issue, as only one file is showing as different, but could you check it first. Thanks.

Adam-Dupaski:
@MarkvanMents should be fine with only the 1 file, will merge. @cenkiravul please update your fork to avoid sending old commits.

PR #1344, opened 2018-08-28
Remove solved known issue

arjan-h:


Adam-Dupaski:
@arjan-h Is there a fix for this KI? Because there is a ticket number, the KI should not be removed without a documented fix citing that same ticket number. Please note this KI also appears in the 7.17 RN.

Adam-Dupaski:
Fixed via #1345.

PR #1361, opened 2018-09-19
7.18.0 retraction

reinouts:
Text as supplied by Nilco


Adam-Dupaski:
@reinouts Merged manually via email, thanks for sending anyway :)

PR #1372, opened 2018-09-26
Update dev-best-practices.md

bartpoelmans:
Type in paragraph 4.7:
- Becauase -> because


MarkvanMents:
Thanks again for the feedback.

Overriding CLA check and merging.

PR #1387, opened 2018-10-14
Update sap-destination-with-rest.md

michal-keidar:


Adam-Dupaski:
@MarkvanMents would be good to update the shared glossary with the feedback in these PRs.

PR #1391, opened 2018-10-16
Update mxcloudv4.md with mda size limitations

joshp123:
All `.mda` files are zip files. Due to limits in the `ZIP` file format, if you want to use a file size > 4GB (either compressed or uncompressed), or > `2^16` files, you need to use the `ZIP64` format. Our version of CloudFoundry does not seem to support `ZIP64`. Thus, deploying an `.mda` greater than these sizes to the MX Cloud will fail.


Adam-Dupaski:
@joshp123 please sign the [CLA](https://www.clahub.com/agreements/mendix/docs).

PR #1392, opened 2018-10-16
[for 7.20] Fix documentation ("Show control bar" property of the data grid)

r0b1n:
- Fix documentation as the actual behavior differs. Hiding control bar (buttons) is not affecting paging buttons, they are still shown.
- Rename "Show control bar" to "Show control bar buttons" in the modeler.


olenakosiakevych:
We have merged the fix before freeze for 7.20.

PR #1401, opened 2018-10-23
Update index.md

hansthen:


MarkvanMents:
@hansthen - I'm afraid I don't understand this release note at all.

Did you release a new version of something - if so what.  If it was a new release, did it potentially solve the problem for other users as well? I think that the amount of information here is just going to worry the users that new software is going out with specific code for a specific customer.

Adam-Dupaski:
@hansthen Yes, we need more specific details here, otherwise there is no reason to publish such a release note. Release notes should detail the issue/fix, but they do not need to specify whether something was fixed for a specific customer. This is already over a week old - can we update or should we close the PR? Thanks.

hansthen:
Yes please close.

MarkvanMents:
@hansthen Thanks for the confirmation. Will close this PR.

PR #1409, opened 2018-11-05
Proofread - added tags and full name of IIS

MarkvanMents:
@hansthen - there is a link here to https://esus1.mendixcloud.com/index.html#mce_temp_url# which is supposed to have information about IIS error responses, but it seems to be a real app doing something else. Any idea what the real link should be, or should I remove it?

Let me know if I should be asking someone else.

Thanks,

Mark


MarkvanMents:
This PR is now so far behind development that I will close it.

Taking the changes and putting them into the file manually as part of restructuring the deployment section of the documentation site.

PR #1410, opened 2018-11-06
Update ## 9 Rename the Administrator User

RVeenvliet:
Added line 115.
"For this to take effect in the Mendix cloud the administrator password needs to be reset in the cloud as well."


RVeenvliet:
Hi Mark,

I just signed the Contributor License Agreement.

knorrie:
There's also #1411 which tries to add the same line in a different wording. I prefer that one since it gives a hint about where to push which button. I also commented a bit more in there. Thanks.

MarkvanMents:
Text merged with PR #1411 which dealt with the same topic.
PR #1411 will be approved.

PR #1470, opened 2018-12-27
Update custom-settings.md

TimvanSteenbergen:
See https://forum.mendixcloud.com/link/questions/92380. It is not always clear what parametername and value to put here. This is making the list of options more complete.


MarkvanMents:
Submitted to runtime team for verification.

PR #1471, opened 2018-12-30

TimvanSteenbergen:
This issue with IsMatch('', 'whatever') is pretty weird, should not have happened or at least be documented. But since it not in the release notes, nore in the docs of communuty commons, we should add it here.

### DIFF
--- a/content/refguide/string-function-calls.md
+++ b/content/refguide/string-function-calls.md
@@ -403,7 +403,7 @@ In `isMatch()`, the regex is implicitly anchored at `^` and `$`.
 
 NB searching an empty string:
 
-* `isMatch('', '.*[0-9].*')` used to return true in 7.0.2 and some older versions, but somewhere it changed and since 7.8.0 it returns false. False is also the answer given by java.util.regex.Matcher.
+* `isMatch('', '.*[0-9].*')` returns false
 
 ## replaceAll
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Submitted to Microflow Editor team for review

zubivan:
Hi @TimvanSteenbergen, thank you for your docs contribution. I think you're right, if that's the case it should be documented. I see you started `False is more complaint with...`, but never finished the sentence. Can you please finish it. And from our side we will check what were exact version where this behaviour got changed.

MarkvanMents:
@TimvanSteenbergen Thanks for your update.

@zubivan - have you got any further in identifying when this happened?

I'd rather have the exact version, if I can, but if it is going to take some time it would be better to release this documentation as is and put the actual version in when we have the information.

Thanks

MarkvanMents:
I have 7.5.1 on my machine - it returns false too.

MarkvanMents:
Just created a new app in 7.0.2 and done the isMatch example above and I still get FALSE!

Have you checked 7.0.2 Ivan?

I'll try 6.10.17 and see what happens there, but it seems that all the versions of Mendix 7 i have return the same thing: false.

This is using the example added to the documentation: `isMatch('', '.*[0-9].*')`

MarkvanMents:
image

image


OK, I've tried it now in 6.10.17 and I STILL get *false*. I am obviously testing something wrongly.

@zubivan can you show me the test returning TRUE in 7.0.2?

PR #1475, opened 2019-01-02
Update known issues for Cloud V4

joshp123:
Added explanation for why restarts may be visible in cloud v4 logs.


MarkvanMents:
@joshp123 - In future PLEASE can you create branches from **development** rather than **master**. I have just pushed your changes to master accidentally. This allows us to check the builds and review the document before it is published (some styles are a bit tricky to get right and don't display properly if you make a mistake.

Thanks

Mark

PR #1476, opened 2019-01-02
Update release notes for CP180.1.0.14335

manreet185:
Release Note for CloudPortal release CP180.1.0.14335 on 3rd Jan,2019


Adam-Dupaski:
@manreet185 Hi, for future reference, we do not use "Cloud Portal" in community documentation. We use "Developer Portal" or "deployment via the Developer Portal."

PR #1481, opened 2019-01-09
Update entities.md

tuhinmitra888:
Change default values of System members.


tuhinmitra888:
Changed default values as I see it is False

image

PR #1488, opened 2019-01-25
lost an array dereference in a recent commit

tieniber:
In a recent commit, we lost a dereference when `token` is used. The variable `token` is an array, so when we need use it, we have to use the second value in the array. So it should be `token[1]`


feddovanede:
@tieniber Looks exactly we discussed!

Adam-Dupaski (approved):

PR #1514, opened 2019-02-14
Update 7.23.md with a typo-fix, and putting "we fixed" in the fixes

meertens:
- Expended vs Expanded
- "We fixed" suggests a bug fix, which can of course be considered as an improvement, but better fits in the fixes section
- A new activity type in nanoflows counts as a new feature to me

### DIFF
--- a/content/releasenotes/desktop-modeler/7.23.md
+++ b/content/releasenotes/desktop-modeler/7.23.md
@@ -13,6 +13,7 @@ description: "The release notes for Mendix Desktop Modeler version 7.23 (includi
 ### New Features
 
 * In [published OData services](/refguide/published-odata-services), clients can now use the `not` operator (as in `$filter=not substringof('j', Name)`).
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@meertens do you work for Mendix?

meertens:
@Adam-Dupaski No. I work at eMagiz and am one of the Mendix experts. At eMagiz, we are usually one of the early adopters of new Mendix versions, and tend to read all the release notes. As someone who writes release notes as well, I thought you'd appreciate the feedback in this form.

Adam-Dupaski:
@meertens I see. Thanks for the feedback, you raise some interesting questions. But besides the typo, nothing here should be changed. The "list operation" note is considered an improvement by the dev team who built that, so our policy is to go with their categorization. And for the "we fixed" improvements, these aren't fixing tickets, that's why we rely on dev team categorization and list them as improvements. The typo will be fixed. Cheers.

PR #1519, opened 2019-02-20
COL-406: add stories api rest doc and restructure the soap part of stories

nchatzi:
Pair developed with @pluizigeEend


JelteMX:
@Adam-Dupaski could you review this PR so we can merge it into this feature branch? We're not publishing yet, so we'll keep it in this branch

PR #1525, opened 2019-03-01
Added documentation for Azure SQL

grootjans:


Adam-Dupaski:
@grootjans thanks for following up on this

PR #1540, opened 2019-03-14
Release notes for Sprintr

beerscheffers:
Release date is on 15th of march


Adam-Dupaski:
Please note that "Sprintr" is not a term to be used in documentation. :)

beerscheffers:
Ah I see, so just leave the Sprintr out of it then? or is there a other term to declare Sprintr specified features and fixes?

Adam-Dupaski:
We just use "Developer Portal" in the docs and RN :)

PR #1546, opened 2019-03-18
SAP Marketplace changes

MarkvanMents:


MarkvanMents:
Self-approving and merging

PR #1550, opened 2019-03-20
Explain that optimization is not possible with custom ranges

CoreRuntime:


Adam-Dupaski:
I will duplicate this for Mx8

PR #1553, opened 2019-03-22
Adding an example because that clarifies a lot

TimvanSteenbergen:
Example 1 Using a left (outer) join to get records of table A that have no association in table B

### DIFF
--- a/content/refguide/oql-left-outer-join.md
+++ b/content/refguide/oql-left-outer-join.md
@@ -14,7 +14,6 @@ The syntax is as following:
 
 ```
 LEFT [ OUTER ] JOIN entity_path [ ON <constraint> ]
-
 ```
 
 `entity_path`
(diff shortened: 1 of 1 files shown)


arjenw:
@TimvanSteenbergen Thanks for your pull request. I think it is a good improvement of our documentation here.

@Adam-Dupaski There are a couple of things I that we can do to polish the documentation here slightly more:
 1. There is a superfluous empty line in the syntax box that can be removed
 2. There are 4 of these pages (one for FULL, INNER, LEFT and RIGHT). They are all quite similar and I propose to merge them into one where we describe the basic syntax and the differences between the types of join.
 3. I propose to write the query example itself slightly different (I changed capitalization of the SQL keywords and put the `SELECT` on top):
```sql
SELECT 
  Customer/Name as Name,
  Customer/<anyotherattribute> as <anyotherattribute>
FROM MyModule.Customer
  LEFT OUTER JOIN Customer/MyModule.Customer_Order/MyModule.Order as Order
WHERE Order/ID IS NULL
```

Adam-Dupaski:
Thanks @arjenw 

1. Will fix.
2. Will contact you offline about this.
3. Will use.

PR #1556, opened 2019-03-25
add scope descriptions for mx:app:stories:read and mx:app:stories:write

disharth:


disharth:
@Adam-Dupaski : Ronny would like to review/aware of these changes.

Adam-Dupaski:
@disharth when should this be published?

disharth:
> @disharth when should this be published?

Ronny's team(Identity Services) can tell about specific dates when are they releasing . I couldn't find a way to tag him here.

Adam-Dupaski:
Should be published along with Stories API.

PR #1562, opened 2019-03-26
Update custom-domains.md

jpulleman:
Feddo and I were discussing this and the current way this text is phrased makes it sound like something Mendix does not support, while in fact it is something that it is not supported in general (e.g. https://www.isc.org/blogs/cname-at-the-apex-of-a-zone/).

We came up with the above to make this more clear and also explain how you could resolve this if you still want to use a custom apex/naked domain.

### DIFF
--- a/content/developerportal/deploy/custom-domains.md
+++ b/content/developerportal/deploy/custom-domains.md
@@ -48,7 +48,7 @@ Create a CNAME (Canonical Name) record and point it to `[YOUR-CUSTOM-DOMAIN].cna
 
 {{% alert type="info" %}}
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Looks good @jpulleman .

Would it be clearer if I used `subdomain.mycompany.com` instead of `www.mycompany.com`?
Let me know - I'll merge it when I hear back from you.

Thanks

MarkvanMents:
@feddovanede perhaps you can give me an answer on the above?

> Would it be clearer if I used subdomain.mycompany.com instead of www.mycompany.com?

feddovanede:
Yes, `subdomain.mycompany.com` might be clearer. It doesn't have to be `www`.

PR #1567, opened 2019-03-29
Stories REST API documentation.

nchatzi:
Add swagger documentation for the Stories Rest API. @pluizigeEend @disharth


content/apidocs-mxsdk/apidocs/stories-api-rest.md
```diff
@@ -0,0 +1,9 @@
+---
+title: "REST API"
+parent: "stories-api"
+layout: swagger
+file: stories-api.yaml
+draft: true
```
Adam-Dupaski: Do we need "draft"? I thought this could be published immediately with everything else.

pluizigeEend: @Adam-Dupaski I'm not sure how the timing of releases is usually handled between new feature and the documentation of that feature. We plan on releasing the feature sometime this week (either Tuesday or Thursday), so is it okay if the documentation for the feature is visible before the feature itself? Not sure how many users would stumble upon it, and then try it out, and then be confused why it doesn't work?

Adam-Dupaski: We try to publish documentation as close to the release as possible. "Draft" would publish but hide the doc, so it's easier to just keep all the docs on this PR until the release and then merge everything together. I will remove "draft" from the metadata.

PR #1605, opened 2019-04-30
Update developer-tools.md

webfryslan:


Adam-Dupaski:
Hi @webfryslan thanks for the pull request, but these are Mendix-recommended tools, so we do not accept suggestions here because they would be difficult to verify and reach consensus on. Thanks.

PR #1606, opened 2019-04-30
Add reference to profile authentication settings

Andries-Smit:


Adam-Dupaski:
Will add for Mx8 Ref Guide as well.

PR #1614, opened 2019-05-06
Additions to offline-first documentation

hnejra:


Adam-Dupaski:
> 
> 
> 🎄 I don't see the need for having nested bullet points here. E.g. this could just be the second sentence of the first bullet point.

This is a style guideline, not a technical change. I will rewrite.

PR #1619, opened 2019-05-07
Update Getting Started With Native Mobile

ConnorLand:
I worked with Danny yesterday to update this document. A few features planned for release will take a little longer to finish, and the language has been changed to reflect that:

* **1 Introduction** – changed per Danny to have more feature-accurate language
* **2 Prerequisites** – browser requirement removed
* **3.2 Downloading and Installing the Make It Native App** – iOS section commented out (as a todo). Mendix believed we would have a running iOS version of this app. However, the App Store has not gotten back to us and we are in the dark as to when the iOS version of the Make It Native app will be approved.

Please let me know if the way I used todo comments in 3.2 is too messy. Maybe there is a cleaner solution.

### DIFF
--- a/content/howto/mobile/getting-started-with-native-mobile.md
+++ b/content/howto/mobile/getting-started-with-native-mobile.md
@@ -8,21 +8,23 @@ tags: ["native", "mobile", "developer", "test"]
 
 ## 1 Introduction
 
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@ConnorLand merging, verify the doc when rendered that the to-do comments are not visible

ConnorLand:
I will verify the comments.

PR #1633, opened 2019-05-10
Draft Release Note for May 13th Developer Portal release.

MarkvanMents:
No Customer-facing implications.

Need to verify whether this needs to be in a Release Note.


MarkvanMents:
Agreed not to put this in the release notes as there are no customer-facing implications.

PR #1643, opened 2019-05-14
Update 7.23.md

RCatersels:
Eclipse is also working again in 7.23.4 so this known issue should be removed. And adjusted in 7.23.3 that it is fixed in 7.24.4.


Adam-Dupaski:
Hi @RCatersels we were working on updating the RN this afternoon and have just documented this fix, so this PR is not necessary. Thanks for the attention and help though!
Cheers,
Adam

PR #1650, opened 2019-05-15
New DB support policy

grootjans:
- Add support policy for database vendor releases, 
- Updated database versions

The policy needs to be reviewed by Koert. The new version list needs to be reviewed by Andrej

### DIFF
--- a/content/refguide/system-requirements.md
+++ b/content/refguide/system-requirements.md
@@ -61,13 +61,16 @@ The browser you use needs to have JavaScript turned on.
 
 ### 5.3 Database Server
 
+Mendix endeavors to support the most recent and patched versions of database vendors. We will strive to add support for new vendor versions 2 minor versions after support has been added. Dropping support for a database will be announced in the release notes at the date a vendor drops support. We will drop support two minor versions later.
+
+Current support:
 * [IBM DB2](db2) 11.1
-* [MariaDB](mysql) 5.5, 10.0, 10.1
-* [Microsoft SQL Server](/developerportal/deploy/mendix-on-windows-microsoft-sql-server) 2008, 2008 R2, 2012, 2014, 2016, 2017
+* [MariaDB](mysql) 5.5, 10.1, 10.2, 10.3
+* [Microsoft SQL Server](/developerportal/deploy/mendix-on-windows-microsoft-sql-server) 2016, 2017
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@MarkvanMents can you please pick this up for the required verification?

MarkvanMents:
@grootjans 

Are you confirming the policy with Koert and asking Andrej for review?

grootjans:
@MarkvanMents yes, working on the confirmation.

MarkvanMents:
Great. I'll merge once we have the confirmation (unless it is important to get it out before).

grootjans:
@MarkvanMents I got a question about the wording. I used 'endeavor', but perhaps 'aims' might be a better substitution given the audience. As George Orwell wrote 'Never use a long word where a short one will do'. What do you think?

MarkvanMents:
@grootjans
Interesting question.
There is (to me and, luckily backed up here https://forum.wordreference.com/threads/endeavour-or-aim.1809938/) a subtle difference between aim and endeavor.
Aim - what we want to do
Endeavor - what we are trying very hard to do (but might not succeed)
In this case I think **aim** is better as _endeavor_ makes it seem like we are trying really hard but we might still fail. Unless, of course, these things really are very difficult and users have to be understanding if we are 3 years behind the latest version.
How about:

> Mendix tries to support the most recent and patched versions of database vendors. We aim to add support for new vendor versions 2 minor versions after the vendor has released it.

Question: is this 2 Mendix minor versions later (i.e. if a vendor releases a new version in May 2019 when Mendix is on 8.0, then we will put it in Mendix 8.2) or is it 2 vendor minor versions (i.e. they release 11.2 and we will put 11.2 in our product when they reach 11.4). I can read it both ways.

grootjans:
@MarkvanMents fantastic answer, and great suggestion. I like it.

grootjans:
@MarkvanMents about the versions:
This is 2 Mendix minor versions later. (i.e. if a vendor releases a new version in May 2019 when Mendix is on 8.0, then we will put it in Mendix 8.2)

MarkvanMents:
@grootjans 

Any progress with reviews by Koert and @ako ?

grootjans:
@MarkvanMents I got the go-ahead from Koert and @ako.

PR #1651, opened 2019-05-15
Add "known issues" section to backups

joshp123:
Describe effects observed when customers have a large number of files in their backups.

### DIFF
--- a/content/developerportal/operate/backups.md
+++ b/content/developerportal/operate/backups.md
@@ -109,7 +109,11 @@ Backup Details | Cloud Version | Description
 
 At the bottom of the screen, you can click **Delete** to delete this particular backup.
 
-## 4 Read More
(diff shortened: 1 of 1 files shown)


MarkvanMents:
@joshp123 I've made a couple of minor changes.
This is about S3 files, I assume. Will the user understand this?
I'll publish this later today unless I get any more changes from you.

joshp123:
Hi Mark,

Thanks for the feedback/changes.
Yes, this about the files that are stored on S3. I chose to leave out the S3 part, as generally we simply mention files; the fact that they are stored on S3 is specific to cloud v4.

Josh

MarkvanMents:
Hi Josh,
I'll just leave it as files then and merge the changes.
Thanks very much for supplying the text.
Mark

PR #1655, opened 2019-05-15
Clarify behavior of autocommit

MarkvanMents:
Expand on the risk of using AutoCommit instead of explicitly committing.


MarkvanMents:
@dudk - can you have a look at this and see if it reflects the autocommit issue you had.
Who should look at this from Data Storage/Runtime?

PR #1663, opened 2019-05-16
Update service-integration.md

AndreasLennevi:
Quite significant updates - spent 1 day on this only
See email for pictures


MarkvanMents:
Hi @AndreasLennevi 
Adam is not in until Monday - can I assign this for him to pick up then, or does this need to be merged today?

AndreasLennevi:
I knew he would be back on Monday so that is what I expected. It’s ok to wait I think

MarkvanMents:
OK - I'll assign it to Adam for when he returns.

PR #1675, opened 2019-05-22
Rewording tables vs index graph for clarity

udh-mendix:
There's some confusion in the documentation for the tables vs index graph, and I've created this MR to fix that.
More details in https://mendix.atlassian.net/browse/DVAS-256


svanderburg:
Changes look fine to me. The situation that the table size actually refers to the size of the entire database is odd, but that's how it is currently implemented. IMO it's better to be transparent about it.

MarkvanMents:
@svanderburg @udh-mendix I have tried to make the situation even clearer by emphasizing that the total memory is the *tables* line. I have also tried to justify having the index line there and explained why it might be useful.
Can you confirm that I haven't lost anything (or added anything untrue or conflicting) by my rewording.

Thanks

udh-mendix:
@MarkvanMents Thanks for the rewording! Looks good to me.

PR #1684, opened 2019-05-24
Update central-data.md

AndreasLennevi:
Did half of this one - Very happy with the direction here - Innovapost helped me crystalize this subject - now it will be great and partly disruptive, since very few people do this now.


AndreasLennevi:
Great work !!!
I start second round of Central Data now

PR #1697, opened 2019-06-03
Fix read only method in widget how to RE: Feedback Ticket 433712

jobvs:
Closes ticket number 433712


ConnorLand:
These minor code fixes will resolve Feedback item 433712. The user's code suggestion was correct. Steps have been taken to try and ensure that these code errors happen less frequently.

PR #1708, opened 2019-06-05
Update documentation on fixing float and currency deprecations in mendix 8

ester-nl:


Adam-Dupaski:
Publish for Beta 3

MarkvanMents:
@ester-nl 

You have removed the information about the conversion of the underlying database from this document. Is this no longer relevant, or was this an oversight?

@grootjans, I think you wrote the original - can you look at the changes and tell me if this still works from the data-storage point of view?

Thanks

ester-nl:
> @ester-nl
> 
> You have removed the information about the conversion of the underlying database from this document. Is this no longer relevant, or was this an oversight?
> 
> @grootjans, I think you wrote the original - can you look at the changes and tell me if this still works from the data-storage point of view?
> 
> Thanks

All of the deprecation error fixes will now happen at once, so fixing the Attributes (and subsequently the domain model and underlying database) will also happen together with all the other deprecation fixes. so yes, that particular sentence is no longer relevant

PR #1725, opened 2019-06-07
Update workflow-int-data-transfer.md

AndreasLennevi:
Not quite ready review - only got half way this time


Adam-Dupaski:
@AndreasLennevi please let me know when it's ready to review

AndreasLennevi:
We leave it as it is now
Just process all changes – in worst case do a little on Monday

I will stop making changes for now

PR #1728, opened 2019-06-12
MDS-135130 add changes for the MindSphere modules released in beginni…

KellerStefan:
…ng of June 2019


MarkvanMents:
@KellerStefan Can you just confirm that my changes are OK. They are only for language and formatting. Then I can publish.

I can also copy the files to the Mendix 7 documentation if they need to be identical.

Thanks

PR #1736, opened 2019-06-14
Mendix Cloud: Deprecate the X-Client-Certificate header

knorrie:
Hi, it's me again.

Here's the actual release notes change for deprecating the http header (which led to documenting the headers in the first place).

Also, two improvements for the http headers page itself.

I'm not sure about the choice for the word 'stable' in the release notes. The meaning has to be: "these ones are documented and you can be sure we won't just change them". Maybe there's a better word.

Thanks.

### DIFF
--- a/content/developerportal/deploy/mendix-cloud-request-headers.md
+++ b/content/developerportal/deploy/mendix-cloud-request-headers.md
@@ -38,6 +38,8 @@ The following headers are set by the Mendix Cloud. If any of these are present i
 | **SSL-Cipher**                              | `ECDHE-RSA-AES256-GCM-SHA384`                                | The TLS ciphers used for the HTTPS connection. |
 | **SSL-Client-S-DN**                         | `CN=Hans van Kranenburg,OU=RnD,O=Mendix,C=NL`                | The Subject DN string of the client certificate for an established TLS connection according to [RFC 2253](https://tools.ietf.org/html/rfc2253). |
 
+There can be additional headers set by the Mendix Cloud, not being documented in the list above. The presence of these headers and the values they have must not be relied on. They are not subject to the deprecation handling as shown below.
+
 ### 2.3 Deprecated request headers inserted by the Mendix Cloud
 
 The following headers are set by the Mendix Cloud, but are deprecated and will be removed in the future.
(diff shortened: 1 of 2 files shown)


content/releasenotes/developer-portal/index.md
```diff
@@ -14,6 +14,13 @@ For updates on the status of Mendix Cloud V4, Mendix Cloud V3, and other deploym
 
 ## 2019
 
+### June 15th, 2019
+
+#### Mendix Cloud Announcement
+
+* All **HTTP Request Headers** set by the Mendix Cloud which available to app developers are documented in [Mendix Cloud HTTP Request Headers](/developerportal/deploy/mendix-cloud-request-headers).
```
MarkvanMents: @knorrie if this was an API, I'd say it was a published/public API. However, this isn't quite the right word. So I've tried to get around it by making the above change.
Will this make sense to developers?

content/developerportal/deploy/mendix-cloud-request-headers.md
```diff
@@ -38,6 +38,8 @@ The following headers are set by the Mendix Cloud. If any of these are present i
 | **SSL-Cipher**                              | `ECDHE-RSA-AES256-GCM-SHA384`                                | The TLS ciphers used for the HTTPS connection. |
 | **SSL-Client-S-DN**                         | `CN=Hans van Kranenburg,OU=RnD,O=Mendix,C=NL`                | The Subject DN string of the client certificate for an established TLS connection according to [RFC 2253](https://tools.ietf.org/html/rfc2253). |
 
+There may be additional headers set by the Mendix Cloud, which are not documented in the list above. The presence of these headers, and the values they have, must not be relied on. They are for internal use only and are not subject to the deprecation handling described below.
```
MarkvanMents: @knorrie  I've tried to give a clue as to why they might be there - hope this is true and will make sense to the reader.

MarkvanMents:
@knorrie Does this have to be published now, or can it wait until Monday?
Let me know if you are happy with it.

content/releasenotes/developer-portal/index.md
knorrie: Yes, great.

content/developerportal/deploy/mendix-cloud-request-headers.md
knorrie: Yes, thanks, I like this.

knorrie:
Any time. Monday is also fine. No hurries. Have a nice weekend!

PR #1743, opened 2019-06-18
Maria resources

MariaShaposhnikova:


content/refguide/resources.md
```diff
@@ -0,0 +1,43 @@
+---
+title: "Resources"
```
MarkvanMents: The issue here is that Studio Pro calls these things *Documents*. Perhaps we can look at this in the meeting?

MarkvanMents: image

MariaShaposhnikova: In UI they are also called "resources" when you add them.

MarkvanMents: No - you are right - they can be called resources, or perhaps resource documents?
perhaps put a picture to show what you mean?

PR #1746, opened 2019-06-19
Update Condition+Property.md

samanthaO1987:
Added a "," in the first line for better readability.


Adam-Dupaski:
Hi @samanthaO1987 thanks for the PR, but unfortunately I don't agree with your change, as that is a dependent clause and the comma is unnecessary.

PR #1763, opened 2019-06-28
Add warning about PIN feature

feddovanede:
The documented implementation will not work when the PIN feature for hybrid apps is used.


ConnorLand:
Recieved, will evaluate now.

PR #1773, opened 2019-07-03
Document nanoflow call action

ekrem-aktas:
Bonus: also add JavaScript call action to activities. 

This feature will be available in 8 GA, so all of this PR must be held until that release.


ConnorLand:
This all looks good. I'll just need to check how the new activity order in the parent page aligns with the new doc order.

PR #1784, opened 2019-07-08
Update gfm-syntax.md

jacobboer:
With a real code example instead of a picture it looks much better.


Adam-Dupaski:
Hi, a picture is used here to include the backticks without rendering the code.

jacobboer:
Why not have both in that case? The picture to be totally clear about the input and the code to show the result...

PR #1785, opened 2019-07-08
Update gfm-syntax.md

jacobboer:
With picture and result


Adam-Dupaski:
The example is for writing the Markdown, not for how it is rendered. Duplication of the code is unnecessary.

PR #1787, opened 2019-07-08
Trends graphs [DEP-1852]/[TW-591]

MarkvanMents:
Documentation for trends V3 and trends V4 needs to be updated to:
* Make it in line with the new trends pages due to be released as part of CP-186
* Add new anchors for placeholders on the trends pages in the product
* Implement Documentation feedback item 264456 from 5/10/2017

Planned implementation is July 25th


content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
-The requests graph shows the number of requests that are sent from the client or systems that integrate with your application using web services. The number of requests per second is split up by request handlers.
+The requests graph shows the number of requests that are sent from the client or systems that integrate with your application using web services.
+
+![](attachments/trends/no-ext-reqs.png)
+
+The number of requests per second is split up by request handlers.
 
-"xas" lists general queries for data in data grids, sending changes to the server and triggering the execution of microflows. "ws" shows the number of web service calls that were done. "file" shows the number of file uploads and downloads. The "/" should not list any requests, because static content is directly served to the user by the front-facing web server, which is placed between the user and this application process.
+* **/** should not list any requests, because static content is directly served to the user by the front-facing web server, which is placed between the user and this application process
+* **api-doc/** ?????????????
```
MarkvanMents: How should I describe these other request categories?

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
+
+This graph shows the **total number of threads** that exist inside the running JVM process.
+
+![](attachments/trends/jvm-thread-count.png)
+
+Besides the threadpool that is used for external HTTP requests, as shown above, this includes the threadpool used for database connections, internal processes inside the Mendix Runtime, and optional extra threads created by the application itself, for example, using a threadpool in a custom module or custom Java code.
+
+### <a name="Trends-appcpu"></a>4.9 Application Node CPU Usage
+
+The CPU graph shows the CPU utilization in percentage, broken down into different types of CPU usage.
```
MarkvanMents: This says that the CPU utilization is in percentage - however the graph scales to 400.

I think we need to explain why it doesn't scale to 100 - is it the number of instances, or something else?

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
+
+![](attachments/trends/app-cpu.png)
+
+The most important value in here is 'user', which shows the amount of CPU time used for handling requests at Mendix Runtime and executing microflows and scheduled events. CPU usage of the database is shown in a separate graph below.
+
+### <a name="Trends-appdiskstatsthroughput"></a>4.10 Application Node Disk Throughput
+
+Disk throughput shows the amount of data that is being read from and written to disk. If there's more than one disk partition in the system, the /srv partition generally contains project files and uploaded files of the application, while /var generally holds the database storage.
```
MarkvanMents: Will this graph show the database storage, or is this now removed?

MarkvanMents: This comment about the /var partition appears throughout the document...

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
+
+This graph displays the amount of data that is stored on disk in absolute amounts.
+
+![](attachments/trends/app-disk-usage-bytes.png)
+
+If there's more than one disk partition in the system, the /srv partition generally holds project files and uploaded files of the application, while /var generally holds the database storage.
+
+### <a name="Trends-appdiskstatsutilization"></a>4.12 Application Node Disk Utilization in Percentage (%)
+
+Disk utilization shows the percentage of time that the disk storage is busy processing requests. This graph should be interpreted in combination with other graphs, like CPU iowait, disk iops, and number of running requests. For example, a combination of a moderate number of IO operations, low amount of disk throughput, visible cpu iowait, filled up memory disk cache, and reports of long running database queries in the application log could point to a shortage of system memory for disk cache that leads to repeated random reads from disk storage.
```
MarkvanMents: Is there a graph of CPU iowaits? What is it now called?

MarkvanMents: Have I chosen the correct other graphs?

MarkvanMents: What is filled up memory disk cache, and which graph shows it?

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
+
+### <a name="Trends-appdiskstatsiops"></a>4.13 Application Node Disk I/Os
+
+The Disk IO statistics show the number of disk read and write operations that are done from and to the disk storage. It does not show the amount of data that was transferred.
+
+![](attachments/trends/app-disk-ios.png)
+
+### <a name="Trends-appload"></a>4.14 Application Node Load
+
+This value is commonly used as a general indication for overall server load that can be monitored and alerted upon. The load value is a composite value, calculated from a range of other measurements, as shown in the other graphs on this page. When actually investigating high server load, this graph alone is not sufficient.
```
MarkvanMents: What is the range of this graph? 0-1? The example shows a low load, I assume. Would be nice if the y-axis was shown to 3 decimal places.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,260 @@ You can find the trends by following these steps:
...
+![](attachments/trends/app-disk-latency.png)
 
-### <a name="Trends-appmxruntimethreads"></a>4.7 Total Number of Threads in the JVM Process
+### <a name="Trends-appdiskstatsutilization"></a>4.16 Application Node Disk Utilization
 
-This graph shows the total number of threads that exist inside the running JVM process. Besides the threadpool that is used for external HTTP requests, as shown above, this includes the threadpool used for database connections, internal processes inside the Mendix Runtime, and optional extra threads created by the application itself, for example, using a threadpool in a custom module or custom Java code.
+Disk utilization shows the percentage of time that the disk storage is busy processing requests. This graph should be interpreted in combination with other graphs, like CPU iowait, disk iops, and number of running requests. For example, a combination of a moderate number of IO operations, low amount of disk throughput, visible cpu iowait, filled up memory disk cache, and reports of long running database queries in the application log could point to a shortage of system memory for disk cache that leads to repeated random reads from disk storage.
```
MarkvanMents: This is the same as the description of 4.12. Should we take this bit on how to analyze the graphs out and put it in a section at the end of the document so that it doesn't become repetitive, and it is easier to maintain?

content/developerportal/operate/trends.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Trends in Mendix Cloud v3"
```
MarkvanMents: Can we replace the V3 and V4 trends document with a single trends document for both versions.
This would make it easier to maintain if the pages are identical, or almost identical.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,276 @@ You can find the trends by following these steps:
...
 ## 5 Database Statistics
 
 The database statistics show the amount of database queries and mutations, as well as the total size of the database.
 
-### <a name="Trends-dbpgstatdatabaseVERSIONmain"></a>5.1 Mutations
+### <a name="Trends-appmxruntimeconnectionbus"></a>5.1 Number of Database Queries Being Executed
+
+This graph shows the number of database queries that are executed by your Mendix application.
+
+![](attachments/trends/no-db-queries.png)
```
MarkvanMents: Is the y-axis really millions of statements per second. Or is it mill-statements per second?

Originally said that the queries were broken down into microflows, but that does not seem to be the case.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,276 @@ You can find the trends by following these steps:
...
-This graph shows the number of database objects that were actually changed by database queries from the application. A single database operation that affects more than one object, will show up as single database query as measured from the application side, but will show the number of object actually changed here, as measured from inside the database.
+The queries are broken down into queries that actually modify data (insert, update, delete), queries that fetch data (select), and the number of transactions (like microflows from which these queries originate).
 
-### <a name="Trends-dbpgtableindexsizeVERSIONmain"></a>5.2 Index vs. Table Size
+### <a name="Trends-appmxruntimepgtableindexsize"></a>5.2 Database Table vs. Index Size
 
 This database size graph shows the distribution between disk space used for storing indexes and actual data. Remember, indexes actually occupy memory space and disk storage, as they're just parts of your data copied and stored, sorted in another way! Besides your data, indexes also have to be read into system memory to be able to use them.
 
-### 5.3 Application Node
+![](attachments/trends/db-table-vs-index.png)
```
MarkvanMents: Does the whole index get read into memory, or does the database also page indexes?

content/developerportal/operate/trends.md
MarkvanMents: > 
> 
> Is there a graph of CPU iowaits? What is it now called?

Found it - it is *5.6 Database Node CPU Usage*

content/developerportal/operate/trends.md
MarkvanMents: No - graphs are different, need to be documented separately.

content/developerportal/operate/trends.md
MarkvanMents: y-axis is MILLI statements per second. Need to document - see link from Knorrie

content/developerportal/operate/trends.md
MarkvanMents: I've moved this to the top of the document as a "How to read the graph" section.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
+
+Non-Persistable Entities live in the JVM memory and are garbage collected regularly. If you have a memory leak, the number of objects in memory will grow over time. This might be a problem. 
 
 ### <a name="Trends-appmxruntimesessions"></a>4.3 User Accounts and Login Sessions
 
 The sessions graph shows the number of logged-in named and anonymous user sessions for your application, next to the total number of existing login accounts in the application.
 
+{{% todo %}}[Which value is which?]{{% /todo %}}
+
+![](attachments/trends/user-accounts-logins.png)
```
MarkvanMents: Not clear which is the value for `total number of existing login accounts`.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
+
 ### <a name="Trends-appmxruntimejvmheap"></a>4.4 JVM Object Heap
 
 The JVM Object Heap graphs shows the internal distribution of allocated memory inside the application process for objects that are in use by microflows, scheduled events, and all other data that flows around inside the Mendix runtime process.
 
-One of the most important things to know in order to be able to interpret the values in this graph, is the fact that the JVM does not immediately clean up objects that are no longer in use. This graph will show unused memory as still in use until the so-called garbage collector, which analyzes the memory to free up space, is run. So, this graph does not show how much of the JVM memory that is in use before a garbage collection will have to stay allocated after the garbage collection cycle, because the garbage collection process will only find out about that when it's actually running.
+![](attachments/trends/jvm-heap.png)
```
MarkvanMents: Should we describe all the information in the graph. `eden space` for example?

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
 
 ### <a name="Trends-appmxruntimesessions"></a>4.3 User Accounts and Login Sessions
 
 The sessions graph shows the number of logged-in named and anonymous user sessions for your application, next to the total number of existing login accounts in the application.
 
+{{% todo %}}[Which value is which?]{{% /todo %}}
+
+![](attachments/trends/user-accounts-logins.png)
+
 ### <a name="Trends-appmxruntimejvmheap"></a>4.4 JVM Object Heap
```
MarkvanMents: Review the description - not clear whether it is just memory used by objects, or whether `all other data` is other sorts of data.

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
+This graph is primarily present to provide more insight in situations where the part of the real used memory *outside* the JVM Object Heap is growing too much, causing problems with memory shortage in the operating system.
+ 
 More information on this graph is available in a Tech Blog post: [What's in my JVM memory?](https://tech.mendix.com/linux/2015/01/14/whats-in-my-jvm-memory/)
 
-### <a name="Trends-appm2eeserverthreadpool"></a>4.6 Threadpool for Handling External Requests
+### <a name="Trends-appmemory"></a>4.6 Application Node Operating System Memory
+
+The memory graph shows the distribution of operating system memory that is available for this server.
+
+![](attachments/trends/node-os-memory.png)
```
MarkvanMents: Should we describe any of the other memory figures?

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
+
+{{% alert type="info" %}}
+CPU usage of the database is shown in [Database Node CPU Usage](#Trends-dbcpu), below.
+{{% /alert %}}
+
+### <a name="Trends-appdiskstatsthroughput"></a>4.10 Application Node Disk Throughput
+
+Disk throughput shows the amount of data that is being read from and written to disk.
+
+![](attachments/trends/app-disk-throughput.png)
```
MarkvanMents: Do we need to say something about srv/ and var/ or something else or remove?

MarkvanMents: This applies throughout the document

content/developerportal/operate/trends.md
```diff
@@ -41,133 +43,296 @@ You can find the trends by following these steps:
...
+
+### <a name="Trends-appdfabs"></a><a name="Trends-appdf"></a>4.11 Application Node Disk Usage (in Bytes)
+
+This graph displays the amount of data that is stored on disk in absolute amounts.
+
+![](attachments/trends/app-disk-usage-bytes.png)
+
+If there's more than one disk partition in the system, the /srv partition generally holds project files and uploaded files of the application, while /var generally holds the database storage.
+
+### <a name="Trends-appdiskstatsutilization"></a>4.12 Application Node Disk Utilization in Percentage (%)
```
MarkvanMents: Is this really the right description? It doesn't seem to match the example graph?
It looks like a percentage of _disk_ rather than a percentage of _time busy_?

content/developerportal/operate/trends.md
```diff
@@ -8,166 +8,327 @@ tags: ["Trends","v3","Mendix Cloud","Developer Portal"]
...
+
+For a single database operation that affects more than one object, this graph shows the number of objects actually changed, as measured from inside the database. However, the [Number of Database Queries Being Executed](#Trends-appmxruntimeconnectionbus) graph will only show a single database query.
+
+### <a name="Trends-appmxruntimepgstatactivity"></a>5.4 Number of Database Connections
 
-The most important value in here is 'user', which shows the amount of CPU time used for handling requests at Mendix Runtime and executing microflows and scheduled events. CPU usage of the database is shown in a separate graph below.
+The database connections graph shows the number of connections to the PostgreSQL server.
 
-#### <a name="Trends-appmemory"></a>5.3.2 Memory
+![](attachments/trends/no-db-connections.png)
```
MarkvanMents: Should we describe some of the categories here, or are they obvious?
Is there anything that the user should do about the _aborted_ idle in transaction if there is anything shown here?

content/developerportal/operate/trends.md
```diff
@@ -8,166 +8,327 @@ tags: ["Trends","v3","Mendix Cloud","Developer Portal"]
...
+This should go up and down with the usage of the application. The number of connections is limited to 50.
 
-### 5.4 Database Node
+### <a name="Trends-dbmemory"></a>5.5 Database Node Operating System Memory
 
-These are infrastructure level metrics with regards to the database node.
+The memory graph shows the distribution of operating system memory that is available for this server.
 
-#### <a name="Trends-dbcpu"></a>5.4.1 CPU
+![](attachments/trends/db-os-memory.png)
```
MarkvanMents: Do we need to explain more - there are lots of memory categories here.

content/developerportal/operate/trends.md
```diff
@@ -8,166 +8,327 @@ tags: ["Trends","v3","Mendix Cloud","Developer Portal"]
...
+![](attachments/trends/db-disk-throughput.png)
 
-## 6 Both Application and Database Node
+### <a name="Trends-dbdfabs"></a><a name="Trends-dbdf"></a>5.8 Database Node Disk Usage (in Bytes)
 
-The following infrastructure metrics are available for both the application node and the database node.
+This graph displays the absolute amount of data that is stored on disk.
 
-### <a name="Trends-appdiskstatsiops"></a><a name="Trends-dbdiskstatsiops"></a>6.1 Disk IOPS
+![](attachments/trends/db-disk-usage-bytes.png)
```
MarkvanMents: Do we need to explain the different categories of disk usage?

content/developerportal/operate/trends.md
```diff
@@ -8,166 +8,327 @@ tags: ["Trends","v3","Mendix Cloud","Developer Portal"]
...
 
-The Disk IO statistics show the number of disk read and write operations that are done from and to the disk storage. It does not show the amount of data that was transferred.
+### <a name="Trends-dbdiskstatsutilization"></a>5.9 Database Node Disk Utilization in Percentage (%)
```
MarkvanMents: Is this the right anchor - it is also used in 5.13, below?
I have also changed the description, which seemed to describe utilization rather than usage..

content/developerportal/operate/trends.md
```diff
@@ -8,166 +8,327 @@ tags: ["Trends","v3","Mendix Cloud","Developer Portal"]
...
+![](attachments/trends/db-os-memory.png)
 
-#### <a name="Trends-dbpgstatactivityVERSIONmain"></a>5.4.3 Database Connections
+The most important part of this graph is the **cache** section. This type of memory usage contains parts of the database that have been read from disk earlier. It is crucial to the performance of an application that parts of the database data and indexes that are referenced a lot are always immediately available in the working memory of the server, in the cache. A lack of disk cache on a busy application will result in continuous re-reads of data from disk, which takes several orders of magnitude more time, slowing down the entire application.
 
-The database connections graph shows the number of connections to the PostgreSQL server. This should go up and down with the usage of the application. The number of connections is limited to 50.
+### <a name="Trends-dbcpu"></a>5.6 Database Node CPU Usage
```
MarkvanMents: CPU also goes to 400%

content/developerportal/operate/trends.md
knorrie: Maybe refer to regular documentation about default request handlers that exist in the Mendix Runtime? For these images, the *-doc ones should be disabled in Mendix Cloud environments (they are by default), and show zero requests, since they leak a lot of interesting information about the internals of the application (yes, also in test environments, because the test url of a customer production app is very easy to guess).

content/developerportal/operate/trends.md
knorrie: Yes, 400 means that 4 cpu cores are available to schedule work on.

knorrie:
Keep in mind that understanding everything that is to be seen on the graphs, including the top 20 suspicious patterns, and including knowing how to correlate information in all the graphs to pinpoint problems is something that you can easily create a full week of training material about. IMO it's ok if the documentation for the trends is on the level of a "general introduction to give an idea what kind of stuff is displayed".

content/developerportal/operate/trends.md
knorrie: The application node disk will probably not "hold database storage".

Throughput is the amount of data that is read or written from/to disk per second. An example of something that can be seen here is (correlating with disk usage in bytes) the fact that inefficient queries with sorting start swapping temporary sort results to disk, causing a table mountain in this graph.

knorrie:
Also, the graphs were mainly reordered, and not changed.

content/developerportal/operate/trends.md
knorrie: I asked for this graph to be removed, since it's not useful for the user. The very long sentence is interesting, but it would fit at the Throughput section as well.

* Yes, iowait is displayed in the cpu graph. It's the pink color and if there's % of it, then a cpu was doing nothing, waiting for disk storage to retrieve information into memory.
* I don't understand the question "Have I chosen the correct other graphs"
* The memory usage graph shows "cache", also pink color. This is memory used to cache disk reads so that "hot" data is kept in memory. If there's a lack of space for this, you'll see disk read throughput go up, and cpu iowait go up. Having plenty of disk cache to store "hot" data is very beneficial for the performance of an application.

content/developerportal/operate/trends.md
knorrie: This graph should be removed, it's not useful for the user at all. I asked for that, but DEP team didn't want to remove anything.

content/developerportal/operate/trends.md
knorrie: For all the v3 graphs, they're rrdtool graphs, where units are automatically chosen based on the value. m = 'milli', which doesn't really make sense, but it can be interpreted as 0.02783 per second == 100 per hour

content/developerportal/operate/trends.md
knorrie: No, only stuff that's actually used by queries gets read into memory. So if I need to get object K from a table, and I have an index on A-Z, the database only needs to quickly read a few disk blocks to get that specific part of the index for K, which will tell it where exactly in the big pile of data the real K object is.

content/developerportal/operate/trends.md
knorrie: named users

content/developerportal/operate/trends.md
knorrie: You can steal text from here if you want:
https://github.com/mendix/m2ee-tools/blob/master/doc/munin.md

content/developerportal/operate/trends.md
knorrie: other sorts of data? I don't follow.

content/developerportal/operate/trends.md
knorrie: Sorry, I don't follow.

content/developerportal/operate/trends.md
knorrie: There's a bunch of default OS level graphs from munin that are just show in our page. This is in some cases not very optimal, since e.g. it shows incomprehensible labels like "xvdb" instead of just "hey, this is your database storage!". Also filesystem usage in bytes, with useless stuff (for the customer) like /run, /dev/shm, which all just clutter the whole thing.

This can be improved (it's not that hard) but the chance of this showing up on the top part of the roadmap is quite slim, unfortunately.

content/developerportal/operate/trends.md
knorrie: This graph is completely useless for customers, it should be removed.

content/developerportal/operate/trends.md
knorrie: See m2ee munin documentation.

content/developerportal/operate/trends.md
knorrie: Just mention that the most important are 'apps', which in this case is the postgresql 'application' ¯\_(ツ)_/¯, which shows what amount of OS memory is occupied by the processes actually doing the select queries etc. And, the cache part, which needs to hold all "hot" data, to keep everything working fast. (And maybe that a lot of concurrent database connections and queries push away disk cache, which can result in a vicious circle of system overload, when the customer wants to run an app that is too heavy on an environment that is to small)

content/developerportal/operate/trends.md
knorrie: See above, it would be nice if this graphs was improved a bit to just show what's relevant and not all other clutter.

content/developerportal/operate/trends.md
knorrie: Disk utilization graph only adds more confusion for the end user, it should be removed.

content/developerportal/operate/trends.md
knorrie: 4 cpu

knorrie:
Also note that I only added info about Mendix Cloud v3 graphs.

For Mendix Cloud v4, the graphs are totally different and they still need a lot of work to be usable. A bunch of them are blatantly incorrect or just miss important data. I would call them unusable for a serious production deployment currently, but that's my own unpopular opinion.

content/developerportal/operate/trends.md
RCatersels: How could we help to get it higher on the roadmap? Next week we have a roadmap session as MVP's with the product owners.

MarkvanMents:
@knorrie - thanks for the comprehensive review. I'll make changes where necessary.

content/developerportal/operate/trends.md
MarkvanMents: @mbyregowMx - I'll wait to see what is going to happen with this graph before changing anything.

content/developerportal/operate/trends.md
MarkvanMents: Added links to other information.

content/developerportal/operate/trends.md
MarkvanMents: @knorrie. The text says
> shows the internal distribution of allocated memory inside the application process for objects that are in use by microflows, scheduled events, and all other data that flows around inside the Mendix runtime process.

I'm not clear what `all other data` means in the sentence above. Is it *variables*, *mappings*, *image collections*, *microflow call stacks*, etc. Basically, is the JVM heap everything that the app needs to store, or is it just objects?

content/developerportal/operate/trends.md
MarkvanMents: @knorrie. The graph shows lots of OS memory uses (page_tables, cache, swap), are any of these important to the user, or is the only one they need to look at **apps**?

content/developerportal/operate/trends.md
MarkvanMents: I'll move this comment about /srv and /var into the general *how to read a graph* section and add more information if anyone says it is necessary.

content/developerportal/operate/trends.md
MarkvanMents: Section added to *how to read a graph* explaining the unit suffix

content/developerportal/operate/trends.md
MarkvanMents: @knorrie. The m2ee munin documentation looks very comprehensive. Is it publicly available, and should I put a link to it in this document for users who want more detail - or would that be a mistake?

content/developerportal/operate/trends.md
MarkvanMents: I'll leave this for now and return to it before publication.

content/developerportal/operate/trends.md
MarkvanMents: I've moved the long sentence to a `How to read all the graphs` section as it applies to several graphs.

@mbyregowMx I'll wait to see if there is a conclusion as to whether this should be included in the Trends page.

content/developerportal/operate/trends.md
MarkvanMents: @mbyregowMx, I've added some text. I'll wait to see what is in the final release.

content/developerportal/operate/trends.md
MarkvanMents: Explained in the `How to read the graphs` section.

content/developerportal/operate/trends.md
MarkvanMents: @mbyregowMx will wait for final release.

content/developerportal/operate/trends.md
knorrie: Yeah, "all other data" is not very helpful as an explanation :D What do you think of the documentation about heap and process memory here: https://github.com/mendix/m2ee-tools/blob/master/doc/munin.md#jvm-heap-memory-usage

(I will just keep throwing that link at you) ;-]

content/developerportal/operate/trends.md
knorrie: Most important is indeed that the green "apps" doesn't fill up the complete OS memory, and that the 'committed' line never exceeds the total amount of it. For the app OS less disk cache is needed. But, this is a graph from the default munin set. We could create adjusted simpler ones (just like I mentioned for disk space usage). That would be nice, but it's not high on the roadmap.

content/developerportal/operate/trends.md
knorrie: @RCatersels Hiring. Hiring is the biggest challenge for my team. We need people who understand both infrastructure and programming. All of them seem to either already work at FAANG or have another well-paid job not wanting to leave. For the work, stability and security is no 1. prio, then fixing potential time bombs for the future before they become an issue, and then nice to haves.

But this could be a fun thing for "delighter days", I think something like that is coming up again at the end of the summer here. Last time (was it around mendix world) I did the postgresql graphs work (which is finally gonna be included in the cloud portal overview now :| ) It's fun to work on.

content/developerportal/operate/trends.md
knorrie: You can link to it, it will just be there. The target audience of that documentation is "whoever wants to try run their own server for hosting a mendix app" (just running a mendix app, without riding the cloud hype waves).

knorrie:
Can I see the full new page somewhere? I've been reacting on snippets now. I can also help with collecting graph examples that show interesting (healthy) amounts of stuff and patterns going on.

MarkvanMents:
@knorrie - If you go to the `Files Changed` and find the trends.md file (at the bottom of the list somewhere) then the … options allow you to `View File` which gives you the current version in all its glory (missing out the extra {{% formatting commands).

Here for example is the latest one:
https://github.com/mendix/docs/blob/2eeac6e3a7ee9cdfe0b3c9562c95e5b7ee2b98c4/content/developerportal/operate/trends.md

Hopefully, that is what you wanted.
If you just want to look at diffs, including diffs from a single commit, then you can also do that.

This document now has so many diffs that it is not shown by default :-)

content/developerportal/operate/trends.md
MarkvanMents: Looks good, I'll take what is there.

MarkvanMents: @knorrie - Reaching the limit of my technical knowledge of the runtime here, but do data and variable widgets on pages also add to the Java Object heap? Your munin document talks about Java actions and microflows, but I'm not sure how the runtime deals with pages?

I suppose I'm saying that I don't know what is Java and what is Javascript/something else.

Perhaps there is a *runtime 101*  which I should be reading to get this information rather than bothering you? Any pointers?

In the meantime, I'll stick to just mentioning Java actions and microflows.

content/developerportal/operate/trends.md
MarkvanMents: I'll mention these two factors. At least it will help users navigate the graph. Thanks.

content/developerportal/operate/trends.md
MarkvanMents: Great - I've added a link in my section 3.2 Interpreting the Graphs

content/developerportal/operate/trends.md
knorrie: Very good questions.

If the web client needs contents for e.g. a datagrid, it will fire off an xpath query to the runtime. The mendix runtime (java) will translate the xpath into a database query, apply security rules and send the result back to the web client asap, after which all memory used is released again. While handing over the resulting data (mendix objects) from database via runtime to web client, the transfer "streams" as much as possible, so that memory usage in the runtime process is kept to a minimum. (i.e. doing an xpath that retrieves a billion objects will use roughly the same amount of jvm heap memory as getting one object).

When doing a "retrieve objects from database" in a microflow, the same happens, but instead of getting the result out again as fast as possible, the result ends up in memory used by the microflow. So, retrieving a billion objects in a microflow action will actually exhaust your JVM memory heap.

Does this make sense?

knorrie: Note: for 100% accuracy, ask data storage team. This is based on how things were in 2007, when I wrote parts of it for Mendix 2.0.0 together with Jonathan V. But I don't think the basics of it ever changed :-D

knorrie:
Ok thanks. Yes, this definitely needs other graph samples, which actually show something going on instead of a totally idle app... ¯\_(ツ)_/¯

content/developerportal/operate/trends.md
MarkvanMents: That is a great explanation. I'll check with the data storage team and let you know if they have totally rewritten it :-D.

From your description it sounds like I can say that 
> data queries will use only a small proportion of the JVM heap, and this amount is not dependent on the amount of data requested.

*(in other words, don't worry that your high memory usage might be due to the way you designed your page)*

knorrie: Yes, it's one of the brilliant things of using a MVCC database. Doing a query with a huge resultset, and still being able to slowly read the result from postgresql, because it will just make sure that the whole view on the data as it was when your query started stays preserved in the database until after you're done.

Sadly, this mechanism cannot be used in microflows. :@ Instead, you need to revert to getting batches (with offset/limit), where the (possibly very complex query) has to be executed over and over again, with all joining and sorting. Aaaand, this results in very interesting performance problems which can be clearly seen on the graphs. Some customers end up having apps that are hammering cpu and disk throughput and everything top speed for hours and hours because of this.

Ok, I'll go do something else now... (it's not a simple thing to implement).

knorrie:
I have put it on my todo for monday to gather new graphs examples for the page.

knorrie:
I replaced most of the graphs in a commit in branch knorrie/MvM-MetricsUpdateTrends

https://github.com/mendix/docs/blob/knorrie/MvM-MetricsUpdateTrends/content/developerportal/operate/trends.md

Tell me if you like it and want to have it, or pick the commit to your branch yourself.

Some additional remarks:
* Object Cache graph will not be shown for Mendix 7+ apps
* Application Node Disk Throughput now has information that is about *Database* disk throughput. The text should move there (or be deleted, or combined with the text for Database Node Operating System Memory). Application disk traffic is not super interesting, it should be a very boring graph. What it displays is getting/storing uploaded files and maybe doing a deployment is visible as a very small spike.
* Number of Database Queries Being Executed still has text about "against the number of transactions", however the graph does not have this transactions line any more (it was not useful, and it's not proper to related it to the others on the same scale).
* For Number of Database Connections it's useful to say something about the difference between active (a microflow or client xpath request is using the database *right now*), idle in transaction (the connection is in use by a microflow, but it's currently executing a microflow activity that is not using the database) and idle (connection is open and available to quickly allocate to a microflow or xpath request that needs it).
* I have not provided new graphs for the ones that I strongly propose to remove.

content/developerportal/operate/trends.md
mbyregowMx: Hi Mark, no graphs are removed as part of this change except for object cache graph which is hidden for apps with Mendix version higher than 7. All other graphs still exists and are just rearranged in the order mentioned in the ticket.

MarkvanMents:
@knorrie - those graphs are great. I like this one: 
image

Is it easier for you to merge, or for me? I'm happy to do it if that is easier for you.

Thanks - now I just have to find the same for V4 graphs!

content/developerportal/operate/trends.md
mbyregowMx: @MarkvanMents - As of now no graphs will be removed, so please keep the documentation for all the graphs for now.

MarkvanMents:
@knorrie - I've done the merge myself - thanks again for all your work on this.

MarkvanMents:
Hi @mbyregowMx and @knorrie 
The documents are now complete (only missing anchor for the V4 Application disk usage in percent). I've incorporated @knorrie 's final comments and formatted the V4 document to match the structure.
I am doing a final proof from a language point of view. Can you do a final technical review of V4 (I think we've got V3 right now) and make sure I have all the sections and anchors you expect.
It is ready for publishing tomorrow - I'll be able to change things after we publish tomorrow if we find the documentation and page don't match in some way.

MarkvanMents:
@knorrie,
We say in the text that the / line should not list any requests. However, the example you give me has up to 13 requests per second. Is this a weird app, or have I put the wrong information in?

image

MarkvanMents:
This is now ready for release tomorrow.

PR #1793, opened 2019-07-10
Additional step in managing-app-signing-keys

BartLuijtenMx:
The process for managing android keys in managing-app-signing-keys was not complete: the automated process does not automatically select the uploaded key. Added two steps to the howto section to add this.


ConnorLand:
Thank you @BartLuijtenMx. This was a valuable contribution to the documentation. The build process is complicated, and these final steps take users more explicitly to the finish line. Good contribution.

PR #1803, opened 2019-07-16
Update bottom bar color properties

jobvs:


MarkvanMents:
@ConnorLand - I think this is one of yours?

ConnorLand:
@MarkvanMents got it, thanks.

PR #1806, opened 2019-07-16
Adding without getting a duplicate object

TimvanSteenbergen:
Adding without getting a duplicate object is not totally obvious. Current add-description leaves you experimenting first.

### DIFF
--- a/content/refguide/change-list.md
+++ b/content/refguide/change-list.md
@@ -7,6 +7,8 @@ tags: ["studio pro", "List"]
 
 {{% alert type="warning" %}}
 This activity can be used in both **Microflows** and **Nanoflows**.
+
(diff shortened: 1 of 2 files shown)


MarkvanMents:
Passed to technical team to fully understand the concepts involved in lists

MarkvanMents:
@TimvanSteenbergen thanks for drawing our attention to this. It seems that there is more within Change List that needs documenting more clearly and I am reviewing the behavior with the development team.
I will publish an updated document reflecting your changes shortly.

MarkvanMents:
@TimvanSteenbergen - Thanks to your PR the teams are more aware of the issues and are making plans to collaborate on deciding how it should work.
As always, I can't promise when the improvements will make it to the product.
Thanks again for helping to improve Mendix.

PR #1812, opened 2019-07-18
Update syntax highlighting library

r0b1n:
- Updated Prism.js library. 
   - Include `tsx`.
   - Disabled line numbers as they were not visible.
 - Updated CSS to prevent blue borders around text in code blocks.


MarkvanMents:
Thanks for doing the research and making a PR @r0b1n - this looks great. I need to work out how to test it (not that I don't believe it will work but I'm not actually technically responsible for the site and I need to work out who is).
I hope we can apply it soon and improve all the code snippet formatting.

PR #1813, opened 2019-07-19
Ticket 85976 - Add Known Issue for 7.23.6

MarkvanMents:


MarkvanMents:
@ironthorn
I've added a KI about zero-length associations. Can you let me know if there is enough info here. Should I add a workaround or is it too simple to bother mentioning?

PR #1827, opened 2019-07-24
Update configuring-your-theme.md

wesleysieses:


ConnorLand:
This is for GA then, right @wesleysieses ?

wesleysieses:
> This is for GA then, right @wesleysieses ?

No this can also be done right away :)

PR #1830, opened 2019-07-26
Teammate Review

MariaShaposhnikova:


MariaShaposhnikova:
First of all, the document on this complicated topic is nicely structured and well-written. It guides users through the main terms, actions, and problems they might face. 
These are the main things that I changed while reviewing this doc:
- I deleted italics used to emphasize things. We do not have this kind of usage in our style guide, I believe
- I added introductory sentences before steps
- I corrected some links to compile more with our style guide
- I am not sure if that can be confusing or not, but there are two variants of the same term: "licensed node" and "licensed cloud node". I left only the first one
- Moved the warning into the Introduction, because in **Overview** it seemed out of place because we are explaining concepts there 
- Renamed **Overview** to **Basic Concepts and Overview**. (I think **Overview** is a bit too general.)

content/developerportal/deploy/licensing-apps.md
```diff
@@ -12,12 +12,16 @@ tags: ["App", "Node", "Developer Portal", "Deploy", "Link", "Unlink", "Licensed"
 
 To run Mendix apps in production on the Mendix Cloud, they must be linked to a licensed node. Without a licensed node you can still deploy your app and test it, however it will only run for a couple of hours before shutting down, cannot be scaled and has limited operational information available.
 
-This page explains how you can link your app to a licensed cloud node. It covers:
+This page explains how you can link your app to a licensed node. It covers the following:
```
MariaShaposhnikova: The part "it covers the following" is a bit confusing. The list below makes you think you will find a particular section where Free Apps are explained, for example. But the list shows two topics touched in the document. I would rewrite the list and add links to corresponding sections if possible.

content/developerportal/deploy/licensing-apps.md
```diff
@@ -12,12 +12,16 @@ tags: ["App", "Node", "Developer Portal", "Deploy", "Link", "Unlink", "Licensed"
...
 
-* Free Apps
+* Free Apps
```
MariaShaposhnikova: The term is  written either "Free App" or "free app". I think we should stick to one.

content/developerportal/deploy/licensing-apps.md
```diff
@@ -29,25 +33,19 @@ When you deploy an app to the cloud, whether it is a free app or a licensed app,
 
 This is the Mendix **environment**.
```
MariaShaposhnikova: I understand that bold is used to highlight a definition/explanation here, but here what is on out guide: Do not use bold for emphasis. Use italics or ALL CAPS (for example, “NOT”). 
Which also makes my deleting italics a useless change, because I always opened the **Italics** section to read our rules on italics, and this usage of it is _NOT_ mentioned there (cries in the corner).

content/developerportal/deploy/licensing-apps.md
```diff
@@ -69,27 +67,29 @@ For the specific case of swapping two apps between licensed nodes, see the guida
...
+1.  Go to the [Developer Portal](http://home.mendix.com).
 
 2.  Select the app which you want to unlink.
 
 3.  Go to **Environments** in the left navigation panel.
 
-    If the environments tab shows the following message, then your app is not currently linked to a node and you can go straight to the section [Connect Your App to a Licensed Node](#connect-app).
+    If the **Environments** tab shows the following message, then your app is not currently linked to a node and you can go straight to the [Connect Your App to a Licensed Node](#connect-app) section.
 
     ![](attachments/licensing-apps/link-node.png)
```
MariaShaposhnikova: I would highlight the link that users should click, because there are several they can click and it makes you search for it on the screenshot. 
Same comment applies to "unlink-free-app" image in right under this one.

content/developerportal/deploy/licensing-apps.md
```diff
@@ -69,27 +67,29 @@ For the specific case of swapping two apps between licensed nodes, see the guida
 
 When you remove an app from its environment you may want to take the data (database and file store) with it. It is a good idea, in any case, to take backups before performing major activities.
```
MariaShaposhnikova: I can't add a comment earlier, but it refers to **4 Linking Your App to a Licensed Node** one paragraph above.
Since there are many sub-sections and the process is quite long, I think it would be nice to briefly list all steps you need to (or can) take to succeed (with links to corresponding sections):
1. Back up your data.
2. Unlink from the current environment.
3. Connect to a licensed node.
4. Restore the backup.

content/developerportal/deploy/licensing-apps.md
MarkvanMents: Highlighted images as suggested

content/developerportal/deploy/licensing-apps.md
MarkvanMents: Very good idea - done

MarkvanMents:
@MariaShaposhnikova 

Thanks you for the review. Lots of positive suggestions, and it makes the document feel much more consistent, both internally and consistent with all our other documents.

PR #1835, opened 2019-07-29
clarification of replacement value

BrunoHolliger:


MarkvanMents:
Request confirmation from development team that this is the case.

content/refguide/string-function-calls.md
```diff
@@ -416,7 +416,7 @@ Replaces all occurrences of a regular expression with another string.
     * Type: string
 * The regular expression to match
     * Type: string
-* The replacement value
+* The replacement value (literal, which means it does not support backreference or captures)
```
zubivan: It's correct. Under the hood we're using Java's `String.replaceAll`. May I suggest saying `The literal replacement value` here, or, similarly to Java docs, `the string to be substituted for each match`.

MarkvanMents: Suggested text used, keeping additional information about backreferences etc.

PR #1838, opened 2019-07-30
Appstore 6.1.0 release

pluizigeEend:
Adds Release Notes.


MarkvanMents:
@pluizigeEend 

Hi Ann - you seem to have also got all your other changes into this PR - I think you must have branched from your own branch - or perhaps something else.

I'll go with what we have and see how it goes tomorrow - I think it is too late to start thinking about doing it differently.

I'll just do a bit of editing so that the information has the same feel as the other release notes.

I'll let you know if I need any more information

PR #1850, opened 2019-08-02
Add info on Studio upgrade

MariaShaposhnikova:


MariaShaposhnikova:
Hi @MarkvanMents k! Can you please have a look at info that I added. I am not sure about the section title.

MarkvanMents:
@MariaShaposhnikova ,
As discussed, I have moved Studio to the top, so that the user doesn't have to struggle to find the information.
I still find the concept of a *Studio* app having a Mendix version very strange. I wonder if we should perhaps talk about _compatibility with_ a version, rather than _having_ a version.
I still think that the Studio team should consider moving to versions which map 1-1 with Studio Pro version - but that is not really an issue for me to resolve.
Hope my changes look OK!

PR #1855, opened 2019-08-05
Some small widget-api content fixes

mijay:
- property vs prop - they are different things
- zip is a format, not an extensions


r0b1n:
While we are here, may we fix `key` Capitalization in here? Using `Key` is not correct. https://github.com/mendix/docs/pull/1855/files#diff-a6bdeadfa8e493618547830dc4f6aaa6L232

Adam-Dupaski:
@mijay I agree with the "zip" format differentiation - I've updated our Style Guide to reflect this; we still italicize file formats. I changed this via another commit (in which I made changes to other docs with this issue).

@ConnorLand please look at all commits in this PR to familiarize yourself with what was necessary.

Adam-Dupaski:
@r0b1n Fixed, thanks for pointing this out.

PR #1856, opened 2019-08-05
Include the automatic conversion of AtlasUI resources

mlaponder:
With Mendix 8 GA we added an automatic conversion, if possible. This PR adds this to the atlas migration guide.


Adam-Dupaski:
@ConnorLand familiarize yourself upon return

PR #1857, opened 2019-08-05
Add multiple constraint known issue to Mendix 8 GA

ishan1729:


ishan1729:
This pull request needs an internal RnD review first.

passalaqua:
I think I like this wording a little better:

> We return incorrect results for XPath queries when the same entity member (either an attribute or an association) is compared against different values.

What do you think?

Adam-Dupaski:
@ishan1729 which SDM verified this public known issue?

ishan1729:
> We return incorrect results for XPath queries when the same entity member (either an attribute or an association) is compared against different values.

Looks good to me. I will apply this suggestion.

Adam-Dupaski:
@ishan1729 is your review done and is this ready to be published?

passalaqua:
> @ishan1729 is your review done and is this ready to be published?

I reviewed it. We're ready to publish.

PR #1868, opened 2019-08-08
Feature/mds143976 sso worker token

HeikoOsigus:
SSO module offers now new functionality to fetch a Token for tasks which are executed in the background. E.g. a microflow which is called by a scheduled event.


HeikoOsigus:
Seems that the CI pipline had some trouble... How can i restart the pipline without make a new commit?

PR #1883, opened 2019-08-13
VisibilityMap not needed

BrunoHolliger:


ConnorLand:
Diego checked this and agreed it should be deleted. I checked there were no more instances of VisibilityMap.

PR #1906, opened 2019-08-16
Upgrade node to v12 and gulp to v4

akaRem:


.nvmrc
```diff
@@ -1 +1 @@
-8.1.3
+12.0.0
```
r0b1n: Use the latest `12`, not `12.0.0` ?

PR #1908, opened 2019-08-19
Rewrite info on HTTP headers

MariaShaposhnikova:


MariaShaposhnikova:
Approved by Feddo, he says the info is accurate.

PR #1919, opened 2019-08-23
Restructure and add info on empty cells

MariaShaposhnikova:


MariaShaposhnikova:
@DanielDamMx , can you please review the changes?

PR #1921, opened 2019-08-26
Added Release Notes for the SAP OData Connector.

zlogic:


zlogic:
Yes, we've released it just now.

PR #1949, opened 2019-09-04
Update 8.1.md

jpulleman:
Discussed with Ylber.


MarkvanMents:
@jpulleman 
Can you add a PUBLIC-KNOWN-ISSUE label and a bit more detail to the JIRA issue https://mendix.atlassian.net/browse/SPC-982 which Paul says relates to this issue?

That then closes the loop and makes sure that we remember to document the fix correctly.

Thanks.

MarkvanMents:
Label added to SPC-982

PR #1965, opened 2019-09-11
How to Pluggable widgets Editability and Visibilty

Andries-Smit:
The sample https://github.com/mendix/text-box-sample has been updated too


content/howto/extensibility/create-a-pluggable-widget-one.md
```diff
@@ -165,64 +165,64 @@ Open the **(YourMendixProject)/CustomWidgets/TextBox** folder in your IDE of cho
...
 	```tsx
 	export class preview extends Component<TextBoxPreviewProps> {
 		render(): ReactNode {
 			return <TextInput value={this.props.textAttribute} />;
 		}
 	}
 	```
-	
-	Before moving on from this step, you should remove the import lines concerning "Hello World" sample text from *TextBox.webmodeler.tsx* and *TextBox.tsx*, as these lines are no longer in use.	
-7.  Add a test widget to the project home page:<br />
```
ConnorLand: @Andries-Smit I think you removed this extra space. Was there a reason for this? Those extra spaces were put in place because when images follow number items, we need to use an extra space after the number to keep the image in-line.

PR #1982, opened 2019-09-19
Feature/hide elements iframe

JelteMX:
- Cleaned up the Gulpfile (had some issues while building locally)
- Adding an option to hide certain elements when you visit a page with `?show=iframe` in the URL
- Adding an option to disable all clicks when you visit a page with `?nolinks=1`

These last two options are being used in a ES project internally


JelteMX:
@Adam-Dupaski technically this is sound, can you merge this after succesfull build?

PR #1985, opened 2019-09-20
Added Best Practice bullet in section 2.4.3

WiekeBouten:
Added a bullit regarding best practices in data synchronization, based on experiences with 2 real applications and user feedback. People forget to synchronize, so automize it as much as possible to prevent data loss

### DIFF
--- a/content/refguide/offline-first.md
+++ b/content/refguide/offline-first.md
@@ -125,6 +125,7 @@ To ensure the best user experience for your Mendix application, follow these bes
 * Because network connections can be slow and unreliable and mobile devices often have limited storage, avoid synchronizing large files or images (for example, by limiting the size of photos)
 * Use an `isDeleted` Boolean attribute for delete functionality so that conflicts can be handled correctly on the server
 * Use before- and after-commit microflows to pre- or post-process data, or perform additional server-side logic using microflows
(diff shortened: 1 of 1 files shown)


content/refguide/offline-first.md
```diff
@@ -125,6 +125,7 @@ To ensure the best user experience for your Mendix application, follow these bes
 * Because network connections can be slow and unreliable and mobile devices often have limited storage, avoid synchronizing large files or images (for example, by limiting the size of photos)
 * Use an `isDeleted` Boolean attribute for delete functionality so that conflicts can be handled correctly on the server
 * Use before- and after-commit microflows to pre- or post-process data, or perform additional server-side logic using microflows
+* Help your user remember to synchronize their data so it is processed as soon as possible: you can check for connectivity and automatically synchronize in your 'Save' nanoflows, or remind a user to synchronize while using a notification or before logging out to ensure no data is lost
```
ConnorLand: @ekrem-aktas What exactly is a "save nanoflow"? We might need to use a more precise term.

WiekeBouten: @ConnorLand Is there an action required from my side? "your 'Save' nanoflow" could be replaced by "the nanoflow that commits your object"

ConnorLand: I was initially asking Ekrem, but your change works. Many thanks.

PR #1988, opened 2019-09-23
Add description of consistency errors

MariaShaposhnikova:


MariaShaposhnikova:
@MarkvanMents , I have implement your review and I would be very grateful if you can have a look at content/refguide/consistency-errors-pages.md (if you have time ofc). I am particularly interested in what you think of introduction. Is the goal of the document clear after I extended the intro? In this case, I will copy it for other consistency errors docs. 
Thanks in advance.

PR #1990, opened 2019-09-24
Add server-side paging and sorting how-to

mlaponder:
This is the server-side paging and sorting how-to.


ConnorLand:
@mlaponder Because I added so many image files, I am going to shut down this PR and upload the finished draft straight to development. I will send you a link. If you have any tech fixes, just let me know and I will make them on my end. I don't think there will be any more tech issues though, because my app worked after following your instructions.

PR #2002, opened 2019-09-30
Update Make It Native release notes to include v.1.1.0

smarki:


ConnorLand:
Android version out Sept 30. iOS Version out Oct 2 (probably). Will publish RNs as appropriate.

PR #2013, opened 2019-10-08
Local Push Notifications (Native)

ConnorLand:
I am opening this branch to collaborate and track changes with @mmehmetAliIzci , @IIsaku , and possibly @Andries-Smit . The first flow is done. I will be working on the subsequent sections during this week.


content/howto/mobile/native-local-notifications.md
```diff
@@ -0,0 +1,368 @@
...
+
+Explaining the workaround (all the steps/substeps above yes?):
+
+* Since the Dataview is on your homepage and nanoflow is set as a data source, it creates an empty notification object. This object will be referred to as XwithGUID.
+* We trigger a ACT_PassGuidToNotification: JSAction DisplayNotification creates a notification with respected params: title,subtitle,body,actionName,Guid. This is step happens before displaying notification.
+* User taps the notification, before executing ON_tapNotification(because you set it in 4.b), you pass **XwithGUID** to your widget then set its GUID attribute programatically. Since you set the GUID attribute of the widget to **XwithGUID**, now you can use this in your **ON_tapNotification** nanoflow.
+* Your widget executes **ON_tapNotification** that expects a notification object with GUID, which in this case will be **XwithGUID**. After this, you can safely use **ON_tapNotification**'s notification parameter for retrieval purposes.
+
+## 4 Scheduling a Notification and Cancelling It
+
```
mmehmetAliIzci: We should inform users that scheduled notifications wont show the notification itself if the app is on foreground. They put the app in the background (go to desktop?) to see the notification. Because mobile guidelines don't encourage developers to show "Push notification while the user is in the app". This is against the UX patterns because it distracts the users "While they are already in the app" and Scheduled/Remote Notification's sole purpose is making users go into the app.  

Although notification is not shown, this doesn't mean that notification is not triggered. We can still use "on Recieve" hook where can be set on Notification widget itself. This hook will allow Mendix developers to show their custom modal/page/pop-up when user receives the notification while they are on the app/

content/howto/mobile/native-local-notifications.md
```diff
@@ -0,0 +1,368 @@
...
+pic
+
+3. Click **OK**.
+
+Start and load the app on your mobile device and tap the nanoflow button. You will see a notification. Go to your device's start screen to see the notification badge on your app:
+
+pic here (from phone)
+
+### 3.3 Set an Action for When a Notification is Tapped or Received
```
mmehmetAliIzci: We should inform users that on Android, while using "Make it Native" app, these options wont work. However if user creates its own Native App via Native Builder, it will start working. So this limitation wont effect production level applications

content/howto/mobile/native-local-notifications.md
mmehmetAliIzci: Done

content/howto/mobile/native-local-notifications.md
mmehmetAliIzci: Creating a table for this

ConnorLand: Mehmet is making a table to show users the different capabilities of remote, local, and scheduled notifications. This will be a large series of tables and will exist as a reference guide. Once he publishes it later, I will make a link in this doc to it.

PR #2015, opened 2019-10-09
Update with new Publish REST video

Criccle:
Update with new Publish REST video


Adam-Dupaski:
@Criccle Thanks for the demos and the PRs!

PR #2059, opened 2019-11-04
Added note that this feature is only avalable for paid subscriptions

rvhofwegen:
Saw this passing by on slack, people searching for this feature. while they didn't have a paid subscription and didn't find this in the documentation


Adam-Dupaski:
Thanks René!

PR #2080, opened 2019-11-12
Add dynamic image support for PW

r0b1n:
Introduced in 8.4.0


ConnorLand:
@r0b1n Can you also please add a list of the supported image file types for pluggable widgets near your new text? It would be best to add this text today please and thank you.

r0b1n:
@ConnorLand I put a link to Images page. Also I added info box about Mendix version which bring support of SVG together with a warning that GIF format is not supported by native android apps.

Adam-Dupaski:
@r0b1n Further on "user" vs. "you" - this brings up a lot of interesting questions that Connor and I have just been discussing, so thanks for the comment. Right now, I see "user" and "you" as equivalent, with "you" being the preference - this is the case in most of our current documentation. This also means that in our documentation, "user" doesn't communicate anything necessarily different than "you." If we need another term for the reader of this specific kind of doc, we need to maybe look at other terms, such as "widget end-user."

ConnorLand:
> @r0b1n Further on "user" vs. "you" - this brings up a lot of interesting questions that Connor and I have just been discussing, so thanks for the comment. Right now, I see "user" and "you" as equivalent, with "you" being the preference - this is the case in most of our current documentation. This also means that in our documentation, "user" doesn't communicate anything necessarily different than "you." If we need another term for the reader of this specific kind of doc, we need to maybe look at other terms, such as "widget end-user."

@r0b1n I will be asking you a few questions later this week to determine if the docs are unclear about how they address certain widget developers and users. There might be a persona that the docs should be clearly addressing which the docs are not making explicit yet.

PR #2088, opened 2019-11-14
Added explanation for passed arguments

BrunoHolliger:
Could you please confirm this and add it to the documentation? In my point of view this seems to be important information. Thanks!

### DIFF
--- a/content/refguide/microflow-call.md
+++ b/content/refguide/microflow-call.md
@@ -28,7 +28,7 @@ The microflow that is called by this activity.
 
 ### 2.2 Arguments
 
-For each parameter of the microflow you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions). Variables are passed as value (not changeable via sub microflow), Lists are passed by reference (if the list is changed in sub microflow, the original list is altered).
+For each parameter of the microflow, you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions). Variables (strings, numbers, etc.) are passed as values (meaning, they are immutable, and not changeable via sub-microflows). Lists and objects are passed as references (meaning, if the list/object is changed in a sub-microflow, the original list/object is altered).
 
 ## 3 Output Properties
 
--- a/content/refguide/microflow-call.md
+++ b/content/refguide/microflow-call.md
@@ -28,7 +28,7 @@ The microflow that is called by this activity.
 
 ### 2.2 Arguments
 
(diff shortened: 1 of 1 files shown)


content/refguide/microflow-call.md
```diff
@@ -28,7 +28,7 @@ The microflow that is called by this activity.
 
 ### 2.2 Arguments
 
-For each parameter of the microflow you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions).
+For each parameter of the microflow you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions). Variables are passed as value (not changeable via sub microflow), Lists are passed by reference (if the list is changed in sub microflow, the original list is altered).
```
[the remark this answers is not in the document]
BrunoHolliger: Hi @zubivan!
Yes, of course, that's fine for me. As long the message is clear, I don't care. :-)
Have a great day,
-Bruno.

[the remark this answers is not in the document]
Adam-Dupaski: @zubivan This is supplementary info, so the technical level is fine. Let's work with the language we have here and aim for technical accuracy. Can you verify my changes?

content/refguide/microflow-call.md
```diff
@@ -28,7 +28,7 @@ The microflow that is called by this activity.
 
 ### 2.2 Arguments
 
-For each parameter of the microflow you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions). Variables are passed as value (not changeable via sub microflow), Lists are passed by reference (if the list is changed in sub microflow, the original list is altered).
+For each parameter of the microflow, you have to supply an argument of the same type. The values of the arguments are expressed using [expressions](expressions). Variables (strings, numbers, etc.) are passed as values (meaning, they are immutable, and not changeable via sub-microflows). Lists and objects are passed as references (meaning, if the list/object is changed in a sub-microflow, the original list/object is altered).
```
zubivan: Maybe we can rephrase it a bit, since objects and lists are also variables. Something like:

> There is a difference in a way argument values are passed to a sub-microflow:
> * Lists and objects are passed as references (meaning, if the list/object is changed in a sub-microflow, the original list/object is altered)
> * Primitive types (strings, numbers, etc.) are passed as values (meaning, they are immutable, and not changeable via sub-microflows).

zubivan: @Adam-Dupaski what do you think?

Adam-Dupaski: @zubivan Added you changes, thanks for the verification.
@BrunoHolliger Thanks for the PR!

PR #2090, opened 2019-11-14
Update JavaScript Actions JS version

Andries-Smit:
@ConnorLand could you please update the code screenshot  as the code template is now update with The numbers?

image


ConnorLand:
@Andries-Smit we chatted on slack about this. If it's easier, I can swing by your desk and we can figure out the fastest way to put the screenshots up.

PR #2138, opened 2019-12-04
Update value type for native style property in Mendix 8.6

tamelbrun:
See WTF-253


jordanch:
Based on subsequent changes to the client code (making the style property value an array with a single object), i believe this should be closed in favour of https://github.com/mendix/docs/pull/2150

@ConnorLand your proofread commit is incorporated.

PR #2147, opened 2019-12-05
Document file property for PW

r0b1n:
With mx 8.5


ConnorLand:
I'll add borders to the images after I merge.

PR #2150, opened 2019-12-06
Native style property

jordanch:
See WTF-253


ConnorLand:
@jordanch just curious -- what does adding .idea do?

jordanch:
@ConnorLand it's a directory containing JetBrains IDE specific settings, usually created when developing with an IDE from JetBrains. Here's some [further info](https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-).

PR #2160, opened 2019-12-11
Recommend Timeouts are set on HTTP connections

MarkvanMents:
see (ticket 69356) [TW-825]


MarkvanMents:
Agreed merge with Jaap.

PR #2166, opened 2019-12-12
Native Builder 3.1.0 release notes

smarki:


smarki:
Still adding some more things

ConnorLand:
This app was meant to come out with 8.5, but has been delayed by a day or two. I will merge a day or two after 8.5

PR #2188, opened 2020-01-03
Adds info about how to commit association changes.

pluizigeEend:
The Add and Remove buttons don't commit the changes on their own.
It is expected that the surrounding object (the Customer data view
as shown in the docs right above this text) has a Save button.
When the user saves the changes, the association changes that were
done from the Add and Remove buttons are committed.

Since this may not be apparently obvious, I added a short note in the
docs.

### DIFF
--- a/content/refguide/reference-set-selector.md
+++ b/content/refguide/reference-set-selector.md
@@ -37,6 +37,8 @@ The reference set selector looks a lot like a [data grid](data-grid) and consequ
 *   The **Add** button adds an association to an existing object. You will need to specify the page which opens when you want to add a new association. For more information, see [Add Button](control-bar#add-button).
 *   The **Remove** button removes the association to an object, but does not change or delete the object itself
 
+Note that you will need to have a **Save** button surrounding your reference set selector to save the association changes.  
+
 ## 2 Properties
 
 An example of reference set selector properties is represented in the image below:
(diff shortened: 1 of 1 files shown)


MarkvanMents:
@pluizigeEend 
Hi Ann - Thanks for the addition. Well spotted.

I assume that you could also commit the owning entity in a microflow if you wanted - it doesn't have to be a save button (just to be completely technically accurate!)

I'm a bit surprised that it doesn't auto-commit - but I don't really understand how that works, and it is better not to leave these things to chance.

pluizigeEend:
> I assume that you could also commit the owning entity in a microflow if you wanted - it doesn't have to be a save button (just to be completely technically accurate!)

That sounds correct to me as well. I can update the text to reflect that. (You don't _need_ to have Save button, but it is .... encouraged behaviour, I guess?)

> I'm a bit surprised that it doesn't auto-commit 

I was as well, which is what prompted me to make this change 😁

MarkvanMents:
@pluizigeEend 

Thanks for the clarification - I'll make a quick change and then you can check that it is still OK.

MarkvanMents:
@pluizigeEend 

I've made the change, and formatted it in a blue box as I think it is an important bit of information.

Let me know if it makes sense to you.

pluizigeEend:
@MarkvanMents Looks great! 👍 Sounds good to me :)

PR #2195, opened 2020-01-07
Document using hanatrial-schema as a database

MarkvanMents:


cstevandy:
@MarkvanMents it looks good. Good to go

PR #2197, opened 2020-01-08
Id's are marked with isAttribute=false

joostverhoog:
In Mendix 8.6.0, Id's in published OData metadata are marked with isAttribute="false".

Also: the note that long numbers may get truncated holds for Id's as well.


Adam-Dupaski:
Add namespace as tag and other tags

PR #2202, opened 2020-01-09
Invalid path parameter leads to 404, not 400

joostverhoog:
After "Is 12 a valid value for parameter petId?", I've added a choice. Because when it's a path parameter, the result is 404, not 400.


joostverhoog:
To be reviewed from a technical perspective first

PR #2213, opened 2020-01-15
Added prefix for constants

BrunoHolliger:
Thanks for checking.


Adam-Dupaski:
Hi @BrunoHolliger , thanks for the PR and suggestion, but I have verified this and we do not advocate or teach this as a best practice. I will thus have to close this PR.
-Adam

PR #2223, opened 2020-01-22
Describe what specific keys does for request handlers

emirmx:


content/developerportal/operate/datadog-metrics.md
```diff
@@ -149,13 +149,13 @@ You can specify which request handlers, microflows, and activities are reported
...
-* `ServiceRequestHandler` – OData requests
-* `RestRequestHandler`
-* `ProcessorRequestHandler`
-* `ClientRequestHandler`
-* `FileRequestHandler`
-* `PageUrlRequestHandler`
+* `WebserviceRequestHandler` - SOAP requests (`mx.soap.time`)
+* `ServiceRequestHandler` – OData requests (`mx.odata.time`)
+* `RestRequestHandler` - REST requests (`mx.rest.time`)
+* `ProcessorRequestHandler` - REST, ODATA, SOAP doc requests
```
MarkvanMents: Hi @emirmx 
Thanks for the PR.
Do you know which statsd/Datadog namespace these request handler doc request stats go to (i.e. like mx.soap.time above).

content/developerportal/operate/datadog-metrics.md
```diff
@@ -149,13 +149,13 @@ You can specify which request handlers, microflows, and activities are reported
...
+* `ClientRequestHandler` - `/xas` requests
```
MarkvanMents: Do the /xas requests need to be described a bit better - for example like they are here: https://docs.mendix.com/developerportal/operate/trends-v4#a-name-trends-appmxruntimerequests-a-4-1-number-of-handled-external-requests

content/developerportal/operate/datadog-metrics.md
```diff
@@ -149,13 +149,13 @@ You can specify which request handlers, microflows, and activities are reported
...
+* `FileRequestHandler` - File upload/download requests
+* `PageUrlRequestHandler` - `/p` requests
```
MarkvanMents: Is there a statsd namespace for these other request handlers too?

MarkvanMents:
@emirmx 
Thanks for finding out this information, it is a real improvement to the documentation. I have asked a couple of specific questions to see if there is even more information we can give. If I know the namespaces for all the request handler stats, then I could perhaps turn this list into a table?

If you don't know the answers, then we can publish this as it is for now and research more information later.

content/developerportal/operate/datadog-metrics.md
emirmx: ProcessorRequestHandler, ClientRequestHandler, FileRequestHandler and PageUrlRequestHandler all go to `mx.client.time`.

content/developerportal/operate/datadog-metrics.md
emirmx: Yes. We may also have a link to that table.

content/developerportal/operate/datadog-metrics.md
emirmx: ProcessorRequestHandler, ClientRequestHandler, FileRequestHandler and PageUrlRequestHandler all go to mx.client.time.

MarkvanMents:
@emirmx 

Thanks for the extra information.

image

Just two more questions:
1. Are the times returned in milliseconds (I assume so, but just to be safe!)
2. As all the request handlers have the same namespace, is there a way of identifying what they refer to more accurately - see picture for options provided by Datadog. For example, I assume xasid will identify the target for xas requests, but I can't see any tags which would label file requests or page requests.

content/developerportal/operate/datadog-metrics.md
MarkvanMents: Great

emirmx:
1. Times should be in millis.
2. Unfortunately there is no additional information currently.

> @emirmx
> 
> Thanks for the extra information.
> 
> image
> 
> Just two more questions:
> 
>     1. Are the times returned in milliseconds (I assume so, but just to be safe!)
> 
>     2. As all the request handlers have the same namespace, is there a way of identifying what they refer to more accurately - see picture for options provided by Datadog. For example, I assume xasid will identify the target for xas requests, but I can't see any tags which would label file requests or page requests.

MarkvanMents:
Thanks.
I've added the additional information you've provided. Can you quickly make sure I've got it right, then I can publish.

emirmx:
> Thanks again for adding to this document.

You're welcome. We should have provided the documentation in the first place :-)

PR #2233, opened 2020-01-27
8.6 RN

Adam-Dupaski:


content/releasenotes/studio-pro/8.6.md
```diff
@@ -38,7 +38,7 @@ Native layouts have a new **Layout type** property, which is similar to the one
 	{{% alert type="info" %}}This is based on an [upvoted idea from Bart Rikers](https://forum.mendixcloud.com/link/ideas/558) submitted to the [Mendix Idea Forum](https://forum.mendixcloud.com/link/ideas). Thanks, Bart!
 	{{% /alert %}}	
 * We optimized the waiting time that occurs after you commit to a Studio-enabled branch.
-* We rebuilt the [Connector](/refguide/view-menu#connector) to give it the same look and feel as the new [Project Explorer](//refguide/project-explorer). Highlighting via the **Filter** now works in the Connector. Should you find any issues, you can switch back to the old version of the Connector via **Edit** > **Preferences** > [New Features](/refguide/preferences-dialog#new-connector).
+* We rebuilt the [Connector](/refguide/view-menu#connector) to give it the same look and feel as the new [Project Explorer](//refguide/project-explorer). Highlighting via the **Filter** now works in the Connector. You can switch to the new version of the Connector via **Edit** > **Preferences** > [New Features](/refguide/preferences-dialog#new-connector).
```
Adam-Dupaski: We use "Should you find..." in https://docs.mendix.com/releasenotes/studio-pro/8.5 and https://docs.mendix.com/releasenotes/studio-pro/8.4. Why does this need to be changed here @MariaShaposhnikova ?

MariaShaposhnikova: The default has changed, the new version is not on by default.

Adam-Dupaski: The "old" version has always been the default, which means "The default has changed" is not correct. Something is inconsistent here with the other release notes and the documentation for this. See: https://github.com/mendix/docs/pull/2225/files

PR #2259, opened 2020-02-07
Mention downtime when deploying to a cluster

MarkvanMents:
@feddovanede and @cwesdorp 
Can you review what I've written and confirm that it is OK.
I notice that it doesn't say that the **After startup** is only performed by the cluster leader. Can either of you confirm whether this should be added, or shall I go to Runtime to confirm?

### DIFF
--- a/content/refguide/clustered-mendix-runtime.md
+++ b/content/refguide/clustered-mendix-runtime.md
@@ -37,22 +37,24 @@ Scaling out can be done using the Pivotal App Manager. For details on using the 
 
 Mendix Runtime has the concept of a cluster leader. This is a single node within a Mendix Runtime cluster that performs cluster management activities. These are the activities:
 
-* `Session cleanup handling` – each node expires its sessions (meaning, not being used for a configured timespan) and the cluster leader removes the sessions persisted in the database
+* **Session cleanup handling** – each node expires its sessions (meaning, not being used for a configured timespan) and the cluster leader removes the sessions persisted in the database
 	* In exceptional cases (for example, a node crash), some sessions may not be removed from the database, in which case the cluster leader makes sure this removal still happens
-* `Cluster node expiration handling` – removing cluster nodes after they have expired (meaning, not giving a heartbeat for a configured timespan)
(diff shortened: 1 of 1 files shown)


content/refguide/clustered-mendix-runtime.md
```diff
@@ -37,22 +37,24 @@ Scaling out can be done using the Pivotal App Manager. For details on using the
...
 
 These activities are only performed by the cluster leader. If the cluster leader is not running, the cluster will still function. However, the activities listed above will not be performed.
 
-The Cloud Foundry Buildpack determines which cluster node becomes the cluster leader and which becomes a cluster slave.
+The Cloud Foundry Buildpack determines which cluster node becomes the cluster leader and which become cluster slaves.
 
 ## 5 Cluster Startup
 
-The cluster leader is responsible for performing the database synchronization. So, if a new app deploy has been detected, all the cluster slaves will wait until the cluster leader finishes the database synchronization. When the database synchronization has finished, the cluster slaves and cluster leader will automatically become fully functional.
+The cluster leader is responsible for performing the database synchronization. This means that, in general, there will some downtime when you deploy a new version of the app while the cluster leader determines whether database synchronization is required.
```
feddovanede: there will _be_ some downtime.

feddovanede: As far as I know, in the Mendix Cloud there will _always_ be some downtime, as all application nodes are stopped first and then restarted.

MarkvanMents: Good catch

MarkvanMents: I'll follow this up.

According to Chris, there customers who have worked out a way of doing blue-green deployments. I'll try to find out who they are and where they are deploying, but I don't want to say _always_ if someone has found a way around it.

This is about clustering the runtime, so the reader may not be using Mendix Cloud?

MarkvanMents:
Thanks for the review, Feddo.

content/refguide/clustered-mendix-runtime.md
MarkvanMents: Have clarified this and amended text accordingly.

PR #2260, opened 2020-02-07
Update odata-representation.md

lightpeak:
Addition about many-to-many with links. Every link is a new call if you're using for example PowerBI. The result is that it takes a very long time to retrieve and is resource hungry. A better way is to publish the many-to-many through a RESTService. Thanks

### DIFF
--- a/content/refguide/odata-representation.md
+++ b/content/refguide/odata-representation.md
@@ -40,6 +40,8 @@ This means that you can only expose an association when the entity on the other 
 
 Using this method, you can expose both sides of the association and you can expose many-to-many associations.
 
+However it is important to know that **each link is a new call**. This means that large datasets take a long time to retrieve and use way more resources. If you're using external software like PowerBI to retrieve the data, a better solution for exposing the many-to-many associations might be a published RESTService. The JSON can then be converted into tables and records.
+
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@lightpeak please sign the [CLA](https://www.clahub.com/agreements/mendix/docs).

lightpeak:
Signed the CLA!

lightpeak:
> @lightpeak please sign the [CLA](https://www.clahub.com/agreements/mendix/docs).

Signed the CLA!

Adam-Dupaski:
@joostverhoog please add comments/content when possible.

Adam-Dupaski:
@lightpeak Thanks for the suggestion, but we have decided to not merge this for the following reasons:
- "Each link is a new call" is not accurate
- This document is about representation of OData, but your suggestion is about implementation
- We do not want to document for a specific user scenario in regards to a third-party tool like PowerBI
Cheers.

RCatersels:
True but info on how to expose multiple multiple references to a BI tool is relevant and not only for Microsoft BI.

Adam-Dupaski:
@RCatersels please see first point as main reason for not merging. We would push this info towards the Forum rather than documentation.

PR #2274, opened 2020-02-17
Add Animation style documentation

Andries-Smit:


ConnorLand:
@Andries-Smit when does this widget go live?

PR #2275, opened 2020-02-17
Clarify PersistentSessions setting and its usage.

hugomendix:


MarkvanMents:
Hi @hugomendix 
Thanks for the pull request.
You've added some information relating to sticky sessions. The only reference I can find for sticky sessions is in the on-premises configuration documents. Can you let me know if I've missed it somewhere.
Are sticky sessions relevant for anyone else (Mendix Cloud/SAP/etc.). If so, should we have some documentation about what they are, what the default is, and how to change it? 
If this is a non-Mendix concept, perhaps we could just link to an external document - providing users have some way of configuring whether sessions are sticky or not.

hugomendix:
Hi @MarkvanMents,
I was referring to sticky sessions as a general mechanism that one could implement in a load-balancer (see for instance https://en.wikipedia.org/wiki/Load_balancing_(computing)#Persistence). AFAIK we generally do not have sticky sessions in our cloud, so it would indeed only apply to on-premise setups. Since Mendix 7 we advise to just not disable persistent sessions in a clustered setup, which is why I've also trimmed down the warning on 'negative impact' which sounds much too scary ;-) Note that some old (<=Mx6) customers may have a custom setup in our cloud that does provide sticky sessions.

MarkvanMents:
@hugomendix 

So would it be better to say:
> In a clustered environment you must have persistent sessions. The only exception is for on-premises installations which have implemented sticky sessions.

We don't document for versions of Mendix below 6 as these are no longer supported, so your other use case shouldn't be an issue.

PR #2277, opened 2020-02-17
Mendix Cloud: TLSv1.3 support

pommi:


MarkvanMents:
@pommi - I've changed the text a bit - but haven't changed the meaning (I hope).
2 questions.

1. Do you know when TLS1.3 roll out will be complete?
2. Am I right in my assumption that it is already live on EU Free and this RN should be published today?

pommi:
@MarkvanMents Thanks for processing this so quickly. We are planning to finish the rollout to all clusters on Wednesday 26th.

MarkvanMents:
Added information to release note - said _by February 27th_ so customers don't think that it will definitely be there until the end of 26th.

PR #2282, opened 2020-02-20
Ec proofread published rest operation

ilagordhan:
This is the first proof of this document. Have checked some functionality with the app, but need to do it more thorougly by publishing REST service in a sample project to get to the deeper windows.
Also have questions about formats for parameters etc. 

I would consider re-structuring this doc to correspond to the operation panes


Adam-Dupaski:
New branch and PR required for cleanup

PR #2291, opened 2020-02-21
Make build version align with mx version

Andries-Smit:


ConnorLand:
I understand what we are doing here, but the code sample is labelled above it as an "example". It also already contains info (like the github and appcenter codes) which the user must change. So this snippet already cannot be copy/pasted without the user changing things. Therefore I think we should replace the Mendix version with an obvious placeholder number (such as 1.2.3) and be done with updating it every minor version. Thoughts @Andries-Smit ?

PR #2295, opened 2020-02-24
Update backups

xiwenc:
It's not supported to restore files-only in cloud v4.


MarkvanMents:
@xiwenc 
Hi Xiwen, 
Can you explain why you removed the use case for making a files only backup? Without a use case it isn't clear to me (and perhaps the users) why anyone would ever make a files only backup since you can't use the files without a database.

xiwenc:
@MarkvanMents the use case of making a files only backup has not been removed. Not sure why you think this is being done.

This change removes the incorrect advice suggesting a files-only backup can be restored in V4. That has never been the case.

MarkvanMents:
@xiwenc Thanks for the clarification- I'll merge this.

PR #2306, opened 2020-02-27
AR Docs

ConnorLand:
This PR contains 3 documents:

1. A small parent page **Use Augmented Reality**
2. **Get Started with AR**
3. **Build HoloLens Apps with Mendix**


ConnorLand:
I have added these docs to the **/docs/howto/** folder in [this commit](https://github.com/mendix/docs/pull/2306/commits/d94e762b9cf4a5861cd6cc219c23ed2a23fbe347). The docs being here instead of the **howto/mobile** folder means they would NOT show up in the navigation, but we COULD still send a link to the docs. This would just provide a little more obfuscation for casual readers browsing our site. They would be far less likely to stumble on the docs in this new place.

PR #2316, opened 2020-03-04
8.6.2 RN

Adam-Dupaski:


Robinfr:
We've backported an improvement from 8.7 to 8.6.2. The release notes for 8.7 were as follows:

```
We have improved how we handle the expired sessions of anonymous users in offline applications. When the session of an anonymous user expires, it is now restored in the background, which allows the anonymous user to continue using the app. (Ticket 94834)
```

Adam-Dupaski:
@Robinfr Added, thanks.

PR #2317, opened 2020-03-05
Gulp/Sass feedback 491652

ConnorLand:
I got some feedback recently from someone with an @us.epiuse.com email:

* Step 3 - Powershell needs to be run as admin
* Step 6 - Before running "npm" and "gulp dev" permissions have to be adjusted with "Set-ExecutionPolicy -ExecutionPolicy ByPass -Scope CurrentUser"
* Step 6 - Highlight before instructing user to run "gulp dev" that IF the user does not have an app running locally in Studio Pro prior to running gulp dev, their browser will try to load a page that will fail and powershell will show an error (I spent hours trying to figure this out thinking I had missed something). I did not realize this until I gave up on gulp and tried Calypso (https://docs.mendix.com/howto/front-end/calypso) only to realize it's a gulp wrapper but has additional instructions that clarify how to use it.
* Highlight that TortoiseSVN is optional and not required to run gulp.
TurtoiseSVN instruction need to be revised as well. There are steps missing or unwritten assumptions.

I have tried to implement all that feedback in this PR. However I am not sure how to note that TortoiseSVN is optional.

@JelteMX if you could validate the tech of my PR's diff, as well as advise on the Tortoise point, I would appreciate it.

Thank you.

### DIFF
--- a/content/howto/front-end/sass-eight.md
+++ b/content/howto/front-end/sass-eight.md
@@ -72,7 +72,7 @@ To prepare your app project, follow these steps:
 
 	At this point, you are ready to start working with Sass.
 
-9.  You also need to add the following selected items into the ignore list of TortoiseSVN for each new app project (or else your app project will take too long to commit a change). You will only have to do this ONCE when you set it up for your project :
+9.  You also need to add the following selected items into the ignore list of TortoiseSVN for each new app project (or else your app project will take too long to commit a change). You will only have to do this *once* when you set it up for your project :
 
 	![](attachments/set-up-sass/selected-ignore.png)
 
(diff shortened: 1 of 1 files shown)


ConnorLand:
> Reviewed. Content of steps seems fine, please fix the step numbers (they all changed to `1.`)

Hi and thanks for feedback. I changed the numbers intentionally. The site (I think it's the site?) now styles all numbering automatically if only "1's" are used, so doing this is actually easier as a best practice now.

ConnorLand:
@JelteMX could you commit about Tortoise being optional in the place you find most appropriate please?

ConnorLand:
Looks great. Thanks for your changes @JelteMX . Ready to merge when you approve.

PR #2326, opened 2020-03-10
Clarify Generalization vs 1-1 Associations documentation

MarkvanMents:
The documentation on choosing between generalizations and 1-1 associations is old and sometimes confusing. It has also had some additional information added which should have been in another document.

Have had a review from runtime (Jonathan) and will submit for review by data storage once his comments have been added.


content/refguide/query-over.md
```diff
@@ -92,3 +92,43 @@ Use the constraint `[QueryOver.File_Folder/QueryOver.Folder/QueryOver.Folder_Sub
...
+	{
+		this.B = __B == null ? null : main.proxies.Specialization.initialize(getContext(), __B);
+ 
+		// BEGIN USER CODE
+		return Core.retrieveByPath(getContext(), __B, "Main.Generalization_Specialization", Reverse);
+		// END USER CODE
+	}
+}
+```
+When setting the `Reverse` Boolean to true and using the `Specialization` object as the input, the returned list will contain all the Specializations associated to the Specialization.
```
MarkvanMents: Should this last sentence read 

> the returned list will contain all the **Generalizations** associated to the Specialization.

?

content/refguide/generalization-and-association.md
```diff
@@ -1,33 +1,33 @@
...
 
 ### 2.1 Performance
 
-In order to understand the impact on, and behavior of, the application, you need to understand the basic concepts of [Transactions](http://en.wikipedia.org/wiki/Database_transaction) and [(Database) Isolation Levels](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed). The Mendix Platform uses Transactions, which means that every microflow, commit, and delete will happen in a (database) transaction. The transaction is initialized as soon as the microflow executes its first write to the database. Retrieve activities will never start a transaction.
+In order to understand the impact on, and behavior of, the application, you need to understand the basic concepts of [Transactions](http://en.wikipedia.org/wiki/Database_transaction) and [(Database) Isolation Levels](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed). The Mendix Platform uses Transactions, which means that every microflow, commit, and delete will happen in a (database) transaction.  The transaction is initialized as soon as the microflow executes its first write to the database. Retrieve activities will never start a transaction.
```
MarkvanMents: From @jveldhuizen - needs review by Data Storage 
'The transaction is initialized as soon as the microflow executes its first write to the database. Retrieve activities will never start a transaction.' Can you check this with Data Storage? If I'm correct, transactions are initialized/started as soon as the microflow do the first database activity, like writing or retrieving. The main difference in write and read is that writes lock one or more rows in the database. That is the reason that we advise people to move the write activities as much as possible to the end of a microflow, so that you block other microflows (if they change the same data) as shortest as possible.

content/refguide/generalization-and-association.md
```diff
@@ -1,33 +1,33 @@
...
 
-The Mendix Platform uses the transaction level [Read Committed](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed), which means, as the name suggests, that only committed objects are visible outside the transaction. Any other microflows trying to access an object that is being changed at the same time, will have to wait until the transaction has completed. This is important to know, since this has significant impact on your choice between inheritance or associated objects.
+The Mendix Platform uses the transaction level [Read Committed](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed), which means,  as the name suggests, that only committed objects are visible outside the transaction. Any other microflows trying to access an object that is being changed at the same time, will have to wait until the transaction has completed. This is important to know, since this has significant impact on your choice between inheritance or associated objects.
```
MarkvanMents: From @jveldhuizen - needs review by Data Storage
Read Committed: please check this with Data Storage too. I think, the exact behavior depends on the DBMS. Maybe you don't always have to wait with accessing/reading data which is being changed by other transactions, but you will not see the actual changes done by the other transaction as long as that transaction doesn't have been committed. You will only see the committed ones.

content/refguide/generalization-and-association.md
```diff
@@ -117,7 +82,7 @@ This explanation might not have given you an explicit answer to the question on
 There are a few situations where a clear answer can be given:
 
 * Never use inheritance for entities with:
-  * A high number of transactions on the different sub entities (As a high we consider multiple changes or creates per second)
+  * A high number of transactions on  the different sub entities (As a high we consider multiple changes or creates per second)
```
MarkvanMents: Is this a reason to use one-to-one, or is this a situation where you shouldn't use either?

content/refguide/generalization-and-association.md
```diff
@@ -100,7 +65,7 @@ Making a decision between inheritance and associations is something you should d
...
+Don't just add inheritance because it is easier, or remove it because it is slower. Especially in scenarios were different object types have to go through a similar process it can be worth it to apply inheritance just so you can re-use functionality and increase the consistency and stability of your application. One place you definitely don't want to use inheritance is in a system with a high transaction volume.  Writing and updating records in tables with inheritance is slower than just updating a single table. If there are many new or changed objects loaded through Excel, web services, or any other integration then inheritance can slow the process down significantly.
```
MarkvanMents: Is this advice against using inheritance, or does this apply to both types and the developer should look at solutions which are neither inheritance nor one-to-one associations (e.g. using an entity with all the information, but leaving it blank where it is not relevant).

content/refguide/generalization-and-association.md
basdebakker: I agree with @jveldhuizen. Still, the difference in practice for this isolation level is negligible. The read locks are released after the retrieve action in either case.

content/refguide/generalization-and-association.md
basdebakker: Indeed, some databases (notably Oracle) use multi-version concurrency control to allow readers to see the committed version of an object while another transaction is updating it, while other databases will block the reader until the end of the updating transaction. This may even depend on a setting. Unfortunately, I don't know the exact behavior of all our supported databases.

content/refguide/generalization-and-association.md
```diff
@@ -1,96 +1,63 @@
...
 
 ### 2.2 Creating & Changing Objects
 
-When changing an object, the Mendix Platform will write those changes to the database as soon as you execute the commit activity. The update or insert query will be performed based on the values you have changed. This behavior varies per database, but most likely this will lock the record and prevent other users from reading it until the transaction has been completed (either finished or rolled back).
+When changing an object, the Mendix Platform will write those changes to the database as soon as you execute the commit activity or when you set the **Commit** action on the change object to **Yes**. The update or insert query will be performed based on the values you have changed. This behavior varies per database, but most likely this will lock the record and prevent other users from reading it until the transaction has been completed (either finished or rolled back).
```
basdebakker: Maybe it should be clarified that this commit activity/action only saves the object to the database and does not commit the database transaction. It should have been called "save".

content/refguide/generalization-and-association.md
```diff
@@ -1,96 +1,63 @@
...
+When you use a microflow to retrieve data, however, all data is retrieved (see [Microflows](#microflows), below).
 
-Here is an example inheritance:
+#### 2.3.1 Inheritance
 
-![](attachments/generalization-and-association/limitation.png)
+If you retrieve any specializations, the platform will only retrieve attributes from the generalization objects themselves if this data is required.
 
-In this example, a list of **Specializations** cannot be retrieved when using a standard by-association retrieve in a microflow if the input is the specialization.
+One exception to this is the System.User entity. If you have an overview of **Administration.Account**, the platform will include the System.User table if security is set to production, whether or not you show any System.User attributes. Both tables have a clustered index on the object id, so joining the information in the database is extremely efficient.
```
basdebakker: I think "extremely" should be reserved for extreme circumstances. I suggest "very".

content/refguide/generalization-and-association.md
basdebakker: The former. If the updates are mostly only to one of the objects in the 1-1 association, this may even be faster than using a single big entity, because the object to be updated will be smaller. As usual, a lot depends on the details of your application.

content/refguide/generalization-and-association.md
```diff
@@ -1,96 +1,63 @@
...
 
 Whenever you have a high number of write transactions in your application, it is far better to choose a one-to-one association, since this limits the number of tables that are being changed/locked during a transaction. However, if you do more inserts than updates it might be worth using inheritance again. Inheritance uses one less table to store the relationship; it does not have the association table. Therefore, any inserts require one indexed table fewer to be updated.
 
 ### 2.3 Retrieving Objects
 
-Mendix is optimized to only retrieve the data that is required for the action that is being executed. That means, for example: if you do not show any associated or inherited attributes, those objects will not be included in the retrieve queries.
+When using data widgets on a page, Mendix is optimized to only retrieve the data that is required for the action that is being executed. That means, for example: if you do not show any associated or inherited attributes, those objects will not be included in the retrieve queries.
```
basdebakker: Though probably obvious if you think about it, it could be pointed out somewhere that if associated/inherited attributes are used in access rules that apply to the current user, those will also need to be read. So it's not only about attributes being shown directly.

content/refguide/generalization-and-association.md
```diff
@@ -100,24 +67,31 @@ Making a decision between inheritance and associations is something you should d
...
+Although data retrieval for pages is optimized to only join with entities and retrieve attributes which are used in the data view, microflow retrieve activities are not. In a microflow, *all* columns are retrieved, from generalizations and specializations of the entity. In addition, all associated entities are retrieved where the selected entity is at the parent end of an association.
+ 
+For entities with a lot of attributes this leads to a lot of data being retrieved from the database. For entities with a lot of associations where they are the parent, this also leads to a lot of additional queries.
+ 
+The most efficient retrieval in a microflow is of an object with associations with owner type `Default` where the object is the `child`. If you retrieve this object, no association tables will be read by default, because you are the child. In this case, the association is implemented as one-to-many, which is not always handy. But making a one-to-one association, with owner type `Both` makes the association act like a parent-to-parent association so that a retrieval of the object will always retrieve the associated object.
```
basdebakker: Is it clear to users what the owner type is? The association properties dialog in Studio Pro uses different terminology.

content/refguide/generalization-and-association.md
basdebakker: The former, see also above.

content/refguide/query-over.md
MarkvanMents: I'll assume I'm correct and make the change.

content/refguide/generalization-and-association.md
MarkvanMents: @basdebakker - I'm not sure I completely understand what is being said here. Does the transaction complete when the data is committed, or is the transaction still open until the microflow finishes? For example, if you do a change with a commit does that complete the transaction within the change activity or is the transaction still held open until the end of the microflow?

MarkvanMents: Oh - I think I've found the answer in another of your comments, when you say that *commit* should really have been called *save*.
I'll use that answer!

content/refguide/generalization-and-association.md
MarkvanMents: Thanks for the comment - I'll review for superlatives and remove them where they seem unnecessary.

content/refguide/generalization-and-association.md
MarkvanMents: It is how it is described on the entity, but a good idea to add the 1-* description as well.

MarkvanMents:
@basdebakker 
Thanks for your responses to my questions and Jonathan's comments.
I have tried to address them all in the text.
Would you have time to have one more look at the document before I publish it. Thanks.

content/refguide/generalization-and-association.md
```diff
@@ -17,17 +17,21 @@ When defining closely related structures, it can be difficult to decide on the b
...
 
-The Mendix Platform uses the transaction level [Read Committed](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed), which means,  as the name suggests, that only committed objects are visible outside the transaction. Any other microflows trying to access an object that is being changed at the same time, will have to wait until the transaction has completed. This is important to know, since this has significant impact on your choice between inheritance or associated objects.
+Using the PostgreSQL database, the Mendix platform has the isolation level [Read Committed](http://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed), which means that only committed objects can be read outside the transaction. If another microflow is trying to read an object that is in the middle of being changed, it will have to wait until the transaction has completed. For some other databases the isolation level may be different. This is important to know, since this has significant impact on your choice between inheritance or associated objects.
```
basdebakker: The isolation level we use is the same for all databases. It's the way the database implements the isolation level that can be different.

content/refguide/generalization-and-association.md
```diff
@@ -37,61 +41,61 @@ When you change an object with inheritance the platform will potentially prevent
...
 
 ### 2.3 Retrieving Objects
 
-When using data widgets on a page, Mendix is optimized to only retrieve the data that is required for the action that is being executed. That means, for example: if you do not show any associated or inherited attributes, those objects will not be included in the retrieve queries.
+When using data widgets on a page, Mendix is optimized to only retrieve the data that is required for the action that is being executed. That means, for example, that if you do not show any associated or inherited attributes those objects will not be included in the retrieve queries. XPath constraints may cause additional data to be retrieved if they contain access rules, for example those that apply constraints based on the current user.
```
basdebakker: The model contains access rules for entities, the access rules may contain XPath constraints.

content/refguide/generalization-and-association.md
MarkvanMents: I've tried to reword this section to reflect your comment.

content/refguide/generalization-and-association.md
MarkvanMents: I've changed this too - hope it is more accurate now.

MarkvanMents:
Changes agreed via Slack. Will publish.

PR #2328, opened 2020-03-10
Specify required pg_dump version

jeohist:
The version of pg_dump bundled with recent PostgreSQL versions will set the archiver version header to 1.14. This will cause an error when restoring the database on our end, which fails the restore process.

### DIFF
--- a/content/developerportal/operate/restore-backup.md
+++ b/content/developerportal/operate/restore-backup.md
@@ -123,10 +123,8 @@ This contains the *db.backup* file. This is a PostgreSQL dump file created using
 
 {{% alert type="warning" %}}
 If the dump does not use the *custom format* then the restore will fail.
-{{% /alert %}}
 
-{{% alert type="warning" %}}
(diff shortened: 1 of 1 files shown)


content/developerportal/operate/restore-backup.md
```diff
@@ -125,6 +125,10 @@ This contains the *db.backup* file. This is a PostgreSQL dump file created using
 If the dump does not use the *custom format* then the restore will fail.
 {{% /alert %}}
 
+{{% alert type="warning" %}}
+If the dump is created with a version of pg_dump more recent than 9.6.17 then the restore will fail.
```
MarkvanMents: @jeohist Is there a workaround? (Other than using an older version of pg_dump)?
I might change this to be a bit more positive - "Use version 9.6.17 or below" if there is no workaround.

jeohist: There is no workaround unfortunately.

MarkvanMents: Thanks

PR #2337, opened 2020-03-16
microflow entity access bug

MxClaudineMaas:
This is the case in Mendix 7 and 8. It may also be a known limitation of Mendix 7 and 8.
There is a story to get this fixed in Mendix 9.

to verify. See comment of Jan-david on friday at 17:09 *(https://mendixsupport.zendesk.com/agent/tickets/97232 
)


MxClaudineMaas:
I just heard from Feddo that if they will fix it, it can not be in the docs but it should be a known issue. Apologies, closing this.

PR #2353, opened 2020-03-23
Update datadog-metrics.md

TimVredevoort:
The value in *7.2 Datadog Regions* for EU was incorrect. The current value is *datadog.eu*, this should be *datadoghq.eu*


MarkvanMents:
Hi @TimVredevoort 
Thanks for the contribution to the documentation.
Could you please sign the CLA. I know it is a bit of overkill for this change, but it gives us permission to use your work and makes it easier to see that the PR has passed all the checks.
Thanks.

PR #2354, opened 2020-03-23
add toLowerCase()

tieniber:
Added toLowerCase() to both the current location and remoteUrl. In testing with a client running on SAP cloud, the remoteURL used capital letters and the href did not. Adding toLowerCase() solved this issue.


ConnorLand (approved):

ConnorLand:
Thanks for solving this issue @tieniber !

PR #2371, opened 2020-03-31
Added "4.4.1 Preventing offline microflow call issues"

govorov:


ConnorLand:
Hey @govorov . I tried to change some small things. Does it still look technically correct? Also, is the clarity enough in the last two lines of 4.1.1? 

I wonder if they are introducing a new microflow in addition to the app's existing microflows?

govorov:
Hey @ConnorLand 
Thanks for reviewing!

> I wonder if they are introducing a new microflow in addition to the app's existing microflows?

What do you mean? If there's a new type of microflows for offline apps? No, it's all the same as before.
We call it "offline microflow call" if a microflow is reachable from a nanoflow that is called from an offline page.

PR #2375, opened 2020-04-01
Document parameterized actions of db-connector

passalaqua:
To be released with Mendix 8.8.0, the new Database-connector now has two extra actions:
- Execute Parameterized Query
- Execute Parameterized Statement

These are similar to the vanilla versions, but take string templates as input in place of a regular (user formatted) string.

### DIFF
--- a/content/appstore/connectors/database-connector.md
+++ b/content/appstore/connectors/database-connector.md
@@ -36,14 +36,18 @@ These are the prerequisites for using this connector:
 
 ### 3.1 Usage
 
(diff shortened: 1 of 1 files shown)


passalaqua:
I should probably say that the *new* version with syntax highlighting (which is not mentioned anywhere in the documentation) will be released with 8.8 (as in, I'll release a new module version compatible with 8.8), however the new feature has been there since module version 2.4.0 (and compatible with mendix version 8.6 and up).

PR #2376, opened 2020-04-01
Added new activities to Nanoflow documentation

Robinfr:
These were missing. Now more aligned with activities documentation.

Since I've also added microflows call, this should only get merged for **8.8** and later


ConnorLand:
All good no edits. Thanks for your use of the serial comma @Robinfr .

PR #2377, opened 2020-04-02
Improvement/loading page

JelteMX:
- Improving loader, no more scroll bars while loading sidemenu
- Moved new header script to bottom of the page and add the correct element to the top, so it doesn't bog down building the page any further. It should now display a correct top bar while it is not loaded (so the page doesn't jump).

### DIFF
--- a/_assets/styles/main.scss
+++ b/_assets/styles/main.scss
@@ -1,6 +1,8 @@
 @import "_sass/core/variables";
 @import "_sass/core/helpers";
 
+@import "_sass/mx-components/mxdock";
+
 @import "_sass/background";
 
 @import "_sass/base";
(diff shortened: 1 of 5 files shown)


JelteMX:
@iannbing I couldn't add you as a reviewer, but could you have a look at it? I think it makes sense. I tried to emulate the header basic structure and add some css, so the topbar with white background and blue top border is already there, even when the header isn't loaded yet.

_assets/styles/main.scss
```diff
@@ -1,6 +1,8 @@
 @import "_sass/core/variables";
 @import "_sass/core/helpers";
 
+@import "_sass/mx-components/mxdock";
```
iannbing, line 4: @JelteMX This line should be removed, right?

PR #2378, opened 2020-04-02
Update studio-apis-for-pluggable-widgets.md

takumagott:
Re-added the stripped down version of `action` property values. Also removed styleObj from Preview Export, as we only want to use `style` from now on.

### DIFF
--- a/content/apidocs-mxsdk/apidocs/studio-apis-for-pluggable-widgets.md
+++ b/content/apidocs-mxsdk/apidocs/studio-apis-for-pluggable-widgets.md
@@ -117,7 +117,11 @@ Name: {EventName}
 Description: {EventDescription}
 ```
 
-### 2.7 Attribute
+### 2.7 Action
+
+When an action is set, an empty object `{}` is passed to indicate that an action has been set. When no client action is set, the passed value will be `null`.
+
+### 2.8 Attribute
 
 A string containing the path of the selected attribute will be passed.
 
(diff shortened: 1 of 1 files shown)


ConnorLand:
@takumagott Earlier the Actions section was taken out because the customers didn't have any business in that part of the API. I take it this new draft is curated, useful information about Actions the customer would want to know? 

Just trying to make sense of the back and forth. Thank you.

takumagott:
In short, the removal was supposed to be temporary until we fixed it in the program. Now we fixed it.
In long:
We removed the part of the docs so customers would not use the previous, complex object structure which contained a lot of information about the action. We couldn't find any use cases for the usage of all that information, but we **could** think of use cases where you want to check whether a value is set or not. At the same time we didn't want to update the docs to this draft yet, because the now documented structure simply didn't exist yet (before, not setting an action resulted in an object with type "NoClientAction" and so on). That's why we wanted to remove the info temporarily until we actually updated the structure.

ConnorLand:
> In short, the removal was supposed to be temporary until we fixed it in the program. Now we fixed it.
> In long:
> We removed the part of the docs so customers would not use the previous, complex object structure which contained a lot of information about the action. We couldn't find any use cases for the usage of all that information, but we **could** think of use cases where you want to check whether a value is set or not. At the same time we didn't want to update the docs to this draft yet, because the now documented structure simply didn't exist yet (before, not setting an action resulted in an object with type "NoClientAction" and so on). That's why we wanted to remove the info temporarily until we actually updated the structure.

Ok great, thanks for the explanation. This PR looks good. I'll pass it through.

PR #2384, opened 2020-04-03
Added known issue for 8.8

Robinfr:
Added known issue about call microflow for offline apps. We will be fixing this soon.


Adam-Dupaski:
@ConnorLand please verify correct KI process is followed, review formatting and acronym usage

ConnorLand:
There may be more scenarios this KI needs to address, so we are on hold now pending further tech investigation.

Adam-Dupaski:
@ConnorLand please format workarounds as consistent with other KIs.

PR #2391, opened 2020-04-07
Mendix in iframes

MarkvanMents:
@feddovanede Here is the PR for the issue with Chrome and iframes discussed on Slack.
Please let me know if you want to make any changes.


feddovanede:
@MarkvanMents Looks correct to me. Thanks!

PR #2392, opened 2020-04-08
Update exposed libraries for Mendix 8.9

jobvs:
In 8.9 we're updating the libraries that can be used in pluggable widgets. Also added some libraries that were missing before.


ConnorLand:
Thanks @jobvs. The list of libraries is getting very long, which is a good thing, but it's a little hard to read. Instead of a long list of links in a doc, could we give **one link** to a repo or somewhere else which also has this info?

jobvs:
Currently there is no other place that has this info. It really is what this doc page is about. Do you have a suggestion for another place? Turning it into a table might help readability?

ConnorLand:
> Currently there is no other place that has this info. It really is what this doc page is about. Do you have a suggestion for another place? Turning it into a table might help readability?

Ok I understand @jobvs . Yeah, a table was going to be my plan B. I'll mess around with it and see if I can make it prettier. Thanks for adding the PR.

PR #2399, opened 2020-04-10
remove appbar2

iannbing:
appbar2 endpoint is deprecated. We need to remove it.

### DIFF
--- a/_assets/js/main.js
+++ b/_assets/js/main.js
@@ -780,31 +780,7 @@
 
     if (window.Intercom && window.__IntercomAppId && window.intercomSettings) {
       $('#back-to-top').addClass('with-intercom');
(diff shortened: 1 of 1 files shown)


JelteMX:
This does mean we will not be able to track users in the future, as Intercom was also depending on these values. We can still send messages back & forth, there is no way to know who's who on Intercom (from docs at least)

Adam-Dupaski:
@iannbing & @JelteMX - can/should we also remove the implementation of Intercom here, since it no longer works on the docs site?

Adam-Dupaski:
Just saw your email @JelteMX (didn't show up here): This does mean we will not be able to track users in the future, as Intercom was also depending on these values. We can still send messages back & forth, there is no way to know who's who on Intercom (from docs at least)

@iannbing is this an issue for your team/unit as far as Intercom usage/ownership?

PR #2406, opened 2020-04-16
Update call-rest form-data due to error in libraries

jonasbschaefer:
Due to a bug in the Mendix libraries the out-of-the-box form-data option does - as of now - not work. It adds redundant quotation marks in the Content-Type header. 
Therefore a hint to the option to write a form-data request yourself is worth including in the documentation

### DIFF
--- a/content/refguide/call-rest-action.md
+++ b/content/refguide/call-rest-action.md
@@ -97,6 +97,8 @@ This option allows you to generate a multipart/form-data request for multiple pa
 
 FileDocuments and images are are also supported for this option when used as variables in microflow expressions.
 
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
Hi @jonasbschaefer , what version of Studio Pro are you using?

Adam-Dupaski:
Hi @jonasbschaefer thanks for the PR. After investigating this issue, we see this is already planned as a fix for 8.9 (to be released next week), and it will also be fixed in an 8.6 patch. Therefore, documenting this workaround is unnecessary and not something we do regularly in the documentation anyway. In the future, if you think you have found such a bug, please send a Mendix Support ticket: https://support.mendix.com/hc/en-us
Cheers,
Adam

PR #2416, opened 2020-04-21
Modified SAP OData Connector Document

mayankmendix:


MarkvanMents:
Sent to Nidhi for review.

PR #2426, opened 2020-04-24
Add note for starting a native app for the first time

jobvs:
The long time running known issue (UICORE-211) that it takes a while to start native apps for the first time will be removed in 8.10 as the situation has improved a lot. It can still be a little slower than usual, so this adds a note to the getting started guide to set expectations for first time users.


ConnorLand:
Ok @jobvs no problem. Could you give me an estimate of how long these "first starts" can take vs regular ones? I'd like to be able to right "Your first time starting a native app may take up to three minutes" or something like that. Specificity can help users manage expectations more.

jobvs:
Sounds good. It's usually 30-40 seconds.

PR #2434, opened 2020-04-29
Add popup menu refstyle

mmehmetAliIzci:


ConnorLand:
@mmehmetAliIzci can you go through this PR and put all elements and other raw code mentions in code tics ( these: ``) please? This will help your PR be consistent with all the work of previous devs.

Note the other entries as examples here https://docs.mendix.com/refguide/native-styling-refguide. All the things they have in blue code tic formatting need to be done for this PR as well. I've added some tics in a recent commit as an example.

The basic rule is that examples of raw code gets code tics.

PR #2446, opened 2020-05-04
Update description of associations via feedback item

Adam-Dupaski:


content/refguide/associations.md
```diff
@@ -6,11 +6,17 @@ tags: ["domain model", "association", "studio pro"]
...
+### 1.1 Ownership {#ownership}
+
+The value of an association can only be viewed or edited from the object of the entity that is the [owner](association-member-properties#owner) of the association. Ownership in an association is signified by an arrow (note that the arrow does not signify direction). Either one entity or both entities can be the owner of the association. If one entity is the owner, there is an arrow that points from the owner to the other entity. If both entities are owners, there is a line between the two entities but no arrow. This is the only way the arrow can be controlled.
+
+It is important to understand why ownership exists, because it is not related to the database and should not be explained from that point of view. Ownership is implemented in Mendix so that you can change relationships dynamically rather than being stuck with your first design. For example, if you design something as a [one-to-many association](#one-to-many) and then need it to be a [many-to-many association with default ownership](#many-to-many), you do not need to rebuild your database, because Mendix handles it for you.
```
MarkvanMents, line 17: I would probably cut the second half of the first sentence.

```suggestion
It is important to understand why ownership exists. Ownership is implemented in Mendix so that you can change relationships dynamically rather than being stuck with your first design. For example, if you design something as a [one-to-many association](#one-to-many) and then need it to be a [many-to-many association with default ownership](#many-to-many), you do not need to rebuild your database, because Mendix handles it for you.
```

content/refguide/associations.md
```diff
@@ -6,11 +6,17 @@ tags: ["domain model", "association", "studio pro"]
...
+## 1 Introduction {#intro}
 
 An association describes a relation between entities. In the domain model, an association is represented by a line or arrow between two entities.
 
-The value of the association can only be viewed or edited from the object of the entity that is the _[owner](association-member-properties#owner)_ of the association. Either one entity or both entities can be the owner of the association. If one entity is the owner, there is an arrow that points from the owner to the other entity. If both entities are owners, there is a line between the two entities.
+### 1.1 Ownership {#ownership}
+
+The value of an association can only be viewed or edited from the object of the entity that is the [owner](association-member-properties#owner) of the association. Ownership in an association is signified by an arrow (note that the arrow does not signify direction). Either one entity or both entities can be the owner of the association. If one entity is the owner, there is an arrow that points from the owner to the other entity. If both entities are owners, there is a line between the two entities but no arrow. This is the only way the arrow can be controlled.
```
MarkvanMents, line 15: I think that the first sentence refers to modifying the association in an app. Perhaps that needs to be clarified a bit. And I'm not sure it is completely true to say that you can only view the association from the owner, although it is true for editing widgets.

Perhaps someone in data storage can look at this paragraph and clarify it. If this can't be done immediately it could be an additional PR?

PR #2464, opened 2020-05-12
Remove "not supported" from the documentation

fazzdev:
Having "not supported" in the documentation is not maintainable. When you have one, then you should have the complete list, which is difficult to come up with. It is better to document what we support instead, assume the rest is not supported.


Adam-Dupaski:
@fazzdev good points, thanks for the PR

PR #2475, opened 2020-05-18
Added documentation for the new sync config option

Robinfr:
Documentation for the new sync config option: "Nothing (preserve data)" that should be used in conjunction with the "Sync to device" action.

### DIFF
--- a/content/refguide/offline-first.md
+++ b/content/refguide/offline-first.md
@@ -68,8 +68,8 @@ Depending on the use-case, more fine-grained synchronization controls might be r
 
 - All Objects, download all objects applying the regular security constraints.
 - By XPath, only download the objects that are matching the [XPath Constraints](xpath-constraints) in addition to the regular security constraints.
-- Nothing (clear data), don't download any objects automatically, but do clear the database when performing a synchronization. This can be useful in cases where the objects should only be uploaded, for example a `Feedback` entity.
-- Nothing (preserve data), don't download any objects automatically, and don't clear the database when performing a synchronization. This can be useful in cases where you want have full control over the download phase and should be used in combination with the [Sync to Device](sync-to-device) action.
+- Nothing (clear data), don't download any objects automatically, but do clear the data stored in the database for this entity when performing a synchronization. This can be useful in cases where the objects should only be uploaded, for example a `Feedback` entity.
+- Nothing (preserve data), don't download any objects automatically, and don't clear the data stored in the database for this entity when performing a synchronization. This can be useful in cases where you want have full control over the download phase and should be used in combination with the [Sync to Device](sync-to-device) action.
 
 If you have custom widgets or JavaScript actions which use an entity that cannot be detected by Studio Pro in your offline-first profile (because its only used in the code), you can use customizable synchronization to include such entities.
 
--- a/content/refguide/offline-first.md
+++ b/content/refguide/offline-first.md
@@ -62,11 +62,14 @@ After the download process (and thus synchronization) is completed, the widgets 
 
 ### 2.3 Customizable Synchronization {#customizable-synchronization}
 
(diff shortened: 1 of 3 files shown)


content/refguide/offline-first.md
```diff
@@ -62,11 +62,14 @@ After the download process (and thus synchronization) is completed, the widgets
...
 
-Depending on the use-case, more fine-grained synchronization controls might be required. Therefore, an app's  synchronization configuration can be customized to override the default configuration. It is possible to limit what is downloaded by using [XPath Constraints](xpath-constraints), which is applied in addition to the security constraints.
+Depending on the use-case, more fine-grained synchronization controls might be required. Therefore, it is possible to change the download behaviour for an entity. You can choose between the following options:
 
-Furthermore, it is possible to disable downloads for an entity. This can be very useful in cases where the objects should only be uploaded, for example a `Feedback` entity.
+- All Objects, download all objects applying the regular security constraints.
+- By XPath, only download the objects that are matching the [XPath Constraints](xpath-constraints) in addition to the regular security constraints.
+- Nothing (clear data), don't download any objects automatically, but do clear the database when performing a synchronization. This can be useful in cases where the objects should only be uploaded, for example a `Feedback` entity.
```
ekrem-aktas, line 71: "... do clear the database ..."

It feels like this option influences the whole offline db, what about something like: `... do clear the data stored in the database for this entity when...."

Robinfr: Good point, will change it.

Robinfr: Changed ✔️

ConnorLand:
Hi @Robinfr and @ekrem-aktas . As I slacked Robin, I believe these changes to the different sync config options should go to this existing Navigation reference guide https://docs.mendix.com/refguide/navigation#customize (Section 4.3.2). Do you think that's best? 

If so, can you please copy/paste the text I proofed, drop is in the Navigation refguide, and then commit to this PR again? That way the new language will end up in the right place and I can merge it easily on release day.

Robinfr:
Hmm it seems we have the same information in multiple places. I'm fine with moving it and referencing it from the Offline first guide? What do you think @ekrem-aktas ?

ConnorLand:
> Hmm it seems we have the same information in multiple places. I'm fine with moving it and referencing it from the Offline first guide? What do you think @ekrem-aktas ?

Indeed we do want the info only in one place. However, it would have to stay in the Navigation refguide and be referenced by Offline First because documentation is tied to UI. These options are offered as a part of the navigation profile UI, so the documentation for those options needs to be in that location.

ekrem-aktas:
> Indeed we do want the info only in one place. However, it would have to stay in the Navigation refguide and be referenced by Offline First because documentation is tied to UI. These options are offered as a part of the navigation profile UI, so the documentation for those options needs to be in that location.

This form is useful only in the context of offline synchronization, otherwise we have to explain what does an "Upload" or "Download" mean in the navigation documentation. It is a bit technical, and we explain them in the offline-first docs. We'll we changing this form in the near future, where there will be more options, more tied to context of synchronization.

I'd have a screenshot & short description of the form and refer to offline-first docs, where the options are detailed.

ConnorLand:
> > Indeed we do want the info only in one place. However, it would have to stay in the Navigation refguide and be referenced by Offline First because documentation is tied to UI. These options are offered as a part of the navigation profile UI, so the documentation for those options needs to be in that location.
> 
> This form is useful only in the context of offline synchronization, otherwise we have to explain what does an "Upload" or "Download" mean in the navigation documentation. It is a bit technical, and we explain them in the offline-first docs. We'll we changing this form in the near future, where there will be more options, more tied to context of synchronization.
> 
> I'd have a screenshot & short description of the form and refer to offline-first docs, where the options are detailed.

Robinfr:
@ConnorLand I've made the changes as requested. I've removed the options from the navigation and now refer to the offline first documentation for more information.

PR #2477, opened 2020-05-18
[DAT-2105] Improve public documentation on db case sensitivity

passalaqua:
This centralizes information about which dbs are case sensitive/insensitive/configuration dependent when used with Mendix.

NN

### DIFF
--- a/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
+++ b/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
@@ -22,7 +22,7 @@ When setting up a new database for Mendix, most of the settings can be left to t
 
 In the database options, the default properties need to be evaluated. When choosing a collation, pay attention to the type of collation you are going to use. Mendix uses UTF-8 for all data evaluation. Depending on your exact locale, you will most likely want to choose one of the `SQL_Latin1_General_` collations. The exact encoding will depend on your OS. For an *en_US* installation, this will be `CP1`.
 
-The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Name](https://docs.microsoft.com/en-us/sql/t-sql/statements/windows-collation-name-transact-sql).
+The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](/refguide/case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Name](https://docs.microsoft.com/en-us/sql/t-sql/statements/windows-collation-name-transact-sql).
 
 As a recovery model, Mendix only requires the **Simple** mode. The functionality offered in the **Full** recovery model option is not used by Mendix; usage of this recovery model will not hurt Mendix, but it could increase the data usage of all the transactions and might slow down any rollbacks in case of an error.
 
--- a/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
+++ b/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
@@ -22,7 +22,7 @@ When setting up a new database for Mendix, most of the settings can be left to t
 
 In the database options, the default properties need to be evaluated. When choosing a collation, pay attention to the type of collation you are going to use. Mendix uses UTF-8 for all data evaluation. Depending on your exact locale, you will most likely want to choose one of the `SQL_Latin1_General_` collations. The exact encoding will depend on your OS. For an *en_US* installation, this will be `CP1`.
 
-The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
+The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Name](https://docs.microsoft.com/en-us/sql/t-sql/statements/windows-collation-name-transact-sql).
 
 As a recovery model, Mendix only requires the **Simple** mode. The functionality offered in the **Full** recovery model option is not used by Mendix; usage of this recovery model will not hurt Mendix, but it could increase the data usage of all the transactions and might slow down any rollbacks in case of an error.
 
--- a/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
+++ b/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
@@ -22,7 +22,7 @@ When setting up a new database for Mendix, most of the settings can be left to t
 
 In the database options, the default properties need to be evaluated. When choosing a collation, pay attention to the type of collation you are going to use. Mendix uses UTF-8 for all data evaluation. Depending on your exact locale, you will most likely want to choose one of the `SQL_Latin1_General_` collations. The exact encoding will depend on your OS. For an *en_US* installation, this will be `CP1`.
 
-The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations, see [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
+The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx) and [Case Sensitive Database Behavior](case-sensitive-database-behavior).
 
 As a recovery model, Mendix only requires the **Simple** mode. The functionality offered in the **Full** recovery model option is not used by Mendix; usage of this recovery model will not hurt Mendix, but it could increase the data usage of all the transactions and might slow down any rollbacks in case of an error.
 
--- a/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
+++ b/content/developerportal/deploy/setting-up-a-new-sql-server-database.md
@@ -22,7 +22,7 @@ When setting up a new database for Mendix, most of the settings can be left to t
 
 In the database options, the default properties need to be evaluated. When choosing a collation, pay attention to the type of collation you are going to use. Mendix uses UTF-8 for all data evaluation. Depending on your exact locale, you will most likely want to choose one of the `SQL_Latin1_General_` collations. The exact encoding will depend on your OS. For an *en_US* installation, this will be `CP1`.
 
-The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx) and [Case Sensitive Database Behavior](case-sensitive-database-behavior).
+The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
 
 As a recovery model, Mendix only requires the **Simple** mode. The functionality offered in the **Full** recovery model option is not used by Mendix; usage of this recovery model will not hurt Mendix, but it could increase the data usage of all the transactions and might slow down any rollbacks in case of an error.
 
--- a/content/refguide/case-sensitive-database-behavior.md
+++ b/content/refguide/case-sensitive-database-behavior.md
@@ -11,6 +11,7 @@ Case sensitive string operations are those that consider upper and lower case as
 Operations that affect strings in queries (for instance, inside an XPath constraint) can be case sensitive or case insensitive depending on the database vendor, version, and configuration used by the Mendix application.
 
 It is also important to note that each implementation of case insensitive operation may treat the case normalization of letters differently.
+As a general rule, it is based on locale which is either specified explicitly on database/table, or the database infers it from the OS.
 This normalization of letters, or case folding, is further described in the [W3C case folding wiki page](https://www.w3.org/International/wiki/Case_folding).
 
 For the purposes of this document we can divide these operations three categories:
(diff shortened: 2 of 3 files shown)


passalaqua:
To the maintainers: I believe @nimatrueway will still review part of these changes. Once they are done, we will inform you that this is content accurate.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+
+### 2.7 SAP HANA
+
+Case sensitive by default. Cannot be configured otherwise.
+
+It does not support sorting or comparison on attributes of strings of unlimited length.
+
+## 3 Overview of Default Case Sensitivity
+
+This table presents the default case sensitivity by different database types:
```
basdebakker, line 66: It would be good if the table showed the default for configuration dependent values.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+
+Case sensitive by default, but its behavior depends on collation. Setting collation to `SQL_TEXT_UCC` makes all comparisons case-insensitive.
+
+### 2.2 POSTGRESQL
+
+Case sensitive with no possible configuration. String functions are case insensitive as they are implemented using the `ILIKE` SQL operator.
+
+### 2.3 MARIADB & MYSQL
+
+Case insensitive by default, but its behavior depends on collation. The default collation is `utf8_general_ci` when the `utf8` character set is used or `latin1_swedish_ci` when the `latin1` character set is used. Changing this collation will affect case sensitivity.
```
basdebakker, line 38: Shouldn't the table then show "configurable"? And why are we using swedish, is that a MySQL default?

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+Sometimes other factors such as settings specific environment or session variables.
+Both the available collation and any other possible settings are database vendor (and sometimes version) specific.
+
+## 2 Behavior of Case Sensitivity by Database Type
+
+The following section describes the default behavior of the following databases in Mendix.
+
+### 2.1 HSQLDB
+
+Case sensitive by default, but its behavior depends on collation. Setting collation to `SQL_TEXT_UCC` makes all comparisons case-insensitive.
```
basdebakker, line 30: What about the string functions, which are shown as "case insensitive" in the table?

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+
+Case sensitive with no possible configuration. String functions are case insensitive as they are implemented using the `ILIKE` SQL operator.
+
+### 2.3 MARIADB & MYSQL
+
+Case insensitive by default, but its behavior depends on collation. The default collation is `utf8_general_ci` when the `utf8` character set is used or `latin1_swedish_ci` when the `latin1` character set is used. Changing this collation will affect case sensitivity.
+
+### 2.4 SQL SERVER
+
+Case insensitive by default, but its behavior depends on collation. The default recommended collation is `SQL_Latin1_General_CP1_CI_AS`. For more information, see our guide on [Setting Up a new SQL Server Database](setting-up-a-new-sql-server-database) and the [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
```
basdebakker, line 42: So what does the collation do with the string functions?

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+
+### 2.5 DB2
+
+Case sensitive. Cannot be configured otherwise.
+
+It also does not support sorting on attributes of strings of unlimited length.
+
+### 2.6 ORACLE
+
+Case sensitive by default, but its behavior depends on collation. The default collation is `binary`, but a `binary_ci` is available for case insensitive operations.
```
basdebakker, line 52: So what does the collation do with the string functions?

content/refguide/case-sensitive-database-behavior.md
passalaqua: Based on MySQL's own documentation, if you use `latin1`, then `latin1_swedish_ci` will be the default collation: https://dev.mysql.com/doc/refman/8.0/en/charset-we-sets.html

content/refguide/case-sensitive-database-behavior.md
passalaqua: Nima and I considered that possibility, but in some cases there really isn't a default (like in SQL Server, it seems). I would prefer people not to assume they're using the default and check their own configs instead (for instance, in cases where customers don't actually change the db collation but it's inherited from the environment, which might have been configured differently).

Still, I'm fine either way.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+
+Case sensitive by default. Cannot be configured otherwise.
+
+It does not support sorting or comparison on attributes of strings of unlimited length.
+
+## 3 Overview of Default Case Sensitivity
+
+This table presents the default case sensitivity by different database types:
+
+| Case Sensitive Operation/Database Types  | DB2 | HSQLDB | MARIADB/ MYSQL | ORACLE | POSTGRESQL | SAP HANA | SQL SERVER |
```
nimatrueway, line 68: Here's the updated table

| Database      | Comparison | Sort   | String Functions |
| ------------- | ---------- | ------ | ---------------- |
| DB2           | S          | S[1]   | S                |
| HSQLDB        | I          | I      | I                |
| MARIADB/MYSQL | C          | C      | C                |
| ORACLE        | C[1]       | C      | I                |
| POSTGRESQL    | S          | S      | I                |
| SAP HANA      | C[1]       | C[1]   | C                |
| SQL SERVER    | C          | C      | C                |

content/refguide/case-sensitive-database-behavior.md
nimatrueway: This section needs update @passalaqua, we explicitly set collation of string attributes to case-insensitive (See `HsqldbMapper:51`), ergo everything about Hsqldb becomes case-insensitive regardless of global settings.

content/refguide/case-sensitive-database-behavior.md
nimatrueway: Collation directly affects string-functions @passalaqua.

content/refguide/case-sensitive-database-behavior.md
nimatrueway: Collation directly affects string-functions @passalaqua

content/refguide/case-sensitive-database-behavior.md
nimatrueway: The fact that the default collation depends on the locale of OS makes it complicated, that's why I suggested to leave it as "C". We can explain further in their vendor-specific section though, which now I'm starting to think it should be a good idea.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,80 @@
...
+tags: ["studio pro"]
+menu_order: 20
+---
+
+## 1 Introduction
+
+Case sensitive string operations are those that consider upper and lower case as different letters.
+Operations that affect strings in queries (for instance, inside an XPath constraint) can be case sensitive or case insensitive depending on the database vendor, version, and configuration used by the Mendix application.
+
+For the purposes of this document we can divide these operations three categories:
```
nimatrueway, line 13: I think somewhere in here we should indicate that as a general rule, sort/comparison is case-sensitive unless database collation sets it otherwise; and string-functions are case-insensitive unless its not implemented by Mendix yet due to complexity.

content/refguide/case-sensitive-database-behavior.md
nimatrueway: > So what does the collation do with the string functions?

Collation directly affects string-functions @passalaqua

MarkvanMents:
> 
> 
> To the maintainers: I believe @nimatrueway will still review part of these changes. Once they are done, we will inform you that this is content accurate.

Thanks for letting us know. I'll keep an eye on this. If I don't seem to be responding once it is ready for publication, Slack me!

content/refguide/case-sensitive-database-behavior.md
passalaqua: Do you also think it's better to flip the db/operation column/rows? I think it's nicer and easier to read, but would be different from the page about null values. (I can also just change there.)

nimatrueway: I hadn't notice that I rotated the table up until now. :smile: Glad that you liked it.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,90 @@
...
+
+### 2.4 MARIADB & MYSQL
+
+All operations depend on collation.
+The default collation is `utf8_general_ci` when the `utf8` character set is used or `latin1_swedish_ci` when the `latin1` character set is used.
+These are both case insensitive.
+
+### 2.5 ORACLE
+
+All operations depend on collation.
```
nimatrueway, line 51: String-functions are always performed **case-insensitive** by us actually.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,90 @@
...
+The default collation is `binary`, which is case sensitive, but a `binary_ci` is available for case insensitive operations.
+
+It is possible to set different collations for sorting and comparison operations by setting different values to the `NLS_SORT` and `NLS_COMP` "Linguistic Sort Parameters".
+
+Does not support comparison on attributes of strings of unlimited length.
+
+### 2.6 SAP HANA
+
+All operations depend on collation.
```
nimatrueway, line 60: Please reflect the discussion we had in slack about SapHana.

https://mendix.slack.com/archives/CDW3BCPH7/p1589963535006700?thread_ts=1589899829.008200&cid=CDW3BCPH7

passalaqua: Done and pushed a new version with those changes.

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,97 @@
...
+
+Sorting and comparison depend on collation.
+The default collation is `binary`, which is case sensitive, but a `binary_ci` is available for case insensitive operations.
+
+It is possible to set different collations for sorting and comparison operations by setting different values to the `NLS_SORT` and `NLS_COMP` "Linguistic Sort Parameters".
+
+Does not support comparison on attributes of strings of unlimited length.
+
+String functions are implemented by converting all letters to uppercase and is, therefore, sensitive to how the database is configured to convert letters.
+It is possible to set a different behavior for converting letters to uppercase by setting different values to the `NLS_UPPER` parameter.
```
nimatrueway, line 63: `NLS_UPPER` is a function rather than an environment variable to be set and we don't use it in Mendix. We use `Upper` instead.

nimatrueway:
Approved from data-storage side. It's in your hands now @MarkvanMents.

MarkvanMents:
I'll look at this tomorrow.
Thanks for working on improving the documentation!

content/developerportal/deploy/setting-up-a-new-sql-server-database.md
```diff
@@ -22,7 +22,7 @@ When setting up a new database for Mendix, most of the settings can be left to t
 
 In the database options, the default properties need to be evaluated. When choosing a collation, pay attention to the type of collation you are going to use. Mendix uses UTF-8 for all data evaluation. Depending on your exact locale, you will most likely want to choose one of the `SQL_Latin1_General_` collations. The exact encoding will depend on your OS. For an *en_US* installation, this will be `CP1`.
 
-The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations, see [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
+The last two options identify how sorting and uniqueness is interpreted. For example, the collation option `_CS_` indicates that the collation sorting style will be case-sensitive. For more information on collations and case sensitivity, see  [Case-Sensitive Database Behavior](case-sensitive-database-behavior) and the Microsoft documentation [Windows Collation Sorting Style](https://msdn.microsoft.com/en-us/library/ms143515.aspx).
```
MarkvanMents, line 25: Is there a better Microsoft document to link to? This one says it has been retired and won't be updated, but I can't find a newer version.

basdebakker: This looks like a good replacement: https://docs.microsoft.com/en-us/sql/t-sql/statements/windows-collation-name-transact-sql?view=sql-server-ver15

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,97 @@
...
+Operations that affect strings in queries (for instance, inside an XPath constraint) can be case sensitive or case insensitive depending on the database vendor, version, and configuration used by the Mendix application.
+
+It is also important to note that each implementation of a case-insensitive operation may treat the case normalization of letters differently.
+As a general rule, normalization is based on locale. This is either specified explicitly in the database/table, or the database infers it from the Operating System.
+This normalization of letters, or case folding, is further described in [Case Folding](https://www.w3.org/International/wiki/Case_folding) on the *World Wide Web Consortium* wiki.
+
+For the purposes of this document we can divide case sensitive operations into three categories:
+
+* Sorting: Indicates the order in which you want objects to be retrieved (alphabetically ascending or descending).
+* Comparing: These are operations that involve equality directly in queries (for instance, in the query `\\Entity[Attribute = 'a']`)
```
MarkvanMents, line 20: Is this just equality, or would it also have an impact on > and < (or can you not do these on strings?)

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -0,0 +1,97 @@
...
+### 2.5 ORACLE
+
+Sorting and comparison depend on the configured collation.
+The default collation is `binary`, which is **case sensitive**, but a `binary_ci` is available for case insensitive operations.
+
+It is possible to set different collations for sorting and comparison operations by setting different values to the `NLS_SORT` and `NLS_COMP` "Linguistic Sort Parameters".
+
+Does not support comparison on string attributes of unlimited length.
+
+String functions are implemented by converting all letters to uppercase using the database's `UPPER` function and is, therefore, insensitive to the `locale` in which it is executed.
```
MarkvanMents, line 63: Can we say explicitly that these are **case insensitive** so we use the same phrase as elsewhere, or is this not the same thing?

content/refguide/case-sensitive-database-behavior.md
basdebakker: Indeed, this is also > and < (and <=, >=, !=).

content/developerportal/deploy/setting-up-a-new-sql-server-database.md
MarkvanMents: Thanks - updated in all documents.

content/refguide/case-sensitive-database-behavior.md
basdebakker: This is somewhat complex. Indeed it makes the process case insensitive, and it would be good to mention this explicitly. But the exact meaning of case insensitive could still depend on the locale, and this phrase is saying that here it doesn't.

MarkvanMents: How about
```suggestion
String functions are implemented by converting all letters to uppercase using the database's `UPPER` function and is, therefore, **case insensitive** and insensitive to the `locale` in which it is executed.
```

content/refguide/case-sensitive-database-behavior.md
MarkvanMents: Thanks - changed to clarify the scope.

content/refguide/case-sensitive-database-behavior.md
basdebakker: Looks good to me. You may also want to change "is" to "are" and "it is" to "they are" as "functions" is plural.

MarkvanMents: Changes made

MarkvanMents:
Hi @passalaqua , @nimatrueway , and @basdebakker 
Thanks for this great piece of documentation.
I've made a few minor language and presentation changes - can you do a final review (or trust me) and then I can publish it.

MarkvanMents:
Thanks - merging now.

PR #2485, opened 2020-05-25
Added a point penalty section

jandevriesmendix:
It explains that Mendix can give point penalties to users who abuse our systems.


Adam-Dupaski:
Good to see you back in the repo @jandevriesmendix !

PR #2496, opened 2020-05-29
Adding service node

feddovanede:
The <service> node is missing in the text, even though it is shown in the image. Leaving out this part causes the OnReceive event of the Notifications widget not to be triggered when the app is in the foreground. Adding the <service> node, as it is shown in the image, resolves this issue. The <service> node was listed in an older version of this document, but probably got deleted by accident.
See also ticket #101377.


ConnorLand:
Merging this now. Mehmet is out of the office, but when he returns Thursday I will figure out why these were missing from the final draft code snippet. Thanks!

PR #2497, opened 2020-06-02
[DAT-2081] Make string-functions perform case-insensitively on SapHana

nimatrueway:
RN
Make string-functions perform case-insensitively on SapHana

### DIFF
--- a/content/refguide/case-sensitive-database-behavior.md
+++ b/content/refguide/case-sensitive-database-behavior.md
@@ -40,8 +40,7 @@ String functions are **case insensitive** as they are implemented using the `ILI
 
 ### 2.3 DB2
 
-**Case sensitive** for sorting, comparing, and in string functions.
-They cannot be configured.
+All operations are **case sensitive**. They cannot be configured.
 
 Does not support sorting on string attributes of unlimited length.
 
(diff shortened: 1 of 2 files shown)


ishan1729:
I approve of the changes.

MarkvanMents:
Hi @nimatrueway ,
Looks great. Three questions:
1. When will this be released, or is it the current status?
2. Does this apply to all versions of Mendix or is it version specific?
3. Does the user have to do anything to an existing SAP Hana database to get it to apply?

Thanks

nimatrueway:
Hi @MarkvanMents !

> When will this be released, or is it the current status?
> Does this apply to all versions of Mendix or is it version specific?

The change will be in effect since the next minor release, namely **8.11** (and any release after that ofcourse).

> Does the user have to do anything to an existing SAP Hana database to get it to apply?

They will notice the change right away. No extra configuration is required.

MarkvanMents:
I think that we need to say what the situation was in Mx <8.11 for users who may be using earlier versions of Mendix. I'll add something.

MarkvanMents:
@nimatrueway,
Hi. I've added a quick note to both the documents that if users have a version before 8.11 they will get case sensitive behavior.
Let me know if you have any comments, otherwise I'll assume this PR is ready to go with 8.11.

nimatrueway:
Thank you @MarkvanMents, it became clearer. :+1:

PR #2501, opened 2020-06-04
June 8: New release for Private Cloud

zlogic:
* Added Release Notes for the latest version of Operator and Agent
* Added a document how to upgrade to the latest version of Private Cloud

### DIFF
--- a/content/developerportal/deploy/private-cloud-upgrade-guide.md
+++ b/content/developerportal/deploy/private-cloud-upgrade-guide.md
@@ -6,19 +6,19 @@ menu_order: 90
 tags: ["Upgrade", "Private Cloud", "Cluster", "Operator", "Deploy"]
 ---
 
-# 1 Introduction
+## 1 Introduction
 
 This document describes how an existing installation of Mendix for Private Cloud can be upgraded to the latest version.
 
-# 2 Upgrade steps
+## 2 Upgrade steps
 
-## Upgrading to Mendix Operator v1.1.0{#operator-v1.1.0}
+### 2.1 Upgrading to Mendix Operator v1.1.0{#operator-v1.1.0}
 
 This process will take about 15 to 30 minutes.
 During the upgrade process, the Mendix Operator will have to be stopped and will not process any changes.
 Apps and Environments managed by the Operator will keep running during this process and will be restarted in the cleanup step.
 
-### Preparation
+#### 2.1.1 Preparation
 
 Open a Bash-compatible terminal and set the target namespace where the Operator is currently installed:
 
@@ -26,9 +26,9 @@ Open a Bash-compatible terminal and set the target namespace where the Operator 
 OPERATOR_NAMESPACE={namespace where the Mendix Operator is installed}
 ```
 
-All following commands require the `OPERATOR_NAMESPACE` variable to be set. When opening a new terminal, re-run this command again.
+All the following commands require the `OPERATOR_NAMESPACE` variable to be set. When opening a new terminal, re-run this command again.
 
-### Stop the Operator
+#### 2.1.2 Stop the Operator
 
 To stop the Operator, run the following command:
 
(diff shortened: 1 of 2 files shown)


content/releasenotes/developer-portal/deployment.md
```diff
@@ -10,6 +10,26 @@ These release notes cover changes to [Mendix Cloud](/developerportal/deploy/mend
...
 
+### June 8th, 2020
+
+#### Mendix for Private Cloud — Mendix Operator v1.1.0
+
+To upgrade an existing installation of Private Cloud to this version, follow the [Upgrade instructions](/developerportal/deploy/private-cloud-upgrade-guide#operator-v1.1.0).
+
+* Mendix apps now run as Kubernetes Deployments instead of StatefulSets. This will allow you to use rolling updates, reducing downtime. In addition, this helps avoid situations where a StatefulSet might become stuck and stop processing any changes.
+* We now allow you to set Kubernetes resource requirements in addition to resource limits. Apps no longer require the maximum amount of CPU and memory, improving utilization of cluster resources.
+* We improved the default liveness and readiness probe configuration.
```
MarkvanMents, line 21: Will the user understand what there are?

(I don't!)

content/developerportal/deploy/private-cloud-upgrade-guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+To start the updated version of the Mendix Operator, run:
+
+```shell
+kubectl -n $OPERATOR_NAMESPACE scale deployment mendix-operator --replicas=1
+```
+
+#### 2.1.7 Cleanup Steps
+
+Delete StatefulSets from the Namespace where the Operator was installed:
```
MarkvanMents, line 106: Do we need a warning here to say that the app will be restarted (just in case it is production)

content/developerportal/deploy/private-cloud-upgrade-guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+```shell
+kubectl -n $OPERATOR_NAMESPACE delete --all statefulsets
+```
+
+These StatefulSets will be replaced with Deployments when the new version of the Operator is started.
```
MarkvanMents, line 112: But you have already started it in the previous step? Or have I misunderstood something?

content/developerportal/deploy/private-cloud-upgrade-guide.md
zlogic: I've added a warning documenting what will happen here and why the manual cleanup is required.

Basically, when the new version of the Operator is started, it will start a new copy of the app (as a new Deployment) and route all traffic to this new Deployment. But the old version of the app will keep running in the StatefulSet, which the user has to remove manually.

content/releasenotes/developer-portal/deployment.md
zlogic: I've added an explanation that this is Kubernetes objects, and also what how this will benefit end users.

content/developerportal/deploy/private-cloud-upgrade-guide.md
zlogic: I've added a warning note that this cleanup step has to be done manually (the new version of the Operator will ignore StatefulSets completely).

content/developerportal/deploy/private-cloud-upgrade-guide.md
MarkvanMents: @zlogic Looks good.
Does this mean that there will be no downtime, or will users still lose their sessions when they are rerouted to the new deployment?
(This is challenging my knowledge of how K8s works!)

MarkvanMents:
@zlogic - thanks for the changes.
I have just one outstanding question about downtime, which might clarify the exact process to the user. But if there is no clear answer, then we are ready to publish, I think.

content/developerportal/deploy/private-cloud-upgrade-guide.md
zlogic: They will lose their sessions, and in addition app will be unavailable until the new deployment is fully started.

MarkvanMents: OK - I think we can leave this text as it is. This is the way that users would expect Mendix to behave if you start a new environment.

Ready to go - let me know when you release the software and I can publish the documentation.

PR #2507, opened 2020-06-08
Added extra info to associations in microflows

AndreasBlaesius:
I added some extra information to the associations in microflows section. This info is extremely important for developers. A developer needs to know if this is using the DB version or the cached version (including uncommitted changes).
It is also important to know the retrieve behaviour, because this can affect the performance of the application.


Adam-Dupaski:
Verified

PR #2511, opened 2020-06-08
Documented current Private Cloud limitations

zlogic:
To reduce support questions, I've tried to summarize what are the current limitations of Mendix for Private Cloud, and also prerequisites which are confirmed to be supported.

### DIFF
--- a/content/developerportal/deploy/private-cloud-supported-environments.md
+++ b/content/developerportal/deploy/private-cloud-supported-environments.md
@@ -223,7 +223,7 @@ To use S3, the Mendix Operator will need an IAM account with the following polic
 ## 6 Networking
 
 {{% alert type="info" %}}
-DNS, load balancing and the ingress controller should be configured separately, on a cluster-wide scope.
+DNS, load balancing and the ingress controller should be configured first for the whole Kubernetes cluster.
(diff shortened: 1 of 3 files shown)


content/developerportal/deploy/private-cloud-supported-environments.md
```diff
@@ -0,0 +1,258 @@
...
+            ]
+        }
+    ]
+}
+```
+
+## 6 Networking
+
+{{% alert type="info" %}}
+DNS, load balancing and the ingress controller should be configured separately, on a cluster-wide scope.
```
MarkvanMents, line 226: @zlogic Do you mean that you have to configure them outside the Mendix scope, or do you mean that you need to configure each of the three separately? If the former then perhaps:
```suggestion
DNS, load balancing and the ingress controller should be configured first for the whole Kubernetes cluster.
```

zlogic: @MarkvanMents Your understanding is correct, it's outside of the Mendix scope and is the customer's responsibility.

MarkvanMents:
@zlogic Can this be published now, or does someone else need to approve?

zlogic:
@MarkvanMents Yes, can you please publish this?

PR #2515, opened 2020-06-10
Change to make clear which modeler the SDK uses

MxTheo:
Change to make clear which modeler the SDK uses. It took me a while to figure out how to use Mendix Studio Pro for SDK, instead of the Modeler 7.23. Hopefully, with this change, it will be a little bit clearer what you need to change to be able to use other versions of the Studio Pro.


Adam-Dupaski:
Hi @MxTheo, thanks for the PR and suggestion. I contacted a dev about this and they made an improvement here for automation - I merged this via PR #2522, please see that for further info. I'll close this PR, as it's no longer necessary.
Cheers,
Adam

PR #2536, opened 2020-06-22
 as Administrator.

TimvanSteenbergen:
" as Administrator." Minor addition, but likely to save some people some time.


ConnorLand:
@TimvanSteenbergen I've consulted with our developers and they believe your permissions issues stem from something else. When you encountered issues, it was most likely because of your setup. Please see this documentation for ways to optimize your setup and solve permission issues: https://docs.npmjs.com/resolving-eacces-permissions-errors-when-installing-packages-globally

I will look into including this link in the docs so that users in your situation know where to turn. Does that make sense?

ConnorLand:
Closing this, see comment above for more information. Our research indicates that if npm is set up like the link above states, users will not have permission issues and that the linked method is best practice.

PR #2537, opened 2020-06-22
Postgres to Hana database migration steps

MarkvanMents:
Add information specifically about migrating to SAP HANA


MarkvanMents:
Approved by Guarav by email.

PR #2549, opened 2020-07-01
Document change of User Role for Admin(MxAdmin) account

MarkvanMents:
Changes of role are only applied when the admin account password is changed.


MarkvanMents:
Approved over Slack.

PR #2558, opened 2020-07-03
Update committing-objects.md

emirmx:
Fix statement that says events are also triggered for auto committed objects for which they are not

### DIFF
--- a/content/refguide/committing-objects.md
+++ b/content/refguide/committing-objects.md
@@ -106,9 +106,10 @@ If you end up with autocommitted objects, it is always because of a modeling err
 
 During commit the the following will occur:
 
-* Events: For explicitly committed objects all before and after events are executed, and if any before-rollback event returns false, an exception can be thrown
+* Events: For *explicitly committed* objects all before and after events are executed, and if any before-rollback event returns false, an exception can be thrown
 	* If an exception occurs during an event, all the applied changes are reverted with the default error handling behavior
 	* Changes made prior to the commit will be kept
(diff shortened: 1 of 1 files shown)


content/refguide/committing-objects.md
```diff
@@ -104,12 +104,12 @@ If a rollback is triggered for any reason (for example, if the user session is t
 
 If you end up with autocommitted objects, it is always because of a modeling error. At some point in time, an association was set to a new object, the associated object was committed, and all of its associations were committed as well to keep all the data consistent.
 
-For both explicitly committed and autocommitted objects, the following will occur:
+During commit the the following will occur:
 
-* Events: all before and after events are executed, and if any before-rollback event returns false, an exception can be thrown
+* Events: For explicitly committed objects all before and after events are executed, and if any before-rollback event returns false, an exception can be thrown
```
MarkvanMents, line 109: Hi @emirmx 
Thanks for making the PR.
Do you think we should explicitly say that Autocommit will **not** trigger these events? Might avoid the problems we were talking about on Slack today.

emirmx: That may be more clear i.e. we can say `For explicitly committed objects (i.e. not auto committed)` or some other phrase.

MarkvanMents:
@emirmx - hi
I've made it very explicit that autocommits don't trigger before and after events.
If you're happy with that, I'll merge!

emirmx:
> @emirmx - hi
> I've made it very explicit that autocommits don't trigger before and after events.
> If you're happy with that, I'll merge!

I think that's good. Thanks.

PR #2564, opened 2020-07-07
DEP-2632 - Display Mendix SSO constants in Developer Portal

MarkvanMents:


MarkvanMents:
Asked 6/7
Written 7/7
Informed 9:45 7/7 that this is not now being released in current release.
Added flag for Future Release.

MarkvanMents:
After discussion with the Identity Services team this change was not released to production.

PR #2565, opened 2020-07-07
Just correcting a reverted buttoncolor description

TimvanSteenbergen:
Just correcting a reverted buttoncolor description


ConnorLand:
Thanks @TimvanSteenbergen !

PR #2575, opened 2020-07-09
Clarify that regular expression resources can only be used in validation rules

MarkvanMents:
There was confusion over the purpose of the regular expression documentation (https://docs.mendix.com/refguide/regular-expressions)

It was used to:
1. Describe a regular expression resource which can only be used in entity validation rules
2. Explain the structure of a regular expression used in both regular expression resources and in the regular expression strings used as parameters to microflow and nanoflow expressions.

Reference: JIRA issue https://mendix.atlassian.net/browse/TW-1053

### DIFF
--- a/content/refguide/expressions.md
+++ b/content/refguide/expressions.md
@@ -31,7 +31,7 @@ An overview of the operators which can be used in expressions is shown below.
 
 ### 1.2 Regular Expressions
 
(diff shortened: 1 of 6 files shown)


MarkvanMents:
No review comments received - will publish as is.

PR #2579, opened 2020-07-13
Release Mendix for Private Cloud Operator 1.3.0

zlogic:
Anticipated release date: July 15

* Updated the Release Notes
* Added documentation how to upgrade to the latest Mendix for Private Cloud version
* Updated documentation and screenshots to include new configuration options


content/developerportal/deploy/private-cloud-upgrade-guide.md
```diff
@@ -68,21 +70,29 @@ kubectl apply -f https://installergen.private-cloud.api.mendix.com/privatecloud/
...
+{{% /alert %}}
+
 Run the following commands to switch to the latest component versions:
 
 ```shell
 kubectl -n $OPERATOR_NAMESPACE patch operatorconfiguration mendix-operator-configuration --type merge -p \
 '{"spec":{
-    "sidecarImage":"quay.io/digital_ecosystems/mx-m2ee-sidecar:1.1.0",
+    "sidecarImage":"quay.io/digital_ecosystems/mx-m2ee-sidecar:1.2.0",
+    "metricsSidecarImage":"quay.io/digital_ecosystems/mx-m2ee-metrics:1.1.0",
```
MarkvanMents, line 95: Is there a reason for having the version of mx-m2ee-metrics different from the version of the mx-m2ee-sidecar.

It might help dependency management to try to keep the versions in step, except for patches, even if the only change is the version number.

This way leaves it open to mistyping and confusion for the user. (They might think that it is a documentation error, for example).

zlogic: We're planning to discuss our versioning strategy this week - the version numbers are becoming difficult to manage.

Originally the plan was to not have any version alignment - so that we can rapidly release a bugfix for just one component.

MarkvanMents: Up to you, of course, but I would go for having all components on the same minor release level. You can still use patch release for bugfixes provided you make sure that all the 1.3.* versions work together.
You can always bump a version without making any changes, but at least it helps users (and you) to know that all the versions will work together. I would imagine this might help support too - if an issue occurs you can immediately spot that there might be a problem running 1.3.0 Operator with 1.2.1 Agent.

MarkvanMents:
Asked to publish over Slack.

PR #2583, opened 2020-07-14
Update xpath-weekday-from-datetime.md

Andy-Fleming:
Noticed that the dates in the example are out by one day. I've tested this will all weekdays...

### DIFF
--- a/content/refguide/xpath-weekday-from-datetime.md
+++ b/content/refguide/xpath-weekday-from-datetime.md
@@ -6,11 +6,11 @@ tags: ["studio pro"]
 
 ## 1 Overview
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Thanks for the Pull Request,
Have asked development team to confirm this, and will then publish it.
Request for confirmation made in #datastorage_totw Slack channel today.

MarkvanMents:
Hi @Andy-Fleming,
The developers here have tried to reproduce this case but have found that the day of the week is being returned as described in the original documentation (i.e. Wednesday (today) is day 4).

They suggest that there might be something else going on and ask that you raise a support ticket from the **Submit a request** button on https://support.mendix.com so that it can be investigated further.

As the results we get are different, I will close this pull request. If there is an issue, Mendix support will contact me with further documentation updates.

Thank you for taking the trouble to suggest an update to our documentation. We appreciate your desire to make Mendix documentation better.

Yours

Mark van Ments.

PR #2585, opened 2020-07-15
WIP: Text box web & native on input event property

IIsaku:
The core text box widget will include an on input event property starting from MX 8.13.


IIsaku:
Please label this PR as work-in-progress. It's not certain whether we will release this for MX 8.13

ConnorLand:
@IIsaku is this going into 8.13, or is it still being debated?

PR #2588, opened 2020-07-15
Update community-commons-function-library.md

robertprice:
RegexReplaceAll is no longer deprecated in Community Commons, so I have removed this flag from the documentation.

This function was deprecated for a short while, but I flagged up the maintainer that it provided difference functionality so the deprecation was rolled back.


MarkvanMents:
Confirmed with developer.

PR #2593, opened 2020-07-17
Host feedback locally

iannbing:
The feedback functionality is in the process of deprecation. For the time being, the feedback functionality on docs.mendix.com will be hosted by docs.mendix.com itself.

This PR includes the necessary files for running feedback on docs.mendix.com without downloading resources from sprintr.home.mendix.com. And we also add an alert on [Feedback page](https://docs.mendix.com/appstore/widgets/mendix-feedback) to inform users.


MariaShaposhnikova:
Hi @JelteMX ! Will you have time to have a look at this PR?

PR #2599, opened 2020-07-22
Update 8.11.md

jds-mendix:
Add release notes regarding the same site policy of cookies.


MariaShaposhnikova:
Hi @jds-mendix ! Your release notes have a reference for a future section in docs. Will you add this section as a separate PR before the release or shall I delete the reference for now?

PR #2603, opened 2020-07-23
Added additional info about the delete behavior

AndreasBlaesius:
It was not clear if delete behavior is working on database level (only with committed objects) or on runtime level (also with uncommitted objects). This information can sometimes be very important, so I added it.


MarkvanMents:
Thanks for your submission @AndreasBlaesius - and for making Mendix documentation better.

PR #2604, opened 2020-07-23
Add 7.23.15 release notes 

MariaShaposhnikova:


content/refguide7/system-requirements.md
```diff
@@ -61,12 +61,12 @@ The browser you use needs to have JavaScript turned on.
 ### 5.3 Database Server
 
 * [IBM DB2](db2) 11.1 for Linux, Unix, and Windows
-* [MariaDB](mysql) 5.5, 10.0, 10.1, 10.2, 10.3
-* [Microsoft SQL Server](/developerportal/deploy/mendix-on-windows-microsoft-sql-server) 2008, 2008 R2, 2012, 2014, 2016, 2017, 2019
+* [MariaDB](mysql) 10.1, 10.2, 10.3
```
Adam-Dupaski, line 64: @dahfjkg Updated this doc also for the deprecations.

dahfjkg: Great! I've been doing the same in a separate MR. I see that in refguide for Mendix 8 we still mention MariaDB 5.5. I think that it should be removed as well: MariaDB 5.5 is no longer supported by the vendor. Checking with the team.

dahfjkg: We will change the documentation for Mendix 8 separately, so this change should be enough for Mendix 7

PR #2624, opened 2020-07-28
Use new constants for configuring audit trail module

passalaqua:
As part of https://github.com/mendix/AuditTrailModule/pull/19 we are moving from an NPE to microflow constants to configure the module.


passalaqua:
Unfortunately I do not know when this will be released. Perhaps I should have made it clear that previous versions were using the old method while newer versions use the new method? How should I proceed, dear doc colleagues?

Adam-Dupaski:
@passalaqua could we write something like this:

"Changing the audit behavior is easily done by altering the values of the following microflow constants (for module version ___ and above; please note that if no value is set, the default is used) or the default values of the **Configuration** entity in the module's domain model (for module version ___ and below):"

passalaqua:
> @passalaqua could we write something like this:
> 
> "Changing the audit behavior is easily done by altering the values of the following microflow constants (for module version ___ and above; please note that if no value is set, the default is used) or the default values of the **Configuration** entity in the module's domain model (for module version ___ and below):"

I think that should work, although I am thinking if we should emphasize that it is a non-persistable entity. Or do you think it's clear enough as is?

The dividing version should be 7.5 (I'd keep patch version out, since any patches for 7.5 might not have that yet, although I don't remember seeing any patch releases there). Up to and including 7.5 should have the NPE. Anything above that should have the new mechanism, provided we merge the PR I mentioned above.

Adam-Dupaski:
@passalaqua I added the text above with the versions, please verify. I'm not sure how important it'd be to state "the non-persistable **Configuration** entity" - I don't think we usually do that, but that will have to be your call for this context. I can merge whenever you let me know about the release.

passalaqua:
@Adam-Dupaski Honestly, I am not sure about the clarity part. I'm not one to judge that clearly (hah!). I'm ok if you leave that out for now, anyways. I'm just pushing a small quick fix to `LogLineDateFormat` and I'll let you know when the release is coming (once we know, might take a little while, btw, as there are a few other low hanging fruits that should be nice to release together).

passalaqua:
Version 7.6.0 has been released! (A month ago, but I forgot to check this page for updates. D:)

PR #2633, opened 2020-08-04
Editable "Inherited from snippet call" (8.13)

tamelbrun:
The option "Inherited from snippet call" was added to the Editable property in 8.13 to clarify the behaviour inside snippets, similar to earlier improvements made to the Read-only style property. (WTF-511)


ConnorLand:
This PR updates documentation in accordance with the new editable prop for widgets called from snippets https://mendix.atlassian.net/browse/TW-1097

No screenshots in documentation need to be updated, since all screenshots which show widget props show the "default" value next to all widgets' editable props. This default state remains unchanged.

PR #2648, opened 2020-08-10
Improve documentation on client certificates

MarkvanMents:
Try to clarify the meanings of various combinations of certificates.


mbyregowMx:
**client certificate will only be accepted if it is signed by both certificates.**

is not correct, client certificate signed by any of the checked intermediate certificate is valid and accepted.

MarkvanMents:
Hi @mbyregowMx 

in https://mendix.atlassian.net/browse/DEP-3030 there is a comment that says:

> 2.	What is the difference in functionality between test 1 and test 2?
> Subca1 is only signed by interne.root.ca. If subca1 has a check mark, what is the added value of also putting a check mark at interne.root.ca?
> In first case **client certificate signed by both root and the intermediate is allowed access**
> In second case client certificate signed by only the intermediate is allowed.

So this comment means _either_ root and the intermediate - not _both_ root and intermediate?

mbyregowMx:
Ah, sorry for the confusing statement, I meant certificate signed by either Root or the intermediate will be valid if both are checked in the ARP.

MarkvanMents:
@mbyregowMx Have made the change. Is this OK now?

MarkvanMents:
Approved by Deployment and Support.

PR #2700, opened 2020-09-03
[DAT-1203] Update string case sensitivity of DB2

nimatrueway:
Since next minor version release of Mendix 8.x, we made xpath string functions case insensitive for db2 like other databases.

### DIFF
--- a/content/refguide/case-sensitive-database-behavior.md
+++ b/content/refguide/case-sensitive-database-behavior.md
@@ -40,7 +40,7 @@ String functions are **case insensitive** as they are implemented using the `ILI
 
 ### 2.3 DB2
 
-As of version [8.14.0](/releasenotes/studio-pro/8.14), sorting and comparing are **case sensitive**, except if collation is configured to **case insensitive**. For versions below 8.14, all operations are **case sensitive** and they cannot be configured. For more information, see the [Making DB2 Case-Insensitive](db2#making) section of *DB2*.
+As of version [8.14.0](/releasenotes/studio-pro/8.14), sorting and comparing are **case sensitive**. For versions below 8.14, all operations are **case sensitive**. There is an exception for both version groups where collation is configured to **case insensitive**, which affects all operations. For more information, see the [Making DB2 Case-Insensitive](db2#making) section of *DB2*.
 
 Does not support sorting on string attributes of unlimited length.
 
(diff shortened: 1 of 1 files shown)


snajahi:
This change has been approved by data-storage team.

Adam-Dupaski:
@nimatrueway can this be published immediately, or for which version release?

nimatrueway:
@Adam-Dupaski unfortunately we do not have a versioned documentation, given that what do you suggest we do? The change is visible since 8.14 and all older versions maintain the previous behaviour.

Adam-Dupaski:
@nimatrueway 8.14 is not out yet, so this will be published upon release. We should make it clear that this applies to 8.14 and above in text. I will tag you with my suggestion.

Adam-Dupaski:
@nimatrueway please verify: https://github.com/mendix/docs/pull/2700/commits/159e080b6d5921437619d90f70fe9099481f1737

content/refguide/case-sensitive-database-behavior.md
```diff
@@ -40,7 +40,7 @@ String functions are **case insensitive** as they are implemented using the `ILI
 
 ### 2.3 DB2
 
-All operations are **case sensitive**. They cannot be configured.
+As of version [8.14.0](/releasenotes/studio-pro/8.14), sorting and comparing are **case sensitive**, except if collation is configured to **case insensitive**. For versions below 8.14, all operations are **case sensitive** and they cannot be configured. For more information, see the [Making DB2 Case-Insensitive](db2#making) section of *DB2*.
```
nimatrueway, line 43: Hey @Adam-Dupaski

>  and they cannot be configured

this was actually wrong since the beginning. so the statement

> except if collation is configured to **case insensitive**.

applies to both before 8.14.0 and after 8.14.0

Adam-Dupaski: @nimatrueway so remove `except if collation is configured to case insensitive.` ?

Adam-Dupaski: @nimatrueway please verify - remove `except if collation is configured to case insensitive.` ?

nimatrueway: Hi @Adam-Dupaski, sorry for late response. I meant something like this:

> As of version [8.14.0](/releasenotes/studio-pro/8.14), sorting and comparing are **case sensitive**. For versions below 8.14, all operations are **case sensitive**. There is an exception for either version group where collation is configured to **case insensitive** that affects all operations. For more information, see the [Making DB2 Case-Insensitive](db2#making) section of *DB2*.

Adam-Dupaski (approved):

PR #2702, opened 2020-09-03
Web Socket Endpoints

MarkvanMents:
A description and example of how to use websocket endpoints in Mendix 8.13

### DIFF
--- a/content/refguide/websockets-in-runtime.md
+++ b/content/refguide/websockets-in-runtime.md
@@ -3,7 +3,7 @@ title: "WebSockets"
 category: "Mendix Runtime"
 description: "A description of how to use websockets in the Mendix Runtime"
 # menu_order: 99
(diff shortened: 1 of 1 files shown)


MarkvanMents:
@jds-mendix Here is the PR for the Web Socket endpoints.
Do we need to add a bit more information about where to put this code (e.g. explicitly say somewhere 'create a Java Action') and how to call it? Or is this information clear enough for anyone who is going to be using it?

jds-mendix:
@MarkvanMents I think that is clear. If it isn't, using web sockets would be a really bad idea. This is a 'low level' / 'advanced' feature.

Btw, in the very last line of the document I believe it should be 'If the client sends ...' not 'If the client send...'.

MarkvanMents:
@jds-mendix Thanks for the response and the proofreading.
See attached image for where it will appear in the documentation.
I've made a couple of minor changes - including correcting the error you spotted. Are you happy for me to publish?
image

PR #2709, opened 2020-09-09
Improve documentation on Mendix Studios target

MarkvanMents:
Updated to include information about Studio Pro.
See TW-1108.

### DIFF
--- a/content/developerportal/deploy/studio-deployment-settings.md
+++ b/content/developerportal/deploy/studio-deployment-settings.md
@@ -6,28 +6,25 @@ description: "Describes how to set deployment targets for Mendix Studio and Stud
 tags: ["node","developer portal","deploy", "Studio", "settings", "target", "Mendix Studios Target"]
 ---
 
-We have modified the [Deploy to Licensed Cloud Node](/refguide/project-menu#deploy) flow in Mendix Studio Pro. When **Deploy to Licensed Cloud Node** is selected in Mendix Studio Pro, your application will automatically be deployed to the *Mendix Studios Target* environment of your licensed node. This deployment will include a restart of that environment. It is now only possible to deploy Mendix Cloud v4 applications which have a [Mendix Studios Target](/developerportal/deploy/studio-deployment-settings#target) set. Please note that you will get an error message if you deploy from Studio Pro when no **Mendix Studios Target** is set.
-
 ## 1 Introduction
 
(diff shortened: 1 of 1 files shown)


NielsTrots:
@MarkvanMents thanks for this. Can you also include the results of enabling this action? 
And in the release notes, please explain our end users what the idea is behind certain changes in the platform. For Developers it can make sense in their reality but for end users totally not. Especially when we change something that worked fine before and was accepted by the community as standard behavior. People need to understand why we do things.

MarkvanMents:
Hi Neils,
I'm not sure what you mean by "the results of enabling this action".

I explain at the end:
> You have selected the target environment and can now deploy your app to that environment from either Studio or Studio Pro.

MarkvanMents:
I also updated the release note this morning from a request from Feddo (https://github.com/mendix/docs/pull/2703)

Does this need any extra information or has Feddos text resolved your issue? The release note now explains that by selecting a target this is where your app will be deployed.

feddovanede:
I think that the release notes now more clearly explain what the result of the change in the flow is: clicking the Deploy to Licensed Node now creates a deployment package _and_ automatically deploys it to the Target environment.

RCatersels:
Now only the option to turn off automatic deployment or even disable this option all together. I still think that this "feature" has not been thought through. We want deployments under control and not by mistake because some new developer just pressed the wrong button in the modeler.

MarkvanMents:
@RCatersels 
Hi Ronald,
Thanks for reaching out about this feature.
Unfortunately, as a technical writer I can only document the features which are released and explain them to customers.
I have forwarded your comments to the teams involved and will let you know if I get any feedback from them.
I realize that you would like to communicate the issues which this change is causing you. I think that the best channels would be via Mendix support and the Mendix forum.

MarkvanMents:
@RCatersels 
Hi Ronald,
I see you have also discussed this in the forum: https://forum.mendix.com/link/questions/102390. I believe the development teams are working on improving the experience and preventing the issues you mention.
I think the best I can do here is clarify that the target doesn't have to be set and explain how to get your MDA into the Developer Portal.

MarkvanMents:
@NielsTrots 
This document talks about using the **Run** command in Studio Pro.
Do you know whether the same applies to the **Deploy to Licenced Cloud Node** option?
image

MarkvanMents:
@NielsTrots 
Hi Niels, is this now OK to publish?

NielsTrots:
@MarkvanMents looks good, go for it. Thanks!

PR #2712, opened 2020-09-11
Deprecate HTTPS SSL/TLS weak ciphers in Cloud v4

pommi:


MarkvanMents (approved):
Hi @pommi 
I've changed the structure a bit to keep the day of the announcement as today, otherwise new release notes will get lost underneath if we say that it is an announcement for December. I think it is clear.

Can this be published now, or do I need to wait.

Thanks, Mark.

pommi:
Go ahead and publish it. :+1:

PR #2725, opened 2020-09-21
Remove alert for not-supported feature.

fazzdev:
Found from a forum post (https://forum.mendixcloud.com/link/questions/102481). Actually, we do support nested-object or nested-array when exporting Json.


Adam-Dupaski (approved):

fazzdev:
> 
> 
> Thanks @fazzdev . Maybe would be good as the SME here if you posted on the Forum this is not supported and doc has been updated?

Good idea! and done.

PR #2733, opened 2020-09-22
Update best-practices-javascript-actions.md

JelleDekker:
Updated part about adding external libraries based on answer provided to my question by Andries here https://forum.mendixcloud.com/link/questions/102877


Andries-Smit:
There are multiple ways of importing, relative, absolute and module. https://webpack.js.org/concepts/module-resolution/
R&D have to share with the current best way is..

Side notes:
* Have having the entire `node_modules` folder in the action folder is not always a good idea. (svn does not like many files)
* Double check what happens. as on some level node_modules folder ignored when committed to svn

Adam-Dupaski (approved):

PR #2741, opened 2020-09-23
Document Response Headers added by Mendix

MarkvanMents:


feddovanede:
Should we also list the value of these headers?

MarkvanMents:
> 
> 
> Should we also list the value of these headers?

Is the value always the same?
In the buildpack (https://github.com/mendix/cf-mendix-buildpack/blob/master/buildpack/nginx.py#L201) it does `add_header Cache-Control "no-cache";`
But when we looked at the HTTP response in a browser it was `cache-control: no-store`

```
HTTP/2 200 OK
server: nginx
date: Fri, 11 Sep 2020 09:43:29 GMT
content-type: application/json;charset=utf-8
content-length: 1048
cache-control: no-store
content-encoding: gzip
x-frame-options: sameorigin
x-vcap-request-id: 680bbaa1-802a-4caa-4684-4debb3c55ef3
strict-transport-security: max-age=31536000
```

@pommi - you gave me this information. Can you answer these questions?

1. Is it worth having these HTTP response headers in the documentation?
2. Can you explain why Feddo sees a different value from the one which is set in the buildpack?
3. If we document the headers, should we also note the default values?

Thanks

pommi:
> @pommi - you gave me this information. Can you answer these questions?
>     1. Is it worth having these HTTP response headers in the documentation?

Not all of them I guess. Some might be worth mentioning, because it's either set by the infrastructure, buildpack or Mendix Runtime.

* **cache-control** is either set by the Mendix Runtime or buildpack. The buildpack only sets this header when `$request_uri ~ ^/((index[\w-]*|login)\.html)?$` (index or login html pages)
* **x-frame-options** is technically set by the buildpack, but can be configured by the customer in the Portal.
* **x-vcap-request-id** is a UUID set by Cloud Foundry to track requests through Cloud Foundry
* **strict-transport-security** is a header set by our TLS terminating webservers

The only value that is always the same is **strict-transport-security**.

>     2. Can you explain why Feddo sees a different value from the one which is set in the buildpack?

See my above answer about index and login html pages.

>     3. If we document the headers, should we also note the default values?

I don't think thats required, maybe except for **strict-transport-security**

MarkvanMents:
@feddovanede and @pommi 

I've included the information which Pim provided. Does this look OK to release now?
We can maintain it if customers have further questions related to these headers.

Thanks

Mark

feddovanede:
Yes, I think this is good enough to release. We can indeed update if customers have questions.

PR #2743, opened 2020-09-24
Remove Documentation on using id() in XPath

MarkvanMents:
Documentation is wrong - this function does not work as described.
See JIRA issue DAT-2233 for discussion


MarkvanMents:
No review comments - publishing.

PR #2753, opened 2020-09-28
Tabbed UI for consumed web service

joostverhoog:
In Mendix 8.15, the consumed web service screen in Studio Pro is no longer a pop-up.


ilagordhan:
Merging into my branch for further editing.

PR #2765, opened 2020-10-01
DML review of consumed-odata-service.md 

paulmendix:
Some review changes.

**Important note**
Currently it contains the sentence` OData version – the OData version: Mendix supports consuming OData 3 and OData 4` and I proposed a change. This is an important change. We don't want to make customers think we have full support for OData V4 yet.


ilagordhan (commented):
Great feedback Paul!
Can't believe that that "virtual" escaped the fine toothcomb.

ilagordhan:
SDM review feedback and updates.

PR #2772, opened 2020-10-02
Deployment Release Notes for 5 October

MarkvanMents:


mbyregowMx:
Hi @MarkvanMents  in content/developerportal/deploy/studio-deployment-settings.md
the **Build Only** works for studio pro, so if **Build Only** is selected you would see no target selected error when try to publish from Studio
I think same need to be adjusted in the release notes to mention that **Build Only** works for only studio pro.

ManpreetRiar:
The release is done @MarkvanMents . Please go ahead to publish the release notes. Kindly check Madhu's comment above.

MarkvanMents:
Made updates requested and publishing.

PR #2777, opened 2020-10-07
OPC UA Connector Documentation

MarkvanMents:
Publication not expected before mid-January 2021
Preparing documentation for a new connector - OPC UA.
Need collaboration to get documentation up to standard.


content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+* Username (String) – the username used if the authentication type is `CREDENTIALS`
+* Password (String) – the password for the username used if the authentication type is `CREDENTIALS`
+* AuthenticationType (Enumeration) – the type of authentication required for this server: `NONE`, `CREDENTIALS`, or `CERTIFICATE`
+* CertificatePassword (String) – the certificate password required when using the `CERTIFICATE` type of authentication
+* CertifcatePasswordEncrypted (String) – ???????????????
+* Result (String) – the latest result from an action which makes a call to the OPC UA server
+* ValueToWrite (String) – the value to write to the OPC UA server when using the **Write** action
+* NamespaceIndex (Integer) – the index an OPC UA server uses for a namespace URI
+* Identifier (String) – the identifier for a node in the address space of an OPC UA server — the OPC UA client connector only supports identifiers with *IdentifierType* of `String`
+* EnableTestMode (Boolean) – ?????????????
```
MarkvanMents, line 91: What is the EnableTestMode

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+#### 3.2.1 **Browse** a List of Nodes
+
+The **Browse** action allows you to browse the nodes within 
+
+![Parameters for the browse action](attachments/opc-ua/browse-action.png)
+
+* Opc ua server cfg – an object of entity type OpcUaServerCfg containing the configuration of the server to which the request is made
+* Namespace index – the index an OPC UA server uses for a namespace URI
+* Identifier – the identifier for a node in the address space of an OPC UA server 
+* Is root – ?????????
```
MarkvanMents, line 108: What does setting this do?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+* Opc ua server cfg – an object of entity type OpcUaServerCfg containing the configuration of the server to which the request is made
+* Namespace index – the index an OPC UA server uses for a namespace URI
+* Identifier – the identifier for a node in the address space of an OPC UA server 
+* On message microflow – defines a microflow to be run every time a message is received from the subscribed service
+* Use return value – `Yes` returns an object of type **Subscription** which defines the new subscription and can be used in the microflow, `No` does not return an object
+* Variable name – the name assigned to the variable containing the return value
+
+{{% alert type="info" %}}
+If you redeploy your app, the subscriptions will no longer receive data from the server and will need to be set up again. ??????????
```
MarkvanMents, line 140: I think you mentioned something about losing subscriptions if you redeploy. Is this rthe right description?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+For each OPC UA server, the following information will need to be stored in a OpcUaServerCfg object.
+
+* ServerID (String) – a name you give to the server so that you can identify it easily
+* URL (String) – the URL of the OPC UA server
+* Username (String) – the username used if the authentication type is `CREDENTIALS`
+* Password (String) – the password for the username used if the authentication type is `CREDENTIALS`
+* AuthenticationType (Enumeration) – the type of authentication required for this server: `NONE`, `CREDENTIALS`, or `CERTIFICATE`
+* CertificatePassword (String) – the certificate password required when using the `CERTIFICATE` type of authentication
+* CertifcatePasswordEncrypted (String) – ???????????????
```
MarkvanMents, line 86: What is the difference between **CertificatePassword** and **CertifcatePasswordEncrypted**?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+* Use return value – `Yes` creates a string variable containing the value read from the node, `No` does not return a variable
+* Variable name – the name assigned to the variable containing the return value
+
+{{% alert type="info" %}}
+All values are read as strings, you will need to convert them if you need a numeric value.
+{{% /alert %}}
+
+#### 3.2.3 **Subscribe** to Updates of Data from a Node
+
+You can subscribe to receive a notification every time the value of a node changes. This creates an object of type **Subscription** which is associated with the OPC UA service and contains details of the subscription and the item which is being monitored. You should assign a microflow to process the data each time a notification is received. The frequency of the notifications is controlled by the UPC UA server.
```
MarkvanMents, line 128: Is the frequency controlled by the server? The OPC UA documentation implies you can pass a frequency in your subscription request.

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+Information about nodes which are subscribed to is stored in the **Subscription** entity associated with the **OpcUaServerCfg** server configuration entity.
+
+{{% image_container width="300" %}}
+![The subscription entity](attachments/opc-ua/subscription.png)
+{{% /image_container %}}
+
+An object is created for each subscription and contains the following information:
+
+SubscriptionID (String)
```
MarkvanMents, line 153: I'm not sure about these values. Perhaps if I can get the demo working I will be able to document them better?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+From this page, you can perform the following actions:
+
+* Search – search for a particular server in the list of servers
+* New server – add a new OPC UA server using the **OpcUaServer_NewEdit** page
+* Edit server - change the details of the selected server using the **OpcUaServer_NewEdit** page
+* View server – this button is there for adding further functionality — in the default module it calls a microflow which displays a message
+* Delete – delete all the information about the selected server — you will be asked for confirmation
+
+#### 3.3.2 OpcUaServer_NewEdit
```
MarkvanMents, line 201: I can document the basic options, but I can't really explain how to use the test options effectively.

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+### 4.2 Initial Configuration
+
+1. Add the **OpcUaServer_Overview** page in the **OpcUaClientMx** module to the navigation of the app, either through the **Navigation** settings, or by adding an **Open Page** button to a page which is already in the navigation (for example the home page).
+
+2. Connect the **OpcUaServer_View** page in the **OpcUaClient_ExampleImplementation** to the **View server** button on the **OpcUaServer_Overview** page in the **OpcUaClientMx** module.
+
+### 4.3 Pages
+
+The **OpcUaServer_View** page adds functionality to the **View server** button on the server overview page.
```
MarkvanMents, line 236: It would be useful to have an overview of the pages in this example module and what the user is supposed to be able to see from each one. I can check the language and formatting and perhaps get some screenshots to support it once I know what I'm looking for.

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+
+## 4 OPC UA Client Example Implementation{#example-implementation}
+
+The OPC UA Client example implementation is a sample app based on the [Prosys OPC UA server](https://www.prosysopc.com/). It implements the following functionality:
+
+* Configure one or more OPC UA server connections
+* View the nodes on the server
+* ?????? 
+* Add a subscription to a node
```
MarkvanMents, line 212: What happens when I add a subscription? It doesn't let me specify a microflow to run - what is it doing with the subscribed data and how can I see what I have received?

content/appstore/connectors/opc-ua.md
JosVerbeekMx: The Enable test mode boolean shows four buttons that enables the user to test the following functionality:
- Browse (retrieves top level nodes of the Address space of an OPC UA server).
- Subscribe 
- Write (works only with writable nodes).
- Read (works with both readable/writable nodes).
The result of a test is displayed in the Result field.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: The Root attribute is used in the Tree View widget to determine the top level object.

JosVerbeekMx: This attribute is set in the microflow OpcUaClient_ExampleImplementation.ProcessOpcUaNode. Change the conditions to set this attribute to 'true' for your own implementation.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: Subscriptions are kept client-side and kept alive as long as both client & server are running. If the requestedMaxKeepAliveCount is reached or requestedLifetimeCount is elapsed (both variables in the Java that sets up the subscription) the subscription is dropped by the server and needs to be re-initiated from the client.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: That is correct. This is currently handled by the Java code, we can make this a parameter if necessary.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: This will be replaced by a safer way to store the Certificate password.

content/appstore/connectors/opc-ua.md
MarkvanMents: I wonder if this should be in an associated entity (perhaps non-persistable). It seems to be related to the implementation pages rather than the information required to get the OPC UA feed working.
It is quite difficult for me to explain why it is in this entity and what people should set it to if they are just using the Java actions and domain model.

content/appstore/connectors/opc-ua.md
MarkvanMents: Is the tree view widget part of the the OPC UA client or just in the Example Implementation?
If it is in the latter, then it would be easier to document this attribute as part of an associated entity in that module - or is that not possible with the way App Store modules interact?

content/appstore/connectors/opc-ua.md
JosVerbeekMx: I can replace this boolean with a 'test connection' button. That should give the user information on whether the connection works or not and we won't bother him/her with attributes/fields that are difficult to understand without context.

content/appstore/connectors/opc-ua.md
MarkvanMents: Thanks for the explanation. So the client is sending keep alive requests so the server knows to keep sending the subscribed data. And if the deployment takes too long it won't send them? Or does a redeploy not initiate the keep alive process?

content/appstore/connectors/opc-ua.md
MarkvanMents: OK - I'll leave this until then. Thanks.

content/appstore/connectors/opc-ua.md
MarkvanMents: Thanks.

content/appstore/connectors/opc-ua.md
MarkvanMents: That would be great.
I'm happy to leave this until the next release and just note that it is an implementation detail not to be used by the developer if you are under pressure to publish something quickly.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: Tree view & OpcUaNode are all part of the example implementation. I don't completely understand the question about documenting the attribute as part of an associated entity.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: A Subscription contains the following attributes:
- SubscriptionID: an identifier generated by the OpcUA server. 
- Identifier: the identifier of a node in the address space of the OpcUA server the client is subscribed to.
- NamespaceIndex: the namespace index of a node in the address space of the OpcUA server the client is subscribed to.
NB. These attributes are used for node identification.
- MonitoredItemID: a server side assigned ID, in this implementation mainly used to end a subscription.
- Active: subscription active or not.
- OpcUaNodeID: identifier within the Example Implementation to connect/display a subscription to the correct OpcUa Node object.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: In the Java-action that adds a subscription an 'On message microflow' can be specified. That Java action adds a background thread which calls said microflow. It's up to the user to decide what to do with the payload, as this can differ between implementations. The Example Implementation module contains an example microflow that creates a MonitoredItem object and these objects are shown on the MonitoredItem_Overview page.

content/appstore/connectors/opc-ua.md
JosVerbeekMx: For the Subscribe, Read & Write buttons: provide the NamespaceIndex and Identifier of the node you want to access. For the latter: provide also a value to write. The Browse button just returns the top level nodes from the OpcUA server address space. 

And of course: valid credentials, be it username/password or a certificate.

content/appstore/connectors/opc-ua.md
MarkvanMents: Sorry, got confused.
What I mean is, what effect does setting _Is root_ have in the Browse action. Does it change what is returned by the Browse action - if so, how? 
The Tree widget is part of the example implementation, but the Browse action is part of the connector. I want to be able to document the connector independently of any particular implementation so that the customer will know exactly what difference it makes if they set it in their app.

content/appstore/connectors/opc-ua.md
MarkvanMents: Thanks

content/appstore/connectors/opc-ua.md
MarkvanMents: Thanks, I'll add that.

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+{{% alert type="info" %}}
+All values are read as strings, you will need to convert them if you need a numeric value.
+{{% /alert %}}
+
+#### 3.2.3 **Subscribe** to Updates of Data from a Node
+
+You can subscribe to receive a notification every time the value of a node changes. This creates an object of type **Subscription** which is associated with the OPC UA service and contains details of the subscription and the item which is being monitored. You should assign a microflow to process the data each time a notification is received. The frequency of the notifications is controlled by the UPC UA server.
+
+![Parameters for the subscribe action](attachments/opc-ua/subscribe-action.png)
```
MarkvanMents, line 130: image
image

Hi Jos - there seems to be a difference between the Subscribe Java action used in the OpcUaServerCfg_Subscribe_TEST microflow and the one you get if you drag and drop from the toolbar. The exposed parameters are in a different order.

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+
+#### 3.2.1 **Browse** a List of Nodes
+
+The **Browse** action allows you to browse the nodes within 
+
+![Parameters for the browse action](attachments/opc-ua/browse-action.png)
+
+* Opc ua server cfg – an object of entity type OpcUaServerCfg containing the configuration of the server to which the request is made
+* Namespace index – the index an OPC UA server uses for a namespace URI
+* Identifier – the identifier for a node in the address space of an OPC UA server
```
MarkvanMents, line 107: If you specify an identifier for the node, does that mean that the browse function just browses this node?
And if you leave this blank, will you get a list of node identifiers returned?

content/appstore/connectors/opc-ua.md
MarkvanMents: > OpcUaNodeID: identifier within the Example Implementation to connect/display a subscription to the correct OpcUa Node object.

Is it possible to keep the connector implementation and the example implementation separate? It makes it easier to document.
Alternatively, explain why it is needed, as we already have the node identifier and the namespace index - so doesn't that uniquely identify the node?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+The module/app is designed for usage with multiple servers if necessary. The state for each OPC UA server is kept by the client in an object of entity type **OpcUaServerCfg**. This  needs to be populated before the actions of the OPC UA client connector can be used. 
+
+{{% image_container width="300" %}}
+![OPC UA Server configuration entity](attachments/opc-ua/opcuaservercfg.png)
+{{% /image_container %}}
+
+For each OPC UA server, the following information will need to be stored in a OpcUaServerCfg object.
+
+* ServerID (String) – a name you give to the server so that you can identify it easily
+* URL (String) – the URL of the OPC UA server
```
MarkvanMents, line 81: Does the URL support HTTP(S) and TCP.
Any others?

content/appstore/connectors/opc-ua.md
MarkvanMents: Also, what is the difference between `SubscriptionID` and `MonitoredItemID`? Surely the monitored item is the `NamespaceIndex`+`Identifier`?
And why is it the `MonitoredItemID` which is used to cancel the subscription, not the `SubscriptionID`?

content/appstore/connectors/opc-ua.md
```diff
@@ -0,0 +1,236 @@
...
+#### 3.2.4 **Unsubscribe** from Updates of Data from a Node
+
+If you have subscribed to  item change notifications but no longer want to receive the notifications, you need to unsubscribe from them.
+
+![Parameters for the unsubscribe action](attachments/opc-ua/unsubscribe-action.png)
+
+* Opc ua server cfg – an object of entity type OpcUaServerCfg containing the configuration of the server to which the request is made
+* Subscription ID – the ID of the subscription from which you want to unsubscribe — this is held as the **SubscriptionID** in the **Subscription** entity
+* Monitored item ID – the ID of the item which is being monitored by the subscription — this is held as the **MonitoredItemID** in the **Subscription** entity
+* Use return value – `Yes` ????? what is the return value, `No` does not return a variable
```
MarkvanMents, line 169: What is the return value from unsubscribe? How would a customer use it?

content/appstore/connectors/opc-ua.md
MarkvanMents: Confirmed that we should be using TCP.

content/appstore/connectors/opc-ua.md
MarkvanMents: Added to documentation. Thanks.

content/appstore/connectors/opc-ua.md
MarkvanMents: Added to documentation. Thanks.

content/appstore/connectors/opc-ua.md
MarkvanMents: I can now do this as my app is working properly.
Thanks very much for your help in that.

content/appstore/connectors/opc-ua.md
MarkvanMents: Have documented after our discussion about the tree widget.

jaspervanderhoek:
Mark,
I have a few comments. How would you like me to address them? Is this good or do you want me to do the changes in git and push them to this branch?

1. The image for OPC UA Toolbox is outdated, this is the correct content: opc-ua-toolbox
2. The image for opc-ua-connector-module is outdated, this is the correct content: 
opc-ua-connector-module
3. I just realized I forgot to update the Mendix version in the docs, I bumped the Mx version to Mx 8.17.0 for Mx9 compatibility. 1.1.3 Dependencies & 4.1 Dependencies =  Mendix 8.17.0 or higher

Except for these 3 minor things all looks good and as far as I'm concerned this can be released.

jaspervanderhoek (commented):
Mark,
I have a few comments. How would you like me to address them? Is this good or do you want me to do the changes in git and push them to this branch?

1. The image for OPC UA Toolbox is outdated, this is the correct content: opc-ua-toolbox
2. The image for opc-ua-connector-module is outdated, this is the correct content: 
opc-ua-connector-module
3. I just realized I forgot to update the Mendix version in the docs, I bumped the Mx version to Mx 8.17.0 for Mx9 compatibility. 1.1.3 Dependencies & 4.1 Dependencies =  Mendix 8.17.0 or higher

Except for these 3 minor things all looks good and as far as I'm concerned this can be released.

MarkvanMents:
Hi Jasper, @jaspervanderhoek 
I've made the changes as requested.
Please let me know when you want me to publish the documentation. Presumably when you and Jos (@JosVerbeekMx)  have published the module?
Thanks for your input.
Mark

jaspervanderhoek (approved):
All looks good, I've discussed the release with Jos and I'll be publishing version 1 later this afternoon. 
So I'm fine with you releasing this whenever you have a moment.

One last question to confirm I'm linking the doc correctly, I'm assuming the documentation url will become: https://docs.mendix.com/appstore/connectors/opc-ua is that correct?

MarkvanMents:
Yes, that's the correct URL.
I'll publish this now.

MarkvanMents:
@jaspervanderhoek Document now published at https://docs.mendix.com/appstore/connectors/opc-ua

PR #2780, opened 2020-10-08
8.12.2 RN draft

Adam-Dupaski:


feddovanede:
Should we mention the tickets mentioned in the Jira issue for the SQL query fix?

Adam-Dupaski:
@feddovanede I see `XPath/OQL query` not SQL. Does the note need adjusting? And yes we can add the ticket number, here as well: https://github.com/mendix/docs/pull/2782/files

feddovanede:
I think the description is correct, I incorrectly referred to it as SQL, instead of OQL. Ticket numbers are 100987 and 107355.

PR #2812, opened 2020-10-15
Operator v1.6

Nidhi251289:
We will be releasing this today.


post-install-script
```diff
@@ -0,0 +1,1690 @@
+#!/bin/bash
```
zlogic, line 1: I think this file was added here by accident.

Nidhi251289: fixed.

MarkvanMents (approved):
Hi @Nidhi251289 ,
I've made a few minor changes. Is it OK to publish?
Thanks for making the PR.

Nidhi251289:
Looks good to me. Kindly proceed ahead with publishing.

PR #2820, opened 2020-10-19
Add NT 5.0.5 relase notes

smarki:
4.2.3 is a backport for versions from Studio Pro 8.9 up to 12.0.


ConnorLand (approved):

ConnorLand:
@stmarkidis so **all** of the content mentioned in this RN update applies to both 5.0.5 and 4.2.3? The only way we can think of doing "dual release entries" is if all the content applies to both releases.

smarki:
> @stmarkidis so **all** of the content mentioned in this RN update applies to both 5.0.5 and 4.2.3? We can only do "dual release entries" is if all the content applies to both releases.

Yes, 4.2.x version track 8.9 to 8.12.0 updates. 5.x 8.12.x onwords

ConnorLand (approved):

PR #2822, opened 2020-10-20
Update consumed-odata-service.md

grootjans:
Proposed update so we explicitly indicate that a key change is breaking and cannot be upgraded.


ilagordhan (approved):
Thanks for the addition @grootjans. 
I also have a question about other changes in, for example, the entity set, association type and other factors as given in the consistency errors document that I worked on with Paul R and Faried

PR #2842, opened 2020-10-27
Update contribute-to-the-mendix-documentation.md

oHAMIoo:


Adam-Dupaski:
Hi, thanks for the suggestion, but this is not the appropriate place for this content. Badges are already described here: https://docs.mendix.com/developerportal/mendix-profile/#badges

oHAMIoo:
Thank you

PR #2850, opened 2020-10-28
Update docs because 'Update' button is always shown now

paulmendix:
We changed the functionality so that the Update button is always shown, not just for the Production environment. For this the documentation also needs an update.

This will be available starting from Mendix 8.16.


content/refguide/consumed-odata-service.md
```diff
@@ -22,7 +22,7 @@ The **Consumed OData Service** document that is added to the project displays th
 	* **Update** – this button is displayed if a new service contract has been detected at the same service endpoint that is different to the one that is currently used in the app project (and shown in the **Consumed OData Service** document); this may be a new service version that has minor non-breaking changes to the previous version and therefore deployed by the publisher to the same service endpoint
```
ilagordhan, line 22: Does this mean that Update will always be shown: whether or not there is a changed contract at the end-point?
```suggestion
	* **Update** – this button is displayed so that you can **Update** the contract that is currently consumed (and shown in the **Consumed OData Service** document) in the project. It is good practice that only minor, non-breaking changes are deployed to the same end-point. 
```

content/refguide/consumed-odata-service.md
```diff
@@ -22,7 +22,7 @@ The **Consumed OData Service** document that is added to the project displays th
...
 	*  **Switch** – this button is shown if other registered instances of the same service (with the same name, from the same app) are available in Data Hub and are deployed to different endpoints (for example, to another environment or because of changes that would break existing apps consuming the previous version)
 
-	{{% alert type="info" %}}Studio Pro will show the most relevant option (**Switch** or **Update**) after examining the service name, app, and endpoint in the Data Hub Catalog. This will also be indicated in the Data Hub search pane for the consumed service with an update arrow. For further information on updating and switching services see the [Updating or Switching a Consumed OData service](#updating) section of this document.
+	{{% alert type="info" %}}Studio Pro will always show the **Update** option where you can check if an update is available. This will also be indicated in the Data Hub search pane for the consumed service with an update arrow. For further information on updating and switching services see the [Updating or Switching a Consumed OData service](#updating) section of this document.
```
ilagordhan, line 25:
```suggestion
	{{% alert type="info" %}}Studio Pro will always show the **Update** option for the **Consumed OData Service** where you can check if an update is available. In the Data Hub search and **Project**  pane, when a different contract is detected at the service end-point, this will be indicated with an update arrow for the service. For further information on updating and switching services see the [Updating or Switching a Consumed OData service](#updating) section of this document.
```

ilagordhan (commented):
Please confirm that the update arrow is only displayed in the DH pane if there is a different contract at the end-point.

content/refguide/consumed-odata-service.md
paulmendix: Indeed. The Update button will always be shown. Regardless of having a new contract available (that was always the case) and now also regardless of the environment.

content/refguide/consumed-odata-service.md
paulmendix: I confirm that the blue update arrow in the Data Hub panel is only displayed if there is a different contract available at the end-point.

content/refguide/consumed-odata-service.md
ilagordhan: Thank you very much Paul.

ilagordhan (approved):

ilagordhan:
This pull-request merged into #2896 8.16 release and DH changes branch

PR #2875, opened 2020-11-11
Added a minor change

Grasshooper:
Added a minor change


Adam-Dupaski:
Hi @Grasshooper, thanks for your suggestion, but unfortunately this change is unnecessary (and "Example" would not need to be capitalized), so I have to close this PR.

PR #2878, opened 2020-11-12
Add the Help & Learning feature 

MariaShaposhnikova:


MariaShaposhnikova:
Please merge if someone from the team reaches out to you and say that the feature is in production. 
Note, that I put Nov 13th as a release date in the Studio release notes and the featured doc. Please update if necessary.

PR #2879, opened 2020-11-12
Access restrictions which use both IP and CA

MarkvanMents:
Reverts changes in #2846


ManpreetRiar:
@MarkvanMents The fix is released. Please publish the documentation. Thanks!

PR #2880, opened 2020-11-13
Undocument the ReadCommittedSnapshot option

basdebakker:
We decided that we cannot support customers running with the ReadCommittedSnapshot option set to the non-default value of false, so it would be better to not mention the option in the documentation. This change has been approved by @arjenw .


MarkvanMents:
Hi @basdebakker - looks OK.
Do we need to communicate this in a release note or to any SQL Server customers who may be using this setting? @NielsTrots - do you have a view on this from the support perspective. Is the number of customers who might be affected so small we don't need to bother doing anything else?

basdebakker:
Note that the option remains in the code, so people using it will not be affected directly.

MarkvanMents:
OK - I'll publish this and update it later if I get further feedback.
Thanks @basdebakker

MarkvanMents (approved):
Change approved.
Change also reflected in Mendix 9 documentation.

PR #2883, opened 2020-11-13
Clarify Refresh in Client behaviour

MarkvanMents:
Description of Refresh in client is confusing - as is the behavior. Need to explain it better to the customer.
The option is described in the following documents:

- change-object
- deleting-objects
- committing-objects
- create-object
- rollback-object


content/refguide/change-object.md
```diff
@@ -60,20 +60,24 @@ Nanoflows do not support committing changes without events. Committing while run
 
 ### 3.3 Refresh in Client{#refresh-in-client}
 
-This setting defines how changes are reflected in the pages presented to the end-user.
+This setting defines whether data sources are rerun after data is committed to the database.
 
 Default: *No*
 
+{{% alert type="info" %}}
+To make pages of a Mendix app efficient, many widgets display values from an object value which is cached on the page. Attributes in widgets which use cached data are *always* reflected in the client even if they are not committed and irrespective of the value of **Refresh in client**.
```
feddovanede, line 68: How do you know which ones those are?

MarkvanMents: @feddovanede 
You don't! My current understanding is that it is everything apart from Data Grid, however the developers do not want to commit to them always working the same way.

Thus:
>When testing your app, ensure that the desired data is being displayed by the widgets you have chosen.

MarkvanMents:
No negative response received - publishing

MarkvanMents:
Cherry Picked to Mx 9

PR #2885, opened 2020-11-13
Update parse-and-format-date-function-calls.md

RantingHuman:
Added information to dateTimeToEpoch and epochToDateTime indicating that they are only available from Mendix 8.15 onwards.


MarkvanMents (approved):
@RantingHuman 
Thanks for your submission to the Mendix documentation. It will definitely help users of Mendix version 8 to know which expressions they can use.

PR #2887, opened 2020-11-17
Add missing scope to token requirements

smarki:


JelteMX:
@ConnorLand Related to ticket 111062, please update this doc. Token requirement is correct

ConnorLand:
> @ConnorLand Related to ticket 111062, please update this doc. Token requirement is correct

Will do.

ConnorLand (approved):

PR #2898, opened 2020-11-23
EC-TW-1155 Consume OData Service dev flow drafts

ilagordhan:
This how to and the supplementary advice on using the feature flag is not to be published on Github until directed by PO and PM. 
It may be issued on a case-by-by case basis as a PDF


content/howto/integration/consume-an-odata-service-developer-flow.md
```diff
@@ -0,0 +1,179 @@
...
+title: "Consume an OData Service and Use External External Entities"
+category: "Integration"
+menu_order: 8.5
+tags: ["integration", "OData service", "data hub", "consume", "external entity", "OData service URL"]
+description: "Describes the development workflow for working with oData and external entities using Odata service URL."
+---
+
+## 1 Introduction
+
+Licensed users of Mendix Data Hub can also consume published OData Services during app development without using the [Data Hub pane](/refguide/data-hub-pane) when, for example, trying out published OData Services which may not be registered in the [Data Hub Catalog](/data-hub/data-hub-catalog), or when working with locally deployed apps and their [published OData services](/refguide/published-odata-services).
```
GulamY, line 11: The first sentence is a little bit confusing. The dialog says navigate to data hub pane, but the text says that we are trying to use consumed odata service without using data hub pane. I understand that there is a distinction between `data hub project` pane and data hub pane. But I do not think we draw that distinction in the UI.

PR #2900, opened 2020-11-23
Add information about datadog metrics.

MarkvanMents:


MarkvanMents:
Available information added, and a table inserted giving a summary of the information which is available.
The document will have to be improved as and when more information becomes available.

PR #2925, opened 2020-12-04
Add info about snapshot isolation for sql server

passalaqua:
Both read committed snapshot and allow snapshot isolation need to be on, so I'm adding the second to the same section as the first.


basdebakker (approved):

MarkvanMents (approved):
Thanks for the additional information @passalaqua 
Will merge this now.

PR #2951, opened 2020-12-16
Update all client library versions for Mx9

KevinVlaanderen:


jobvs:
These content changes only apply to Mendix 9. Meanwhile we also want to keep the previous version for Mendix 8. So we'd like to have the whole group of 'Pluggable Widgets API' docs to have separate versions for the 8 and 9 releases.

ConnorLand (approved):

KevinVlaanderen:
@ConnorLand How do we intend to handle the comment raised by Job?

ConnorLand:
> @ConnorLand How do we intend to handle the comment raised by Job?

I am currently creating the versioning. I will be able to move your PR through as is in a few minutes.

PR #2963, opened 2020-12-21
App Service deprecation update

GulamY:
These changes should be synchronized with 8.18 release

Starting from version 8.18.0 app services are no longer supported


Adam-Dupaski:
Replacing this PR with #2999 for Mx8 docs base.

PR #2977, opened 2020-12-28
Update sap-fiori-3-0.md

Paolaufer:
Made a change in the first paragraph and about the Fiori launchpad.


ConnorLand:
Hi @Paolaufer . I confirmed "launchpad" should indeed be lowercase. Was the "SAP Suite" a previously correct term which is no longer being used?

Paolaufer:
Hi,
There is no official product called SAP Suite. There is the SAP Business Suite, but what they were referring to was SAP products in general (see email attached).
Hope this helps.
Regards,
Pao

ConnorLand (approved):

PR #2981, opened 2020-12-30
Update 9.0.md

ironthorn:
Mark and I agree that the third bullet point does not add anything and only confuses people. Let's remove it here as well.


MariaShaposhnikova:
Hi @ironthorn! I think that this PR is identical to [https://github.com/mendix/docs/pull/2980](https://github.com/mendix/docs/pull/2980). Is there a reason why there are two PRs? Did anything go wrong with the first one?

PR #2999, opened 2021-01-11
Deprecate app services

Adam-Dupaski:
Replaces #2963.


Adam-Dupaski:
8.18.2 or later

Adam-Dupaski:
@GulamY @joostverhoog are app services deprecated/removed from support for 8.18.2?

joostverhoog:
After careful consideration, we ended up deciding that the level of support for app services (deprecated) will not change. So this pull request can be closed.

Adam-Dupaski:
@MarkvanMents see above - app services remain deprecated in Mx8, so the Dev Portal page will stay for the foreseeable future.

PR #2982, opened 2020-12-30
Update offline synchronization docs

ekrem-aktas:

PR #3003, opened 2021-01-13
Update offline-first docs - introduce 'Selective Sync'

omnajjar:
This PR is a forward of https://github.com/mendix/docs/pull/2982

This update should be part of `Mx9.0.3`


omnajjar:
This PR needs a review by offline team, once we're done I will unmark it as 'darft'

content/refguide/synchronize.md
```diff
@@ -12,15 +12,43 @@ This activity can only be used in **Nanoflows**.
...
+This mode synchronizes objects partially, based on a selection:
+
+{{% image_container width="600" %}}
+![Synchronize](attachments/client-activities/synchronize-objects-selection.png)
+{{% /image_container %}}
+
+With this mode, only the selected objects or lists are synchronized. The synchronization is bi-directional, 
+meaning both the server database and the local database is updated for the selected objects.
+
+If the selected object is new, and not committed to the local database, it will be ignored during synchronization.
```
govorov, line 42: (This comment is a copy of the same comment in 2982)

"The selected object" is a bit ambiguous as it might be a list, a single object, or a combination of both. Maybe we can make it a bit more clear with something like:

> If the set of objects selected for synchronization contains any objects that have not been committed yet, those objects will be skipped and thus not synchronized.

content/refguide/synchronize.md
```diff
@@ -12,15 +12,43 @@ This activity can only be used in **Nanoflows**.
...
+
+If the selected object has local changes, the following steps are performed:
+
+1. Update the server database with the changes from local database.
+1. Update the local database from server database. This is useful in case the selected object has calculated attributes or has been modified in a before/after event handler microflow.
+
+If the selected object originated from the server (not created on the device), and no longer exists on the server (or is inaccessible due to the access rules), the local changes are not applied and the object is removed from the local database. In this case the value of the variable in the nanoflow for that object becomes `empty`. The server stores the discarded changes in the `System.SynchronizationFailure` entity to prevent data loss.
+
+If the selected object does not have any local changes, synchronization updates the local copy from the server database. If the object was deleted or is no longer accessible due to access rules, the object is removed from the local database.
```
govorov, line 51: Same as above

govorov, line 51: (This is a copy of the same comment from 2982)

"The object was deleted" in combination with "the object is removed from the local database" if a bit confusing. As far as I understand we mean the runtime database in the first part. Let's mention it explicitly then?

> If there is an object that has been deleted from the server or is no longer accessible due to access rules, that object will be removed from the local database too.

omnajjar: Done! ✔️

omnajjar: Done! ✔️

omnajjar:
Also in this PR, this typo is fixed! https://github.com/mendix/docs/pull/2982#discussion_r564570732

PR #3012, opened 2021-01-19
Different default value for http client cleanup

joostverhoog:
In Mendix 9.0.4, there is a new default value for the custom runtime setting http.client.CleanupAfterSeconds


Adam-Dupaski (approved):

Adam-Dupaski:
@joostverhoog the `refguide8` page was changed instead of the correct `refguide` page for 9.0.4 - I made the fix, please verify.

joostverhoog:
@Adam-Dupaski Indeed, good catch!

PR #3025, opened 2021-01-26
8.12.6 RN

Adam-Dupaski:


dahfjkg:
Please add the following for the Datastorage team:
```
The following was done for ticket 11311:
- Update SAP HANA driver to 2.7.9;
- Allow JDBC URL options to set data source properties for SAP HANA;
- Disable transactionalLobs option in SAP HANA.
```

PR #3026, opened 2021-01-27
Task Queue

MarkvanMents:


MarkvanMents:
I see that there is no option in the properties of the **Call Microflow** button on a page to call the microflow in a task queue. Do we need to state this explicitly?

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+
+ * Microflows are always executed in a *sudo* context, even when the scheduling microflow has **Apply entity access** set to *true* (see [Microflow Properties](microflow) for more information).
+ * Only committed persistable entities can be passed as parameters to the microflow. Passing a persistable *New* or *Changed* entity produces a runtime error. Basically, this means an entity must have been committed previously or is committed in the same transaction in which the task is created.
+ - The microflow is not executed immediately. The task is added only to a task queue when (and if) the transaction from which it has been scheduled ends successfully. At that point any cluster node may pick it up.
+ - If the execution fails with an exception, the failure is logged in the `System.ProcessedQueueTask` table.
+
+#### 2.2.2 Through the API
+
+The `Core` class in `com.mendix.core` has been extended with a method [microflowCall](https://apidocs.rnd.mendix.com/8/runtime/com/mendix/core/Core.html#microflowCall(java.lang.String)). It can be used to schedule a microflow for background execution as in the following example:
```
MarkvanMents, line 58: Is this API documentation going to be published for Mx9? I can't find it at present.

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+The context in which a background task runs is still under discussion and may change in the future.
+{{% /alert %}}
+
+### 2.3 Configuration options{#configuration}
+
+The period for a graceful shutdown of queues can be configured.
+
+| Configuration option          | Example value | Explanation                                              |
+|-------------------------------|---------------|----------------------------------------------------------|
+| `TaskQueue.ShutdownGracePeriod` |          10000| Time in ms to wait for task to finish when shutting down.|
```
MarkvanMents, line 81: Where can the user set this configuration option? Would be useful to point them at the right place.

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+
+* that the microflows exist
+* that the parameters match
+* that the queue exists 
+
+If any of these condition checks fail, tasks are moved to `System.ProcessedQueueTasks` with **Status** `Incompatible`. The Runtime will only start after all scheduled tasks have been checked. This should in general not take very long, even if there are thousands of tasks.
+
+### 2.7 Shutdown
+
+During shutdown, the `TaskQueueExecutors` will stop accepting new tasks. Running tasks are allowed a [grace period](#configuration) to finish. After this period, the runtime will send an interrupt to all task threads that are still running and again allow a grace period for them to finish. After the second grace period the runtime just continues shutting down, eventually aborting the execution of the tasks. The aborted tasks will be reset, so that they are re-executed later or on another cluster node.
```
MarkvanMents, line 122: Are these two grace periods the same, and are they both controlled by `TaskQueue.ShutdownGracePeriod`?

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+
+{{% alert type="info" %}}
+Interrupting task threads may cause them to fail. 
+{{% /alert %}}
+
+## 3 Monitoring
+
+To monitor tasks in the Task Queue the `QueueHelpers` module can be used in Mendix 9.
+
+Since this hasn't been released yet, it can be downloaded through [Gitlab](https://gitlab.rnd.mendix.com/runtime/taskqueuehelpers-module) by Mendix employees only.
```
MarkvanMents, line 132: I don't think we should include things in our documentation which are only available to Mendix employees. Can I remove all the references to `QueueHelpers`?

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+
+### 3.1 Logging
+
+A Logger exists specifically for all actions related to Task Queue, named `Queue`.
```
MarkvanMents, line 136: Do you mean **Log Node**?

MarkvanMents (commented):
Please can you address the questions in this document.
Thanks.

MarkvanMents:
Are there any considerations to be taken into account if you use a task queue when calling a microflow from a nanoflow, or does it work just the same?

content/refguide/task-queue.md
```diff
@@ -0,0 +1,186 @@
...
+
+After the task has been executed, it is moved to be an object of the `System.ProcessedQueueTask` entity with `Status` `Completed` or `Failed`. If the task failed with an exception, this is included in the `ErrorMessage` attribute.
+
+Arguments are stored in the `Arguments` attribute as JSON values. Arguments can be any primitive type ([variable](variable-activities))or a committed persistent object, which is included in the `Arguments` field by its Mendix identifier. Upon execution of the task, the corresponding object is retrieved from the database using the Mendix identifier. For this reason the persistent object must be committed before the task executes, because otherwise a runtime exception will occur.
+
+When a node crashes, this is eventually detected by another cluster node, because it no longer updates its heartbeat timestamp. At this point the other node will reset all tasks that were running on the crashed node. The reset performs the following actions:
+
+* create a copy of the task as a `System.ProcessedQueueTask` object with `Status = Aborted`
+* set the `Status` back to `Idle`
+* increment the `Retried` field
```
MarkvanMents, line 171: Is there any limit to the number of times the task will be retried?
There probably isn't any need, I suppose I'm thinking of a situation where the microflow crashes the node - but I'm not sure that that can happen.

content/refguide/task-queue.md
hugomendix: Yes, for now that should be good. We'll add it back as soon as the module is in the AppStore.

content/refguide/task-queue.md
hugomendix: Yes, it will/should be. I couldn't find any API docs for Mendix 9 (only 6 to 8). Where can I find them?

content/refguide/task-queue.md
hugomendix: This is a custom runtime setting: `Project` -> `Settings` -> `Configurations` tab -> double click one -> `Custom` tab -> new setting named "TaskQueue.ShutdownGracePeriod".

content/refguide/task-queue.md
hugomendix: Yes, both are defined by the shutdown grace period. So the maximum time the runtime waits for tasks to end is actually twice the configured period.

content/refguide/task-queue.md
hugomendix: Yes, indeed

content/refguide/task-queue.md
MarkvanMents: I'll ask my team and also on Slack.

content/refguide/task-queue.md
hugomendix: No, there is no limit. In theory this means that the runtime will try to execute the task eternally if it gets aborted each time. In practice this is extremely unlikely, because nodes don't get shutdown all the time. I'd say it is already quite unlikely that it reaches 2 retry attempts.

content/refguide/task-queue.md
MarkvanMents: Clarified in document.

content/refguide/task-queue.md
MarkvanMents: Removed.

content/refguide/task-queue.md
MarkvanMents: Changed and added to Log Nodes document

content/refguide/task-queue.md
MarkvanMents: Thanks for the explanation - I don't think we need to document this.

MarkvanMents (commented):
@hugomendix, This seems OK to me now.

I had two other questions.
 
> Are there any considerations to be taken into account if you use a task queue when calling a microflow from a nanoflow, or does it work just the same?
>
> I see that there is no option in the properties of the Call Microflow button on a page to call the microflow in a task queue. Do we need to state this explicitly?

I'll assume that there is nothing that needs to go in the documentation to address them.

MarkvanMents:
9.0.3 is published.

PR #3027, opened 2021-01-27
8.18.1 RN

Adam-Dupaski:


dahfjkg:
Please add the same RN that was added to 8.12.6:
```
* We updated the [SAP HANA](/refguide8/saphana) driver to 2.7.9.
* We now allow JDBC URL options to set data source properties for SAP HANA.
* We disabled the `transactionalLobs` option in SAP HANA.
```

PR #3035, opened 2021-02-01
line 90 -> making 'Business Owner' bold

Jule4ka:
by chance found a typo, which resulted in BO being not bold.


ilagordhan (commented):
Thank you very much for the feedback.

PR #3036, opened 2021-02-01
Add missing qr codes and links for legacy apps

smarki:
When moving from 8 to 9 somehow the QR codes and links weren't included.


ConnorLand:
I will merge this, then look into what caused this to be omitted during the migration. Thank you very much @stmarkidis

ConnorLand (approved):

PR #3040, opened 2021-02-03
Update supported platforms private cloud

cstevandy:
There is inconsistency for supported platforms of mendix for private cloud between the page https://docs.mendix.com/developerportal/deploy/private-cloud and https://docs.mendix.com/developerportal/deploy/private-cloud-supported-environments.

Updated the page by adding the link to the latest supported platform.


MarkvanMents (approved):
Thanks Christian (@cstevandy ).
Much better to have the list in one place.

PR #3060, opened 2021-02-12
Add support to Postgres 13

passalaqua:
Starting with Mendix 8.18.2 we will support Postgres 13.


passalaqua:
Please note that this should only be merged once 8.18.2 is out and if DAT-2271 is completed before then.

passalaqua:
Just as an update: Story is complete. Once 8.18.2 is released, this PR can be merged!

PR #3064, opened 2021-02-16
Add Java actions to Task Queue

hugomendix:
These will be added in Mendix 9.0.5.


Adam-Dupaski:
@MarkvanMents this is for task queues

MarkvanMents:
Hi Hugo (@hugomendix ),
Thanks for the PR - look good.
There is a lot of duplication between the two sections. In particular, the context information appears in both sections as well as the information about the API returning immediately and so there being no return value.
Do you think it would make it easier for customers if this information was pulled out into the general description of what Task Queues are, so we only have the information once, or is it important to have it in both sections in case users miss it?

hugomendix:
Hi Mark,
Yes, I agree about the documentation and think it would be good to extract it to a single place. Do you want to do that or would you like me to take a shot at it?

MarkvanMents:
I'm happy to do it. Unless you particularly want to?

hugomendix:
No, just go ahead...

MarkvanMents:
Have updated the document to create a new general section explaining context and other important things up front.
@hugomendix , please let me know if it looks OK.

MarkvanMents (approved):

PR #3066, opened 2021-02-17
Added picture of blank file

AnnYe-Fluithaas:
Hopefully reduces misunderstanding of what the "RequiredLib" file should look like.


AnnYe-Fluithaas (commented):
An uncertainty about the boxing for the image - I noticed that the style of the framing is different than what was used for "public-private.jpg". Most noticeable difference I saw was the shading.
I don't know which program was used for the drawing of the box in the other image.

I'm not sure if this visual inconsistency is something that should be addressed?

Adam-Dupaski:
Thanks @AnnYe-Fluithaas - I'll add the image border in SnagIt after merging

PR #3087, opened 2021-02-26
Add information about ScheduledEventInformation entity.

MarkvanMents:


content/refguide/scheduled-events.md
```diff
@@ -17,13 +17,30 @@ A scheduled event is added to your module as a document (right-click your module
 Scheduled events can be tested locally, but they will not be run if your app is deployed as a Free App. See the Free App section of [Mendix Cloud](/developerportal/deploy/mendix-cloud-deploy#free-app) for more information on Free App limitations.
 {{% /alert %}}
 
-## 2 Common Properties
+A `ScheduledEventInformation` object is created every time the scheduled event is run. This records the following:
+
+* **Name** – the **Name** of the scheduled event as entered in [Common Properties](#common-properties)
+* **Description** – the **Documentation** of the scheduled event as entered in [Common Properties](#common-properties)
```
MarkvanMents, line 23: Is this correct? Or is the description pulled from elsewhere (where). If they are the same, why do they have different names?

content/refguide/scheduled-events.md
```diff
@@ -17,13 +17,30 @@ A scheduled event is added to your module as a document (right-click your module
...
+* **StartTime** – the time this run of the scheduled event started
+* **EndTime** – the time this run of the scheduled event ended
+* **Status** – the current status of this run of the scheduled event:
+    * **Running** – the event is current running and has not yet ended
+    * **Completed** – the event has completed successfully
+    * **Error** – the event completed with an error — you can find the error on the error log using the *StartTime* and *EndTime* to locate it if necessary
+    * **Stopped** – the scheduled event was stopped before completion, for example by the app being stopped
```
MarkvanMents, line 30: Is this the correct interpretation of **Stopped**?

jds-mendix: Yes, that's correct. Scheduled events are marked as stopped during shutdown or if a cluster node 'disappears'.

content/refguide/scheduled-events.md
jds-mendix: This is correct for scheduled events coming from the model, for api-scheduled ones the description is whatever you pass as description.

content/refguide/scheduled-events.md
```diff
@@ -17,13 +17,30 @@ A scheduled event is added to your module as a document (right-click your module
 Scheduled events can be tested locally, but they will not be run if your app is deployed as a Free App. See the Free App section of [Mendix Cloud](/developerportal/deploy/mendix-cloud-deploy#free-app) for more information on Free App limitations.
 {{% /alert %}}
 
-## 2 Common Properties
+A `ScheduledEventInformation` object is created every time the scheduled event is run. This records the following:
```
jds-mendix, line 20: Technically only for `UserScheduledActions` not for `SystemScheduledActions`.

jds-mendix: This includes all events scheduled in the model. Via the API, you can do both and the distinction is not clear from the API docs.

MarkvanMents:
@jds-mendix 
Hi Jan David,
Thanks for the clarifications.
Since this document is talking about setting up a scheduled event in the model, I think I'll leave most of the subtlety about using the API out of this - otherwise I'll have to add links to the APIs for every action the customer can do through the API.

Since the differences are only there when using the API, perhaps the differences should be described there. Looking quickly at the API definition it is still not clear to me what the distinction between user and system is, but then I must be missing something because the [schedule method](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#schedule(java.lang.String,java.util.Date)) doesn't even seem to have any way of adding Names or Descriptions. Let me know what I've missed.

If you want me to get involved with improving the API documentation to make this distinction clearer, please let me know.

Can you confirm that this is good enough to go out as it is and we can improve it later?
Thanks

MarkvanMents:
Approved by Jan David, but (https://mendix.slack.com/archives/C3T260XGA/p1614327529035100) but following up on other suggestion before publishing.

MarkvanMents:
Added information about scheduled event documentation not being visible as discussed in JIRA [TW-1225]
Have not added information about SystemScheduledActions as this doesn't seem to be documented anywhere. Will follow up separately - discussion so far is:

> … I would suggest to add that there are SystemScheduledActions and that they don’t result in ScheduledEventInformation in the database. Just one sentence along those line might save some poor dev a day or two

content/refguide/scheduled-events.md
MarkvanMents: Cannot find other documentation on this difference - will pick this up later and document properly.

PR #3088, opened 2021-02-26
Update check-box-set-selector.md

insead215:
Added the following note to 2.2:
At least one _Display Attribute_ (with Header text, Attribute etc.) must be designated, otherwise the widget might block the page or popup from loading.


ConnorLand:
Hi @insead215 . Thank you for your PR. This is an older widget. Therefore could you provide me with a few more details about how you discovered this behavior?

insead215:
Hi @ConnorLand . Thanks for looking into this. I placed the widget in my design and accidently forgot to add a Display Attribute. However, I was able to compile, run and test the application, which then proceeded to hang when I came to the page in which I used the checkbox widget.

ConnorLand (approved):

PR #3096, opened 2021-03-02
Added forgotten releaste notes to 9.0.4

joostverhoog:


Adam-Dupaski (approved):

Adam-Dupaski:
I have removed the 2 fixes that were already released in 8.18.0, which came out before 9.0.4. We do not duplicate notes between the last minor version and a new major version beta/GA. Assumption is that those have been fixed for 9.0 - this has been verified with AppDev unit directors. Only new fixes go in the new major version beta/GA RN.

PR #3105, opened 2021-03-04
Case sensitive microflow names in config

grootjans:
Based on a ticket we recently had with a customer, where a config change led to metrics no longer being submitted.


MarkvanMents (approved):
Thanks for the submission Robbert Jan @grootjans.
Makes my job much easier when someone else does all the work.
Approving and merging now.

PR #3100, opened 2021-03-03
[WC-109] Add documentation for refactored progress bar widget

Keraito:
Similar to https://github.com/mendix/docs/pull/3086:

The progress bar widget has been converted to a pluggable widget that will only be supported on MX 9 and above. I kept the old widget documentation as well, since mx developers can still use these widgets with mx 7 or 8.

PR #3107, opened 2021-03-04
[WC-110] Add documentation for refactored progress circle widget

Keraito:
Similar to #3100:

The progress circle widget has been converted to a pluggable widget that will only be supported on MX 9 and above. I kept the old widget documentation as well, since mx developers can still use these widgets with mx 7 or 8.


MarkvanMents:
One for you Connor?

ConnorLand (approved):

PR #3118, opened 2021-03-11
Free apps are no longer limited for Data Hub

joostverhoog:
In Mendix 9.0.5, free apps are no longer limited for Data Hub


ilagordhan (approved):
Thank you for the informative discussion. Hope that this better reflects what is implemented.

PR #3119, opened 2021-03-11
Free apps are no longer limited for Data Hub

joostverhoog:
In Mendix 8.18.3, free apps are no longer limited for Data Hub


ilagordhan (approved):
Thanks Joost for the explanation. Hope this better reflects what is implemented.

PR #3124, opened 2021-03-12
Clarify backup structure

jeohist:
We no longer require the use of subfolders for files in Cloud v4, though we still generate backups in that format to be compatible with Cloud v3. I don't know whether the phrasing I've used is clear enough, but at least this should make life a lot easier for customers migrating from on-prem / Mendix Private Cloud to Cloud v4.

### DIFF
--- a/content/developerportal/operate/restore-backup.md
+++ b/content/developerportal/operate/restore-backup.md
@@ -137,11 +137,17 @@ The dump must be created with pg_dump version 9.6.17 or below. If it is created 
 
 ### tree folder
 
-This contains the files which are stored in external file storage. The file has the name of the uuid used within Mendix to identify the resource.
+This contains the files which are stored in external file storage. Each file has the name of the uuid used within Mendix to identify the resource.
 
-#### Cloud v3
+#### Mendix Cloud V4
 
-The situation for Cloud v3 is slightly different. Each file is stored in a second level location:
+For Mendix Cloud V4 the files are stored in a flat structure.
+
+#### Mendix Cloud V3
+
+The situation for the deprecated Mendix Cloud v3 is slightly different. If you need to restore a backup to Mendix Cloud V3, you will need to implement the structure described below. The restore functionality in Mendix Cloud V4 will also recognize this structure if you are restoring a backup taken on V3 to a V4 environment.
+
+Each file is stored in a second level location:
 
 ```
 /tree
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi Joey (@jeohist)
This section is specifically about _the format of the backup file_, so even for Cloud V4 the format of the backup will be as described in the documentation, if I understand what you say above.
Are you saying that when the backup is restored to Cloud V4 that this folder structure is not implemented as it is stored in the backup file but that the files are flattened out and all put in the same folder in the blob storage? In other words, that the restored backup won't look the same as the backup?
What happens when the backup is restored to a non-Mendix cloud - such as Mx4PC? Does it flatten the structure or keep it?

MarkvanMents:
@jeohist - can you let me know about the comment above. I'd like to improve the documentation, but I need to be clearer what we are saying about the structure of the backup file and how it relates to the structure in the deployed app.
Thanks

jeohist:
@MarkvanMents This PR was intended more as a way to show that the tree structure is optional for Cloud v4. I don't know whether this also applies to Mendix for Private Cloud. The reason for this PR was to make it easier for people to migrate from Mendix for Private Cloud to Cloud v4 (putting backups into the tree structure is a bit of an annoyance), maybe we can rewrite the text to reflect that.

MarkvanMents:
@jeohist So is the structure of the S3 files in the backup just the same as the structure that it had on disk (whatever that was)? Or does the Mendix backup create this structure in case the backup needs to be restored to a V3 node?

I was under the (probably mistaken) impression that the backup was created in this format on purpose, but are you saying that the backup was only in this format because it was a V3 backup and I would have seen something different for a V4  backup?

(I'm all for making things easier for customers - which is why I want to make it as clear as possible).

Thanks for getting back to me on this.

MarkvanMents:
@jeohist 
Hi Joey,
Can we have a quick call about this. I think we can resolve it in 10 minutes rather than having discussions here and in Slack?

MarkvanMents (approved):
@jeohist 
Hi Joey,
I've extended the description a bit, based on our meeting yesterday.
I'm happy to publish - do you have any more comments or is it good to go?
Thanks for submitting the PR and taking the time to explain it to me.

MarkvanMents (approved):
Going for it.

PR #3164, opened 2021-03-25
Add release notes for App Templates app release.

Membad:


Adam-Dupaski:
Hi @Membad - are there some more interesting details that can be added here? Otherwise, I question if release notes are really necessary.

Adam-Dupaski (approved):

PR #3170, opened 2021-03-25
Disabled  should be set with readOnly

Andries-Smit:
At this point of the tutorial it is still the disabled based on the readOnly property
`disabled={this.props.textAttribute.readOnly}`

Later in the onChange section the `isReadOnly` function introduced.


ConnorLand:
Thanks Andries! I have made these changes in both the Mx8 and Mx9 versions of the doc.

PR #3176, opened 2021-03-26
Add differences in XPath query engines for Mendix 9

basdebakker:


basdebakker:
> Would it be better to use less abstract entities in examples? For example, here we use customers and addresses: https://docs.mendix.com/refguide/xpath-constraints

Personally, I find the shorter names easier to read. What does the documentation team think?

MarkvanMents (approved):
Thanks for the submission, Bas (@basdebakker )
Merging and publishing this now.

PR #3185, opened 2021-03-29
update JS actions docs with async/await

bahkostya:


content/howto/extensibility/best-practices-javascript-actions.md
```diff
@@ -333,23 +333,23 @@ Explaining the callback code:
...
-        .then(function(jsonData) {
-            return jsonData.name; // Get the data
-        })
-        .catch(function(error) { Promise.reject("Failed to get user information") });
-    });
     // END USER CODE
 }
 ```
-
+TODO
```
Robinfr, line 352: There is a TODO here. But I think the part below here still needs to change a bit. We might not have to mention that you don't have to call `resolve` since there is no `Promise.resolve` in any example and we simply `return` the result in the updated examples.

What do you think?

content/howto/extensibility/best-practices-javascript-actions.md
```diff
@@ -333,23 +333,23 @@ Explaining the callback code:
 This function uses the browser promise API:
```
Robinfr, line 333: There is no mention of promises in this code tho aside from the usage of `async/await` so I'm not sure that we need to mention this. Maybe instead replace it with: "This function uses the Fetch API:"

content/howto/extensibility/best-practices-javascript-actions.md
```diff
@@ -481,3 +489,4 @@ Not all capabilities are recommended for use. Consider the side effects that an
 * JavaScript basics:
 	* [Mozilla JavaScript Basics](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/JavaScript_basics)
 	* [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+    * [Async functions](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await)
```
Robinfr, line 492: This doesn't seem aligned with the other items in the list.

Robinfr (changes requested):
Reviewed, needs a few changes 😄

content/howto/extensibility/best-practices-javascript-actions.md
bahkostya: I agree, do you think we should remove the block below completely? 

> As this is a part of a promise already, it does not need to call a `resolve` function or return a promise itself — this makes error handling on promises easier, as a single error function can be used to catch errors at all three steps: fetch, parse JSON, and accessing the data (for a more detailed explanation, see the [MDN promise documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise))

bahkostya: Or can be changed to something like:

> As this is an async function, error handling at all three steps: fetch, parse JSON, and accessing the data can be done inside a single `try...catch` block (for a more detailed explanation, see the [MDN documentation for error handling with async/await](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await#adding_error_handling)).

Robinfr: I would replace it with the second block. I think in this instance, explaining more is probably better 😄

content/howto/extensibility/best-practices-javascript-actions.md
bahkostya: fixed

content/howto/extensibility/best-practices-javascript-actions.md
bahkostya: yep, by accident had settings in IDE set to spaces, but this project uses tabs, fixed now

Robinfr (approved):

Robinfr:
@ConnorLand ready for your review 😄

ConnorLand (approved):

PR #3254, opened 2021-04-14
Introduce 'isSyncing' special check

omnajjar:
This update should be part of `Mx9.2.0` docs

### DIFF
--- a/content/refguide/special-checks.md
+++ b/content/refguide/special-checks.md
@@ -133,9 +133,9 @@ An example of an input can be the following:
 isSynced($currentObject)
 ```
 
-## 6 Checking Whether a Synchronization Is In Progress {#is-syncing}
+## 6 Checking Whether a Synchronization Is Running {#is-syncing}
 
-This function is available only in expressions for [conditional visibility or editability](common-widget-properties), as only they are evaluated on client side.
+This function is available only in client-side expressions (expressions in [nanoflows](nanoflows) and [pages](pages)).
 
 ### 6.1 Input Parameters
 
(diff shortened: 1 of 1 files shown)


content/refguide/special-checks.md
```diff
@@ -132,3 +132,27 @@ An example of an input can be the following:
 ```java
 isSynced($currentObject)
 ```
+
+## 6 Checking Whether a Synchronization Is In Progress {#is-syncing}
+
+This function is available only in expressions for [conditional visibility or editability](common-widget-properties), as only they are evaluated on client side.
```
govorov, line 138: You can use the function in any client-side expression, not only conditional visibility :thinking:
For example a text block or whatever else.

content/refguide/special-checks.md
```diff
@@ -132,3 +132,27 @@ An example of an input can be the following:
...
+
+No input parameter is required for this check.
+
+### 6.1 Output
+
+The output is described in the table below:
+
+| Value                                                        | Type    |
+| ------------------------------------------------------------ | ------- |
+| Returns `true` when there is an in progress [synchronize](synchronize) action running and `false` otherwise. This check can be used to avoid performing [simultaneous synchronizations](synchronize#limitations).| Boolean |
```
govorov, line 150: Maybe something like this?

`Returns true when there is a synchronization process running, false otherwise`

To focus more on the fact that the sync itself is happening at a given moment, regardless of the nanoflow activity.

omnajjar: Yes, indeed it is more focused✔️

content/refguide/special-checks.md
omnajjar: Updated! ✔️

govorov (approved):

ConnorLand (approved):

PR #3255, opened 2021-04-15
Update Standard XL Plan

jeohist:
We do not offer any database plans with 3GB of RAM, it should be 4GB.
This looks like a typo as it's correct in the Premium plans.


MarkvanMents (approved):

MarkvanMents:
Thanks for the change, @jeohist

PR #3268, opened 2021-04-19
Delete database-size-reduction.md

jeohist:
This document is not applicable to Cloud v4 (in fact, it could cause the opposite to happen). We should remove it, or at least mark it as applying to Cloud v3 only.


MarkvanMents:
Will check with support to see if they need it for their extended V3 support.
Otherwise am happy to delete.

MarkvanMents:
Hi Joey (@jeohist )
I checked with support and they would like this document left on the site for now.
I have therefore added a warning that it should only be used with V3 and removed other links to the document (https://github.com/mendix/docs/commit/2aa9b97d53012787449954bcf6f93e7e80641222)

PR #3279, opened 2021-04-20
Draft: Internet Protocol Version 6 (IPv6) support for applications in Mendix Cloud v4

Infra-Red:
Signed-off-by: Andrei Krasnitski <andrei.krasnitski@mendix.com>

### DIFF
--- a/content/developerportal/deploy/access-restrictions.md
+++ b/content/developerportal/deploy/access-restrictions.md
@@ -71,7 +71,7 @@ Click **Save** to save the current certificate profile.
 
 You can specify a number of different IP ranges. Click **New** to add a new one, or use the **Edit** or **Delete** button to modify an existing IP range.
 
-For each IP range, you can specify a **Name** for the range, and a range of addresses. Mendix Cloud v4 supports both IPv4 or IPv6 format addresses.
+For each IP range, you can specify a **Name** for the range, and a range of addresses. Mendix Cloud v4 supports only IPv4 format addresses.
 
 ![](attachments/accessrestrict/ip-range.png)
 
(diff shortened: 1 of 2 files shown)


MarkvanMents (changes requested):
HI Andrei (@Infra-Red )
There are two changes in this PR, one announcing a change for 10 May, but the documentation change is only valid after 10 May.
I'm happy to make an announcement in the release notes, but I can't change the documentation until the feature is actually implemented.
Are you happy for me to remove the change to access-restrictions.md and just publish the release note?

Infra-Red:
Fine with me, please go ahead and publish only announcements. I will provide additional pull request with documentation changes after the release. Thanks!

MarkvanMents:
Thanks Andrei (@Infra-Red )
I've proofread the release note.
Two Questions:
1. Do customers need to restart their apps on 10 May to get IPV6 access, or is it implemented automatically?
2. it is not clear to me whether users can set up IPV6 ranges in Access Restrictions _before_ 10 May? If they can, I think we should make that clearer. It would mean they can set them up today and know that there will be zero downtime for their apps on 10 May.

Infra-Red:
Hi @MarkvanMents. Thank you for your feedback.
1. It is implemented automatically on the platform level.
1. Customers can configure IPV6 ranges in Access Restrictions before 10 May. We will also inform Technical Contacts of the affected environments directly.

MarkvanMents (approved):
Thanks for this PR Andrei, @Infra-Red.
I have included your answers in the text and will publish this now.
Let me know if you think anything could be clarified.
Mark

PR #3299, opened 2021-04-29
Private Cloud release notes May 4

zlogic:
Updated documentation and release notes for the upcoming Private Cloud release.

### DIFF
--- a/content/developerportal/deploy/private-cloud-cluster.md
+++ b/content/developerportal/deploy/private-cloud-cluster.md
@@ -703,19 +703,23 @@ When choosing the **Existing docker-registry secret**, you will need to add this
 
 Check the **Enable Proxy** checkbox if a proxy is required to access the public internet from the namespace; you will be asked for the proxy configuration details.
 
-List all local (including cluster-local) IP addresses and domains in the **No proxy for** field. For more information about how to use this field, see the [documentation](https://pkg.go.dev/golang.org/x/net/http/httpproxy):
+List all local (including cluster-local) IP addresses and domains in the **No proxy for** field. The format is listed below:
 
-> It specifies a string that contains comma-separated values specifying hosts that should be excluded from proxying.
->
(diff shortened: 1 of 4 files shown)


content/developerportal/deploy/private-cloud-cluster.md
```diff
@@ -699,11 +699,29 @@ For **Amazon Elastic Container Registry**, you will need to configure registry a
...
+List all local (including cluster-local) IP addresses and domains in the **No proxy for** field. The format is listed below:
+
+Hosts which should be excluded from proxying are specified as:
+
+* a string containing comma-separated values
+* each value is
+	* an IP address prefix (`1.2.3.4`)
+	* an IP address prefix in CIDR notation (`1.2.3.4/8`)
+	* a domain name
+	* a special DNS label (`*`) indicates that no proxying should be done
```
MarkvanMents, line 715: I don't understand what putting `*` here would do.
We are listing the IPs for which no proxying should be done. If we list `*` is that a double negative? In other words, if we put `*` here than proxying has to be done for all IP addresses, even if we have added some other IP addresses here?

MarkvanMents (changes requested):
Hi Dmitrii (@zlogic )
Thanks for the comprehensive release notes.
I have one question about the httpproxy exclusion list - if you could address that to make it clearer for the customer?

content/developerportal/deploy/private-cloud-cluster.md
zlogic: This is documentation from a third-party library, and the way I understand this works is:

* `*` is a double-negative, which means "do not use the proxy for any IP addresses or domains".
* If `*` is on the **No proxy for** list, adding any domains or IP addresses will not have any effect. `*,example.com,1.2.3.4/8` means "do not use the proxy for any IP address - and also don't use the proxy for `example.com` and `1.2.3.4/8`". Which is identical to just using `*`.

MarkvanMents: OK - updated.

MarkvanMents (approved):

zlogic:
Thank you @MarkvanMents! Can you please publish the release notes now?

PR #3308, opened 2021-05-05
Update deploy-api.md

RCatersels:
JSON example was wrong. Booleans should be false and not "false" because then they become a string. Furthermore should the picture of the domain model be updated. Environment object is missing all kinds of attributes. Furthermore does it not reflect the Backup API V2. And the documentation should change again when V3 cloud gets depricated due to attribute name changes depending on V3 or V4 cloud.


MarkvanMents (approved):
Thanks for the submission Ronald (@RCatersels )
Great to have our documentation improved.
Yours
Mark

PR #3320, opened 2021-05-10
Added documentation on the dropping of IE Support

Robinfr:
We've dropped support for IE11 in Mendix 9, meaning we no longer support any version of Internet Explorer. This document makes clear what users that still use that version will see, as well as what to do if your users are still stuck using IE.


Robinfr:
@ConnorLand can be reviewed 😄

Adam-Dupaski (approved):

PR #3324, opened 2021-05-10
Update content/refguide/call-rest-action.md

ilagordhan:
# Conflicts:
#	content/refguide/call-rest-action.md

From the original MR from JV for call-rest-action change for 9.2


ilagordhan:
Original pull request reverted which was mistakenly merged.

PR #3334, opened 2021-05-12
Update ux-theming commands based on latest version 1.7.0

JelteMX:
UX-theming was updated to 1.7.0, with that the commands to run ux-theming were changed. It isn't that important, as it is not an officially supported tool, but I wanted to make sure people use the right commands.


ConnorLand (approved):

ConnorLand:
Thank you for your npm additions @JelteMX .

PR #3341, opened 2021-05-17
Update environments.md

RVeenvliet:
Document provides wrong information by only saying they can be combined. They can be combined, but are not checked both. It's either working based on IP or based on certificate.


MarkvanMents (approved):
Hi Roy (@RVeenvliet)
This is, in fact, mentioned in the document linked from here, which covers access restriction profiles in more detail (https://docs.mendix.com/developerportal/deploy/access-restrictions#2-access-restriction-profiles). But it is a good idea to add it here too.
Thanks for your contribution to the documentation. It's great to get feedback from developers who are using it.
Mark

PR #3345, opened 2021-05-17
[RTI] Clarify section about scheduled events

atennapel:
I clarified the section about scheduled events.


MarkvanMents (approved):
Thanks Albert @atennapel 
Thanks for making the change and improving the documentation.

PR #3348, opened 2021-05-18
Authentication module name IIoT not IoT

HeikoOsigus:
update screenshot for mindpshereauthenticator.png


MarkvanMents (approved):
Thanks for the review/update Heiko (@HeikoOsigus )
Will merge it into the main PR.
I've spotted some other Iot/IIot changes in the main MindSphere document, so I'll change those as well!

PR #3349, opened 2021-05-18
Updated the OQL demo link

TimvanSteenbergen:
Updated the OQL demo link to the version for Mx 9+


Adam-Dupaski:
Thanks @TimvanSteenbergen for maintaining these demos!

PR #3351, opened 2021-05-19
Update mendix-cloud.md

RCatersels:
I find it strange that an upgrade of Postgres is advised since the cloud is still on 9.6. So installing an higher version is imho ill advice. Updating PGAdmin also solves the issue and should be the perferred method.

### DIFF
--- a/content/releasenotes/developer-portal/mendix-cloud.md
+++ b/content/releasenotes/developer-portal/mendix-cloud.md
@@ -26,7 +26,7 @@ For information on the current status of deployment to Mendix Cloud and any plan
 
 #### Improvements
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
@pommi 
Hi Pim,
This was your change. Do you agree that this is a good alternative to the original suggestion?

pommi:
I get what @RCatersels is saying, but we advise to upgrade the PostgreSQL **client** software. PGAdmin can be seen as one, like all other PostgreSQL clients, mentioned on https://wiki.postgresql.org/wiki/PostgreSQL_Clients.

We could change the text a little bit by changing:
```
, for example PostgreSQL 12 or 13.
```
To:
```
, for example PostgreSQL client 12 or 13, pgAdmin 4.12+.
```

MarkvanMents:
@pommi 
Does this look OK now, Pim?

pommi:
Looks good to me :+1:

MarkvanMents (approved):
Thanks for the suggestion Ronald (@RCatersels)
The development team would like to keep the other options, so I've added your suggestion rather than replacing the existing ones.
Thanks for the submission and improving the Mendix documentation. Thanks for taking the trouble to provide a link as well, that is always helpful to our customers.
Yours
Mark

PR #3281, opened 2021-04-20
V3 to V4 Migration Tool

MarkvanMents:

PR #3377, opened 2021-05-26
Reinstate V3 to V4 Migration Tool

MarkvanMents:
Updates to support the new V3 to V4 migration tool.
Reverts mendix/docs#3374 and reinstates mendix/docs#3281


content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,204 @@ There are a few other differences between the way you develop and deploy apps in
...
+    * Custom environment variables
+    * Preferred maintenance window
+    * Tags
+
+    {{% alert type="info" %}}You do not need to set up your **Custom Domains**. When the app is migrated to V4, it will be given the same name as your current app and so your Custom Domain will automatically pick up the V4 app.{{% /alert %}}
+
+### 4.3 Replicating the Data
+
+Now that you have your new Mendix Cloud V4 environment, you can start replicating the data.
+
```
ravi4mendix, line 108: New changes to be added:

```
### Two ways to transfer database data

The system is capable of transferring database data from v3 to v4 in the following two ways:

1. Live replication using AWS Database Migration Service (DMS).
2. PostgreSQL dump and restore.

AWS DMS allows changes on v3 to be replicated to v4 as they happen. This means smaller downtime during final migration.
However, it is important to note that sometimes AWS DMS replication fails to transfer the data correctly. When this happens we automatically detect the failure and switch to database dump and restore mode. In this case you will see the "Force database dump restore" checkbox marked automatically on the UI. Dump-restore mode could also be forced from the UI by ticking the “Force database dump restore” checkbox.

### Process with AWS DMS
```

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,204 @@ There are a few other differences between the way you develop and deploy apps in
...
+3. The target v4 app node will be filled with the name `<current-app-name>-v4`.
+
+4. Select the source and target environments. In most cases, these will have the same name. For example, you want to replicate data from the V3 `production` environment to the V4 `production` environment.
+
+    ![Select V3 node and environment](attachments/migrating-to-v4/select-v3-node-to-migrate.png)
+
+5. Ensure that the app in the target V4 environment is stopped. You cannot replicate data into an app that is running.
+
+6. Click **Replicate data and files**.
+
```
ravi4mendix, line 122: Add a Note here:

```
**NOTE** : Do not mark the "Force database dump restore" checkbox
```

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,204 @@ There are a few other differences between the way you develop and deploy apps in
...
+    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the V3 appnode.
+
+    The replication will continue to run, so any changes to the data while your V3 app is still running are reflected in the replicated data.
+
+7. Click the **Migration** icon to see the status of your replications.
+
+    ![Icon for migration page](attachments/migrating-to-v4/migration-page.png)
+
```
ravi4mendix, line 130: New changes to be added here

```
8. Once initial data transfer is complete, progress will be at 100% but data will be kept synchronized from v3 to v4.

### Process with PostgreSQL dump and restore

1. Follow steps 1-5 mentioned above in [Process with AWS DMS](#process-with-aws-dms) .

2. Mark the "Force database dump restore" checkbox and then click **Replicate data and files**.

    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the V3 appnode.

    This is a one time database copy from v3 to v4 node unlike live replication. This is recommended in the following cases:

    * Database replication with AWS DMS fails to replicate the data correctly
    * Database replication with AWS DMS fails with validation errors

3. Once data transfer is complete, progress will be at 100%.
```

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,204 @@ There are a few other differences between the way you develop and deploy apps in
...
+The migration page lists all the ongoing migrations including the following information:
+
+* the source and target environments
+* UUID of the source production environment
+* whether the tool is current **Replicating** data to the V4 environment or is **Stopped**
+* what proportion of the source data has been replicated, listed for both files and database – this will sometimes show zero
+    * if replication isn't running
+    * during the replication process when the process checks whether files are synchronized
+* UUID of the target production environment
+* a button to **Stop replication** if the tool is currently replicating, or **Start replication** if the tool is currently stopped
```
ravi4mendix, line 140: Line 140 to be updated like below

```
* a button to **Stop replication** with "interrupt" option if the tool is currently replicating, or **Start replication** if the tool is currently stopped
```

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,204 @@ There are a few other differences between the way you develop and deploy apps in
...
+### 4.4 Testing the Replicated Data
+
+Once you have replicated your data, you should test it, to ensure that everything is working as expected.
+
+{{% alert type="info" %}}You should wait before testing to ensure the replication progress has reached 100%, otherwise you may find some data is missing.{{% /alert %}}
 
-To migrate your app from a v3 node to a v4 node in the Mendix Cloud, follow the steps in the sections below.
+1. Go to the migration page of the Migration Tool.
+
+2. Click **Stop** next to the migration you want to test.
```
ravi4mendix, line 152: New changes to be added here

```
2. Click **Stop replication** next to the migration you want to test.

    **Note:** If replication is interrupted (i.e. “Interrupt” checkbox is marked) data on v4 could be incomplete, especially if this is done before progress has reached 100%.
   ```

**Review by ravi4mendix — CHANGES_REQUESTED** [review]
Adding changes for the new features that are added to the tool. Will share new updated images shortly.

**Review by ravi4mendix — CHANGES_REQUESTED** [review]
select-v3-node-to-migrate
migration-status

Attaching new screenshots to replace existing ones with the same names.

File: content/developerportal/deploy/migrating-to-v4.md (hunk 2 above)
Comment (Adam-Dupaski) [reply content/developerportal/deploy/migrating-to-v4.md:122]: https://github.com/mendix/docs/pull/3377/commits/6623947a8d6f583b8237528f5e777545e9efe467

File: content/developerportal/deploy/migrating-to-v4.md (hunk 4 above)
Comment (Adam-Dupaski) [reply content/developerportal/deploy/migrating-to-v4.md]: https://github.com/mendix/docs/pull/3377/commits/44772b3fbcfd6f72e324fde9c9e148bdf29b5a4a

File: content/developerportal/deploy/migrating-to-v4.md (hunk 5 above)
Comment (Adam-Dupaski) [reply content/developerportal/deploy/migrating-to-v4.md]: https://github.com/mendix/docs/pull/3377/commits/9ae99d23c7f854bb919294f5cf17ab6d362a151a

File: content/developerportal/deploy/migrating-to-v4.md (hunk 1 above)
Comment (Adam-Dupaski) [reply content/developerportal/deploy/migrating-to-v4.md:108]: https://github.com/mendix/docs/pull/3377/commits/c1ed2798a77249460a15992812bd71515131fb2f

File: content/developerportal/deploy/migrating-to-v4.md (hunk 3 above)
Comment (Adam-Dupaski) [reply content/developerportal/deploy/migrating-to-v4.md:155]: https://github.com/mendix/docs/pull/3377/commits/d57ca1d61a9688f81f7d58cae31dc2ef57d1fe77

**Comment by Adam-Dupaski** [comment]
> 
> 
> select-v3-node-to-migrate
> 
> migration-status
> 
> Attaching new screenshots to replace existing ones with the same names.

https://github.com/mendix/docs/pull/3377/commits/ca310d135a6340c92282db401ee989dac4c9d4aa

**Comment by Adam-Dupaski** [comment]
@MarkvanMents please review upon return.

File: content/developerportal/deploy/migrating-to-v4.md [hunk 6]
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the V3 appnode.
+
+    The replication will continue to run, so any changes to the data while your V3 app is still running are reflected in the replicated data.
+
+7. Click the **Migration** icon to see the status of your replications.
+
+    ![Icon for migration page](attachments/migrating-to-v4/migration-page.png)
+    
+8. Once initial data transfer is complete, progress will be at 100%, but data will be kept synchronized from v3 to v4.
+
```
ravi4mendix, line 143: @Adam-Dupaski can you please move the lines 158-171 right here.

```
### Process with PostgreSQL dump and restore

1. Follow steps 1-5 mentioned above in [Process with AWS DMS](#process-with-aws-dms) .

2. Mark the "Force database dump restore" checkbox and then click **Replicate data and files**.

    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the V3 appnode.

    This is a one time database copy from v3 to v4 node unlike live replication. This is recommended in the following cases:

    * Database replication with AWS DMS fails to replicate the data correctly
    * Database replication with AWS DMS fails with validation errors

3. Once data transfer is complete, progress will be at 100%.
```

Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/4d04e307fbbc2f26b9c242c0712f5e5442ad80b2

igor-mendix:
4.3.3, point 2: "The replication process will copy all the data in the database, including files based on FileDocument entities, such as images, which are stored in the storage of the v3 app node." - this is confusing. It seems like it only applies to dump-restore mode, but it's not related to DB copy mode at all because this is done by file sync part.
In the same place, regarding use case scenarios. We don't expose information about the kind of DMS failure that was experienced. For now it just transitions to "failed" state, so we should simplify this. Maybe something like "DMS fails after several attempts or there's critical data validation failure in which case dump-restore mode will be forced automatically"

Also it is worth noting that dump-restore will be run twice: once on replication activation, second on replication stop (if "interrupt" flag is not set).

"The migration page lists all the ongoing migrations, including the following information:" this should be a separate header. Right now it seems to be a part of dump-restore mode.

In this section it is also worth noting that file and row counts are not realtime and may be slightly off if the v3 app is running. And maybe also note that for file sync it is refreshed on every sync run (5 minutes pause between stop of the previous one and start of the new one, not exactly every 5 minutes), and for dump-restore it's refreshed twice when actual process is run on start and on stop as mentioned above.

Maybe also a note about displayed times: they depict how much time the last operation took (be it file sync or dump-restore). These timings add to downtime, but as file sync and dump-restore are run in parallel, only the bigger one will contribute directly to the downtime.It is also recommended to stop replication at least once because upon stop there are more operations that take additional time. If the replication is stopped, last operation timing will include these additional operations thus being more accurate.

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+{{% alert type="warning" %}}
+The final migration will only check that *production* data has been completely transferred to the new V4 app.
+
+If you want to keep all your test and acceptance data, you need to review the migration status yourself. We recommend that you do this before initiating the final migration, but you can migrate non-production data after the final migration if necessary.
+{{% /alert %}}
+
+{{% alert type="info" %}}
+We recommend that you allow at least an hour between starting to replicate the data (section 4.3, above) and starting the final migration. This is because we reduce the TTL for the DNS records at the beginning of the process, but this will take some time to propagate and allow us to switch quickly between your V3 and V4 app.
+{{% /alert %}}
+
```
ravi4mendix, line 208:
```
It is important to note that if you have forced database dump restore during the initial replication or it was automatically turned on due to a detected failure with AWS DMS replication in that case the final migration will have considerable downtime.
The displayed times on the `Migration Status` page depict how much time the last operation took (be it file sync or database dump restore). These timings add to downtime, but as file sync and dump-restore are run in parallel, only the bigger one will contribute directly to the downtime.
 ```

**Review by ravi4mendix — CHANGES_REQUESTED** [review]
Some more updates

File: content/developerportal/deploy/migrating-to-v4.md [hunk 8]
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+#### 4.3.3 Process with PostgreSQL Dump and Restore {#dump-restore}
+
+1. Follow steps 1-5 above in the [Process with AWS DMS](#aws-dms) section.
+
+2.  Select the **Force database dump restore** check box, then click **Replicate data and files**.
+
+    The replication process will copy all the data in the database, including files based on `FileDocument` entities, such as images, which are stored in the storage of the v3 app node.
+
+    This is a one-time database copy from the v3 to v4 node, unlike live replication. This is recommended in the following cases:
```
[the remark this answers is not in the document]
igor-mendix: There's a mixup here a bit. It is forced automatically only on validation failure. But it is also recommended to switch to dump-restore mode if there are multiple failures.

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+
+4. Select the source and target environments. In most cases, these will have the same name. For example, you want to replicate data from the V3 `production` environment to the V4 `production` environment. Please note you should not check the **Force database dump restore** check box.
+
+    ![Select V3 node and environment](attachments/migrating-to-v4/select-v3-node-to-migrate.png)
+
+5. Ensure that the app in the target V4 environment is stopped. You cannot replicate data into an app that is running.
+
+6. Click **Replicate data and files**.
+
+    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the V3 appnode.
```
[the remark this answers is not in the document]
igor-mendix: Files are stored on the appnode. IMO we should remove "files based on ..." part altogether and just say "...copy all files and data in the database"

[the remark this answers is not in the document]
ravi4mendix: IIRC we also store files on database if the microflow is designed in such a manner. That is the reason I have mentioned this. Is it not true?

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
ravi4mendix: Fixed it now.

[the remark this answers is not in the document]
igor-mendix: So I'd replace the 3 rows Ravi mentioned with:
"This is a one-time database copy from the v3 to v4 node, unlike live replication. Database dump and restore will be forced automatically if AWS DMS data integrity check fails after multiple attempts. DMS may fail for other reasons but it does not trigger fallback to dump-restore mode because it is usually recoverable if replication is restarted. If DMS keeps failing then it is advised to force dump-restore mode manually by ticking the relevant checkbox."

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: Also worth adding:
"It is also recommended to stop replication at least once because upon stop there are more operations that take additional time. If the replication is stopped, last operation timing will include these additional operations thus being more accurate."

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/9908a3e6b93c05a6ca47c5448210ee7d3c236763

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+
+    This is a one-time database copy from the v3 to v4 node, unlike live replication. This is recommended in the following cases:
+
+    * Database replication with AWS DMS fails to replicate the data correctly
+    * Database replication with AWS DMS fails with validation errors
 
-To migrate your app from a v3 node to a v4 node in the Mendix Cloud, follow the steps in the sections below.
+3. Once data transfer is complete, progress will be at 100%.
+
+The migration page lists all the ongoing migrations, including the following information:
```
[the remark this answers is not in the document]
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/e35ae2451e62fc3355f4be1441de8d975fa87044

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+
+* The source and target environments
+* The UUID of the source production environment
+* Whether the tool is current **Replicating** data to the V4 environment or is **Stopped**
+* What proportion of the source data has been replicated, listed for both files and database – this will sometimes show zero
+    * If replication isn't running
+    * During the replication process when the process checks whether files are synchronized
+* The UUID of the target production environment
+* A button to **Stop replication** with an **Interrupt** option if the tool is currently replicating or a **Start replication** option if the tool is currently stopped
```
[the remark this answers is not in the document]
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/0212b6555d689a7bf9b6965e2f094d49a61fde28

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+*  The **Migrate button** which triggers the [final migration](#final-migration)
+
+    ![Status of the Migration](attachments/migrating-to-v4/migration-status.png)
```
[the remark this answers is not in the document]
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/ad675f5d3e842c01a04fce8935f296afa6a9af5a

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,233 @@ There are a few other differences between the way you develop and deploy apps in
...
+    The replication process will copy all the data in the database, including files based on `FileDocument` entities, such as images, which are stored in the storage of the v3 app node.
+
+    This is a one-time database copy from the v3 to v4 node, unlike live replication. This is recommended in the following cases:
+
+    * Database replication with AWS DMS fails to replicate the data correctly
+    * Database replication with AWS DMS fails with validation errors
 
-To migrate your app from a v3 node to a v4 node in the Mendix Cloud, follow the steps in the sections below.
+3. Once data transfer is complete, progress will be at 100%.
+
```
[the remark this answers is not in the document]
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/43138c59eb55eef3c83d406d9f856520663c742e

content/developerportal/deploy/migrating-to-v4.md
Adam-Dupaski: https://github.com/mendix/docs/pull/3377/commits/e6f292ba50876ddd5ef1ca6847c9bd7372de2dc8

Adam-Dupaski:
@MarkvanMents can pick up more updates from the 2nd.

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
igor-mendix: This I don't know :(

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,249 @@ There are a few other differences between the way you develop and deploy apps in
...
+    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the v3 appnode.
+
+    The replication will continue to run, so any changes to the data while your v3 app is still running are reflected in the replicated data.
+
+##### 4.3.1.2 Process with PostgreSQL Dump and Restore {#dump-restore}
+
+1. Follow steps 1-5 above in the [Process with AWS DMS](#aws-dms) section.
+
+2.  Select the **Force database dump restore** check box, then click **Replicate data and files**.
```
MarkvanMents, line 142: Does the v3 app need to be stopped for this to work, or can you do a dump when the v3 app is running?

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
MarkvanMents: Is the information displayed on the migration page the same for both AWS DMS and dump/restore?

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,249 @@ There are a few other differences between the way you develop and deploy apps in
...
 
-To migrate your app from a v3 node to a v4 node in the Mendix Cloud, follow the steps in the sections below.
+* The source and target environments
+* The UUID of the source production environment
+* Whether the tool is current **Replicating** data to the v4 environment or is **Stopped**
+* What proportion of the source data has been replicated, listed for both files and database – this will sometimes show zero
+    * If replication isn't running
+    * During the replication process when the process checks whether files are synchronized
+* The UUID of the target production environment
+* The **Stop replication** button with an **Interrupt** option if the tool is currently replicating
```
MarkvanMents, line 171: What is the difference between interrupt and not interrupt? Can we be clearer for the customer?
My assumption is:
1) Without interrupt, the stop button will not stop immediately but will continue until replication is 100%
2) For AWS DMS with interrupt, the stop button will stop the replication immediately, leaving the replication possibly incomplete and inconsistent
3) For dump/restore with interrupt, the incomplete dump will be thrown away.

Can you confirm or correct this?

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,249 @@ There are a few other differences between the way you develop and deploy apps in
...
+This will not copy any new data from the v4 environment back to the v3 environment. Any data added to the v4 database before the rollback is effectively lost.
+{{% /alert %}}
+
+### 4.7 Issues
+
+The following issues might occur, or you might decide to "Rollback" a successful migration.
+
+#### 4.7.1 Data Replication Fails or Times Out
+
+If the replication process fails during the final migration, or it times out (the timeout is fixed at 20 minutes), then the apps and environments will not be renamed and the apps in the original v3 environments will be restarted.
```
MarkvanMents, line 263: Is this only relevant for AWS DMS, or is there a 20 minute timeout on dump/restore as well?

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,249 @@ There are a few other differences between the way you develop and deploy apps in
...
+
+4. Select the source and target environments. In most cases, these will have the same name. For example, you want to replicate data from the v3 `production` environment to the v4 `production` environment. Please note you should not check the **Force database dump restore** check box.
+
+    ![Select v3 node and environment](attachments/migrating-to-v4/select-v3-node-to-migrate.png)
+
+5. Ensure that the app in the target v4 environment is stopped. You cannot replicate data into an app that is running.
+
+6. Click **Replicate data and files**.
+
+    The replication process will copy all the data in the database, and files based on `FileDocument` entities, such as images, which are stored in the storage of the v3 appnode.
```
MarkvanMents, line 134: Are you still happy with keeping this about `FileDocument` entities? There doesn't seem to be a resolution to this discussion (https://github.com/mendix/docs/pull/3377#discussion_r678380965).

MarkvanMents (commented):
@ravi4mendix
Hi Ravi,
I've now done a final proofread - the changes are in this commit: https://github.com/mendix/docs/pull/3377/commits/f1f277dbc2548e8db84cec2b07f3010f8e4c09af.
I have added a few questions, but we can publish this now and update the document later if customers have issues around these subjects.
Let me know whether changes are needed, or whether you want me to publish.
Thanks.

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: We can do the dump while it's still running. The only point when the app has to be stopped is final migration.

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
igor-mendix: It is pretty much the same. The only difference is that there's indication what mode (DMS or dump-restore) is used.

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: 1. Correct
2. Correct
3. The dump being taken on v3 is streamed directly to restore on v4, thus if it's interrupted only part of data will be present on v4. Which basically means the database is incomplete and unusable.

igor-mendix: The whole idea behind exposing the interrupt feature is to give a user the ability to correct mistakes. For example, if he selected wrong environments.

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: Oh my, we totally missed this. Thanks for catching it. We'll have to think how to better organize this on the backend.

igor-mendix: Right, now I remembered. In case of dump-restore mode, we were thinking to add measured time of the last dump-restore operation to these 20 minutes. This measured time is shown in the interface.

This is not implemented yet, but I'll do it today if there's no opposition.

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: I think it would be nice to make it a bit more explicit? How about the following?
```suggestion
* The **Stop replication** button with an **Interrupt** option if the tool is currently replicating
    * if you click **Stop replication** the replication will stop when all the data is replicated
    * if you check the **Interrupt** option, replication will stop immediately when you click **Stop replication** leaving the v4 database in an incomplete and probably inconsistent state
```

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: Thanks - don't need to add anything here then.

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
MarkvanMents: Great - no change needed there either.

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: So can I say:
```suggestion
If the replication process fails during the final migration, or it times out (the timeout is fixed at 20 minutes for AWS DMS — for dump restore it is 20 minutes longer than the duration of the previous replication), then the apps and environments will not be renamed and the apps in the original v3 environments will be restarted.
```

igor-mendix: "duration of previous replication" may not be right away clear to the customer. I'd say something like "for dump-restore it is set to 20 minutes plus measured operation time, which is displayed in the UI".

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: "if you check the **Interrupt** option, replication will stop immediately" - I'd say "replication will terminate running operation as soon as possible". It's not exactly immediately because it'd still do some cleanup before marking it fully stopped.

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: Have updated the suggestion above ☝️

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: Have updated the suggestion above ☝️

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
igor-mendix: It seems we're mixing up stuff here. So if we have `FileDocument` entities in the database (which are basically files, but stored in the database if I understand correctly) then we can mention that. But I don't think this is necessary because database is transferred fully. Why should we emphasize one specific record type? This only complicates things.

In addition to `FileDocument` entities that reside on the dbnode (and are transferred to RDS), there are also ordinary files that reside on the appnode (and are transferred to S3). But we don't mention that. Neither in this paragraph, nor in DMS-related list.

This is why my proposal to simplify this to having only "...copy all files and data in the database" stands.

content/developerportal/deploy/migrating-to-v4.md
igor-mendix: I've added a comment in the original thread. @ravi4mendix what do you think?

content/developerportal/deploy/migrating-to-v4.md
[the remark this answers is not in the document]
igor-mendix: Or maybe better to say "...copy all file and database records".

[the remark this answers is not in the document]
MarkvanMents: OK,
The thing about a `FileDocument` entity is that it is just a pointer to a file on disk. So when the app wants the content, it actually pulls it from a file, rather than from the table.
But I suppose the customer could have done something else which created files which are _not_ FileDocuments, so I'll make it more general.

content/developerportal/deploy/migrating-to-v4.md
MarkvanMents: [Repeat of my other response]
The thing about a `FileDocument` entity is that it is just a pointer to a file on disk. So when the app wants the content, it actually pulls it from a file, rather than from the table.
But I suppose the customer could have done something else which created files which are _not_ FileDocuments, so I'll make it more general.
How about:
```suggestion
    The replication process will copy all the data in the database as well as all the files which are stored on the v3 appnode.
```

MarkvanMents: Discussion moved to here: https://github.com/mendix/docs/pull/3377#discussion_r681749667

content/developerportal/deploy/migrating-to-v4.md
```diff
@@ -55,13 +46,251 @@ There are a few other differences between the way you develop and deploy apps in
...
+* The source and target environments
+* The UUID of the source production environment
+* Whether the tool is current **Replicating** data to the v4 environment or is **Stopped**
+* What proportion of the source data has been replicated, listed for both files and database – this will sometimes show zero
+    * If replication isn't running
+    * During the replication process when the process checks whether files are synchronized
+* The UUID of the target production environment
+* The **Stop replication** button with an **Interrupt** option if the tool is currently replicating
+    * if you click **Stop replication** without checking the **Interrupt** option, the replication will stop when all the data is replicated
+    * if you check the **Interrupt** option, replication will stop as soon as possible when you click **Stop replication** leaving the v4 database in an incomplete and probably inconsistent state
```
ravi4mendix, line 173: Can we add one more piece of information here
```
if you check the **Interrupt** option, replication will stop as soon as possible when you click **Stop replication** leaving the v4 database in an incomplete and probably inconsistent state. In the UI you will see the state as failed when you select this option.
```

ravi4mendix: Thanks for fixing this. Let's publish it now. We can always improve it. :)

PR #3383, opened 2021-05-27
NBUI release notes

smarki:
The version number is only known the moment we successfully build the release version. I will update it tomorrow before when the release is live.


ConnorLand:
> The version number is only known the moment we successfully build the release version. I will update it tomorrow before when the release is live.

Understood, will target the 28th for merging this PR.

ConnorLand (approved):

PR #3384, opened 2021-05-27
Specified the Windows JDK Version

GoPrice:
Specified the Windows JDK version that should be used when signing android applications. It is small details like this that can save a lot of time.

### DIFF
--- a/content/refguide/managing-app-signing-keys.md
+++ b/content/refguide/managing-app-signing-keys.md
@@ -74,7 +74,7 @@ Android apps can be developed and deployed to Android devices without signing th
 
 To generate a keystore for Android, follow these steps:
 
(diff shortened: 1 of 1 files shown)


feddovanede (commented):
I have the keytool.exe file in my Java 11 SDK folder as well? Does it really have to be Java 8?

smarki:
Android Studio 4 comes bundled with JDK 11. It's relative to the AS installation folder. So I am pretty sure 11 works fine. 
Also if you have Studio Pro installed you get openjdk 11 installed with it.

ConnorLand:
Thanks for your pull request @GoPrice . Because options beyond **JDK 1.8.0_144** are valid, I am going to close this PR. 

Thank you for your contribution to the documentation.

PR #3399, opened 2021-06-01
Mapping redirects outdated, use `aliases` feature instead. 

r0b1n:
Remove files and associated gulp task.


MarkvanMents:
Hi @NatasaKralj ,
Added you to this because it is affecting the stack. Don't expect you to necessarily understand it. We can discuss it tomorrow.

MarkvanMents (approved):
Thanks, Roman @r0b1n 
I'll merge it in the evening when no one else is using the repo, just in case.
Would have taken me ages to work out what to do with gulp, so you've saved me a lot of time.
Thanks.
Mark

r0b1n:
@MarkvanMents is that mapping for assets (css and js files) still being used? Maybe we could get rid of that part also?

MarkvanMents:
@r0b1n I'll have to have a look. I'm still learning what is in the stack and trying to work out exactly what we need.
yarn.lock has 600+ dependencies in it, for example, but I don't know whether that is a lot or a little for what we do.
I think it may still be being used as I haven't changed anything around css or JS.

Adam-Dupaski:
@MarkvanMents please verify the redirects file is backed up in Dropbox just for reference

MarkvanMents:
> 
> 
> @MarkvanMents please verify the redirects file is backed up in Dropbox just for reference

Done - in Team folder/Projects/Docs Stack/Files taken offline

MarkvanMents:
Hi Roman, @r0b1n 
Not sure exactly what you mean by the **mapping for the assets** - only just started on the journey to understand what is going on.
However, the files in the **js** and **styles** folders of *_assets* are still being used, as far as I can see, so we can't remove them if that was what you meant.

r0b1n:
@MarkvanMents there is a [section for assets](https://github.com/mendix/docs/blob/development/static/nginx.conf#L45-L48) similar to the page redirects section we removed in this PR. If the page redirect section is not working, then I assume the assets redirect section is also not working. Then there is no need to keep it and create `mappings/assets.map` file. Assets themselves should remain in place of course.

MarkvanMents:
Thanks for doing this for me, Roman (@r0b1n)
It really helps to start clearing out the things which we don't need, so we can concentrate on what we do need.

PR #3407, opened 2021-06-07
WDSL, REST and ODATA endpoints are not registered unless correspondin…

dfedorenko:
…g services exist

### DIFF
--- a/content/refguide/runtime.md
+++ b/content/refguide/runtime.md
@@ -15,7 +15,7 @@ You can extend the functionality of the runtime by writing Java actions. The API
 ## Integration API documentation
 
 {{% alert type="info" %}}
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi Dmytro, @dfedorenko 
1. Is this for Mx9.3? There is a related PR, #3409, which is flagged for 9.3 only.
2. Could you check whether you have signed the cla (see end of this page) as this will set the status of the PR to green.

MarkvanMents:
Hi @ilagordhan 
This seems to be related to integration too. Does it look OK?

MarkvanMents:
@joostverhoog can you confirm that this is for Mx9.3?
I also note that the example URL uses localhost, were the example in #3409 uses domain name. Would it be better to use domain name here as well, perhaps the same example as in #3409 ?

joostverhoog:
This pull request can be closed. #3409 covers all of it

MarkvanMents:
Thanks for sorting this out.

PR #3409, opened 2021-06-07
The api-doc endpoint is only available when there is at least one service

joostverhoog:
In Mendix 9.3.0, the api-doc endpoint is only available when there is at least one service

### DIFF
--- a/content/refguide/runtime.md
+++ b/content/refguide/runtime.md
@@ -87,7 +87,7 @@ You need a license to run an application in production mode. Without a license, 
 You can extend the functionality of the Runtime Server by writing Java actions. For more information,  see the [Runtime API](/apidocs-mxsdk/apidocs/#runtime) section of *API Documentation*.
 
 {{% alert type="info" %}}
-Links to available API documentation such as WSDLs for published web services are available on the URL path `/api-doc` (for example: `http://localhost:8080/api-doc/`).
+If the app contains published services, links to available API documentation such as [OpenAPI documentation](open-api) for [published REST services](published-rest-services), links to [published OData services](published-odata-services), and WSDLs for [published web services](published-web-services), are available on the URL path `/api-doc` (for example: `https://myapp.mendixcloud.com/api-doc/`).
 {{% /alert %}}
 
 ## 5 Main Documents in This Category


MarkvanMents:
@ilagordhan Does this look OK to you?
Looks OK as far as I can tell.

content/refguide/runtime.md
```diff
@@ -87,7 +87,7 @@ You need a license to run an application in production mode. Without a license,
 You can extend the functionality of the Runtime Server by writing Java actions. For more information,  see the [Runtime API](/apidocs-mxsdk/apidocs/#runtime) section of *API Documentation*.
 
 {{% alert type="info" %}}
-Links to available API documentation such as WSDLs for published web services are available on the URL path `/api-doc` (for example: `http://localhost:8080/api-doc/`).
+If the app contains published services, links to available API documentation such as [OpenAPI documentation](open-api) for [published REST services](published-rest-services), links to [published OData services](published-odata-services), and WSDLs for [published web services](published-web-services), are available on the URL path `/api-doc` (for example: `https://myapp.mendixcloud.com/api-doc/`).
```
ilagordhan, line 90:
```suggestion
If the app contains published services, links to available API documentation such as [OpenAPI documentation](open-api) for [published REST services](published-rest-services), links to [published OData services](published-odata-services), and WSDLs for [published web services](published-web-services), are available at [API Documentation](/apidocs-mxsdk/apidocs/).
```

ilagordhan: The link that was inserted as an example does not exist, and if I understand this correctly, then users should be referred to the API docs which lead them to the relevant APIs

ilagordhan (commented):
The given link does not work and users should be referred to the the API docs as suggested.

content/refguide/runtime.md
MarkvanMents: No,
We are talking about the api documentation for the service published by the app.
The user of the api (which might be Data Hub, for example) can read the documentation of the api by going to the URL `{applicationURL}/api-doc/` 

This will only exist for a valid application URL, but this is an example URL (myapp.mendixcloud.com)

ilagordhan: Thanks - I missed the "for example"
```suggestion
If the app contains published services, links to available API documentation such as [OpenAPI documentation](open-api) for [published REST services](published-rest-services), [published OData services](published-odata-services), and WSDLs for [published web services](published-web-services), are available on the URL path `/api-doc` (for example: `https://myapp.mendixcloud.com/api-doc/`).
```

ilagordhan: The suggestion looks good to me!

ilagordhan (approved):

MarkvanMents:
Thanks @ilagordhan and @joostverhoog .
Marked for publication with 9.3.

PR #3419, opened 2021-06-09
Fix broken links

Luyao-Zhang-1:
The previous URL (https://developer.mindsphere.io/apis/advanced-assetmanagement/api-assetmanagement-api-swagger-3-9-0.html) contains the version of the API. The version now is 3.15.0, so the URL is also changed accordingly. To prevent the links from becoming outdated again in the future, it is better to direct the user to the **Available API Versions**, and then user can just choose the available API version for **Europe 1**.


MarkvanMents:
@Luyao-Zhang-1 - good spot.
I see that the first reference to this, on line 22, has already been changed.
Because this is MindSphere documentation, I'll just get someone from MindSphere to check that this is the best change.

MarkvanMents (approved):
Hi Luyao (@Luyao-Zhang-1 )
I didn't get a response from the MindSphere people, so I'll just approve this and merge it anyway.
Thanks for picking this up.

MarkvanMents:
I'll leave you to delete the branch, rather than doing it for you.

PR #3420, opened 2021-06-09
Remove Old Search Fields

MarkvanMents:
New MxDock has search built in.
This had the consequence that attempting to search using the old search fields triggered the MxDock search which moved the focus unexpectedly and was annoying for the users.
If you performed a search before MxDock was fully loaded, the old search function was triggered, giving a different UX for search.

This PR:
1. Removes search fields in the body of the pages on docs.mendix.com
2. Removes the references to the old search function at search-static.mendix.com


MarkvanMents:
Tested locally with @NatasaKralj

PR #3311, opened 2021-05-06
Update backups-api.md

RCatersels:
JSON example was different then you actually get.

PR #2315, opened 2020-03-03
Backups api v2

xiwenc:
Earlier this week the new Backups API V2 was released. Now we are updating the public documentation to encourage users to use V2.

@MarkvanMents could you review this?

PR #3423, opened 2021-06-10
Clarify backup API

MarkvanMents:
Customers were confused about the use of V2 of the backup API.
Updating terminology to match the API, and introducing the process to help the customer.

### DIFF
--- a/content/apidocs-mxsdk/apidocs/backups-api.md
+++ b/content/apidocs-mxsdk/apidocs/backups-api.md
@@ -7,7 +7,21 @@ menu_order: 10
 
 ## 1 Introduction
 
-The Backups API V2 allows you to manage data snapshots of applications hosted in the Mendix Cloud V4. Data snapshots consist of a Postgresql database dump and file objects referenced from the database. You can create new snapshots and restore or download them. Uploading snapshots is currently only supported via the [Developer Portal](/developerportal/operate/backups). Unlike the [older V1 API](backups-api-v1), this new V2 API is focused on asynchronous operations of long-running tasks.
+The **Backups API V2** allows you to manage backups of the data in your app hosted in the Mendix Cloud V4.
+
+If you want to download a backup of your data, you need to perform three steps:
+
+1. Create a *snapshot*, or find the `snapshot-id` of an existing one
+2. Create an *archive* file from a snapshot
+3. Download the archive using the URL provided 
+
+Data **snapshots** consist of a PostgreSQL database dump and all the file objects referenced from the database.
+
+Database **archives** are a zip file which contains all the data in the snapshot, or the database and files separately if you prefer. 
+
+You cannot currently upload an archive through this API. This function is currently only supported via the [Developer Portal](/developerportal/operate/backups). However, you can use this API to restore data from an existing environment snapshot.
+
+V2 of this API is focused on working with snapshots and archives asynchronously, as these can be very long-running tasks for large quantities of data. The [older V1 API](backups-api-v1), in contrast, works synchronously. 
 
 {{% alert type="info" %}}
 This article is only applicable to applications deployed in **Mendix Cloud V4**. You can check which version of the Mendix Cloud you are using in the [Developer Portal](/developerportal/deploy/environments-details).
@@ -17,14 +31,13 @@ This article is only applicable to applications deployed in **Mendix Cloud V4**.
 
 The Backups API requires authentication via API keys that are bound to your Mendix account (for more information, see [Deploy Authentication](deploy-api#authentication)). In addition to the **API Access** permission, the **Backups** permission is also required to manage backups.
 
-
 ## 3 API Calls
 
-### 3.1 List Environment Backups
+### 3.1 List Environment Snapshots
 
 #### 3.1.1 Description
 
-Lists the backups of an environments. By setting the `offset` parameter you can page through the list of backups belonging to an environment.
+Lists the snapshots of an environments. By setting the `offset` parameter, you can page through the list of snapshots created for an environment.
 
 ```bash
 HTTP Method: GET
@@ -59,7 +72,7 @@ Mendix-ApiKey:  26587896-1cef-4483-accf-ad304e2673d6
 An object with the following key-value pairs:
 
 *   _snapshots_ (List): List of snapshot objects.
-*   _total_ (Long): There are in total this amount of snapshots for the requested environment.
+*   _total_ (Long): The total number of snapshots for the requested environment.
 *   _offset_ (Long): The offset value of the current request.
 *   _limit_ (Long): The limit value of the current request.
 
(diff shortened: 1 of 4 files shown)


content/developerportal/operate/backups.md
```diff
@@ -92,14 +92,14 @@ You can download one of the following:
 See [Download a Backup](download-backup) for more information.
 
 {{% alert type="info" %}}
-As the download files are generated "on the fly" (meaning, while in progress), it is not possible to estimate the file size before downloading. Your browser will not show a progress bar.
+As the download archive is generated "on the fly" (meaning, while in progress), it is not possible to estimate the file size before downloading. Your browser will not show a progress bar.
```
joshp123, line 95: This isn't 100% accurate actually. When we start creating the archive, we don't know how big it is. After it's been created, however, we do. (I don't think that we display it, however).

(The method of creating the archive is the same as when we use the API, so we request an archive to be created, and then when it's finally finished, we show the URL to the customer).

content/apidocs-mxsdk/apidocs/backups-api.md
```diff
@@ -241,34 +256,32 @@ An object with the following key-value pairs:
...
-   "url":"https://..."
+   "status_message":"Completed backup creation",
+   "model_version":"1.0.0.7",
+   "expires_at":"2020-05-18T16:00:18.000Z",
+   "finished_at":"2020-02-18T16:00:19.000Z",
+   "updated_at":"2020-02-18T16:00:19.000Z",
+   "snapshot_id":"51dc7872-771e-4c3e-853b-352359444db6",
+   "created_at":"2020-02-18T16:00:18.000Z",
+   "comment":"Manually created snapshot",
+   "state":"completed"
```
udh-mendix, line 272: the API bits are confusing to me, did you verify the API responses?

MarkvanMents: No, I haven't actually used the API, all this information came from the original PR #2315 which was created by the developers. I assumed that this had been tested.

MarkvanMents: These only look like they have changed because I changed the order or sections 3.2-3.5 so that creating the snapshot came before creating the archive, and Git sees them as a change rather than a move.

content/apidocs-mxsdk/apidocs/backups-api.md
```diff
@@ -641,7 +653,7 @@ URL: https://deploy.mendix.com/api/v2/apps/<ProjectId>/environments/<Environment
 
 *   _ProjectId_ (String): Unique project identifier. Can be looked up via the [developer portal](/developerportal/deploy/environments-details) or [apps API](deploy-api#list-apps).
 *   _EnvironmentId_ (String): Unique environment identifier. Can be looked up via the [developer portal](/developerportal/deploy/environments-details) or [environments API](deploy-api#list-environments).
-*   _SnapshotId_ (String): Identifier of the backup.
+*   _RestoreId_ (String): Identifier the request to restore the data.
```
MarkvanMents, line 656: This is the only API entry I changed - for this call you need the restore id, and that is also correct in the example.

content/apidocs-mxsdk/apidocs/backups-api.md
MarkvanMents: I'm not sure that the responses are right, though, as we have a PR from a customer pointing out at least one difference: #3311

udh-mendix: thats pending on verification. I'll file a BAK ticket for that so it can be picked up soon.

content/developerportal/operate/backups.md
udh-mendix: We do display the size of backup in CP  after its created. I'm fine with this rename.

udh-mendix:
LGTM.

content/developerportal/operate/backups.md
joshp123: My comment is not about the rename, it's about the overall statement.

I would suggest something like:
`As the download archive is generated on request, it's not possible to estimate the file size before requesting a download`

However, this isn't the most important thing.

MarkvanMents:
@joshp123 and @udh-mendix 
Hi Josh and Umang,
This PR is still around - I hope I've picked up all the terminology issues and explained things a bit better to the customer.
I see that Umang has added a LGTM - sorry I didn't follow up on that earlier. Can you confirm that this can be published, or is there something else needed?
Thanks
Mark

content/developerportal/operate/backups.md
joshp123: I just checked, this isn't accurate, we do show a progress bar when downloading backups:

image

joshp123:
@MarkvanMents - Ali will be picking this up instead of Umang now. He'll double check the outstanding comments here.

MarkvanMents:
@joshp123 Thanks for picking this up.

aliozcan:
Mark 👋 , I'll start with verifying outputs. Will add comments afterward.

content/apidocs-mxsdk/apidocs/backups-api.md
MarkvanMents: Does this need to be checked before we publish, or can we change it afterwards?

content/developerportal/operate/backups.md
MarkvanMents:
```suggestion
As the download archive is generated on request, it is not possible to estimate the file size before requesting a download.
```
I'll change it to this then Josh, @joshp123

MarkvanMents (commented):
@aliozcan,
Do you know when you will be able to review this?
I'd quite like to get this sorted and published, even if we have to make changes later.
I think it is already an improvement on what is already in the documentation.
I've suggested one change below in response to @joshp123.
Let me know what you want to do - I have some leave coming up from Friday so if we don't publish this soon you will have to wait another fortnight.
Thanks
Mark (@MarkvanMents )

content/apidocs-mxsdk/apidocs/backups-api.md
aliozcan: Hey Mark, I've managed to get API responses. I can verify it quite quickly. However, this part is not clear to me.
You mention reordering sections 3.2 and 3.5 but this change corresponds to section 3.3 (Check download job of a Backup for an Environment) I guess, and the output for section 3.3 is correct.

MarkvanMents: I reordered all the sections between 3.2 and 3.5 to put the APIs in the order the customer needs to call them. So that will affect 3.3 as well.
Does that solve the issue?

aliozcan: Hey Mark, I can check the API responses quickly. However, this change seems quite big so I think we can change it afterwards.

aliozcan:
Hey Mark, I left a comment regarding the output verification. That change is huge and we can alter it later. As for the file size changes, your suggestion looks good to me.

content/apidocs-mxsdk/apidocs/backups-api.md
aliozcan: These changes are quite big so I think we can change it afterwards as you said.

MarkvanMents (commented):
@aliozcan 
Shall I merge this then and you will pick up a full review of the output later.
I think that this change will improve/clarify the existing documentation so it is worth getting it out there.
Thanks.

MarkvanMents:
Agreed with Ali over Slack. Merging.

PR #3425, opened 2021-06-11
Add Module-level theme settings

MehdiGolchin:
Add a new page for Module-level theme settings.

### DIFF
--- a/content/howto/front-end/module-level-theme-settings.md
+++ b/content/howto/front-end/module-level-theme-settings.md
@@ -9,9 +9,11 @@ tags: ["build", "app", "studio", "studio pro", "style", "styling"]
 
 ## 1 Introduction
 
-Atlas UI gives you a powerful tool to build a modern user experience. However, the flexibility of Mendix does not stop you from making your design system from scratch. This document will introduce you to the advanced options that module-level theme settings provide. If you wish for further guidance on front-end best practices, see [Implement Best Practices for UX Design](ux-best-practices).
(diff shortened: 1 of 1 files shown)


content/howto/front-end/module-level-theme-settings.md
```diff
@@ -0,0 +1,60 @@
+---
+title: "Module-level theme settings"
```
ConnorLand, line 2: Hi @MehdiGolchin . Because this document is describing information more than instructing a course of action, I think it belongs more in Reference Guide than How-To's. Could you please do the following?

1. Move the document into the refguide folder? 
2. Give it the correct parent document. 
* If this styling information applies *only* to native apps, I suggest "native-mobile) as the parent doc.
* If it applies to all *mobile* apps, I suggest "mobile". 
* If it applies to all Studio Pro apps, I am not sure if there is an ideal home for it -- let me know if this is the case and I'll talk to my colleagues.

content/howto/front-end/module-level-theme-settings.md
```diff
@@ -0,0 +1,60 @@
...
+{{% alert type="info" %}}
+Please note that by excluding a UI element, you can no longer use it across your project.
+{{% /alert %}}
+
+### 2.1 Available Properties
+
+
+| UI Element | Description |
+| ------------------------------------ | ----------- |
+| `layouts` | `true` to exclude all the layouts of a specific module, default `false`. |
```
ConnorLand, line 37: @MehdiGolchin I asked Chris Hodges if there were any additions we could make to this doc from a product direction perspective. He offered the idea of "examples are needed for the 5 available properties, as well as a final one which shows a combination - section 2.2"

Could you try adding those code samples please?

ConnorLand (approved):

PR #3443, opened 2021-06-16
Change absolute references to relative ones (TW-1344)

Luyao-Zhang-1:


CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
 For guidelines on contributing to the Mendix documentation, please see these two documents:
 
-* [How to Contribute to the Mendix Documentation](https://docs.mendix.com/developerportal/community-tools/contribute-to-the-mendix-documentation)
-* [Documentation Writing Guidelines](https://docs.mendix.com/developerportal/community-tools/documentation-writing-guidelines)
\ No newline at end of file
+* [How to Contribute to the Mendix Documentation](content/developerportal/community-tools/contribute-to-the-mendix-documentation)
```
MarkvanMents, line 3: Hi @Luyao-Zhang-1 
The links should not contain "content". This is the folder where the original markdown is kept within GitHub.
When the site is built, the HTML generated from the content folder is put in the **root** of the site.
So here, the URL is `https://docs.mendix.com/developerportal/community-tools/contribute-to-the-mendix-documentation` and the relative link is `/developerportal/community-tools/contribute-to-the-mendix-documentation`.
In other words, if I am looking at a page `https://docs.mendix.com/refguide/something` I can go to a file in the same folder by just using the file name `(something-else)`, but to go to a different space, like developerportal, I need to go back to the root of the site `(/developerportal/something-new)`. But I don't need to put the whole 'https://docs.mendix.com' because then if I'm looking at a different version (development build or, in the future, a localised build) it will always take me to the *production* site.
```suggestion
* [How to Contribute to the Mendix Documentation](/developerportal/community-tools/contribute-to-the-mendix-documentation)
```

Luyao-Zhang-1: @MarkvanMents and I talked about this. We see that the full references should be kept in this file because this file will not be published on our docs site.

Luyao-Zhang-1: We should avoid using `../` and `../../` in the cross-references, because this can lead to problems if we ever change the current folder structure. In the commit #b84dcad, I removed all the `../` and `../../` in the cross-references.

MarkvanMents (changes requested):
Hi @Luyao-Zhang-1 
Looks good. just a couple of things on the two documents which are for the GitHub repo rather than our site.
Adam is assigned to this, so I'll let him do the final review.

Adam-Dupaski:
Reviewed, merging

PR #3447, opened 2021-06-17
Clean up numbered links and anchors

NatasaKralj:
- The vast majority of changed links were of `{#X-name-of-heading}`, with X being the heading number. Those have been cleared of number references.
- There were two instances of full URLs used, instead of a relative cross-reference (that I found), those have also been fixed.
- There were two instances of a heading with `<a name="DeployToNextEnvironment"></a>` before the heading name, fixed to the standard with no number reference. This seemed safe to do, as they were headings.
- One numbered cross-reference was left, as it had a note it was mapped `#5-amazon-s3-storage-service-settings`


Adam-Dupaski:
@NatasaKralj changes look good, but there are a few conflicts to fix against dev. Can you resolve these above? Most seem to be just selecting the anchor/link option on your branch (the one that should be published). Let me know if you have questions.

NatasaKralj:
@Adam-Dupaski I think I got them, using the resolve conflict on GitHub.

PR #3457, opened 2021-06-21
Clarify redeployment of app

feddovanede:
Added additional clarification that redeployment is necessary the first time you set up Datadog for your app.


MarkvanMents (approved):
Thanks @feddovanede

PR #3458, opened 2021-06-21
Add warnings to deprecated components in Marketplace

Luyao-Zhang-1:


content/appstore/widgets/radio-button-list.md
```diff
@@ -6,6 +6,10 @@ tags: ["marketplace", "marketplace component", "widget", "radio button list", "p
 #If moving or renaming this doc file, implement a temporary redirect and let the respective team know they should update the URL in the product. See Mapping to Products for more details.
 ---
 
+{{% alert type="warning" %}}
+The [Attribute Radio Button list](#attribute-radio-button-list) widget is deprecated. The [Association Radio Button list](#association-radio-button-list) widget is still supported.
```
Adam-Dupaski, line 10: Can you standardize the recommendation part as compared with `google-maps.md` for consistency?

Luyao-Zhang-1: This situation is a bit different. With the Radio Button List widget, previously you could make an Attribute Radio Button List or an Association Radio Button list - now the former is deprecated and you can still use the latter, while the Google Maps module is completely deprecated, and you can use the Google Maps widget as a substitute.
Maybe I change the text to this: "This widget is partly deprecated - the Radio Button List is deprecated, but the Association Radio Button is still supported."?

Adam-Dupaski: Thanks for the explanation. Would it be best to call attribute and association "types" of radio button lists? In that case, you could say: "This widget is partly deprecated - the Attribute Radio Button List type is deprecated, but the Association Radio Button type is still supported." ?

Luyao-Zhang-1: Thanks. It is a good idea.

Adam-Dupaski (approved):

Luyao-Zhang-1:
Done in [PR #3458](https://github.com/mendix/docs/pull/3458#pullrequestreview-689243337).

PR #3475, opened 2021-06-24
Remove show truncated events from calendar docs

diego-antonelli:


ConnorLand (approved):

ConnorLand:
I confirmed this is the only instance of "truncated events" in docs, so this deletion looks good.

PR #3491, opened 2021-06-30
Change real names to general code

Luyao-Zhang-1:


MarkvanMents:
Hi Luyao,
As discussed, I would consider using fake emails with the correct format. The example is supposed to look like a real JSON payload and it is useful to the customer to see exactly what it looks like, with the correct format for the email addresses.

Adam-Dupaski (approved):

PR #3501, opened 2021-07-02
Adding some trouble-shooting information

TimvanSteenbergen:
Adding some trouble-shooting information for rollbacks, since this is always (unless a restart happened) an indication of a problem, sometimes a big problem. In such cases it is nice to have a helpful documentation.

### DIFF
--- a/content/developerportal/operate/trends-v4.md
+++ b/content/developerportal/operate/trends-v4.md
@@ -293,6 +293,9 @@ Value | Explanation
 **xact updated** | Number of tuples updated per second.
 **xact deleted** | Number of tuples deleted per second.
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Assistance requested in Slack to review this.

MarkvanMents (approved):
Hi Tim (@TimvanSteenbergen )
The development team thought that some of your diagnosis was a bit too specific to one particular case. I have therefore kept the general advice on where to look for issues but removed the specific guidance about the Community Commons module.

Thanks very much for highlighting where we can help our customers focus on solving problems. I'm sure this will help lots of people who find high numbers of rollbacks during normal operation.
Mark

PR #3508, opened 2021-07-05
adding the word favicon

TimvanSteenbergen:
Since that is what people are looking for when they want/need to change the favicon. See https://forum.mendix.com/link/questions/108107


MarkvanMents (approved):
Thanks for this change, Tim (@TimvanSteenbergen )
Yes we should definitely use the term customers are looking for. Thanks for taking the trouble to make a PR.
Approving.

PR #3517, opened 2021-07-09
Add Workflow Commons module to the Marketplace Guide

MariaShaposhnikova:


content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,89 @@
...
+    2. User role does the following:
+        1. Executes workflows by completing user tasks.
+        2. Views personal performance in MyTaskDashboard.
+
+## 4 Usage
+
+### 4.1 Page Templates
+
+#### 4.1.1 My Initiated Workflows
+Allows users who initiate workflows to keep track of their progress. A user is also able to `Withdraw` a workflow if it is still `In progress`.
```
MariaShaposhnikova, line 79: Is **Withdraw** a button?

content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,89 @@
...
+
+#### 4.1.2 Workflow Task Dashboard
+Shows users who execute tasks, for example, a task-based dashboard.
```
MariaShaposhnikova, line 82: Is this page template only for users who _execute_ task, can it be an admin who just _monitors_ them?

content/appstore/modules/workflow-commons.md
MitchelMx: Withdraw is indeed a button, but you could also consider it an action to be executed on an `In progress` workflow.

content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,89 @@
...
+
+The [Workflow Commons](https://marketplace.mendix.com/link/component/117066) module provides an out-of-the box starter experience for users who would like to build a workflow in their app.
+
+### 1.1 Typical Use Cases
+
+This module allows Mendix developers with little or no experience in building workflows to start with standardized options for implementing and monitoring workflows.
+
+### 1.2 Features
+
+* Dashboards for monitoring workflows, for administrators, initiators, and executors of workflows
```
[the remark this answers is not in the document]
MitchelMx: Executors are any users that have the ability to chose an outcome/complete a user task.

content/appstore/modules/workflow-commons.md
MitchelMx: Admins have the Workflow Dashboard, Workflow Task Dashboard is specifically for executing users.

content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,89 @@
...
+3. Open the **MendixSSOUserVariant** folder.
+   4. Select all pages and the microflow in this folder (select multiple items by selecting the first item and then keep <kbd>Ctrl</kbd> pressed to select other items).
+   2. Right-click and choose **Exclude from project**.
+4. Open the **SystemUserVariant** folder.
+   7. Select all pages and the microflow in this folder.
+   2. Right-click and choose **Include in project**.
+
+## 3 Configuration
+
+1. Add the following pages to your navigation:
```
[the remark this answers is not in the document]
MitchelMx: Yes these are all pages that the module provides ready to use.

content/appstore/modules/workflow-commons.md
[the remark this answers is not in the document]
MariaShaposhnikova: I will add this explanation to the document.

content/appstore/modules/workflow-commons.md
MariaShaposhnikova: Thanks, this helps to choose styling, I don't think we need code styling here in this case.

content/appstore/modules/workflow-commons.md
MariaShaposhnikova: Thanks, this helps to choose styling, I don't think we need code styling here in this case.

content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,94 @@
+---
+title: "Workflow Commons"
+category: "Modules"
+description: "Describes usage and implementation the Workflow Commons module that is used with workflows."
```
PaulJongenMx, line 4: "implementation the" I think this does not run correctly. Expect something like: "implementation of the"

content/appstore/modules/workflow-commons.md
```diff
@@ -0,0 +1,94 @@
...
+### 1.2 Features
+
+* Dashboards for monitoring workflows for administrators, initiators, and executors (users who can chose an outcome/complete a user task) of workflows
+* Page templates to generate task pages, task admin, an inbox, and different dashboards
+* Attachments layer on top of workflows
+* Comments section to use with workflows
+
+### 1.3 Limitations
+
+* Requires [Mendix SSO](https://marketplace.mendix.com/link/component/117212) to fully function directly after the installation. If you do not want to use Mendix SSO, you will have to copy contents to your own module and use templates to build pages.
```
PaulJongenMx, line 25: This section can be removed (because it's explained in 2.1 how to use WFC without MxSSO).

MariaShaposhnikova: We have this section in our template, but I think we can skip this indeed, as it is explained in Dependencies.

PR #3520, opened 2021-07-12
Update administrator.md

feddovanede:
Add description of data restore process

### DIFF
--- a/content/refguide/administrator.md
+++ b/content/refguide/administrator.md
@@ -38,12 +38,25 @@ Default: *1*
 {{% alert type="info" %}}
 Since the value of the password is general knowledge, it is safer to change this to a custom password.
 
-This password is only used when Mendix is running locally. You can change the password for your other licensed [environments](/developerportal/deploy/environments-details) in the Developer Portal.
+This password is only used when Mendix is running locally. Changing the password in your model will not update the password in your Cloud environments. You can change the password for your other licensed [environments](/developerportal/deploy/environments-details) in the Developer Portal.automatically when the 
 {{% /alert %}}
 
 #### 2.2.1 Free Apps
 
-You cannot set a password for MxAdmin for Free Apps published to the cloud. To log in for the first time to a Free App deployed to the cloud, do the following:
+The MxAdmin user is not created automatically when you deploy your app. For licensed environments the MxAdmin user is created when you change the password for the first time.
+
+When your Free App has never been deployed, and the database still needs to be created, the data snapshot in your project will be restored to the database of your Free App, if a data snapshot is present. You can use this process to add the MxAdmin to your Free App by doing the following:
+1. In Studio Pro, go to **App > Security**.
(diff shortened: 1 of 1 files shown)


content/refguide/administrator.md
```diff
@@ -38,12 +38,25 @@ Default: *1*
...
 Since the value of the password is general knowledge, it is safer to change this to a custom password.
 
-This password is only used when Mendix is running locally. You can change the password for your other licensed [environments](/developerportal/deploy/environments-details) in the Developer Portal.
+This password is only used when Mendix is running locally. Changing the password in your model will not update the password in your Cloud environments. You can change the password for your other licensed [environments](/developerportal/deploy/environments-details) in the Developer Portal.automatically when the 
 {{% /alert %}}
 
 #### 2.2.1 Free Apps
 
-You cannot set a password for MxAdmin for Free Apps published to the cloud. To log in for the first time to a Free App deployed to the cloud, do the following:
+The MxAdmin user is not created automatically when you deploy your app. For licensed environments the MxAdmin user is created when you change the password for the first time.
```
MarkvanMents, line 46:
```suggestion
The MxAdmin user is not created automatically when you deploy your app as a Free App. For licensed environments the MxAdmin user is created when you change the password for the first time, for example through the [Environment Details](/developerportal/deploy/environments-details) for the Mendix Cloud.
```
@feddovanede , is this a difference between licensed and Free Apps which we should mention here: https://docs.mendix.com/developerportal/deploy/mendix-cloud-deploy#free-app?

content/refguide/administrator.md
```diff
@@ -38,12 +38,25 @@ Default: *1*
...
+
+When your Free App has never been deployed, and the database still needs to be created, the data snapshot in your project will be restored to the database of your Free App, if a data snapshot is present. You can use this process to add the MxAdmin to your Free App by doing the following:
```
MarkvanMents, line 48:
```suggestion
When your Free App *has never been deployed, and the database still needs to be created*, any data snapshot you have added to your project will be restored to the database of your Free App. You can use this process to add the MxAdmin user to your Free App by doing the following:
```

MarkvanMents (changes requested):
Good suggestion, Feddo (@feddovanede ).
I've made a couple of suggestions - you can commit them if you agree.
I also had a question about whether we should highlight this as another difference on our list of differences between free and licensed apps.
I can publish this once we have agreed what to do with my changes.

content/refguide/administrator.md
feddovanede: We could. But the MxAdmin user is also not created automatically when you deploy to a licensed node. The only difference is that for licensed nodes there is a **Change Admin Password** button that will _create_ the MxAdmin user if it doesn't exist yet. If it does exist, it just updates the password. This button is not available for Free Apps.

feddovanede: Look good to me!

MarkvanMents (approved):

PR #3528, opened 2021-07-13
Update app-roles.md

Maismaus:
Had some issues where a technical contact was not assigned on app creation. After contact with MX support, they explained that the technical contact is only assigned on first deployment.


Adam-Dupaski:
Thanks @Maismaus

PR #3531, opened 2021-07-14
SAP Release Note 14 July

MarkvanMents:
And documentation update for unbinding services.


MarkvanMents:
Added more information following discussion on [MCDEP-610]
@NielsTrots - Niels, a heads-up that this adds an instruction to contact support about unbound services. I don't think this will be a lot of work, but let me know/make a PR if there is a better way to describe the request which will make it clear to support what is being asked.
Merging this PR.

PR #3534, opened 2021-07-15
Switching icons for PWA docs

diego-antonelli:


ConnorLand:
Here we are attempting to use SVGs for clear red, green, and yellow icons a user can comprehend quickly. These SVGs are from the Glyphicons Halflings family, which are used elsewhere on docs (such as the Edit button). The plan is to use these SVGs first, and later we will see if Hugo can support custom icon font libraries. 

Icon font libraries would make using such icons a bit easier and cut down on the amount of Markdown used.

ConnorLand (approved):

PR #3538, opened 2021-07-16
Update url-redirector.md

RoelandSalij:
URL redirector is not available in the Marketplace.


Adam-Dupaski:
@RoelandSalij thanks, I will just remove this doc.

RoelandSalij:
But, but, but.... not being available in the Marketplace (which is a weird decision but alas) does not equal 'totally removed from this universe'?

Adam-Dupaski:
@RoelandSalij true (I guess), but we don't document components that are not platform-supported. For example, Mendix 4 still exists in this universe, but it's not supported, so we don't publish/support that documentation ;)

PR #3546, opened 2021-07-19
Update: Introduce new sync mode 'Synchronize Unsynchronized Objects'

omnajjar:
Docs update for Mx 9.4.0


ConnorLand (approved):

ConnorLand (approved):

ConnorLand:
@omnajjar is waiting for approval and then we can remove the Draft label.

ekrem-aktas (changes requested):
Some comments:

* At the beginning, we say: "The Synchronize action has two modes:". No longer true.
* On section 6 there's mention of the following: "...regardless the of type (full or selective)." We can remove the full or selective part.
* Should we explicitly note that the new synchronization mode and selective synchronization do not respect the Synchronization configuration?
* Synchronization Behavior section seems like only applies to SS. Maybe we can extend it a bit with something like: "This section describes the behavior for "Synchronize unsynchronized objects" and "Synchronize selected objects" modes."
* The first sentence in the synchronization behavior section implicitly applies to SS only. Shall we make it explicit?

ekrem-aktas (approved):

Adam-Dupaski (approved):

PR #3555, opened 2021-07-21
Explicitly mention Free Apps are not supported

feddovanede:
Explicitly mention Free Apps are not supported


Adam-Dupaski (approved):

Adam-Dupaski:
@MarkvanMents FYI upon return.

PR #3560, opened 2021-07-23
Native Mobile Builder release notes v1.0.107

jogboms:


MariaShaposhnikova:
Note: this change will be visible to users when 8.18.9 is out and the PR should be merged together with this release.

PR #3573, opened 2021-08-04
database only restore API flag

lazv85:
Restore API is changed that It can accept db_only flag as a query parameter.

### DIFF
--- a/content/apidocs-mxsdk/apidocs/backups-api.md
+++ b/content/apidocs-mxsdk/apidocs/backups-api.md
@@ -577,11 +577,21 @@ URL: https://deploy.mendix.com/api/v2/apps/<ProjectId>/environments/<Environment
 **Query Parameters**
 
 *   _source\_snapshot\_id_ (String): Identifier of the snapshot which will be restored. This value is required and must belong to a snapshot within the same application, although it could be a different environment.
+*   _db\_only_ (Boolean): Boolean flag. It might have `true` or `false` value. It determines if it is a database only restore operation. It defaults to false if not present.
 
 **Example Request**
 
 ```bash
 POST /api/v2/apps/b5f19af7-7453-465e-b9a1-d7556f524c1e/environments/d436e0cd-6200-4ac5-b858-849a6ddbb56a/restores?source_snapshot_id=5f8ace23-19df-4134-bd67-c338142a6097
+Host: deploy.mendix.com
+
+Content-Type: application/json
+Mendix-Username: richard.ford51@example.com
+Mendix-ApiKey:  26587896-1cef-4483-accf-ad304e2673d6
+```
+
+```bash
+POST /api/v2/apps/b5f19af7-7453-465e-b9a1-d7556f524c1e/environments/d436e0cd-6200-4ac5-b858-849a6ddbb56a/restores?source_snapshot_id=5f8ace23-19df-4134-bd67-c338142a6097&db_only=true
 Host: deploy.mendix.com
 
 Content-Type: application/json


RCatersels:
From this remark it looks like this should already be working. But just tested it and the API gives a bad request when trying to use this boolean. Or am I to early and is this not yet live? Much wanted feature though. Can not wait to work with this.
Regards,
Ronald

lazv85:
Hi Ronald, 

you are a bit early, should go to prod presumably with next deployment of the CP.

content/apidocs-mxsdk/apidocs/backups-api.md
```diff
@@ -577,6 +577,7 @@ URL: https://deploy.mendix.com/api/v2/apps/<ProjectId>/environments/<Environment
 **Query Parameters**
 
 *   _source\_snapshot\_id_ (String): Identifier of the snapshot which will be restored. This value is required and must belong to a snapshot within the same application, although it could be a different environment.
+*   _db\_only_ (Boolean): Boolean flag. It might have `true` or `false` value. It determines if it is a database only restore operation. It defaults to false if not present.
```
MarkvanMents, line 580: I think it might be better to say what the true value means. Perhaps
```suggestion
*   _db\_only_ (Boolean): Boolean flag. Set this to *true* if you are doing a database only restore operation. It defaults to *false* if not present.
```
What do you think?

MarkvanMents (changes requested):
image
@lazv85 
Hi Slava,
Thanks for the PR, I assume this is linked to the CP release which is now due tomorrow?
Can you sign the CLA - not so important for Mendix employees, but it helps us as we don't see errors when we look down the list of PRs.
Can you also have a look at my review comment and see if it makes sense. You can commit it if it looks OK ( I think you should have the authority to do that.
Mark

MarkvanMents:
> 
> 
> From this remark it looks like this should already be working. But just tested it and the API gives a bad request when trying to use this boolean. Or am I to early and is this not yet live? Much wanted feature though. Can not wait to work with this.
> Regards,
> Ronald

@RCatersels 
Hi Ronald,
We have our docs on GitHub to allow customers to contribute and provide feedback - which is great. I know you have done this in the past and it is really helpful to get contributions from customers to improve our documentation.

However, any Pull Requests which are not merged to master are not yet supported. This could mean that the feature is not ready, or that the documentation may not be accurate as it is still being reviewed. We try to use labels such as _Future release_ or _Work in Progress_ but these are mainly for our internal use and, in this case, the label had not yet been added.

It is great to get feedback from customers who are using the GitHub repo, but do bear the above in mind before you use anything you find in the docs repo which is not yet published to the site.
Yours
Mark

MarkvanMents (approved):

MarkvanMents:
@lazv85 
Hi Slava,
Was this part of the CP release on Friday, or is it a separate release?
Please can you confirm if this can be published now - or let me know when it will be released (via Slack if you'd rather).
Thanks
Mark

lazv85:
@MarkvanMents 
Hi Mark,

It was promised to be deployed this week.


Thanks
Slava

joshp123:
Hey @lazv85 , @MarkvanMents! @aliozcan and I were just discussing this feature (great to see that it's going live soon!), and I thought that it would be a good idea to update the documentation for the DB only restore to indicate that, since it does not restore files, this is a dangerous operation and can cause your environment to be left in a broken state. Therefore, people using this flag should be cautious and only use it if they are certain that this is what they want.

Would you be able to help out with this? Thanks!

content/apidocs-mxsdk/apidocs/backups-api.md
```diff
@@ -577,6 +577,7 @@ URL: https://deploy.mendix.com/api/v2/apps/<ProjectId>/environments/<Environment
 **Query Parameters**
 
 *   _source\_snapshot\_id_ (String): Identifier of the snapshot which will be restored. This value is required and must belong to a snapshot within the same application, although it could be a different environment.
+*   _db\_only_ (Boolean): Boolean flag. Set this to *true* if you are doing a database only restore operation. It defaults to *false* if not present.
```
MarkvanMents, line 580: @joshp123 
Hi Josh, how about adding the following big orange warning?
```suggestion
*   _db\_only_ (Boolean): Boolean flag. Set this to *true* if you are doing a database only restore operation. It defaults to *false* if not present.

    {{% alert type="warning" %}}Setting this to `true` will not restore any of your files and so there is a risk that data will be missing from your app or that your app will not work as expected. Use this option with caution.{{% /alert %}}
```

joshp123: Thanks @MarkvanMents , that looks great!

MarkvanMents:
Hi Slava (@lazv85 )
I believe this is going to be released tomorrow.

PR #3599, opened 2021-08-12
More complete description by example

TimvanSteenbergen:
Another finding in real live programming where a parseDecimal  got followed by a Decision '$var > 0' which started throwing an exception once we set parseDecimal's default to 'empty' instead of '0'.


Adam-Dupaski (approved):

Adam-Dupaski:
Thanks @TimvanSteenbergen

PR #3604, opened 2021-08-16
Improvement: make it clear that the synced objects are the ones on the device

omnajjar:


omnajjar:
Will pass it to the team for a review, and once ll good I'll unmark it as a draft!

content/refguide/synchronize.md
```diff
@@ -32,7 +32,7 @@ The behavior of this mode can be configured through [synchronization configurati
 ![Synchronize](attachments/client-activities/synchronize-unsynchronized-objects.png)
 {{% /image_container %}}
 
-With the **Unsychronized objects** mode, all objects with committed changes are synchronized. The synchronization is bi-directional, meaning both the server database and the local database are updated for these objects. For more information, see the [Synchronization Behavior](#synchronization-behavior) section below.
+With the **Unsychronized objects** mode, all objects on the device with committed changes are synchronized. The synchronization is bi-directional, meaning both the server database and the local database are updated for these objects. For more information, see the [Synchronization Behavior](#synchronization-behavior) section below.
```
ekrem-aktas, line 35: We have offline PWA profiles as well, doesn't mentioning the "device" sound like it is valid for only native apps?

omnajjar: Yeah indeed, it won't be interpreted to a browser as well! .... what do you think then of: 
`... all objects with committed changes locally are synchronized...`

ekrem-aktas: How about something like this: "... all objects that have been committed to the offline database are synchronized"?

omnajjar: Yes, that's more specific and inclusive! 👌🏼   ... will update it!

omnajjar: Updated! ✔️

ekrem-aktas (approved):
LGTM

ConnorLand (approved):

PR #3608, opened 2021-08-17
Add a warning about the possible error

Luyao-Zhang-1:


Luyao-Zhang-1:
@ConnorLand could you please take a look at this PR? It is for [TW-1204](https://mendix.atlassian.net/browse/TW-1204).

ConnorLand:
> @ConnorLand could you please take a look at this PR? It is for [TW-1204](https://mendix.atlassian.net/browse/TW-1204).

It looks good, but because we aren't recommending any specific workarounds I might just describe the error plainly:

"If an entity in a list view has an attribute populated by a microflow, it will cause an error."

Luyao-Zhang-1:
> > @ConnorLand could you please take a look at this PR? It is for [TW-1204](https://mendix.atlassian.net/browse/TW-1204).
> 
> It looks good, but because we aren't recommending any specific workarounds I might just describe the error plainly:
> 
> "If an entity in a list view has an attribute populated by a microflow, it will cause an error."

@ConnorLand Thanks for your feedback! I will describe the error plainly then in this case.

PR #3612, opened 2021-08-18
Add description for new `Use QuerySegment` feature of consumed odata

dfedorenko:
…service


content/refguide/consumed-odata-service.md
```diff
@@ -128,6 +128,7 @@ Click the **Properties** tab for the consumed OData service which displays the p
 * **Application ID** – the unique identifier of the application that the service was published from in the Data Hub Catalog
 * **Metadata** – the contents of the metadata file defining the service
 *  **OData version** – the OData version: can be OData 3 or OData 4
+*  **Use QuerySegment** – instructs application to use OData's QuerySegment notation ([OData specification](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody)). Feature is not available if consumed service explicitly indicates that is it not supported. 
 
 ## 4 Updating or Switching a Consumed OData Service {#updating}
```
joostverhoog, line 134: An attachment in /content/refguide8 was changed. This feature is for Mendix 9, right? I expected the attachment change in /content/refguide/

content/refguide/consumed-odata-service.md
```diff
@@ -128,6 +128,7 @@ Click the **Properties** tab for the consumed OData service which displays the p
 * **Application ID** – the unique identifier of the application that the service was published from in the Data Hub Catalog
 * **Metadata** – the contents of the metadata file defining the service
 *  **OData version** – the OData version: can be OData 3 or OData 4
+*  **Use QuerySegment** – instructs application to use OData's QuerySegment notation ([OData specification](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody)). Feature is not available if consumed service explicitly indicates that is it not supported.
```
joostverhoog, line 131: We should add a note to indicate the version in which this feature was introduced, so that people using an older version of Studio Pro 9 will not be confused by this setting not being there

joostverhoog, line 131: I would prefer to start with a low-code-like explanation about what happens, and the consequences, and at the end link them to the details in the OData spec.

So something like: When set to 'No', this app retrieves data using GET requests. When set to 'Yes', it uses POST requests. The URL's of these GET request may become very long, whereas the POST requests have short URL's. Make sure that the service supports this feature. You can't set the value to 'Yes' for OData 3 services, or if the service indicates that it doesn't support it. See the --deeplink--OData specification--/deeplink-- for more details.

joostverhoog (changes requested):

content/refguide/consumed-odata-service.md
joostverhoog: Looks good now

joostverhoog: Looks good now

joostverhoog (approved):

Adam-Dupaski (approved):

Adam-Dupaski:
@joostverhoog can this be published right away?

dfedorenko:
@Adam-Dupaski this for Studio Pro 9.6.0. Also added this in property description.

PR #3647, opened 2021-08-31
Add pie/doughnut chart to NMR widget list

jordanch:


jordanch:
**Please do not merge** this until the widget is released to the marketplace. I will respond here when this is done. I would add a label unfortunately I don't have the permissions to do this.

ConnorLand (approved):

PR #3655, opened 2021-09-01
DataHub-Catalog-Release-Notes-1.45

elize-vdr:


content/releasenotes/data-hub/index.md
```diff
@@ -9,6 +9,12 @@ These release notes cover changes made to the [Mendix Data Hub](/data-hub/).
 
 ## 2021
 
+### September 1st, 2021
+
+#### Fixes
+* Permanently removed the `DELETE` ServiceVersion API Call. This was removed from the Spec in a previous release, but a DELETE was still possible for those who used it and knew about it before, this is now permanently removed and can no longer be called.
```
reinouts, line 15: As an end user I would be interested in the recommended way to delete endpoints / service versions now that this API has been removed.

elize-vdr: The DELETE was already removed in a previous release and handled there, it is only the code that was not removed and users that knew about it could still use it. From the user's perspecitve DELETE already no longer exists, so in my mind I thougth we should not say anything that we neglected to remove the code, since it does not show us in a good light,

content/releasenotes/data-hub/index.md
```diff
@@ -9,6 +9,11 @@ These release notes cover changes made to the [Mendix Data Hub](/data-hub/).
 
 ## 2021
 
+### September 1st, 2021
+
+#### Fixes
+* Removed the `SecurityClassification` from the `Endpoints` JSON structure of the API. The creation of Endpoints as `Public` via the API is no longer allowed due to security reasons.
```
Adam-Dupaski: @elize-vdr which API is "the API" here? There are 3 options: https://docs.mendix.com/apidocs-mxsdk/apidocs/data-hub-apis

elize-vdr:
> @elize-vdr date is listed as today - this was already released?

You are right, type-o, I will fix

content/releasenotes/data-hub/index.md
elize-vdr: You are right, type-o, I will fix

elize-vdr: I changed it to "Register API"

Adam-Dupaski (approved):

PR #3657, opened 2021-09-01
Two extra examples for length

TimvanSteenbergen:
Since you can never have enough examples and since i had to find out for myself how function length would respond to an empty variable. Now I know: no error, returned value is 0

### DIFF
--- a/content/refguide/string-function-calls.md
+++ b/content/refguide/string-function-calls.md
@@ -112,19 +112,13 @@ The output is described in the table below:
 | -------------------- | ------- |
 | Length of the string | Integer |
 
-### 4.3 Example
+### 4.3 Examples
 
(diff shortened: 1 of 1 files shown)


content/refguide/string-function-calls.md
```diff
@@ -112,19 +112,13 @@ The output is described in the table below:
...
-
-The output is the following:
-
-```java
-14
-```
+| Input | Output |
+| --- | --- |
+| `length('thisismystring')` | 14 |
+| `length('$MyString')` and `MyString = 'qwer'` | 4 |
```
Adam-Dupaski, line 120: Thanks for the addition @TimvanSteenbergen . Is the code formatting I've added correct for lines 120 & 121?

PR #3659, opened 2021-09-02
Add note about local notification event limitation

jordanch:
Based on [this ticket](https://mendix.atlassian.net/browse/NC-377), the Notifications widget (available in [Native Mobile Resources 3.1.0](https://marketplace.mendix.com/link/component/109513)) was updated to handle local notification events. However, based on the dependency used in code, "On receive" events are not supported in the local notification context.

This MR updates certain howto's to make this clear.

### DIFF
--- a/content/howto/mobile/local-notif-action.md
+++ b/content/howto/mobile/local-notif-action.md
@@ -26,6 +26,12 @@ Before starting this how-to, make sure you have completed the following prerequi
 
 In this section you will learn to show a page when a user taps a notification.
 
+{{% alert type="info" %}}
+Local notifications do not support **On recieve** events.
+{{% /alert %}}
+
+To get started, do the following:
+
 1.  Drag and drop a **Notifications** widget onto your native home page. 
 
 	{{% image_container width="400" %}}![notifications widget](attachments/native-push/notif-widget.png){{% /image_container %}}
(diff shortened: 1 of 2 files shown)


jordanch:
This is ready to merge once approved.

content/howto/mobile/local-notif-action.md
```diff
@@ -33,7 +33,7 @@ In this section you will learn to show a page when a user taps a notification.
 2. Double-click the widget.
 3. Click **Actions** > **New**. 
 4. Name your action *show_page*.
-5. Select **On open to** > **Show a Page**.
+5. Select **On open to** > **Show a Page**. Note: local notifications do not support **On receive** events.
```
KevinVlaanderen, line 36: Do people have a good understanding of the distinction between 'local' and 'remote' notifications, at this point?
In general, formulating a statement like this, mentioning one case without stating the alternative, can be problematic if people aren't really aware of the distinction in the first place.
So if I read this and I don't really know about local and remote notifications, I might not understand what the implication of this sentence is. If it doesn't work for 'local' notifications, when 'does' it work?

jordanch: I understand what you are saying, although, since this document is 1 part of a group to do with local notifications only, I reckon this distintion should be clear.

It's the _actions_ part of local notifications.

What do you think?

KevinVlaanderen: Hence my first question. If your estimation is that it is clear at this point, then I'm good with that.

KevinVlaanderen (approved):

ConnorLand (approved):

PR #3669, opened 2021-09-06
Add change workflow state

MariaShaposhnikova:


content/refguide/change-workflow-state.md
```diff
@@ -0,0 +1,53 @@
...
+## 3 Action Section {#action}
+
+The **Action** section of the properties pane shows the action associated with this activity.
+
+You can open a dialog box to configure this action by clicking the ellipsis (**…**) next to the action.
+
+You can also open the dialog box by double-clicking the activity in the microflow or right-clicking the activity and selecting **Properties**.
+
+### 3.1 Operation
+
```
PaulJongenMx, line 40: Select the operation that should represent the new state of the workflow instance. 

Abort workflow
This operation will abort the workflow instance from further processing. This brings the workflow instance in its final state; there is no operation to undo an abort operation. The property Reason can be set for administration purposes, as to why the workflow instance is aborted.

Continue workflow
This operation will continue the processing of the workflow instance at the point where it was before a new version of the workflow definition was deployed. The workflow instance will no longer be in state "Incompatible".   If the task is no longer in the new workflow definition this action will result in an error.  

Pause workflow
This operation will halt (pause) further processing of the workflow. User tasks can't be completed when the workflow instance is in "Pause" state. Use the "Resume workflow" action to resume the processing of the workflow.

Restart workflow
This operation will stop continuation of the workflow at it's current task and start from (jump to) the initial task in the workflow definition. This action can be used when a workflow instance has become "Incompatible" after a modified workflow definition has been deployed. Please note that at the restart of the workflow instance, the workflow context object has the same values as before the "Restart workflow" operation was executed. 

Resume workflow
This operation will resume the workflow after it has been paused via the "Pause workflow" operation. User tasks can be completed again. 

The workflow instance state changes are reflected in the System.Workflow.State attribute

content/refguide/change-workflow-state.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+### 3.2 Workflow Object
+
+
```
PaulJongenMx, line 45: The Workflow object property should be a specialization of the System.Workflow entity and identifies the workflow instance the operation should be executed on.

content/refguide/change-workflow-state.md
```diff
@@ -0,0 +1,53 @@
...
+
+### 3.3 Reason
+
+
```
PaulJongenMx, line 49: The Reason property is only used when Change Workflow state Operation Abort workflow is selected and allows you to define a description why this workflow instance was aborted.

MariaShaposhnikova: Implemented

MariaShaposhnikova:
Merging this PR into a PR with all workflow changes for 9.6.

PR #3675, opened 2021-09-07
Document behavior on DDL commands

feddovanede:
See https://mendix.slack.com/archives/C5SASDZBN/p1630679438064000

Hopefully this will improved, but for now we should document the way it currently works.


MarkvanMents:
> 
> 
> I see I only did the Mendix 9 document. The Mendix 8 and Mendix 7 docs should be updated as well, I think?

We normally only do the current version but, since it is you asking, I'll do 7 and 8 as well.

MarkvanMents (approved):

PR #3677, opened 2021-09-07
add known issue for operator 2.1.0

cstevandy:


Adam-Dupaski:
@MarkvanMents if adding this KI, please follow process per SP RN until fix. A Jira issue for the fix should also be provided.

MarkvanMents (approved):

PR #3679, opened 2021-09-07
Add unique constraint limitation

feddovanede:
See https://mendix.slack.com/archives/C5SASDZBN/p1631023791080000?thread_ts=1630679438.064000&cid=C5SASDZBN


basdebakker (approved):
Approved by data storage.

MarkvanMents:
Hi @feddovanede and @basdebakker 

Doesn't this new section (2.4) say exactly the same as section 2.3, except with a different error message and referring to the Mendix way of setting uniqueness rather than the Oracle way?

Can't we combine these two sections, since the cause and solution is exactly the same?

basdebakker:
> Doesn't this new section (2.4) say exactly the same as section 2.3, except with a different error message and referring to the Mendix way of setting uniqueness rather than the Oracle way?

They refer to quite different operations. The first one is about using DISTINCT in queries, the new one is about setting unique validation rules in the domain model.

MarkvanMents:
I still see these two things as related. I suggest making it clearer by putting something like this:

```
### 2.3 Constraints on Long Strings

Oracle has limitations on setting constraints on CLOB (character large object) data. Mendix stores long strings as CLOB objects and this means that Mendix restricts some of the things you can do with your Oracle database if you define string attributes which are longer than 2000 characters. These restrictions are listed below.

The workaround is to use string attributes which are less than 2000 characters if you want to use this functionality.

#### 2.3.1 Selecting DISTINCT Attribute

Using [SELECT DISTINCT](/refguide/oql-select-clause) in OQL queries is not supported for strings longer than 2000 characters.

If you run into this limitation, a message like `Error Msg = ORA-06502: PL/SQL: numeric or value error: character string buffer too small` will be logged.

#### 2.3.2 Uniqueness Constraint

You cannot set a [uniqueness constraint](/refguide/validation-rules#uniqueness) on string attributes longer than 2000 characters.

If you run into this limitation, an exception like `Error Msg = ORA-02329: PL/SQL: column of datatype LOB cannot be unique or a primary key`.
```

basdebakker:
Makes sense. But the part that is common for 2.3 in your proposal also applies to the existing 2.1 and 2.2. So it could be moved to the common part of section 2. Which, incidentally, should be section 3 as there is already another section 2.

MarkvanMents:
Thanks for the suggestion @basdebakker (and pointing out the bug 🥇 
I'll make a commit and ask you and @feddovanede to review.

MarkvanMents:
Hi @basdebakker and @feddovanede 
I've rewritten (and renumbered) that section so that it is clear that all the restrictions are related to the same thing. Hope it works for you.
I've made the changes in one copy, but I'll make the same changes in the Mx 7 and Mx 8 versions once you are happy with them.

Thanks for the addition to the documentation, and the opportunity to update the numbering and organization of the document.

content/refguide/oracle.md
```diff
@@ -23,29 +23,41 @@ When setting-up perform the following steps:
...
 3. Ensure that the user has been granted enough quotas to create the resources they need, or give them an unlimited grant (for example, `GRANT UNLIMITED TABLESPACE TO mendix` where `mendix` is the user/schema that you have created).
 
 {{% alert type="info" %}}
 During the creation of the Mendix database, the number of structural modifications made will depend on the size of your domain model. If this number is quite large, or if there is a large structural change, it may be prudent to increase the value of `OPEN_CURSORS`.
 {{% /alert %}}
 
-## 2 Unlimited and Very Long Strings
+## 3 Unlimited and Very Long Strings
+
+The majority of differences between PostgreSQL and Oracle are in how they handle very long, or unlimited length, strings. Oracle has limitations on the functionality CLOB (character large object) data. Mendix stores long strings as CLOB objects and this means that Mendix restricts some of the things you can do with your Oracle database if you define string attributes which are unlimited or longer than 2000 characters. These restrictions are listed below.
```
basdebakker, line 37: functionality *of* CLOB

content/refguide/oracle.md
```diff
@@ -23,29 +23,41 @@ When setting-up perform the following steps:
...
 
-It is not possible to sort, group, or use aggregate functions such as `count()` on unlimited strings or strings with a specified length greater than 2000 characters. This is because such long or unlimited strings are implemented with the data type CLOB. Consider decreasing the length of the string attribute or removing it from data grids.
+### 3.4 Uniqueness Constraint
 
-### 2.3 Selecting DISTINCT Attribute
+You cannot set a [uniqueness constraint](/refguide/validation-rules#uniqueness) on string attributes longer than 2000 characters.
 
-Selecting DISTINCT attributes of the string type with a size greater than 2000 characters is not supported by Mendix due to a known Oracle limitation of selecting DISTINCT columns with a CLOB data type. If you run into this limitation, you may encounter an exception in the logs with a message like this: **Error Msg = ORA-06502: PL/SQL: numeric or value error: character string buffer too small**.
+If you run into this limitation, an exception like `Error Msg = ORA-02329: PL/SQL: column of datatype LOB cannot be unique or a primary key`.
```
basdebakker, line 59: This sentence is incomplete.

MarkvanMents:
Thanks @basdebakker 
I've corrected those two things.
My proofreading skills seem to be awful today. I blame the heat, Mendix World, and this block of Kryptonite under my desk. At least someone has their wits about them and is picking up my errors. Thanks.
Let me know if you are happy for me to copy this to the other versions (Mx 7 and Mx 8) and publish.

content/refguide/oracle.md
```diff
@@ -23,29 +23,41 @@ When setting-up perform the following steps:
...
 
-It is not possible to sort, group, or use aggregate functions such as `count()` on unlimited strings or strings with a specified length greater than 2000 characters. This is because such long or unlimited strings are implemented with the data type CLOB. Consider decreasing the length of the string attribute or removing it from data grids.
+### 3.4 Uniqueness Constraint
 
-### 2.3 Selecting DISTINCT Attribute
+You cannot set a [uniqueness constraint](/refguide/validation-rules#uniqueness) on string attributes longer than 2000 characters.
 
-Selecting DISTINCT attributes of the string type with a size greater than 2000 characters is not supported by Mendix due to a known Oracle limitation of selecting DISTINCT columns with a CLOB data type. If you run into this limitation, you may encounter an exception in the logs with a message like this: **Error Msg = ORA-06502: PL/SQL: numeric or value error: character string buffer too small**.
+If you run into this limitation, an exception like `Error Msg = ORA-02329: PL/SQL: column of datatype LOB cannot be unique or a primary key`will be logged.
```
basdebakker, line 59: Nitpick, but I suggest a space between "key`" and "will".

MarkvanMents: No - absolutely right. I'll change it.
Thanks again.

basdebakker (approved):
Looks good to me.

MarkvanMents (approved):
Ready to publish

PR #3685, opened 2021-09-09
Taskqueue user execute

hugomendix:
Tasks now run in an equivalent context as the one they were created in. The old behavior is now deprecated and a one-way project setting has been introduced for this.
This change is effective from Mendix 9.6 onwards.

### DIFF
--- a/content/refguide/task-queue.md
+++ b/content/refguide/task-queue.md
@@ -109,7 +109,7 @@ Core.userActionCall("AModule.SomeJavaAction")
   .executeInBackground(context, "AModule.SomeQueueName");
 ```
 
-The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a new, but equivalent context. See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#userActionCall(java.lang.String)) for more information.
+The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a [new, but equivalent context](#context). See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#userActionCall(java.lang.String)) for more information.
 
 ### 2.5 Configuration options{#configuration}
 
--- a/content/refguide/task-queue.md
+++ b/content/refguide/task-queue.md
@@ -127,7 +127,7 @@ This grace period is applied twice during the [shutdown](#shutdown) (described b
 
 Besides scheduling and executing tasks, the Mendix platform keeps track of tasks that have been executed in the background: for example, which completed and which failed.
 
-Internally, a scheduled or running task is represented by the Mendix entity `System.QueuedTask`. In a high performance setting, this entity should *not* be used directly by user code, because the underlying database table is heavily used. For example counting how many `System.QueuedTask` objects exist at the moment will lock the table and might cause a serious slowdown in task processing. You should also not Write directly to `System.QueuedTask`. Instead, mark a task for background execution in the **Call Microflow**  or **Call Java Action** activity or using the Java API.
+Internally, a scheduled or running task is represented by the Mendix entity `System.QueuedTask`. In a high performance setting, this entity should *not* be used directly by user code, because the underlying database table is heavily used. For example counting how many `System.QueuedTask` objects exist at the moment will lock the table and might cause a serious slowdown in task processing. You should also not Write directly to `System.QueuedTask`. Instead, mark a task for background execution in the **Call Microflow** or **Call Java Action** activity or using the Java API.
 
 Tasks that have been processed, that is have completed or failed, are saved as objects of entity type `System.ProcessedQueueTask`. These objects are at the user's disposal. They might be used, for example, to do the following:
 
@@ -157,7 +157,7 @@ The **Status** attribute of `System.QueuedTask` and `System.ProcessedQueueTask` 
 
 * `Idle`: The task was created and is waiting to be executed.
 * `Running`: The task is being executed.
-* `Completed`: The task executed successfully.  A `System.ProcessedQueueTask` is added to reflect this.
+* `Completed`: The task executed successfully. A `System.ProcessedQueueTask` is added to reflect this.
 * `Failed`: The task is no longer executing, because an exception occurred. A `System.ProcessedQueueTask` containing the exception is added to reflect the failure. The task will not be retried.
 * `Aborted`: The task is no longer executing, because the cluster node that was executing it went down. A `System.ProcessedQueueTask` is added to reflect this. The task will be retried on another cluster node.
 * `Incompatible`: The task never executed, because the model changed in such a way that it cannot be executed anymore. This could be because the microflow was removed/renamed, the arguments were changed or the Task Queue was removed.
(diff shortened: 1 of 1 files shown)


RCatersels:
You have no idear how happy you are making me with this feature. Kudo's for changing this.

content/refguide/task-queue.md
```diff
@@ -38,12 +38,11 @@ You can control how many tasks can run in parallel on each node when you create
 
 #### 2.1.3 Context in Task Queues
 
-For microflows and Java actions which are running in a task queue, the context in which the task runs changes slightly in the following ways:
+For microflows and Java actions which are running in a task queue, the circumstances in which the task runs changes slightly in the following ways:
```
MarkvanMents, line 41: Is there a reason that you don't want to use the word `context`?
For me, context is more accurate, has a technical meaning that customers and internal developers understand, but also has a general meaning in English that non-technical people can also understand.
Circumstances feels very abstract and carries, for me, the implication that it is something accidental rather than deliberately designed.

content/refguide/task-queue.md
```diff
@@ -88,7 +87,7 @@ Core.microflowCall("AModule.SomeMicroflow")
   .executeInBackground(context, "AModule.SomeQueueName");
 ```
 
-The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a system context. See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#microflowCall(java.lang.String)) for more information.
+The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a new, but equivalent context. See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#microflowCall(java.lang.String)) for more information.
```
MarkvanMents, line 90: Do you think we could expand this "new, but equivalent context"? Perhaps to say:
> this means that the task will have the same role(s) as the context passed and so will be subject to the same security.

MarkvanMents, line 90: I think we should link to the explanation of context here:
```suggestion
The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a [new, but equivalent context](#context). See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#microflowCall(java.lang.String)) for more information.
```

content/refguide/task-queue.md
```diff
@@ -110,7 +109,7 @@ Core.userActionCall("AModule.SomeJavaAction")
   .executeInBackground(context, "AModule.SomeQueueName");
 ```
 
-The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a system context. See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#userActionCall(java.lang.String)) for more information.
+The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a new, but equivalent context. See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#userActionCall(java.lang.String)) for more information.
```
MarkvanMents, line 112: and here:
```suggestion
The method `executeInBackground` takes two parameters: a context and a queue name. The context is only used for creating the task; executing the task will be done with a [new, but equivalent context](#context). See the [API documentation](https://apidocs.rnd.mendix.com/9/runtime/com/mendix/core/Core.html#userActionCall(java.lang.String)) for more information.
```

content/refguide/task-queue.md
```diff
@@ -138,18 +137,32 @@ Tasks that have been processed, that is have completed or failed, are saved as o
 
 `System.ProcessedQueueTasks` objects are never deleted. The user is free to delete them when desired.
 
-### 2.7 Task status
+### 2.7 Execution Context
```
MarkvanMents, line 140: Add an anchor here for the links above.
```suggestion
### 2.7 Execution Context{#context}
```

content/refguide/task-queue.md
hugomendix: Yes, that was a deliberate change on my part, because I felt context was used ambiguously here:
1. context can refer to the Mendix session context object
2. context can refer to the circumstances that change between invoking in the foreground and the background
The subsequent bullet items are not about changes to the context object. In the other places in the text context is accurate, however, because there it is referring to the Mendix session context.

content/refguide/task-queue.md
```diff
@@ -138,18 +137,32 @@ Tasks that have been processed, that is have completed or failed, are saved as o
...
+
+* When a user is logged in, the task will be executed in a new context for the same named user. This context will be the same as if the user is logged in.
+* When no user is logged in, the task will be executed in a new anonymous context. This context will be for a new anonymous user with the same language and timezone as the original user.
+* When a system session is used to schedule the task (using the Java API), the task will be executed in a new system context.
+
+Projects containing task queues that were created before Mendix 9.6 will get a deprecation warning on the system context. This warning can be remedied in the project settings for the runtime. By choosing *no* for *System context tasks* the deprecation warning goes away and the tasks will executed in an equivalent context to the one they were created in.
```
MarkvanMents, line 148: i.e. here? 
image

MarkvanMents (commented):
Can you have a look at my suggestions/comments?
I have also made some small language changes to the section where you describe context.

content/refguide/task-queue.md
hugomendix: Yes, it will look like this: 
image

hugomendix:
Agree with you changes. Replied to your questions.

content/refguide/task-queue.md
MarkvanMents: Thanks for the screenshots.
I don't think we need them in the document, but perhaps we can help the user to find the setting by expanding the text a bit?

```suggestion
Projects containing task queues that were created before Mendix 9.6 will get a deprecation warning in the log: `The project setting 'System context tasks' is deprecated`.
You can remove this warning in the **Runtime** tab of the app **Settings** in Studio Pro. Set **System context tasks** to *no* to execute tasks in an equivalent context to the one they were created in and remove the warning.
```

content/refguide/task-queue.md
```diff
@@ -138,18 +137,32 @@ Tasks that have been processed, that is have completed or failed, are saved as o
...
+
+{{% alert type="warning" %}}
+After choosing *no*, you cannot switch back to *yes* because executing tasks in system contexts (unless scheduled from a system session) is deprecated.
```
MarkvanMents, line 151:
```suggestion
You will be asked to confirm this change as, after choosing *no*, you cannot switch back to *yes* because executing tasks in system contexts (unless scheduled from a system session) is deprecated.
```

content/refguide/task-queue.md
MarkvanMents: Thanks for the explanation. I hadn't thought about the second point. You mean, for example, that an entity might have been deleted or modified in the meantime?

I wonder if `conditions` might be a better word? But I'll leave it as `circumstances` for now.

MarkvanMents:
Thanks for the quick response.
Just a couple of suggestions to remind customers where the runtime settings are. Otherwise, looks great.
Thanks for the submission.

content/refguide/task-queue.md
hugomendix: Yes, that would be good as well. I'll leave that up to you.

hugomendix: Yes, that would be good as well. I've updated the text.

hugomendix: And with the second point I was referring to the changes as elaborated in the bullet items.

MarkvanMents (approved):
Great,
Thanks for your explanations.
I think this is ready to be published with the other 9.6.0 docs when 9.6.0 is released.

PR #3686, opened 2021-09-09
List value reload method (9.6)

alihcsumer:
Reload method is introduced for list value data sources targeting version 9.6.0.


MarkvanMents:
@ConnorLand - I think you can deal with this, but let me know if it should be someone else.

ConnorLand (approved):

PR #3691, opened 2021-09-10
[RTI-315] Add known issue of runtime statistics

atennapel:
I do not know how to link to 9.6, the page does not exist yet.


MarkvanMents:
Hi Albert (@atennapel )
The links look OK, although they will need an anchor added. Feel free to choose an appropriate name and we can make sure we put the correct anchor in 9.6.md
I've put a label to make sure we check that the links are correct once the 9.6 release notes are published.

atennapel:
Thank you @MarkvanMents , is it fine as it is now then? (9.6#????) Or should I put a random number there?

MarkvanMents:
Thanks for the question.

The numbers are not entirely random. The numbers are either a ticket number from zendesk or, for internal known issues, the JIRA issue number which has been marked with the PUBLIC_KNOWN_ISSUE label.

You should know one of these, so if you put that in this PR and note it in the 9.6 release notes on (GitLab), then the two things should tie up neatly.

atennapel:
Alright, updated. There is no zendesk issue, but the Jira number is RTI-315.

Adam-Dupaski:
@MarkvanMents shouldn't these KIs have been published already (7 days ago)? They will be totally retroactive if only coming out with the 9.6 fix note. This will also make it trickier to create the 9.6 RN and publish these fixes. Best practice is to publish KIs when identified.

MarkvanMents:
@Adam-Dupaski 
Yes you are right. I didn't spot that they are new issues, I read them as fixes to known issues.
I'll update the PR and publish it now.

MarkvanMents:
> 
> 
> @MarkvanMents shouldn't these KIs have been published already (7 days ago)? They will be totally retroactive if only coming out with the 9.6 fix note. This will also make it trickier to create the 9.6 RN and publish these fixes. Best practice is to publish KIs when identified.

Hi @Adam-Dupaski 
Thanks for picking this up. KIs are now published and this only has the links to the solution for publication with 9.6

Adam-Dupaski:
Handling this in the 9.6 RN branch/PR.

PR #3693, opened 2021-09-10
Studio Pro 9.6.0 introduces updatable external entities

joostverhoog:
Some attributes can be updatable, and there is a new microflow activity to send the changes values to the publishing app

### DIFF
--- a/content/data-hub/data-hub-catalog/consume.md
+++ b/content/data-hub/data-hub-catalog/consume.md
@@ -38,7 +38,7 @@ It is good practice for publishers of a service to serve a notice of deprecation
 
 When you use an external entity from a published OData service through the **Data Hub** pane in Studio Pro, you are consuming the dataset from the service (which is published from the app deployed in a specific environment). The OData endpoint for the dataset is used in the consuming app.
 
-External entities may be read-only. In that case it is not possible to change the structural values of attributes or associations between two external entities.
+It is not possible to change the structural values of attributes or associations between two external entities.
 
 When security is enabled for your app, you can define access rules for external entities just as you would for [persistable](/refguide/persistability#persistable) and [non-persistable](/refguide/persistability#non-persistable) entities. You can define access rules based on user roles (for more details, see [Security and Controlling Access to Information](security)).
 
--- a/content/data-hub/data-hub-catalog/consume.md
+++ b/content/data-hub/data-hub-catalog/consume.md
@@ -38,9 +38,9 @@ It is good practice that publishers of a service serve a notice of deprecation o
 
 When you use an external entity from a published OData service through the **Data Hub** pane in Studio Pro, you are consuming the dataset from the service (which is published from the app deployed in a specific environment). The OData endpoint for the dataset is used in the consuming app.
 
-External entities are read-only, so it is not possible to change te structural values of attributes or associations between two external entities.
+External entities may be read-only. In that case it is not possible to change te structural values of attributes or associations between two external entities.
 
-When security is enabled for your app, you can define access rules for external entities just as you would for [persistable](/refguide/persistability#persistable) and [non-persistable](/refguide/persistability#non-persistable) entities. You can only define read access, and also access rules based on user roles (for more details, see [Security and Controlling Access to Information](security)).
+When security is enabled for your app, you can define access rules for external entities just as you would for [persistable](/refguide/persistability#persistable) and [non-persistable](/refguide/persistability#non-persistable) entities. You can define access rules based on user roles (for more details, see [Security and Controlling Access to Information](security)).
 
 You can associate external entities with local entities (both [persistable and non-persistable](/refguide/persistability). However, the external entity cannot be the owner of an association, which means that the association has to be from a local entity to the external entity in the domain model, and the value for the association [owner](/refguide/associations#ownership) must be set to **Default**.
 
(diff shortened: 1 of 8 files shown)


joostverhoog:
Review required, first from the integration team, and then from tech writers

content/refguide/consumed-odata-service-requirements.md
```diff
@@ -26,10 +26,11 @@ This section describes the features of a consumed OData service that are support
 
 Vocabulary annotations can be used in a service to indicate features that are not supported. The following vocabulary annotations are recognized for entity sets:
 
-* **Countable** – marking an entity set as `Countable="false"` prevents the user from adding the entity to the app
-* **Filterable** – marking an entity set as `Filterable="false"` sets all properties as non-filterable
-* **Sortable** – marking an entity set as `Sortable="false"` sets all properties as non-sortable
+* **Countable** – an entity set marked as `Countable="false"` prevents the user from adding the entity to the app
```
osoykan, line 29: >  prevents the user from adding the entity to the app

I didn't get how `Countable` is related to adding an entity. Is that correct?

content/refguide/send-external-object.md
```diff
@@ -0,0 +1,27 @@
...
+
+The **Send External Object** can be used to persist changes to an external object.
+
+{{% alert type="info" %}}
+This activity was introduced in Studio Pro [9.6.0](/releasenotes/studio-pro/9.6).
+{{% /alert %}}
+
+## 2 Updatable attributes
+
+When external entities may have updatable attributes. The values of these attributes can be changes, for instance in a widget or by using the [change object activity](change-object).
```
osoykan, line 20: > The values of these attributes can be changes

_changed_

content/refguide/send-external-object.md
```diff
@@ -0,0 +1,27 @@
...
+
+Use the **Send External Object** activity to send these changes to app that publishes the entity. That app will persist the changes, overwriting the previous values of the attributes.
```
osoykan, line 22: >  send these changes to app

_the app_

content/data-hub/data-hub-catalog/consume.md
```diff
@@ -38,9 +38,9 @@ It is good practice that publishers of a service serve a notice of deprecation o
 
 When you use an external entity from a published OData service through the **Data Hub** pane in Studio Pro, you are consuming the dataset from the service (which is published from the app deployed in a specific environment). The OData endpoint for the dataset is used in the consuming app.
 
-External entities are read-only, so it is not possible to change te structural values of attributes or associations between two external entities.
+External entities may be read-only. In that case it is not possible to change te structural values of attributes or associations between two external entities.
```
osoykan, line 41: Comma: _In that case ,_

to change te  -> _to change the_

MarkvanMents: Just corrected te -> the !

content/refguide/consumed-odata-service-requirements.md
joostverhoog: Yes, it's correct as stated. 
This pull request only changes the wording from "marking an entity set" to "an entity set marked", because a consuming user isn't actively involved in changing the contract.

content/data-hub/data-hub-catalog/consume.md
joostverhoog: I now removed "read-only".
The use of read-only is confusing here. The structure of entities is defined by the service and can't be changed in the consumed app. So in that sense they are read-only. But the values can change now, so in that sense it's not read-only.

joostverhoog:
Review by the integration team is done. This looks good to go.

content/refguide/send-external-object.md
```diff
@@ -0,0 +1,34 @@
...
+
+Use the **Send External Object** activity to send these changes to the app that publishes the entity. That app will persist the changes, overwriting the previous values of the attributes.
+
+Only the changes are being sent, when two users make changes to different attributes, these changes are both applied separately.
+
+This activity can only send objects from an OData service.
+
+Some services may have updatable attributes that are part of the key of the entity. Services published by Mendix apps do not have this, but other services might. It's not recommended to change those attributes.
+
+## 3 Related
```
MarkvanMents, line 30: Presumably the rollback behaviour of **Send External Object** is also different from Commit?
Should we mention this here?

MarkvanMents:
I don't see any discussion of events for writeable external entities.
Is it clear to the developer that they cannot add events to provide validation and other functions to external entities? Or are events available?

MarkvanMents (approved):
I have added a couple of questions, but the PR is approved as these can be answered later.

content/refguide/send-external-object.md
osoykan: There is no rollback for `Sending` an HTTP request. It is against its nature. So, it is a completely different feature in that sense.

osoykan:
No events are not available for external entities. They are for local entities. In the entity detail, there is no `Event Handlers` section.

osoykan (approved):

content/refguide/send-external-object.md
MarkvanMents: Thanks - I'll leave it as it is then.

joostverhoog:
Hi @MarkvanMents, thank you for your review.

Regarding event handlers: those have never been supported for external entities; that section of the dialog is not shown at all. I agree that that's not obvious when you read the refguide. However, improving that seems out of scope for this pull request, which is for the changes in 9.6.0. Maybe we can improve that documentation in a separate pull request?

Regarding rollback: rollback is related to commit. Do we suggest anywhere that it may be related to the send external object activity as well? For things that can't be done I find it difficult to figure out what we should document. Do you have a suggestion?

MarkvanMents:
Hi @joostverhoog 
I mentioned those two things as you do have a section in `send-external-object` which is about related things, and there you talk about using `Commit` to deal with persistable objects.

I agree that it is difficult to know what to write about things that something _doesn't_ do.

I'll merge this now and we can see if customers come back with questions before adding any additional informaiton.

PR #3718, opened 2021-09-20
Specifics on licensing limit error messages

thijsrohof:
Added a section that describes how to notice that an application has exceeded the Free licence limit.


mendixmiles (commented):
Nothing to change / add

Adam-Dupaski (approved):

PR #3723, opened 2021-09-22
Update scheduled-events.md

MxClaudineMaas:
https://mendix.slack.com/archives/C3T260XGA/p1632231097158200

### DIFF
--- a/content/refguide/scheduled-events.md
+++ b/content/refguide/scheduled-events.md
@@ -59,7 +59,13 @@ The `ScheduledEventInformation` objects are not cleared automatically. If you ha
 
 ## 5 Additional information
 
-### 5.1 Calculating Intervals
+### 5.1 Running Concurrently
+
+You cannot run more then 10 scheduled events in parallel.
+
+This limit is hard-coded and is independent of how the app is scaled.
+
+### 5.2 Calculating Intervals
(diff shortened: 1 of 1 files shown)


MarkvanMents (commented):
Hi Claudine, @MxClaudineMaas 
Thanks for the Pull Request - I saw it in the Slack discussions.
I have given the information its own heading, as it didn't seem to be totally related to Long Running Events. I also added additional information I saw in the discussion.
Can you just check that it is still OK. Then I can approve and merge.
Thanks
Mark

MxClaudineMaas:
Hi Mark, 

I would not mention that it is hard-coded; this means that it is hard-coded in our code (which is not something we want to let customers know ;). For customers it means that they just can not change it themselves in any way.

MarkvanMents:
HI Claudine,
I've changed it to
> This limit cannot be overridden and is independent of how the app is scaled.

I think it is useful to say that it cannot be changed, so the customer doesn't ask support to change the limit for them.
Is this OK?

MxClaudineMaas:
Yes good point :), that saves us some tickets indeed :P!

And yes, agreed, thank you.

MarkvanMents (approved):

PR #3729, opened 2021-09-23
Using Deeplink and MendixSSO

Ceballj:
- I would like to suggest including additional information for apps using MendixSSO for SSO, as the value for the LoginLocation constant is different than the one when using SAML for SSO.
- This was discussed with @MxClaudineMaas


Luyao-Zhang-1:
Check with @Ceballj. I will merge this PR. Thanks @Ceballj!

PR #3733, opened 2021-09-27
Added Git Team Server to preferences

grootjans:
@MariaShaposhnikova This is the proposed text for the preferences dialog. The release notes will point to this.


Adam-Dupaski:
@MariaShaposhnikova to verify this goes with 9.6.

MariaShaposhnikova:
Verified by Robbert Jan that this goes in 9.6.

PR #3734, opened 2021-09-27
Add team server git Beta notice

grootjans:
@MariaShaposhnikova We have to wait to confirm that this is in the 9.6


MariaShaposhnikova:
@grootjans , I am a bit confused with what I see in this PR. Did you copy the release notes to add this point to them? "As of this release, it is possible to start projects on the Git edition of Team Server through Studio Pro. How this can be turned on is detailed in the [preferences](refguide/preferences-dialog#new-features). This feature is current in [Beta](/releasenotes/beta-features/)." 

In this case, the base branch should be "ad-9.6-rn" where the release notes are and then we can add only the one point, otherwise, there is a high chance we'll end up with several conflicts.

grootjans:
Sorry, I didn't see it chose the wrong base. I thought it would be smarter than that.

PR #3739, opened 2021-09-27
Update expressions.md

NatasaKralj:
For [TW-1454](https://mendix.atlassian.net/browse/TW-1454)

I don't think it warrants a new subsection. Seems like a necessary general warning for users to keep in mind when writing expressions with multiple statements.


Adam-Dupaski:
@NatasaKralj looks good.

PR #3730, opened 2021-09-24
Update deep-link.md

RCatersels:
See forum question https://forum.mendixcloud.com/link/questions/109785 This is currently missing in the documentation.

PR #3740, opened 2021-09-27
SameSiteCookies 'Strict' prevents SSO to function

TimvanSteenbergen:
Some change to Custom settings to the SameSiteCookie. In the near future I may run into this when we upgrade to Mx9. This looks like a nice place to jot down this notification.
This is not my own wisdom: see forum questions 109976 and 109785. Btw. Sounds like there is likely a better solution in near future if SAML improves their inner working. For now this will have to do.

### DIFF
--- a/content/appstore/modules/saml.md
+++ b/content/appstore/modules/saml.md
@@ -110,7 +110,7 @@ Each IdP (entity descriptor) should have its own configuration set. Every IdP ca
 
 ### 3.3 Setting `SameSite` Cookies to `None`
 
-In Mendix 8, `SameSite` cookies default to `None`. In Mendix 9, however, `SameSite` cookies default to `Strict`. Therefore, if the login page of your Identity Provider (IdP) has popped up before, the browser does not forward the session cookies issued by the `/SSO/` handler. (For the same reason the deeplink also works, if you have already logged in via your IdP before and its login page is therefore not opened.) Currently setting `com.mendix.core.SameSiteCookies` to `None` or `Lax` in the [custom runtime settings](/refguide/custom-settings) will tell the browser that your application accepts this cookie. 
+In Mendix 9, `SameSite` cookies default to `Strict`. Therefore, if the login page of your Identity Provider (IdP) has popped up before, the browser does not forward the session cookies issued by the `/SSO/` handler. (For the same reason the deeplink also works, if you have already logged in via your IdP before and its login page is therefore not opened.) Currently setting `com.mendix.core.SameSiteCookies` to `None` or `Lax` in the [custom runtime settings](/refguide/custom-settings) will tell the browser that your application accepts this cookie. 
 
 ## 4 Advanced Configuration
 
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@Luyao-Zhang-1 saw your slack about this, please also work with Mark to verify for runtime/SSO aspects.

Luyao-Zhang-1:
@TimvanSteenbergen It is only necessary to change the default for SameSite cookies to `none` or `lax` only when you use the Deep Module with it, right?

RCatersels:
No, SAML alone also needs this. See forum question https://forum.mendix.com/link/questions/109785. 
The problem seems to be that in Mendix 9 the SameSite cookie defaults to “Strict” and thus the browser does not forward the session cookie issued by the /SSO/ handler if the login page of your IdP has popped up before (and for the same reason the deeplink also works if you have already logged in via your IdP before and its login page is therefore not opened).

TimvanSteenbergen:
Not to my knowledge, but Ronald knows that SAML needs it too. Thanks Ronald, for mentioning that. Saves me a bughunt when we upgrade to Mx9 :-)

feddovanede:
I created a test app and I'm able to successfully sign in using SAML in both Mendix 8 and Mendix 9.

In Mendix 8 the **SameSite** setting for the cookies are set to _None_
afbeelding

In Mendix 9 the **SameSite** setting for the cookies are set to _Strict_. Still, this doesn't prevent me from signing in.
afbeelding

The cookies are set by the runtime, and they are used by the runtime, so on the exact same URL, so strict should work.

You can test this for your self on https://samlmoduletestmx8-sandbox.mxapps.io/. You can click the **SamlTest.id** button on the login page. The sign in page will provide several test credentials you can use to sign in.

The app is running as a Free App, so the runtime settings are default, as you can't change them.

feddovanede:
You will get into an endless loop if you use a DeepLink. You can test this with https://samlmoduletestmx8-sandbox.mxapps.io/link/test/

RCatersels:
Then it indeed would make sence to only document this in the deeplink. documentation Is there a fix possible so that this will also just going to work when you also have the deeplink module?

feddovanede:
Interestingly enough, the request initiator chain doesn't give show any reason to block the cookie. The deeplink URL is called from /SSO/assertion, which is on the same site as the deeplink URL (see https://web.dev/same-site-same-origin/#site)
afbeelding

My guess is that the cookie is blocked, because /SSO/assertion is called from the IdP, which is on a different site. But it is strange that that request isn't mentioned in the chain?

When you don't use the DeepLink module, the SAML module will redirect you to the _index.html_ page by default. On this request, the cookie is blocked as well.
afbeelding

However, the _index.html_ page executes the _mxui.js_ code, which makes a request to _/xas/_ to get the session data. On this request the cookies is sent, so apparently this starts a new request chain, and as _index.html_ is on the same site as _/xas/_, the cookie is sent.
afbeelding
afbeelding
It could be the chain is restarted because the request to _index,html_ returns a _200_ response, whereas the request to _xas_ returns a _303_ response.

So, this probably has to be fixed structurally in the SAML. For now, I think we should document this, preferably in the documentation of the DeepLink module, I think the steps are the same as for [iframing your app](https://docs.mendix.com/developerportal/deploy/environments-details#iframe): the preferred, most secure way to resolve this issue, is to make sure the IdP and app are in the same domain (and thus the same site). If your Mendix app is on _app.domain.com_, and you open the deeplink _app.domain.com/link/test_, and you are redirected to your IdP to sign in on _idp,domain.com/SSO_, which returns you to _app.domain.com/SSO/assertion_ after a successful sign in, which finally forwards you to  _app.domain.com/link/test_, your requests will all be within the same site, and thus the cookie should be forwarded each time.
If that is not an option, changing the SameSite setting for your app will resolve the issue as well.

RCatersels:
I think that 99% of all the SAML setups have an IdP  with a different domain (and in multi tenant environments this is always the case) so it would be nice if the SAML module could fix this. Will this get automaticly on the backlog or should we file a support request for this?

feddovanede:
@RCatersels I was planning on adding an item on the backlog for this myself. If you want to keep updated on the progress on this issues (which might take some time to investigate and resolve), it might be better if you create a ticket yourself.

Luyao-Zhang-1:
So I will close this PR. The solution to the endless redirect issue will be added to the [Deep Link doc](https://github.com/mendix/docs/pull/3730/files).
Thank you all for the help and discussion.

PR #3751, opened 2021-09-30
Mendix v6 is not supported.

joshp123:
We should not mention that it's possible to run unsupported versions in Mendix Cloud V4.


MarkvanMents (approved):
Looks great, @joshp123 
Will merge immediately.

joshp123:
Thanks a lot, @MarkvanMents ! Appreciate it.

PR #3753, opened 2021-10-01
Clarify difference between Free Apps and Unlicensed Apps

MarkvanMents:
Free App was used in cases where the apps were actually _unlicensed_ apps.
The main difference is a limitation on the number of concurrent users in the unlicensed apps.


MarkvanMents:
Approved by Tom de G. over email.
Publishing.

PR #3755, opened 2021-10-01
Update screenshots for 9.6 redesign

Luyao-Zhang-1:


content/appstore/general/app-store-content.md
```diff
@@ -135,7 +135,8 @@ To add the Email with Template module from the Marketplace via Studio Pro, follo
 
     * **Add as a new module** – if you select this option, new entities are created in your app
     * **Replace existing module** – if you select this option, you need to specify which **Module to replace**
-   
+      
+
```
Adam-Dupaski, line 139: Double space not necessary, just 1 space :)

Adam-Dupaski:
@Luyao-Zhang-1 blue highlighting on email-template-app-explorer.png for example is from your SnagIt, correct? Because we use red highlighting as the standard.

Luyao-Zhang-1:
@Adam-Dupaski It is confusing indeed. That's what you see on the UI actually. You get this blue box around an item, when you select it in the list.

Adam-Dupaski:
> 
> 
> @Adam-Dupaski It is confusing indeed. That's what you see on the UI actually. You get this blue box around an item, when you select it in the list.

Oh okay, wasn't aware. PR looks good then!

Luyao-Zhang-1:
> > @Adam-Dupaski It is confusing indeed. That's what you see on the UI actually. You get this blue box around an item, when you select it in the list.
> 
> Oh okay, wasn't aware. PR looks good then!

Thanks! There are still two deploy docs that need an update. These screenshots are a bit difficult to reproduce. I will work with Mark to get them done. Then I will merge this PR.

MarkvanMents:
Images look good.
The only one I'm not sure about is content/developerportal/deploy/attachments/qianfan-sso/image46.png - you have cropped it so that it doesn't specify that production security is turned on (which is step 1 of that section). Might be OK, have a look at the documentation and see whether it would also help to highlight that.
Otherwise, they all look good to me.

content/developerportal/deploy/qianfan-sso.md
```diff
@@ -291,26 +291,30 @@ The permissions in Yufu are equivalent to the roles in Mendix.
...
+
+5. Set the **Caption** to *logout*.
 
     ![](attachments/qianfan-sso/image45.png)
 
 ### 4.6 Role Settings
 
 1. Open **Security** and set the **Security Level** to *Production*.
 
+    ![security-set-to-production](attachments/qianfan-sso/security-production.png)
```
Luyao-Zhang-1, line 316: @MarkvanMents Thanks for checking and your comment. I made an extra screenshot for the security level setting, and put it under step 1 directly. Then the other screenshot can be a bit smaller.

MarkvanMents: Great - missed that. 👍

Luyao-Zhang-1:
Thank you both for the review. I will now merge the PR.

PR #3763, opened 2021-10-06
Improvements to the page

arjenw:
This changeset contains a serie of improvements for this page. However, this page is still significantly outdated (pictures and mentioned JSON's) and based on the stateful server behavior used Mendix 6 and before. Since Mendix 7 the runtime became stateless, which changed the interaction between client and server significantly.

Merging these proposals is one thing, however, this page requires a significant overhaul to become relevant again. My suggestion is to take this up with the frontend teams to get to the right JSON's and pictures.


MarkvanMents:
Thanks for your submission Arjen.
I'll review it as is, and I have approached the developers to do the more comprehensive review that you suggested.

PR #3764, opened 2021-10-06
[UFC-206] Added custom runtime setting for file document caching (mx9)

Robinfr:
**From security team. This should be merged only after CVE has been published. Same for the other MRs**

Added the `EnableFileDocumentCaching` setting, which can be used to enable file document caching if desired. We are disabling this by default in the upcoming Mendix 9 version.


MarkvanMents:
Hi @Robinfr 
You say that these are changes for versions 7, 8 and 9. I assume these will be released on different days, so is it better to make three different PRs?
Can you confirm that this is for:

- 9.7.0
- 8.18.12
- 7.23.26

Thanks

MarkvanMents:
OK, I've put all three labels on it for now, please confirm which version it is released in.
Is there a reason why this might be released in a patch release? I think users are less likely to see the release note if it is in a patch?

Robinfr:
@MarkvanMents It's for both 9.6.1 and 9.7

PR #3776, opened 2021-10-07
Two versions of IOS app

ChrisdeG:


ConnorLand:
Thank you for pointing out those initial links @ChrisdeG. I have clarified other occurrences in the documentation. Now things are a little more explicit regarding versioning.

Thank you for contributing to the documentation.

PR #3785, opened 2021-10-12
[WC-367] Remove demo apps from widgets' documentation

diego-antonelli:


ConnorLand:
Looks good, I'll merge @diego-antonelli . Just curious, was there a decision to stop supporting these Demo Apps? Will there be an alternative in the future?

PR #3792, opened 2021-10-13
Remove the limitation of associated attributes in DB2 and SAP HANA since it's no longer applicable.

snajahi:
The once documented limitation regarding associated attributes in DB2 and SAP HANA  is no longer applicable. We therefore should remove the related section in the documentation.

### DIFF
--- a/content/howto/general/dev-best-practices.md
+++ b/content/howto/general/dev-best-practices.md
@@ -372,7 +372,7 @@ Apps should keep up with new Mendix releases as much as possible.
 
 When introducing a new [Mendix Marketplace](https://marketplace.mendix.com/) component to an app, carefully consider the support level of the component. Using components that are community supported introduces a maintainability and upgrade risk.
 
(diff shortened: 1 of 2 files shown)


dahfjkg:
Passed team review

MarkvanMents (approved):
Thanks @snajahi 
Approving this ready for 9.8.0

snajahi:
> Hi Pavel, So this change is for 9.8.0? - I'll put a label on it to make sure it isn't merged before! It sounds like we need to keep something in the doc for customers who are using other versions which are not LTS? Perhaps, rather than deleting the section we could add a banner which says
> 
> > This limitation has been removed in versions x.xx.xx and above
> 
> What do you think?

Hi @MarkvanMents, the changes were updated as suggested. They might also require a review by someone from the security team.

MarkvanMents:
> 
> 
> > Hi Pavel, So this change is for 9.8.0? - I'll put a label on it to make sure it isn't merged before! It sounds like we need to keep something in the doc for customers who are using other versions which are not LTS? Perhaps, rather than deleting the section we could add a banner which says
> > > This limitation has been removed in versions x.xx.xx and above
> > 
> > 
> > What do you think?
> 
> Hi @MarkvanMents, the changes were updated as suggested. They might also require a review by someone from the security team.

Done

MarkvanMents:
Revived in #3908

PR #3834, opened 2021-10-28
8.18.13 RN

Adam-Dupaski:


Adam-Dupaski:
Need to verify for 9.7 sec/fix note to be included

PR #3847, opened 2021-11-03
Add MXP009 documentation to Performance Best Practices page (needs to be published with respect to the Studio Pro 9.8 release - Nov 30, 2021)

dogancakmak:
MxAssist Client team implemented a new best practice: MXP009. We need to have a documentation regarding this new best practice.
---
MXP009 will be released with Studio Pro 9.8.

### DIFF
--- a/content/refguide/performance-best-practices.md
+++ b/content/refguide/performance-best-practices.md
@@ -166,7 +166,7 @@ To fix the issue, do the following:
 
 ## 8 Avoid Using Too Many Levels of Inheritance {#mxp009}
 
-Using multiple levels of inheritance and too many specializations on entities can affect performance, especially when you are using XPaths for [entity access rules](access-rules). This generates complex queries adding XPaths for every specialization access rule and leads to slow queries on a large dataset. 
+Using multiple levels of inheritance and too many specializations on entities can affect performance, especially when you are using XPaths for [entity access rules](access-rules). This generates complex queries, adds XPaths for every specialization access rule, and leads to slow queries on a large dataset. 
 
 ### 8.1 Steps to Fix
 
--- a/content/refguide/performance-best-practices.md
+++ b/content/refguide/performance-best-practices.md
@@ -166,7 +166,7 @@ To fix the issue, do the following:
 
 ## 8 Avoid Using Too Many Levels of Inheritance {#mxp009}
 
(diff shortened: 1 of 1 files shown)


jaspervanderhoek:
The steps to fix look good, I personally don't agree with the 'two levels' statement, since we break that rule in our teachings and several modules we publish, however that is what we explicitly teach in the advanced course so we should keep that identical.


We should rewrite the problem statement:
_Using multiple levels of inheritance and too many specializations on entities can affect performance, especially when you are using XPaths for [entity access rules](access-rules). This generates complex queries, adds XPaths for every specialization access rule, and leads to slow queries on a large dataset._ 

The impact of inheritance and access rules is a bit more nuanced. The database doesn't really care if it joins a relationship with an association table of the parent vs the child, they are both Primary Keys that are joined. As an initial rule this is a good start but there are specific cases that are impact differently. 

Straight forward inheritance & security:
- The more levels of inheritance you introduce the more the runtime/database will have to work to join the tables together. Having many levels will introduce more joins which take longer to combine.   You will only experience issues if your (parent) tables have thousands of records in them. 
- Security, in the basic case we assume there are no xpath constraints specified in the access rules. Having different combinations of NoAccess/Read/ReadWrite specified for the different sub classes is costing some time but that is no different than having a single table with as many attributes an complex attribute access  (this falls under the same anti-pattern of having many different access rules, many different module roles all just with a tiny difference in attribute access = slower retrieves).


What does have the potential to have a large impact is having access rules on the sub classes, but this depends on how you design your application. When I show a grid of all _parent_ in a datagrid then the database/runtime will have to apply every single xpath constraint on all these records. However if you don't query on the _parent_ entity (show it in a grid or query on it in a microflow) then having access rules in inheritance isn't an issue.

MariaShaposhnikova:
@dogancakmak , can you please have a look at @jaspervanderhoek comment? I am not sure what we want to include/delete/rewrite in the document especially since we have something very similar in the MxAssist reccommendation itself.

dogancakmak:
@MariaShaposhnikova I have read the comment from @jaspervanderhoek. The main remark is that this best practice is only useful on large datasets. When I reviewed the documentation again, I see that we already imply the case of having large dataset. We can merge it as is unless there is a better way to express the importance of having large dataset in the text.

@MariaShaposhnikova do you think we already imply having large datasets or do we need to change the wording to emphasize more?

MariaShaposhnikova:
@dogancakmak , thanks! I repharsed it a bit to make it more clear.

PR #3851, opened 2021-11-03
Clarify why one should not pin certs

joshp123:


content/developerportal/deploy/certificates.md
```diff
@@ -33,10 +33,14 @@ You may need to use certificates to set up connections from your Mendix Cloud ap
 
 ## 2 Incoming – Certificates
 
-Certificates for *.mendixcloud.com* and *.mxapps.io* are managed by Mendix and can be updated without notice.
+Certificates for *.mendixcloud.com* and *.mxapps.io* are managed by Mendix and will be updated on a regular basis, without notice. This is part of our regular operations and security posture.
 
 {{% alert type="warning" %}}
```
joshp123, line 38: Is there a "more serious" alert level above warning? The severity is to indicate "this will break your stuff, don't do it".

MarkvanMents: We do have _danger_ which will use a red box. However, I don't think it is used anywhere on the site. I think that _warning_ is probably enough if the text is strong.

content/developerportal/deploy/certificates.md
```diff
@@ -33,10 +33,14 @@ You may need to use certificates to set up connections from your Mendix Cloud ap
...
 Do not pin these certificates in your solutions.
```
MarkvanMents, line 39: Make text stronger by making it bold?
```suggestion
**Do not pin these certificates in your solutions.**
```

content/developerportal/deploy/certificates.md
```diff
@@ -33,10 +33,14 @@ You may need to use certificates to set up connections from your Mendix Cloud ap
...
+
+Doing so can interrupt your operations when Mendix updates these certificates.
+
+Pinning bypasses the built-in certificate chains of your operating system or JVM, and is widely considered a bad practice.
```
MarkvanMents, line 43: Just because a bypass can also be a noun and my mind won't let go of that:
```suggestion
Pinning a certificate bypasses the built-in certificate chains of your operating system or JVM, and is widely considered a bad practice. 
```

content/developerportal/deploy/certificates.md
```diff
@@ -33,10 +33,14 @@ You may need to use certificates to set up connections from your Mendix Cloud ap
 
 ## 2 Incoming – Certificates
 
-Certificates for *.mendixcloud.com* and *.mxapps.io* are managed by Mendix and can be updated without notice.
+Certificates for *.mendixcloud.com* and *.mxapps.io* are managed by Mendix and will be updated on a regular basis, without notice. This is part of our regular operations and security posture.
```
MarkvanMents, line 36: Had to look up _security posture_ but it is actually a thing, so am happy to leave it here unless we can think of something with a bit less jargon.

MarkvanMents (commented):
Made a couple of suggestions and comments. But basically looks good to go.

content/developerportal/deploy/certificates.md
joshp123: I never thought of it like that until you mentioned it, but indeed your suggestion is better!

joshp123:
Thanks for the review, @MarkvanMents, I've incorporated your changes. I'd appreciate a merge when you're able to do so!

MarkvanMents (approved):
Thanks for your work Josh.
Merging, as discussed.

PR #3856, opened 2021-11-04
Additional clarification on certificates

joshp123:
Hey @MarkvanMents, I added some additional lines here on the technical side, and also to emphasise that we will automatically update the certificates (as opposed to manually), which I think should make it clearer to customers that this is the future and is better for them.

Let me know what you think!

Thanks.


MarkvanMents (approved):
Looks good.
Shall I publish?

joshp123:
Yes please @MarkvanMents , thank you!

PR #3887, opened 2021-11-17
Update Version Control documents with Git information

MariaShaposhnikova:


content/refguide/using-version-control-in-studio-pro.md
```diff
@@ -247,7 +271,7 @@ If you have multiple development lines, you sometimes want to port changes from
 
 Merging is always done while you have a working copy open. The merge will result in extra local changes in that working copy. It is advisable to commit local changes first before merging extra changes into a working copy. Otherwise, the uncommitted local changes and the changes caused by the merge will be combined and it is very hard to untangle them if you are unhappy with the merge. Studio Pro will warn you if you have uncommitted changes.
 
-Select **Version Control > Merge Changes Here**, then select the appropriate type of merge (for example, **Port fix**).
+Select **Version Control > Merge Changes Here**, then select the appropriate type of merge (for example, **Port fix**). {{% todo %}}[Decide on whether to leave this example on Port fix and make it clear it's about SVN or delete it.]{{% /todo %}}
```
MariaShaposhnikova, line 274: Three options here:
1. We delete information on Port fix from here, till the end of the section (it is one example). 
2. We leave it and make it clear that this is about SVN. 
3. We leave this example and make it clear it is about SVN and we add a different example for Git.

content/refguide/using-version-control-in-studio-pro.md
```diff
@@ -108,7 +108,21 @@ If you also changed Java source code, added widgets or made other changes that a
 
 Committing is only allowed if your working copy is up to date with the repository. If someone else committed a change since the last time you updated, you will have to update first. This is because the revision you create with the commit should incorporate both your changes and the changes by the other person. Updating will combine the latest changes in the repository with your changes. After reviewing the result, and fixing any conflicts, you can commit again.
 
-### 4.3 Updating
+### 4.3 Pushing
```
MariaShaposhnikova, line 111: Is this information enough?

content/refguide/using-version-control-in-studio-pro.md
```diff
@@ -118,13 +132,13 @@ To update the working copy of your app, click the **Update** button in the **Cha
 
 Changes you receive from the repository when updating are combined with the changes you have made to your working copy (if any). Afterwards, your working copy will contain both your changes and the changes you received. As part of the update, the original of your working copy is updated as well.
 
-For example, if the last time you updated you received all changes up to and including revision 40, this means that the original for your working copy is revision 40. Since you started making changes to your working copy, other people on your team have made another four commits (41, 42, 43 and 44). If you now update, you will receive those changes and 44 will be the new *original* to which your changes are compared.
+For example, if the last time you updated you received all changes up to and including revision N, this means that the original for your working copy is revision 40. Since you started making changes to your working copy, other people on your team have made another four commits (X, Y, and Z). If you now update, you will receive those changes and Z will be the new *original* to which your changes are compared.
```
MariaShaposhnikova, line 135: I added letters to indicate the revision number not to have examples only applicable for Git or SVN. Do you think that works?

content/refguide/using-version-control-in-studio-pro.md
grootjans: Currently only the merging a feature branch works in Git, and port fix will be added soon. Port fix will only be able to port a single commit per time, because in contrast with SVN, individual commits need to be moved, resolved and committed. SVN bundles all changes together and thus allow multiple revisions/commits.
Point I am making is that even when we add port fix, the specifics of this examples do not match anymore.
I would propose:
- Use merge feature branch as an example instead of port fix, and mention with an alert that SVN currently supports port fix, and advanced merge.

content/refguide/using-version-control-in-studio-pro.md
grootjans: I would add the following explanation on top:
Pushing is the act of sending your local changes to Team Server. By default this is done when committing, but it is possible to wait for this step until later.

{alert}
When using SVN, changes are always pushed as part of the commit step.
{alert}

content/refguide/using-version-control-in-studio-pro.md
grootjans: I think that is fine 👍

content/refguide/using-version-control-in-studio-pro.md
```diff
@@ -259,8 +283,14 @@ Reverting changes works for changes that have not been committed yet. Changes th
 
 Choose **Version Control > Reverse Merge Changes...**
 
+For SVN, you can choose a range (**Start revision** and **End revision**):
+
 ![](attachments/using-version-control-in-studio-pro/revert-committed-changes.png)
 
+For Git, you can only revert a single revision:
```
grootjans, line 290: Proposal:
For Git, reverting changes is done one commit at a time. If you want to revert multiple commits, you can do that by reverting the latest commit, then the previous one and so on.

content/refguide/using-version-control-in-studio-pro.md
MariaShaposhnikova: I've implemented it but rewritten it a bit. As for the alert, we already have a note at the beginning of the section that says almost the same, so I suggest keeping only one: Push is only available in Git. In SVN, pushing is done together with committing automatically.

content/refguide/using-version-control-in-studio-pro.md
MariaShaposhnikova: Added

content/refguide/using-version-control-in-studio-pro.md
MariaShaposhnikova: Thanks for the explanation, I was wrongly assuming that these settings would be identical in the future. Indeed, the best way is to mention them separately then. I've written it in a slightly different way, please take a look.

PR #3889, opened 2021-11-17
Add MXP010 documentation to Performance Best Practices page (needs to be published with respect to the Studio Pro 9.9 release - Dec 21, 2021)

pavelkucherov:
MxAssist Client team implemented a new best practice: MXP010. We need to have a documentation regarding this new best practice.
_______________________________________________
MXP010 will be released with Studio Pro 9.9.


jaspervanderhoek:
The statements make sense, in my experience duplication of security rules are usually 90% identical and due to the fact that people attempt to put workflow into their security.

For simple duplication I think this makes sense, an nuance you should consider (and I don't know how easy this is to build in also as check in studio) is how to deal with 90% overlap. Most often I have seen duplication in scenario's such as this.  All 'FirstLineSupport' is allowed to access [tickets_from/department/assigned_to/user='me'] (usually a bit more complex), but what is different is that on [Status='Draft'] they have only read access, on [Status='Submitted'] they can write a few fields, [Status='Accepted'] they can edit more, etc, etc.   This makes it simple for users to not have to think about what fields are editable.

The recommendation for this use-case should be something along these lines. 

- Assess your set of security rules and avoid letting your security model drive your workflow or process rules (the security engine isn't optimized for workflow task assignment). 
   - Consolidate the variation you have in your rules, add additional checks in your microflow to validate state changes rather than having all variations in the access rules
   - Consider splitting your entity in multiple entities with 1-1 associations and cascading delete. On these individual entities you can simplify the access profiles and potential limit access to the entire entity rather than dozens of individual fields.

MariaShaposhnikova:
@jaspervanderhoek , thanks for your input! I implemented your suggestion, but rewrote them a bit. I deleted the "workflow" term from it as well to avoid confusion with the workflow editor. Could you or @pavelkucherov have a look at the changes and comment whether they look good to you?

jaspervanderhoek:
@MariaShaposhnikova Looks good to me

pavelkucherov:
@MariaShaposhnikova Looks good to me

content/refguide/performance-best-practices.md
```diff
@@ -177,3 +177,14 @@ Consider the following alternatives:
 * Combine attributes in one entity and add an enumeration instead of setting the [generalization](generalization-and-association)
 * Create entities with a one-to-one association instead of setting the generalization
 * Create a non-persistable entity that inherits from an outcome of your business logic
+
+## 9 Duplicated Access Rules {#mxp010}
```
mrseanryan, line 181: Please add the code, else will be out of sync with the PR for MXP011 which adds Codes for all the other Best Practices...

## 9 Duplicated Access Rules `MXP010` {#mxp010}

pavelkucherov: @MariaShaposhnikova Could you please add the code ?

pavelkucherov: @mrseanryan Code has been added

MariaShaposhnikova: I don't think we can keep them in-sync, we will end up in conflicts anyway. My plan is: this PR goes first, then I solve conflicts in MXP011 and merge it.

mrseanryan: ok - it looks ok now?

I will close this comment, thank you!
Sean

PR #3899, opened 2021-11-19
[RTI-364] Add ScheduledEventsCleanupAge and ProcessedTasksCleanupAge runtime settings

atennapel:
For Mx9.9.


Adam-Dupaski (approved):

content/refguide/custom-settings.md
```diff
@@ -39,9 +39,11 @@ The following custom settings can be configured:
...
 | **com.mendix.core.isClusterSlave** | Set to `true` in a high-availability scenario when this is *not* the [Cluster Leader](clustered-mendix-runtime#cluster-leader-follower). The buildpack will usually enforce this setting, but it may need to be set for some on-premises deployments. | `false` |
 | **com.mendix.core.SameSiteCookies** | The [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) property can be included in all cookies that are returned by the embedded HTTP server. The possible values are `"Strict"`, `"Lax"`, and `"None"`. The default is `"Strict"`. Setting it to `"None"` is typically needed only when an application is embedded in an iframe of another application with a different domain. Newer browsers may require the connection to be secure (HTTPS) when set to `"None"`. If the connection is plain HTTP, then this setting must be changed to `"Strict"` (recommended) or `"Lax"`. | |
+| **com.mendix.core.ScheduledEventsCleanupAge** | This settings specifies (in milliseconds) how old completed scheduled events can be before they are removed from the database. By default this setting is off. | |
```
hugomendix, line 42: - It sounds a big strange to state that the "setting is off". Maybe rephrase to: "When this setting is not specified, no cleanup is performed."
- Typo: "settings" -> "setting"

hugomendix: The term "scheduled events" might be confusing, because the model also contains scheduled events, which are obviously not removed. We could state more clearly that the `System.ScheduledEventInformation` rows are cleaned up from the database.

content/refguide/custom-settings.md
```diff
@@ -39,9 +39,11 @@ The following custom settings can be configured:
...
 | **com.mendix.core.SessionIdCookieName** | Defines the name of the cookie value which represents the session id. Can be useful to change when running in a container which assumes a certain name for the session cookie. | XASSESSIONID |
 | **com.mendix.core.StorageService** | Defines which storage service module will be used. The storage service module takes care of storing the actual files associated with `System.FileDocument` objects, such as uploaded files. Possible values are `com.mendix.storage.localfilesystem`, `com.mendix.storage.s3`, `com.mendix.storage.azure`, and `com.mendix.storage.swift`. | com.mendix.storage.localfilesystem |
 | **com.mendix.storage.PerformDeleteFromStorage** | Defines whether a delete of a Mendix file document should result in an actual delete in the storage service. A reason to not perform an actual delete in the storage service can be when it is also used as a backup service. | true |
+| **com.mendix.core.ProcessedTasksCleanupAge** | This settings specifies (in milliseconds) how old completed processed tasks can be before they are removed from the database. By default this setting is off. | |
```
hugomendix, line 46: Same comments as for scheduled events.

hugomendix (changes requested):

Adam-Dupaski:
@MarkvanMents please take a look at these comments and further changes requested from Hugo

content/refguide/custom-settings.md
atennapel: Updated.

content/refguide/custom-settings.md
atennapel: Updated.

hugomendix (approved):

MarkvanMents (approved):
Proofread - looks good

content/refguide/custom-settings.md
```diff
@@ -39,9 +39,11 @@ The following custom settings can be configured:
...
 | **com.mendix.core.SameSiteCookies** | The [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) property can be included in all cookies that are returned by the embedded HTTP server. The possible values are `"Strict"`, `"Lax"`, and `"None"`. The default is `"Strict"`. Setting it to `"None"` is typically needed only when an application is embedded in an iframe of another application with a different domain. Newer browsers may require the connection to be secure (HTTPS) when set to `"None"`. If the connection is plain HTTP, then this setting must be changed to `"Strict"` (recommended) or `"Lax"`. | |
+| **com.mendix.core.ScheduledEventsCleanupAge** | This setting specifies (in milliseconds) how old rows in the System.ScheduledEventInformation table have to be before they are removed from the database. Only rows with the "Completed" status are cleaned up. When this setting is not specified, no cleanup is performed. | |
```
hugomendix, line 42: Can we added the following note or alert to this?
> When turning on the cleanup after having used scheduled events for a long time, there might be many events to cleanup, which will done when the runtime server starts. This many cause additional database load, but will not block the startup. It is recommended not to do this during a busy period.

content/refguide/custom-settings.md
```diff
@@ -39,9 +39,11 @@ The following custom settings can be configured:
...
 | **com.mendix.core.StorageService** | Defines which storage service module will be used. The storage service module takes care of storing the actual files associated with `System.FileDocument` objects, such as uploaded files. Possible values are `com.mendix.storage.localfilesystem`, `com.mendix.storage.s3`, `com.mendix.storage.azure`, and `com.mendix.storage.swift`. | com.mendix.storage.localfilesystem |
 | **com.mendix.storage.PerformDeleteFromStorage** | Defines whether a delete of a Mendix file document should result in an actual delete in the storage service. A reason to not perform an actual delete in the storage service can be when it is also used as a backup service. | true |
+| **com.mendix.core.ProcessedTasksCleanupAge** | This setting specifies (in milliseconds) how old rows in the System.ProcessedQueueTask have to be before they are removed from the database. Only rows with the "Completed" status are cleaned up. When this setting is not specified, no cleanup is performed. | |
```
hugomendix, line 46: Can we added the following note or alert to this?

>    When turning on the cleanup after having used many tasks over a long time period, there might be many processed tasks to cleanup, which will done when the runtime server starts. This many cause additional database load, but will not block the startup. It is recommended not to do this during a busy period.

content/refguide/custom-settings.md
MarkvanMents: This seems a lot of information in this table.
Would it be better to add a section to https://docs.mendix.com/refguide/scheduled-events#5-additional-information which explains this setting more thoroughly and then link to that?

content/refguide/custom-settings.md
MarkvanMents: This seems a lot of information in this table.
Would it be better to add a section to https://docs.mendix.com/refguide/task-queue#configuration which explains this setting more thoroughly and then link to that?

MarkvanMents (changes requested):
Hi @hugomendix 
Can you have a look at my suggestions. I feel that this is adding information to this table which could perhaps be explained in the documents which cover these two processes specifically.

I think it might be good to add this information there in any case, as customers might not look here for this information if they are implementing a task queue or scheduled event.

content/refguide/custom-settings.md
hugomendix: Yes, that sounds like a good idea. Still, we want to prevent having an angry customer, because he turned on this setting without reading the disclaimer.

content/refguide/custom-settings.md
hugomendix: Yes, that sounds like a good idea. Still, we want to prevent having an angry customer, because he turned on this setting without reading the disclaimer.

MarkvanMents: Can you update the documents I referenced and add them to the PR? Then I can review them.
Thanks.

atennapel:
Closing in favour of: https://github.com/mendix/docs/pull/3948

PR #3900, opened 2021-11-22
Add automatic retry on task queue to Java API

hugomendix:
As of Mendix 9.9.0 we will support the feature to automatically retry a task, for which execution in the task queue failed.


MarkvanMents (approved):
Looks good.
Have made a couple of proofreading changes. Don't think I've changed the meaning at all.
Thanks for the PR Hugo (@hugomendix )

PR #3907, opened 2021-11-23
removed mentioning datadog in Private Cloud Monitoring docs to preven…

cstevandy:
Updated private cloud monitoring, removed mentioning Datadog to prevent misunderstanding that we only support Prometheus not all Datadog component


MarkvanMents (approved):
Looks good. Merging.

PR #3919, opened 2021-11-29
String function calls proofread

NatasaKralj:
The function `findAll` description and example text need technical verification before merging.


NatasaKralj:
Removed 'Needs R&D technical verification', as refguide 8 version already had it written correctly.
Merging changes.

PR #3867, opened 2021-11-11
Remove the limitation of associated attributes in DB2 and SAP HANA in Mx 9 MTS version.

MarkvanMents:
This assumes that https://github.com/mendix/docs/pull/3792 has been merged first.

PR #3926, opened 2021-11-30
Remove the limitation of associated attributes in DB2 and SAP HANA in Mx 9 MTS version.

MarkvanMents:
Second attempt, first one was lost in the merge of #3908 which accidentally merged #3867 as they both had the same initial commit.


MarkvanMents:
Anchor warnings will go away if this is merged after #3924

PR #3942, opened 2021-12-03
[RTI-338] Add Task Queue Retry to docs

atennapel:


atennapel:
Note: this is no longer for 9.9.0 but for 9.10. I still have to update the docs here.

MarkvanMents (approved):
Added note that these are only available after 9.9.0,
tried to clarify exactly what the options are - particularly for exponential retry.
Thanks for the PR, @atennapel

MarkvanMents:
Is this still in draft, or can I convert it to ready for review?

atennapel:
Thanks Mark, it's still in draft currently. Because we missed the 9.9.0 freeze we are now aiming for 9.10.0. So I still need to update the docs because we have changed some more things.

atennapel:
I have updated the MR for Mx 9.10.

MarkvanMents (approved):
Reviewed changes. Looks good.

PR #3943, opened 2021-12-03
Update some country names

MarkvanMents:
Check [SPX-2323]
Should be in 9.9 (and 9.10) but changes may miss the cutoff. Check before publishing.


Adam-Dupaski:
In RN for 9.9 #3970

MarkvanMents:
Changing back to 9.10.0. From [SPX-2323]

> @Mark van Ments Yes, it is confusing. We first implemented the change in 9.9.0, but then rolled it back until the discussion is finalized.
> The change is no longer part of 9.9.0.

Adam-Dupaski:
This will have to be merged with 9.6.4.

PR #3944, opened 2021-12-06
Nk data grid widgets

NatasaKralj:
Not ready for merge, WIP


content/refguide/control-bar.md
```diff
@@ -29,7 +29,22 @@ The control bar can also include a number of selection options and spreadsheet e
 
 Most properties of control bar buttons are the same as properties of buttons. For more information on button properties, see [Buttons](button-properties).
 
-Sections below describe the purpose of each control bar button and their specific properties if any. 
+To add additional buttons to the control bar, do the following:
+
+1. Right-click on the control bar, while it is highlighed.
```
ConnorLand, line 34: Remove comma.

content/refguide/control-bar.md
```diff
@@ -29,7 +29,22 @@ The control bar can also include a number of selection options and spreadsheet e
...
+1. Select **Add button**.
+1. Select the new type of button.
+1. Double click on the new button to customize.
+
+![Add Button to Control Bar](attachments/data-widgets/control-bar-add-button.png)
+
+{{% alert type="info" %}}
+It is not possible to add additional buttons by dragging them from the **Toolbox** into the control bar.
+
+When setting the visibility of control bar buttons, it is not possible to use currently selected object(s) for conditional visibility expressions, or visibility based on the selected object’s attribute.
```
ConnorLand, line 44: I assume "object(s)" is not UI text, because it is not bold. Try the following changes to write in general, hypothetical terms while also avoiding "object(s)", which I consider to be a break in grammar:

Change **object(s)** to **objects**.
Change **object's** to **objects'**.
Change the last **attribute** to **attributes**.

content/refguide/grids.md
```diff
@@ -15,7 +15,18 @@ Grids have the following components in common:
 * [Sort bar](sort-bar) –  allows end-users to sort items in the grid 
 * [Control bar](control-bar) –  allows end-users to control the objects displayed in the grid by means of buttons 
 
-## 2 Read More
+## 2 Choosing a Grid
+
+Choosing which grid to use, to display data on a page, doesn't have to be difficult. A few things to consider are:
```
ConnorLand, line 20: Unordered lists are better suited for ideas than questions. They also usually do not end in punctuation. Therefore I would rewrite the intro as follows:

"If your app needs to display data on a page, different widgets suit different use cases. Determine what type of data you need to display, imagine how you would like it presented, and then consult the options below:"

Then, under that intro, I'd make unordered lists which show the strength of each option:

The **Data Grid** widget is well suited for these cases: (crosslink to Data Grid widget)

* Displaying data as a table
* Employing several rows and columns
* Displaying simple text data

The **Template Grid** widget is well suited for these cases: (crosslink to Template Grid widget)

* Displaying tiles or widgets with each list item
* Show an image
* Add a button to your list item
* Include a custom layout for your items (crosslink to custom layout instructions if possible)

content/refguide/control-bar.md
ConnorLand: That is a good motivation. I recommend achieving that goal without using **(s)** as it breaks number agreement (it's either "a currently selected object" or "currently selected objects"). 

Just for context, several developers try to employ **(s)** and **and/or** in their pull requests, which I edit out. I suppose I'm trying to maintain a precedent here.

cjfhodges (approved):

PR #3953, opened 2021-12-08
added release note for private cloud portal

cstevandy:
We are planning to release a new version of the private cloud portal on 9th Dec 2021, if everything is fine.


MarkvanMents (approved):
Looks good.

PR #3968, opened 2021-12-13
Add naming constraint on metrics (Mendix 9.10)

hugomendix:
Names of customer metrics cannot start with 'mx'.

### DIFF
--- a/content/developerportal/operate/datadog-metrics.md
+++ b/content/developerportal/operate/datadog-metrics.md
@@ -425,7 +425,7 @@ If you have any issues related to accessing Datadog, please contact their suppor
 
 | Metric | Description |
 | --- | --- |
-| jmx.com.mendix.* | Core runtime metrics |
+| jmx.com.mendix.* | JMX metrics for the `com.mendix` domain (core runtime). |
 | mx.database.diskstorage_size | Disk storage available to the application database (this is a fixed value) |
 | mx.activity.time | How long a microflow activity takes to run |
 | mx.client.time | The time to handle a request to a request handler that is used by the web ui |
--- a/content/refguide/metrics.md
+++ b/content/refguide/metrics.md
@@ -279,7 +279,7 @@ The above filter discards metrics which starts with **"Unnamed."**, **"Invalid."
 The following should be taken into account when configuring the metrics registries.
 {{% /alert %}}
 
-1. There are also some internal metrics used by Mendix. Filters will also have an effect on them. For example metrics emitted by Mendix which starts with "com.mendix." or "mx.".
+1. There are also some internal metrics used by Mendix. Filters will also have an effect on them. For example metrics emitted by Mendix which start with "mx.".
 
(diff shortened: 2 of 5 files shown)


content/refguide/metrics.md
```diff
@@ -279,7 +279,7 @@ The above filter discards metrics which starts with **"Unnamed."**, **"Invalid."
 The following should be taken into account when configuring the metrics registries.
 {{% /alert %}}
 
-1. There are also some internal metrics used by Mendix. Filters will also have an effect on them. For example metrics emitted by Mendix which starts with "com.mendix." or "mx.".
+1. There are also some internal metrics used by Mendix. Filters will also have an effect on them. For example metrics emitted by Mendix which start with "mx.".
```
MarkvanMents, line 282: Have all the com.mendix.* metrics disappeared now?
For example, in my datadog documentation I refer to jmx.com.mendix.* metrics which are for the core runtime. Will they get confused with custom metrics if the customer decides to call them something similar?
Or are they irrelevant to this document?

hugomendix: We have no metrics starting with `com.mendix`. The ones mentioned in the Datadog documentation start with `jmx.com.mendix` and there are the JMX metrics for the `com.mendix` domain which are exposed by Datadog. It's a bit confusing that we have 2 kinds of metrics: the JMX ones and the 'normal' ones. The JMX ones are legacy metrics which are exposed using Java MX beans. For all other metrics we use Micrometer now, which can have different back-ends (Prometheus, InfluxDB, StatsD, etc.). Customers can extend these metrics using the Metric actions described in the other modified pages.

content/developerportal/operate/datadog-metrics.md
```diff
@@ -425,7 +425,7 @@ If you have any issues related to accessing Datadog, please contact their suppor
 
 | Metric | Description |
 | --- | --- |
-| jmx.com.mendix.* | Core runtime metrics |
+| jmx.com.mendix.* | JMX metrics for the `com.mendix` domain (core runtime). |
 | mx.database.diskstorage_size | Disk storage available to the application database (this is a fixed value) |
```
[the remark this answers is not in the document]
MarkvanMents: Would be nice to make the list complete. I documented what I could find in Datadog with what teams could tell me.
The problems are:
1. Getting dev teams to own the metrics and described them accurately. This is as far as I got when I asked on 24 November 2020 in Slack.
2. Being told when metrics are add, changed, or removed.

content/refguide/metrics.md
MarkvanMents: Thanks for the explanation.
I'll leave this as proposed as there doesn't seem to be a single source for finding out all the metrics that are/can be produced.

MarkvanMents (approved):
Thanks for the PR, Hugo (@hugomendix).
I think there is more that can be done to document metrics information (from other sources) - but I'll look at this separately.

PR #3986, opened 2021-12-17
Add docs about Styling Editor

takumagott:
This adds a section about the styling editor, as well as the setting which styling files should be shown inside the App Explorer.

This is for the 9.9 release. Thanks in advance!

Also, I'll be on holiday starting Monday, so if there are questions after today, please reach out to someone else in the team, like Fabian van der Veen.


Adam-Dupaski:
@MariaShaposhnikova let me know if you need assistance on this.

PR #3997, opened 2021-12-20
LM-134: Add File Storage metrics

jeohist:
This documents the upcoming File Storage metrics. It depends on <https://github.com/mendix/cf-mendix-buildpack/pull/488> which is pending review, and would be released together with a future Cloud Portal release. As such, this is a Draft Pull Request.

It also removed the bit about `/srv` which is only applicable to Cloud v3.

### DIFF
--- a/content/developerportal/operate/trends-v4.md
+++ b/content/developerportal/operate/trends-v4.md
@@ -48,6 +48,7 @@ You can find the trends by following these steps:
 1. Open your app in the [Developer Portal](http://sprintr.home.mendix.com).
 2. Click **Metrics**.
 3. Select the environment you want to monitor under the tab **Trends**.
+4. Choose the period for which you want to see the trends: **Day**, **Week**, **Month**, or **Quarter**.
 
 ![](attachments/trends-v4/environment.png)
 
--- a/content/developerportal/operate/trends-v4.md
+++ b/content/developerportal/operate/trends-v4.md
@@ -55,11 +55,7 @@ You can find the trends by following these steps:
 
 As with all complex IT systems, there are many interrelated components which can cause performance issues. This document cannot cover all possibilities, but is intended as a general introduction to the information which is displayed and a few ideas about where to look for possible areas of concern.
 
-#### 3.2.1 Disk Partitions
-
-If there is more than one disk partition in the system, the `/srv` partition generally contains project files and uploaded files of the application, while `/var` generally holds the database storage.
-
-#### 3.2.2 Combining Information{#combine-info}
+#### 3.2.1 Combining Information{#combine-info}
 
 You can often get more information about the performance of your app by combining the information from several graphs. Useful graphs for this are:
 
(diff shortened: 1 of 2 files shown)


content/developerportal/operate/trends-v4.md
```diff
@@ -234,6 +230,22 @@ This graph should be interpreted in combination with other graphs. See [Combinin
 
 The disk usage graph shows only the disk usage inside the container. This is usually only relevant if your application creates a lot of temporary files in `/tmp`. This value is not the same as the file document storage.
 
+### <a name="Trends-appnumberoffiles"></a>4.11 Application Number of Files
+
+The **Application Number of Files** graph shows the number of files that are stored in file storage.
```
MarkvanMents, line 235: Should we say these are FileDocument types? Do customers store other sorts of files here? Is it worth saying that it is S3 storage?

content/developerportal/operate/trends-v4.md
```diff
@@ -234,6 +230,22 @@ This graph should be interpreted in combination with other graphs. See [Combinin
...
+
+![](attachments/trends-v4/app-number-of-files.png)
+
+### <a name="Trends-appsizeoffiles"></a>4.12 Application Size of Files (in bytes)
+
+{{% alert type="info" %}}
+This metric was added on DATE to the [Cloud Foundry Mendix Buildpack](https://github.com/mendix/cf-mendix-buildpack). The data will only be visible if you have deployed your Mendix app after that date, and it will only be visible from the moment of deployment.
```
MarkvanMents, line 242:
```suggestion
This metric was added on DATE to the [Cloud Foundry Mendix Buildpack](https://github.com/mendix/cf-mendix-buildpack). The data will only be visible if you have deployed your Mendix app since that date, and it will only be visible from the moment of deployment.
```

Do you mean to say that the graph will only _show_ the size of files from the moment of deployment and before that it will appear that the files have zero size?

MarkvanMents (commented):
Looks good,
Have a couple of questions/suggestions.
Thanks for the submission, Joey (@jeohist )

content/developerportal/operate/trends-v4.md
jeohist: Yeah, FIleDocument types is more accurate. 
I don't know whether we talk about file storage or S3 storage in Mendix Public Cloud docs elsewhere. I'm fine with it either way, maybe S3 is better to distinguish it more explicitly from local app storage (e.g. `/tmp/`).

content/developerportal/operate/trends-v4.md
jeohist: The graph will not show data prior to the date of deployment, it'll show the 'No Data' graph.

MarkvanMents: I'm still not entirely clear.
Say the metric is implemented on 1 January and I redeploy on 10 January.
1. Before 10 January I see **No Data Available.
2. After 10 January I see the graph (for example for the month) and it shows
    * zero for dates up to 1 January
    * ?? between 1 January and 10 January - presumably the correct data, or is this only captured after the redeploy?
    * The correct values from 10 January onwards

content/developerportal/operate/trends-v4.md
MarkvanMents, line 235: How about:
```suggestion
The **Application Number of Files** graph shows the number of files created by entities which are based on `FileDocument` generalizations, that are stored in S3 file storage.
```

content/developerportal/operate/trends-v4.md
```diff
@@ -48,18 +48,15 @@ You can find the trends by following these steps:
 1. Open your app in the [Developer Portal](http://sprintr.home.mendix.com).
 2. Click **Metrics**.
 3. Select the environment you want to monitor under the tab **Trends**.
+4. Choose the period for which you want to see the trends: **Day**, **Week**, **Month**, or **Quarter**.
```
MarkvanMents, line 51: Suddenly realized that this wasn't documented.
Sorry to hijack your PR 😀

content/developerportal/operate/trends-v4.md
jeohist: The graphs would be empty up to the moment of deployment, because they're not gathered until the metrics are implemented in the buildpack. So you'd see nothing until January 10th, and the correct values after.

MarkvanMents: Thanks for the clarification.
I've tried to write that in the suggestion above. Please commit if you are happy with what I've written, or we can discuss it further if it can be improved.

jeohist:
Hi @MarkvanMents , thanks for the reviews! We have decided not to go ahead with displaying these metrics publicly for the time being, but it'd be a waste to throw away everything we've worked on here. Shall I just close the PR and save the branch for the future?

MarkvanMents:
Hi @jeohist 
If this is going to be resurrected in the next couple of months, then you can just keep the PR open with a Future label. But if it is going to be longer than that, probably best to close the PR and keep the branch for later.
I'll leave the decision up to you.
Since the branch is on the mendix/docs repo, if you don't want the branch to be accidentally deleted, I suggest renaming it and adding DONOTDELETE to the name.
Alternatively, make your own fork where we can't accidentally delete your work 😀

jeohist:
@MarkvanMents works for me, thanks!

MarkvanMents:
Closing in favour of #4323 to resolve merge conflicts on new Docs Stack

PR #3998, opened 2021-12-20
Announce mendix.com certificate renewal

jeohist:
We have renewed the *.mendix.com certificate and will replace it in January 2022.

### DIFF
--- a/content/releasenotes/developer-portal/mendix-cloud.md
+++ b/content/releasenotes/developer-portal/mendix-cloud.md
@@ -12,6 +12,18 @@ For information on the current status of deployment to Mendix Cloud and any plan
 
 ## 2021
 
+### December 20th, 2021
+
+#### Improvements
+
+**Update of `*.mendix.com` TLS certificate**
+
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi Joey (@jeohist )
Last time we did this we mentioned a few more details (https://docs.mendix.com/releasenotes/developer-portal/mendix-cloud#september-27th-2019). Do we want to do this again, or shall we assume that people who are playing with certificates outside browsers know what they are doing?

jeohist:
> Hi Joey (@jeohist ) Last time we did this we mentioned a few more details (https://docs.mendix.com/releasenotes/developer-portal/mendix-cloud#september-27th-2019). Do we want to do this again, or shall we assume that people who are playing with certificates outside browsers know what they are doing?

Those details (those) are still there in the Certificate Transparency logs which we're linking to. It's more convenient for us and customers, as all the information they might need is located in a single place.

MarkvanMents (approved):

PR #4004, opened 2021-12-21
Transformed the example requests into valid json

Maismaus:
The example requests were all wrapped in strings ("") which prevented me from copying it directly into Postman and is also invalid json.


MarkvanMents (approved):
Thanks for the Pull Request @Maismaus 
Looks like a useful change to the documentation.
Mark

PR #4034, opened 2021-12-31
Custom HTTP header

RCatersels:
It tooks us quite a while to figur out why a custom header was missing while the other headers did show fine. Only then did we find out that HTTP headers may not contain an underscore. Would be nice if the Mendix modeler would already check this. Now at least it is in the documentation.
Regards,
Ronald

### DIFF
--- a/content/refguide/call-rest-action.md
+++ b/content/refguide/call-rest-action.md
@@ -130,7 +130,7 @@ The **Password** property defines the password that will be used to authenticate
 
 ### 5.4 Custom HTTP Headers
 
-These headers are added to the HTTP request header. Each custom header is a pair with a key and a value (a microflow expression).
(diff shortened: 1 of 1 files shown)


Adam-Dupaski:
@MarkvanMents I think you've worked on docs for HTTP headers before?

MarkvanMents (approved):

MarkvanMents:
Hi Ronald (@RCatersels)
I've updated your suggestion to be a warning box and included a link to the NGINX documentation.
Thanks very much for suggesting this change. I'm sure other developers will find it very useful.
As far as including a check in Studio Pro, I've confirmed with the developers that this is appropriate documentation, but I suggest you add your suggestion to the [Mendix ideas forum](https://forum.mendix.com/p/ideas) as it is more likely to gain traction there than coming through documentation.
Mark

PR #4041, opened 2022-01-05
Add a note about Japanese translation to the landing page

Luyao-Zhang-1:


MarkvanMents (approved):
Looks fine Luyao(@Luyao-Zhang-1).

Nice to use a partial so we can update the note independently.
We could look more into how the landing page is currently styled and used, but with the switch to the new site and landing page it isn't worth it.

PR #4068, opened 2022-01-13
[MOO-494] Add docs for dangling references

govorov:


content/refguide/offline-first.md
```diff
@@ -65,27 +65,17 @@ Synchronization performed using a UI element (for example, a button or an on-cha
...
+2. There might be objects deleted from the device database since the last synchronization. The upload phase checks which objects have been deleted.
+3. If there are any changed or new file objects, their content is uploaded to the server and stored there temporarily. Each file is uploaded in a separate network request.
+4. All the changed and new objects are sent to the server, and the content of the files is linked to the objects. The server performs referential integrity validation of the objects (see the [Dangling references](#dangling-references) section for more info). The objects are committed to the server database. Information about deleted objects is also sent to the server so the server can delete them from its database too. This step is performed in a single network request.
```
smarki, line 77: I miss the significance of why a single network request is used. I am not sure we want to go into detail, but without extra info it's not helpful to a reader?

content/refguide/offline-first.md
```diff
@@ -65,27 +65,17 @@ Synchronization performed using a UI element (for example, a button or an on-cha
...
+The upload phase executes the following operations:
 
+1. As the local database can be modified only by committing or deleting an object, such an object can be either a new object created while offline, or an existing object previously synced from the server. The upload phase detects which objects have been committed to the local database since the last synchronization. The detection logic differs per synchronization type. For Synchronize all, all committed objects in the local database are checked. For Synchronize objects, all committed objects from the list of selected objects are checked.
+2. There might be objects deleted from the device database since the last synchronization. The upload phase checks which objects have been deleted.
+3. If there are any changed or new file objects, their content is uploaded to the server and stored there temporarily. Each file is uploaded in a separate network request.
```
smarki, line 76: Same as below; I am not sure why the separate network request is usefull if left unexplained in this context.

content/refguide/offline-first.md
```diff
@@ -158,10 +148,37 @@ During the synchronization, changed and new objects are committed. An object's s
...
 
-### 2.6.3 Preventing Synchronization Issues {#prevent-sync-issues}
+#### 2.6.3 Dangling references {#dangling-references}
+
+During the synchronization the server performs referential integrity validation of the new or changed objects that are being synchronized to the server.
+This validation ensures that none of the synchronized objects have associations pointing to an object that exists only in the device database.
+If an association doesn’t satisfy this condition, this is called a dangling reference.
+
+For example, when a committed `City` object refers to an uncommitted `Country` object, synchronizing the `City` object alone will yield an invalid `Country` object reference, which is a dangling reference error.
```
smarki, line 161: Maybe simplify the language more. Does commit has the meaning we thing for a user?
In offline context, commit means saving to the database, do we actually verify danglin references then? 

Is commit the right verb to describe the operation of sync and saving to the database?

content/refguide/offline-first.md
```diff
@@ -158,10 +148,37 @@ During the synchronization, changed and new objects are committed. An object's s
...
+
+A dangling reference error is a modeling error.
```
smarki, line 163: These deserve their own sections. Could we split dangling references and how to avoid them to their own page?

smarki (changes requested):

content/refguide/offline-first.md
smarki: Maybe a database integrity section or something

govorov:
Add info blocks mentioning that deletes are not supported in mx < 9.7.

content/refguide/offline-first.md
govorov: > do we actually verify danglin references then?

The validation happens upon sync only. The wording here is a bit misleading, good point.

> Is commit the right verb to describe the operation of sync and saving to the database?

We call it "commit" everywhere in the docs, there's a nanoflow activity called "Commit", etc. 
That's the wording we use. The concept of committing an object is known to those who read the docs. 
We also explicitly mention that sync works at the database level.

content/refguide/offline-first.md
govorov: A single network request implies that the operation is kind of atomic.
There's also a difference between upload and download, where for the latter we use one request per entity.

I do feel we can elaborate more on the differences between single/multiple requests, but I don't really know where to draw the line to be honest.

content/refguide/offline-first.md
ekrem-aktas: "Commit" is the correct word to use indeed: intended as "saving to the offline db" and here specifies the state of the objects. Dangling reference check does not happen at commit time, but when the committed object is synchronised.

content/refguide/offline-first.md
```diff
@@ -158,10 +148,37 @@ During the synchronization, changed and new objects are committed. An object's s
...
+To prevent them from happening during selective synchronization make sure you either select both sides of the association for the synchronization
```
ekrem-aktas, line 169: The suggestion of selecting both objects wouldn't work because `Country` is stated as uncommitted above.

content/refguide/offline-first.md
```diff
@@ -158,10 +148,37 @@ During the synchronization, changed and new objects are committed. An object's s
...
+To prevent dangling reference errors during full synchronization make sure both sides of the association are committed before synchronizing.
+
+When some of the synchronized objects have dangling references, 
+the server will synchronize all other objects except the ones with dangling references,
+for each of which it will create a synchronization error and store it in
+`System.SynchronizationError` entity. The error message in such cases looks like the following:
+
+**Synchronizing an object of type City with guid ... has failed due to a modelling error. 
+The object has a reference to other objects (City_Country) that have not been synchronised to the runtime yet. 
+This breaks referential integrity of the object because it references a non-existing object in the runtime database.
```
ekrem-aktas, line 180:
```suggestion
This breaks referential integrity of the object because it references an object that only exists on the device database.
```

content/refguide/offline-first.md
```diff
@@ -65,27 +65,17 @@ Synchronization performed using a UI element (for example, a button or an on-cha
...
-4. <a name="stepthree"></a>All the changed and new objects are committed to the server, and the content of the files are linked to the objects. Information about deleted objects is also sent to the server so the server can delete them from its database too. This step is performed in a single network request. Any configured before- or after-commit or before- or after-delete event handlers on these objects will run on the server as usual: after the data has been uploaded and before it is downloaded.
+The upload phase executes the following operations:
 
+1. As the local database can be modified only by committing or deleting an object, such an object can be either a new object created while offline, or an existing object previously synced from the server. The upload phase detects which objects have been committed to the local database since the last synchronization. The detection logic differs per synchronization type. For Synchronize all, all committed objects in the local database are checked. For Synchronize objects, all committed objects from the list of selected objects are checked.
```
ekrem-aktas, line 74: > all committed objects from the list of selected objects are checked.

This is a leftover from dangling reference description: we don't "check" anything anymore. But we can say "included" to mention which objects are selected for synchronization. 

We also don't seem to mention "Synchronize unsynced objects" mode.

content/refguide/offline-first.md
ekrem-aktas: We can continue with: "If upload of a file fails, the whole synchronization is aborted without causing any changes in both the server and local database.

content/refguide/offline-first.md
```diff
@@ -65,27 +65,17 @@ Synchronization performed using a UI element (for example, a button or an on-cha
...
+2. There might be objects deleted from the device database since the last synchronization. The upload phase checks which objects have been deleted.
+3. If there are any changed or new file objects, their content is uploaded to the server and stored there temporarily. Each file is uploaded in a separate network request.
+4. All the changed and new objects are sent to the server, and the content of the files is linked to the objects. The server performs referential integrity validation of the objects (see the [Dangling references](#dangling-references) section for more info). The objects are committed to the server database. Information about deleted objects is also sent to the server so the server can delete them from its database too. This step is performed in a single network request.
+5. Any configured before- or after-commit or before- or after-delete event handlers on these objects will run on the server as usual: after the data has been uploaded and before it is downloaded.
```
ekrem-aktas, line 83: * We also execute create event handlers for new objects.
* "... and before local database is updated. This means any further changes you make to the synchronised objects in the event handlers will be updated in the local database in the download phase"

There's one exception to the description above - changing the contents of a file entities aren't reflected back. Maybe important to mention.

content/refguide/offline-first.md
ekrem-aktas: +1 for a separate section, but we have more stuff to put under such a section. That's why I'd do it in a separate story.

ekrem-aktas (changes requested):

content/refguide/offline-first.md
smarki: Exactly that is my point. We know that; will a customer understand that?

smarki: I think a more explanatory sentence; like the operation happens in one request and that's the reason why, similar to Ekrem's suggestion is aggreable with me.

content/refguide/offline-first.md
govorov: ✔️

content/refguide/offline-first.md
```diff
@@ -65,27 +65,17 @@ Synchronization performed using a UI element (for example, a button or an on-cha
 
 ### 2.2 Synchronization Phases
 
-The synchronization process consists of two phases. In the [upload phase](#upload), your app updates the server database with the new or changed objects that are committed. In the [download phase](#download), your app updates its local database using data from the server database.
+The synchronization process consists of two phases. In the [upload phase](#upload), your app updates the server database with the new or changed objects that are committed. In the [download phase](#download), your app updates its local database using data from the server database. Note that synchronization only works on the database level.
```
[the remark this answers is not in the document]
govorov: ✔️

content/refguide/offline-first.md
govorov: Good point. Fixed ✔️

content/refguide/offline-first.md
govorov: This text is exactly what the runtime logs in the console. Do you suggest to change it there as well? The existing version is fine with me, as it's the message shown by the runtime, so "non-existing" is not ambiguous in this case.

content/refguide/offline-first.md
govorov: It's not about dangling in any way, it's about selecting what data to synchronize indeed. If you find "checked" controversial, we can replace it with "selected for synchronization" for instance.

govorov: "Check" here means "scan the db and pick what needs to be synced". "Checked" and "included" for sync are not the same sets: an object can be "checked" (entire database in case of full sync) but not "included" (as we include only dirty) for sync.

Alternalively, we can say something like "for full sync all ditry objects are included". What do you think?

content/refguide/offline-first.md
govorov: Nice catch Ekrem 👍 Added.

content/refguide/offline-first.md
ekrem-aktas: Yes, I'd also change it on the logs as well - but noticed that my version needs adjustment - we need the drop "database" from it because a dangling object can be a new one that has not yet been committed to the offline database.

content/refguide/offline-first.md
ekrem-aktas: Indeed I misunderstood the usage of the check here - with your explanation is makes sense. I'd keep it as-is.

content/refguide/offline-first.md
```diff
@@ -158,10 +155,38 @@ During the synchronization, changed and new objects are committed. An object's s
...
 * The object is not valid according to domain-level validation rules
+* The object has a dangling reference (see the [Dangling references](#dangling-references) section for more info)
 
 {{% alert type="warning" %}}When a synchronization error occurs because of one the reasons above, an object's commit is skipped, its changes are ignored, and references from other objects to it become invalid. Objects referencing such a skipped object (which are not triggering errors) will be synchronized normally. Such a situation is likely to be a modeling error and is logged on the server. To prevent data loss, the attribute values for such objects are stored in the `System.SynchronizationError` entity (since Mendix 8.12).  {{% /alert %}}
 
-### 2.6.3 Preventing Synchronization Issues {#prevent-sync-issues}
+#### 2.6.3 Dangling references {#dangling-references}
+
+During the synchronization the server performs referential integrity validation of the new or changed objects that are being synchronized to the server.
+This validation ensures that none of the synchronized objects have associations pointing to an object that exists only in the device database.
```
ekrem-aktas, line 165: I think we should drop "database" from the end of the sentence.

ekrem-aktas (changes requested):

content/refguide/offline-first.md
govorov: ✔️

content/refguide/offline-first.md
govorov: Let's do this as a separate change set after we change the message in the runtime then. Created a small story.

ConnorLand:
@stmarkidis have changes been made to your satisfaction?

smarki (approved):

ConnorLand (approved):

PR #4081, opened 2022-01-17
Update yarn.lock

MarkvanMents:
Some packages can be updated.
However, there are many which cannot, but these will be removed with the stack update.


MarkvanMents:
Tested in local build.

PR #4084, opened 2022-01-18
9.10 RN

Adam-Dupaski:


content/releasenotes/studio-pro/9.10.md
```diff
@@ -33,7 +33,7 @@ description: "The release notes for Mendix Studio Pro version 9.10 (including al
...
+With this release, we are making this functionality more simple and powerful. From now on, XPath sources of a [list view](/refguide/list-view) widget, [Data Grid 2](/appstore/modules/data-grid-2) widget, and many other widgets have access to the same variables as page expressions! That means XPath constraints can refer to the objects of all surrounding data widgets as well as to their attributes. For example, your list view can show all `Module.Kittens` satisfying the `[Cuteness > $ownerDataView/CutenessThreshold]` constraint. Such a list view will be automatically updated whenever the mentioned `CutenessThreshold` is changed. This makes creating XPath constraints easier and more powerful while reducing the need for microflow or nanoflow-based sources. For more information, see the [XPath Constraint](/refguide/xpath-source#xpath-constraints) section of *XPath Source*.
```
Adam-Dupaski, line 36: @ConnorLand the hyphen is now required after "microflow" as well if you are going to break those up: "microflow- or nanoflow-based sources"

Adam-Dupaski: @ConnorLand I don't think this breakup was necessary, to be clear.

ConnorLand: Good correction. I will add the hyphen. 

I used the breakup instead of the slash because I try and use words over punctuation when possible. I see several instances of punctuation instead of words in my PRs, and oftentimes words are clearer.

The slash is certainly clear though. If you like the slash better, I'm happy to change it back.

Adam-Dupaski: @ConnorLand fair enough, but it's debatable whether this is in fact clearer. Previously, "microflow/nanoflow" functioned as one idea/grouping for a data source, and now they are differentiated, when the differences between a microflow and nanoflow as a data source don't actually matter here AFAIK. So in the end, you have a more complicated clause that requires more punctuation and slows the reader down because it's less minimal. We can keep it in this case, though.

PR #4085, opened 2022-01-18
Corrected time limitation for Unlicensed Apps.

tknippenberg:
Previously mentioned Sleep Mode is only applicable for 'Free' (but licensed) apps. For unlicensed apps there is no Sleep Mode.


MarkvanMents (approved):
Thanks Tom (@tknippenberg ).
I also checked with Private Cloud, and they confirmed that the customer will identify this behaviour on their K8s pods.
Approving and publishing.

PR #4089, opened 2022-01-19
Add more info on missing alerts

joshp123:


joshp123:
FYI @udh-mendix

joshp123:
@MarkvanMents - would you mind taking a quick look, please? Thanks!

content/developerportal/operate/monitoring-application-health.md
```diff
@@ -44,6 +44,8 @@ Any status that could cause your app to stop functioning will result in an alert
 
 There are a number of categories on the status overview that are related to various technical parts of your application environment. Any category that does not display as *OK* needs to be investigated. Each category indicates a different part of the app infrastructure that is monitored.
 
+Some alerts listed in the table below only appear when the alert is triggered. Therefore, an alert not being present in the overview is the same as the alert having an *OK* status.
+
 ### 3.1 Application Status
```
MarkvanMents, line 49: @joshp123 This looks a bit weird, it looks like there should be a `### 3.n` for each of the tables, not just Application Status. Do you agree?

content/developerportal/operate/monitoring-application-health.md
```diff
@@ -44,6 +44,8 @@ Any status that could cause your app to stop functioning will result in an alert
 
 There are a number of categories on the status overview that are related to various technical parts of your application environment. Any category that does not display as *OK* needs to be investigated. Each category indicates a different part of the app infrastructure that is monitored.
 
+Some alerts listed in the table below only appear when the alert is triggered. Therefore, an alert not being present in the overview is the same as the alert having an *OK* status.
```
MarkvanMents, line 47: @joshp123 - A presume you mean all the tables below, not just the first (Application Status) one?
```suggestion
Some alerts listed in the tables below only appear when the alert is triggered. Therefore, an alert not being present in the overview is the same as the alert having an *OK* status.
```

MarkvanMents (commented):
@joshp123 
A couple of minor questions, but otherwise fine.

content/developerportal/operate/monitoring-application-health.md
joshp123: Hmm, it is indeed a bit weird. I think the idea is to reflect the actual alerts page, which has no subsections itself, and looks like this:



So, I would suggest leaving this as is for now.

joshp123:
Thanks for the review, @MarkvanMents ! I have replied to your comments.

content/developerportal/operate/monitoring-application-health.md
MarkvanMents: OK - thanks for the explanation.

MarkvanMents (approved):
Shall I publish this immediately?

joshp123:
Yes please @MarkvanMents , and thank you!

PR #4103, opened 2022-01-21
Datetime fields are not supported as keys

ako:
Call out fact that date and datetime fields cannot be used in odata entity keys.


Adam-Dupaski (approved):

Adam-Dupaski:
@johannahemminger for reference

PR #4130, opened 2022-01-29
Update basic-package.md

TimvanSteenbergen:
Apparently no documentation exists yet on deploying the app for a Basic package, see https://forum.mendix.com/link/questions/112701. So I created it, hope you like it.

### DIFF
--- a/content/developerportal/deploy/basic-package.md
+++ b/content/developerportal/deploy/basic-package.md
@@ -84,22 +84,12 @@ To order your Mendix Basic package, do the following:
 
 After 48 hours, your Mendix Basic package will be provisioned as a new licensed node. You will receive an email at the address entered in your order informing you that this has happened and linking you to the documentation on how to use the node.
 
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi @TimvanSteenbergen 
These instructions are already linked from [Migrate a Free App to Mendix Basic](https://docs.mendix.com/developerportal/deploy/migrate-free-app-to-basic#3-6-deploy-your-app), where section 3.6 links to the same set of instructions in the [Mendix Cloud - deploy](https://docs.mendix.com/developerportal/deploy/mendix-cloud-deploy#3-deploying-a-licensed-app-to-the-mendix-cloud) documentation.

I don't really want to duplicate the same set of instructions in multiple places if I can help it. Do you think that a link to that section of the Mendix Cloud deploy might be a good alternative?

TimvanSteenbergen:
Crap, I have overlooked that doc and link. Could have saved me some time there, by better rtfm-ing. Totally agree: no duplicate content. So: yes a link will do.

MarkvanMents:
Thanks for picking up the question on the forum and responding positively by helping with documenting the solution. Sorry that you ended up having to do more work in this case.

I'll use this PR to make the link to the existing documentation stand out more, as that was obviously a weakness in this case.

MarkvanMents (approved):
Thanks for the submission Tim.

PR #4147, opened 2022-02-03
Preparation for new Docs Stack

MarkvanMents:
Add no_list: false for pages which currently list child pages automatically
Add values for cascading space:, mendix_version:, and priority: to each space.

Based on #4144, which should be merged first.


NatasaKralj (approved):
Looks good

PR #4155, opened 2022-02-07
Made heading #3 more appropriate

syedaliammar:
The heading #3 otherwise looked like it listed the operation possible on external entities, not the ones affected when using them.


Adam-Dupaski:
@johannahemminger please do a language review here

PR #3086, opened 2021-02-25
[WC-97] Add documentation for rewritten badge widget

IIsaku:
The badge widget has been converted to a pluggable widget (v2.0.0) that will only be supported on MX 9 and above. I kept the old widget documentation as well, since mx developers can still use these widgets with mx 7 or 8.

PR #4178, opened 2022-02-15
[WC-1017] Update rich text documentation

raffiar:
Updated documentation for Rich text editor

### DIFF
--- a/content/appstore/widgets/rich-text.md
+++ b/content/appstore/widgets/rich-text.md
@@ -15,8 +15,18 @@ The [Rich Text](https://marketplace.mendix.com/link/component/74889/) widget ena
 * Format selected text
 * HTML output of formatted text
 * Show editor options either on a toolbar or as a bubble
-* Use the custom option to select which editing options you want to show
+* Use the toolbar tab to select which editor preset you want to show
+	* **Basic** - This package contains plugins suitable for quick input fields.
+	* **Standard** - This package contains plugins that can be used for creating standards-compliant content.
+	* **Full** - This heavier package contains plenty of plugins suitable for various different needs.
+	* **Custom** - The simplest way to configure the toolbar as you want.
 * Input and display text is sanitized
+* Use **Enter and Shift+Enter** mode to configure how you want to wrap your content
+	* Paragraph (`<p>`)
+	* Break lines (`<br>`)
+	* Blocks (`<div>`)
+* Use highlight code feature to insert rich code fragments and see a live preview with highlighted syntax.
+* Use advanced filter content to allow or disallow elements in your rich text or use auto configuration.
 
 ## 2 Configuration
 


content/appstore/widgets/rich-text.md
```diff
@@ -15,8 +15,18 @@ The [Rich Text](https://marketplace.mendix.com/link/component/74889/) widget ena
 * Format selected text
```
Keraito, line 15: Since this is a major update, the old and the new widget are quite different. What we normally do is move all the old documentation towards a new section https://github.com/mendix/docs/pull/3086/files#diff-7b7d5624620be27259d245760138c730cd5b21948ff22edff78d54f8154458dfR32 and write new ones for the new version. But you don't necessarily have to write everything from scratch. Stuff that applies from previous version (like intro and stuff) you can just keep).

content/appstore/widgets/rich-text.md
```diff
@@ -15,8 +15,18 @@ The [Rich Text](https://marketplace.mendix.com/link/component/74889/) widget ena
...
+	* **Standard** - This package contains plugins that can be used for creating standards-compliant content.
+	* **Full** - This heavier package contains plenty of plugins suitable for various different needs.
+	* **Custom** - The simplest way to configure the toolbar as you want.
 * Input and display text is sanitized
+* Use **Enter and Shift+Enter** mode to configure how you want to wrap your content
+	* Paragraph (`<p>`)
+	* Break lines (`<br>`)
+	* Blocks (`<div>`)
+* Use highlight code feature to insert rich code fragments and see a live preview with highlighted syntax.
+* Use advanced filter content to allow or disallow elements in your rich text or use auto configuration.
```
Keraito, line 29: Think this looks good, you can prob copy this "Features" section to the new documentation

content/appstore/widgets/rich-text.md
```diff
@@ -15,8 +15,18 @@ The [Rich Text](https://marketplace.mendix.com/link/component/74889/) widget ena
...
 
 ## 2 Configuration
```
Keraito, line 31: For the new "Configuration" section, we normally go over the different widget properties + tabs and discuss them one by one so the user can exactly know what properties configures what. Also stuff like default options and conditional properties.

Examples (check the "configuration" sections): https://docs.mendix.com/refguide/data-grid (slightly more high level: talking about what it results in) and  https://docs.mendix.com/appstore/modules/tree-node (really low level: talking about every property and what it does). Either approach is fine, although I would recommend the second one. The first one is more for complex widgets where there is a lot to explain, while the second approach is fine for less complex widgets

Keraito (changes requested):
Good start, but we have to align a bit with other docs. I've included a few links so you know how it should look or what you can replicate.

Keraito:
@Luyao-Zhang-1 @Adam-Dupaski There are still some changes necessary here (from tech side + to align with other docs), we'll let you know when this is ready to be reviewed by the docs team.

content/appstore/widgets/rich-text.md
```diff
@@ -15,7 +15,79 @@ The [Rich Text](https://marketplace.mendix.com/link/component/74889/) widget ena
 * Format selected text
 * HTML output of formatted text
 * Show editor options either on a toolbar or as a bubble
-* Use the custom option to select which editing options you want to show
```
Keraito, line 18: You accidentally removed this line from the prev docs, can you add it back again? 😄

Keraito (approved):

content/appstore/widgets/rich-text.md
Keraito: Resolved

content/appstore/widgets/rich-text.md
Keraito: Resolved

content/appstore/widgets/rich-text.md
Keraito: Resolved

Keraito:
@Luyao-Zhang-1 all yours now 😄

PR #4201, opened 2022-02-17
[WC-1025] Update carousel documentation

raffiar:
@ConnorLand Mentioning you because of our previous discussion.
This pr needs technical review

### DIFF
--- a/content/appstore/widgets/carousel.md
+++ b/content/appstore/widgets/carousel.md
@@ -38,17 +38,17 @@ The following sections will describe the different available widget properties a
 
 ### 3.2 Display
 
-* **Pagination** - show pagination dots
+* **Pagination** – shows pagination dots
 	* Default: **Yes**
-* **Navigation controls** - show arrows on the left and right side
+* **Navigation controls** – shows arrows on the left and right side
 	* Default: **Yes**
-* **Autoplay** - allow images to start sliding automatically when the widget is ready
+* **Autoplay** – allows images to start sliding automatically when the widget is ready
 	* Default: **Yes**
-* **Delay** - creates a time interval between one item cycling to the next
+* **Delay** – creates a time interval between one item cycling to the next
 	* Default: **1000ms**
-* **Infinite loop** - cycle between images endlessly
+* **Infinite loop** – cycles between images endlessly
 	* Default: **Yes**
-* **Animation** - add a fade animation on each cycle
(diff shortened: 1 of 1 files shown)


content/appstore/widgets/carousel.md
```diff
@@ -12,6 +12,57 @@ The [Carousel](https://marketplace.mendix.com/link/component/47784/) widget disp
...
+* Supports widget as a content
+* Navigate to the next or previous image using arrows or dots
+* Autoplay your images
+* Set the timeout between each slide
+* Use the infinite loop to scroll the slides endlessly
+* Add animation when switching slides
+
+## 2 Usage
+
+The widget requires a context configured on the **Data Source** section in **General** tab. The possible data sources are described below.
```
diego-antonelli, line 24: Sounds a bit confusing the context of a data source.. Maybe we should change to `The widget requires a defined **Data source** available in the **General** tab.`

diego-antonelli: But I guess that is for docs team. Technical check is OK

ConnorLand: I like the way you phrased it in the comment @diego-antonelli . Change is done.

content/appstore/widgets/carousel.md
```diff
@@ -12,6 +12,57 @@ The [Carousel](https://marketplace.mendix.com/link/component/47784/) widget disp
 
 ### 1.1 Features
 
+* Allows you to use the widget as a content area
```
ConnorLand, line 15: Is "content area" correct @raffiar ? I didn't know what "as a content" meant.

raffiar: hmm, that's a good question. What I meant is that you can put any widget in carousel. Like the slide can be not only image (widget) but also text (widget)
What is the best way to explain that?

diego-antonelli: I think we use the term "Composable region"

https://docs.mendix.com/appstore/widgets/tooltip#1-1-trigger
https://docs.mendix.com/appstore/modules/tree-node#general

raffiar: maybe it's better to say `Allows you to use a widget as an item content`?

ConnorLand: I've resolved it by using both language suggestions from Raffi and Diego.

ConnorLand (approved):

PR #4211, opened 2022-02-21
[WC-1016] update color picker documentation

raffiar:
This pr needs technical review

### DIFF
--- a/content/appstore/widgets/color-picker.md
+++ b/content/appstore/widgets/color-picker.md
@@ -8,88 +8,108 @@ tags: ["marketplace", "marketplace component", "widget", "color picker", "react"
 
 ## 1 Introduction
 
-The [Color Picker](https://marketplace.mendix.com/link/component/107044/) enables selecting and adjusting a color value.
+The [Color Picker](https://marketplace.mendix.com/link/component/107044/) enables selecting and adjusting a color value that is stored in a string attribute
 
 ### 1.1 Features
 * Render using the button, input box, and inline modes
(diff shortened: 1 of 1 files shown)


content/appstore/widgets/color-picker.md
```diff
@@ -10,6 +10,59 @@ tags: ["marketplace", "marketplace component", "widget", "color picker", "react"
 
 The [Color Picker](https://marketplace.mendix.com/link/component/107044/) enables selecting and adjusting a color value.
 
+### 1.1 Features
+* Render using the button, input box, and inline modes
```
diego-antonelli, line 14:
```suggestion
* Render using button, input box or inline modes
```

Up to docs reviewer in this case

diego-antonelli (approved):

Luyao-Zhang-1:
@MariaShaposhnikova I think the cross-references in 2.4, 2.5, and 2.6 should link to the the sections in this doc [Properties Common in the Page Editor](https://docs.mendix.com/refguide/common-widget-properties#common-properties), but could you please help to confirm?

MariaShaposhnikova:
@Luyao-Zhang-1 , I see you are referring to tabs instead of sections here. In the refguide, we refer to the Properties pane and then sections in the pane. You can also double-click a widget and will see a dialog box with properties and tabs, so this document is referring to the second option. Properties might be in different order in these two options and sometimes even do not match 100% but I think this is only in the microflow editor. Please double-check.

PR #4221, opened 2022-02-22
Update index.md

georgetasopoulos:
Improvement point added for Release 2022.02.24

### DIFF
--- a/content/releasenotes/developer-portal/index.md
+++ b/content/releasenotes/developer-portal/index.md
@@ -19,7 +19,8 @@ To see the current status of the Mendix Developer Portal and Control Center, see
 ### February 24th, 2022
 
 #### Improvements
-* In all StudioPro versions older than 9.6, GIT enabled projects will not be available. In newer versions of StudioPro, both GIT and SVN projects are supported.
+
+* The apps displayed on the **My Apps** tab that appears when starting Studio Pro are pulled from the [My Apps](/developerportal/#my-apps) page in Developer Portal. In Studio Pro versions below [9.6.0](/releasenotes/studio-pro/9.6), the **My Apps** tab will no longer display [Git for version control](/refguide/branch-line-manager-dialog#byo-server-app) apps, as Git integration is not supported. In Studio Pro versions 9.6.0 and above, both Git-enabled and SVN-enabled apps are displayed.
 
 ### February 10th, 2022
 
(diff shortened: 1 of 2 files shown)


content/releasenotes/developer-portal/index.md
```diff
@@ -16,6 +16,11 @@ To see the current status of the Mendix Developer Portal and Control Center, see
 
 ## 2022
 
+### February 24th, 2022
+
+#### Improvements
+* In all StudioPro versions older than 9.6, GIT enabled projects will not be available. In newer versions of StudioPro, both GIT and SVN projects are supported.
```
Adam-Dupaski, line 22: Hi @georgetasopoulos I am not sure I understand the relation here between these notes and Developer Portal. Where are these apps "available" and where will they appear?

georgetasopoulos: Hi Adam. Every user has his/her apps' overview in the developer portal. This overview includes both GIT and SVN enabled projects (these are the only 2 options when you create a new project). This overview is also available in the start-up screen of StudioPro so that the users can choose which app to work on. What we changed is that we will no longer display GIT projects in StudioPro 9.5 and lower, because they don't support GIT integration. In these older versions, only SVN projects should be visible. For newer versions (>=9.6), both SVN and GIT projects should be visible. I hope it is more clear now. If not, please let me know.

Adam-Dupaski: It's clear, and it's clear that these apps displayed in Studio Pro are pulled from Dev Portal (https://docs.mendix.com/developerportal/#my-apps).

But is this also relevant for https://docs.mendix.com/developerportal/#my-apps ?

Adam-Dupaski: Actually I guess not, since https://docs.mendix.com/developerportal/#my-apps displays all apps...

Adam-Dupaski: Updated the note, please take a look.

georgetasopoulos: Thanks Adam, the text looks ok. However I'm not able to open the hyperlinks. Is this normal? 

image

Adam-Dupaski: Thanks will be ready to publish. The relative links don't work directly from GitHub, but they will work on publishing.

Adam-Dupaski (approved):

PR #4251, opened 2022-03-03
Clarify use of entity access in queued tasks.

reinouts:
See Slack thread here: https://mendix.slack.com/archives/C3T260XGA/p1646300113098139?thread_ts=1646131531.687969&cid=C3T260XGA

### DIFF
--- a/content/refguide/task-queue.md
+++ b/content/refguide/task-queue.md
@@ -170,9 +170,9 @@ If you are using Mendix version 9.9.0 or above, you can use the `com.mendix.core
 
 ### 2.7 Execution Context {#context}
 
-Prior to Mendix 9.6 tasks were always executed in a *sudo* context, even if the scheduling microflow had **Apply entity access** set to *true* (see [Microflow Properties](microflow) for more information). As of Mendix 9.6 this behavior has been deprecated and tasks now run in an equivalent context to the one in which they were scheduled. This has the following effect:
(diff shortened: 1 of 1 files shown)


content/refguide/task-queue.md
```diff
@@ -172,9 +172,16 @@ If you are using Mendix version 9.9.0 or above, you can use the `com.mendix.core
 
 Prior to Mendix 9.6 tasks were always executed in a *sudo* context, even if the scheduling microflow had **Apply entity access** set to *true* (see [Microflow Properties](microflow) for more information). As of Mendix 9.6 this behavior has been deprecated and tasks now run in an equivalent context to the one in which they were scheduled. This has the following effect:
 
-* When a user is logged in, the task will be executed in a new context for the same named user. This context will be the same as if the user is logged in.
+* When the microflow corresponding to the task has **Apply entity access** set to *false*, the task will be executed in a *sudo* context.
```
sayyildiz, line 175: This is already a property of Microflow and already documented here: https://docs.mendix.com/refguide7/microflow#5-2-apply-entity-access . The problem was that the previous implementation of "Task Queue" was ignoring this. Thankfully now it handles correctly.

I think we don't have to explicitly mention it here as it is not related (at least directly) with Task Queue. Also `sudo` context, etc. are internal technical details of Mendix. IMHO we should try to keep those out of public documentation. I really don't know why we used that term in this document before. I just searched for it in the public documentation and I saw that this is the only page that contains `sudo` context.

reinouts: Addressed suggestions by @sayyildiz

MarkvanMents (approved):
@reinouts 
Looks good. I didn't see any discussion of the Java API in the Slack thread, but I assume this has been confirmed by runtime.
I'll merge this immediately as it is current documentation.
Thanks for making this PR.

PR #4256, opened 2022-03-04
Replace Mendix App with Mendix Developer App

jdiehl:
We plan to eventually replace the Mendix App with a new app for collaboration. Since this new app will no longer contain any functionality to test hybrid apps (and this features seems broken in the existing app), we want to point Makers to our Mendix Developer App instead. This update changes the documentation accordingly.


ConnorLand:
Thanks very much @jdiehl . I think the Mendix Developer App might have been under-documented in the past. It looks like this PR fills in some of those gaps.

I'll look through and see if further changes -- such as screenshot updates -- are needed. 

One question; I see the Mendix Developer App does use PhoneGap to test hybrid apps. Does this functionality still work?

jdiehl:
I have tested the App on Android and iOS with a Hybrid Mobile app built in Mendix 8 and both worked. On the other hand, I could not get the Mendix App to work at all (it always crashes when I scan a QR code). Therefore, I believe it is best to point our customers to the other app.

jdiehl:
@ConnorLand I have rebased this pull request on the new release

ConnorLand (approved):

PR #4268, opened 2022-03-08
Consume insertable

joostverhoog:
Studio Pro 9.12 can consume insertable external entities, change them, and send them to be inserted into the external system


johannahemminger (approved):

johannahemminger:
@joostverhoog I see what you mean. I omitted it because I thought it would read a bit better and that it was implied that that was just one possibility, but it can be confusing. I would like to use "for example" instead of "for instance" just to make it a bit more accessible and because that's what we tend to use in our writing, sound good?

johannahemminger:
These changes are being replicated in https://github.com/mendix/docs/pull/4320 to work with the new stack structure.

PR #4290, opened 2022-03-16
Move translation note down into alert

NatasaKralj:


MarkvanMents (approved):
Looks good.
I wonder if there is now too much padding above the note? I think we have two lots of padding there?
But that is a minor issue. I'll approve it.

NatasaKralj:
Thanks for noticing that Mark. I made the top margin smaller. Merging.

PR #4271, opened 2022-03-08
Backport Add new runtime setting UseMimeDecoderForBase64 to 9.6.10

fazzdev:
Backport of changes which are documented for 9.12.0
This is dependent on changes made in #4220 and assumes that 9.12.0 is released before 9..6.10

PR #4298, opened 2022-03-16
Backport Add new runtime setting UseMimeDecoderForBase64 to 9.6.10

MarkvanMents:
Replaces #4297 and #4271 as #4297 would not build properly.


MarkvanMents:
Closing in favour of  #4297 - now seems to be OK

PR #4317, opened 2022-03-17
Provide more details on access rule evaluation

jds-mendix:


sinteur (approved):
Documents the behaviour of new objects properly.

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
```diff
@@ -121,3 +121,17 @@ Because of this XPath constraint, the access rule only applies to orders for whi
...
+
+### 3.1 New Objects
+
+When a new object is created, or when a new object is sent to the runtime server as part of a request, all XPath constraints are assumed to evaluate as `true`. This evaluation result is stored in memory and valid for the lifetime of the request. Committing the object does _not_ lead to access rules or XPath rules being re-evaluated.
+
+### 3.2 Persisted Objects
+
+When a persistable object is passed to the runtime server as part of the request, the object is retrieved from the database. At that time the access rules are evaluated based on the values retrieved from the database. As for new objects the result of this access rule evaluation is stored in memory and not changed for the lifetime of the object or request. In particular changes to attributes, or committing of the object does not cause re-evaluation of access rules.
```
MarkvanMents: I find this confusing.
If I change a persistable object in the client and then ask the runtime to commit the change, I don't want the runtime server to read the object from the database. This would overwrite my changes.
I assume it will only read it from the database _after_ it has committed any changes that I have made to the object.

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
```diff
@@ -121,3 +121,17 @@ Because of this XPath constraint, the access rule only applies to orders for whi
...
+
+Access rules are abstract descriptions of access rights. To apply them they need to be evaluated. Given a user with certain user roles and the state of the database it can be determined if an access rule applies. The Mendix runtime stores the result of access rule evaluations in memory. In general, this evaluation happens on retrieval of objects from the database. The results will stay valid for the lifetime of the object, which is usually the request. Access rules are evaluated differently depending on the object state. More details about their evaluation when accessed through a user context are given below.
+
+### 3.1 New Objects
+
+When a new object is created, or when a new object is sent to the runtime server as part of a request, all XPath constraints are assumed to evaluate as `true`. This evaluation result is stored in memory and valid for the lifetime of the request. Committing the object does _not_ lead to access rules or XPath rules being re-evaluated.
+
+### 3.2 Persisted Objects
```
MarkvanMents: There is nothing about Non-persistable objects here.
Is this because there are no rules on non-persistable objects?
If there are access rules for NP objects, I think we should mention it.

MarkvanMents (changes requested):
Hi Jan-David @jds-mendix 

I have made a couple of comments. Can you see if they make sense and whether they need any update to the text.

If no changes are needed, this is ready to be published.

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
jdsalchow: You are absolutely right, this is incomplete. What actually happens is that the client sends "changes" and an object id. The request is than processed as follows:

1. the object is retrieved from the database according to the id and its access rules are evaluated
2. changes, if any, are applied to the object

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
jdsalchow: NP objects cannot have XPath constraints, so those don't need to be evaluated

jdsalchow: It would not hurt mentioning that here I guess ;)

jdsalchow:
Hi @MarkvanMents, I have  responded. Shall I make changes or do you want to?

MarkvanMents:
> Hi @MarkvanMents, I have responded. Shall I make changes or do you want to?

Perhaps better if you make the changes? Especially the one about when the object is retrieved and how the access rules are then applied? Then I'm sure it is technically correct.

But let me know in the unlikely event that you are struggling to find the words 😀.

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
```diff
@@ -121,3 +121,21 @@ Because of this XPath constraint, the access rule only applies to orders for whi
...
+
+Access rules are abstract descriptions of access rights. To apply them they need to be evaluated. Given a user with certain user roles and the state of the database it can be determined if an access rule applies. The Mendix runtime stores the result of access rule evaluations in memory. In general, this evaluation happens on retrieval of objects from the database. The results will stay valid for the lifetime of the object, which is usually the request. Access rules are evaluated differently depending on the object state. More details about their evaluation when accessed through a user context are given below.
+
+### 3.1 New Objects
+
+When a new object is created, or when a new object is sent to the runtime server as part of a request, all XPath constraints are assumed to evaluate as `true`. This evaluation result is stored in memory and valid for the lifetime of the request. Committing the object does _not_ lead to access rules or XPath rules being re-evaluated.
+
+### 3.2 Persistable Objects
```
jds-mendix: @MarkvanMents I actually meant `Persisted Objects`, i.e. those actually stored in the database, as opposed to `New Objects`. I think this division also works though

MarkvanMents: Ah, I see. I'm so used to correcting people who use Persistable wrongly I didn't see the difference.

I think this might be too subtle for the reader - perhaps I need to say **Objects Stored in the Database**.
```suggestion
### 3.2 Objects Stored in the Database
```

MarkvanMents (approved):
Great update. It's much clearer to me now.
I have suggested a change to the heading to reflect your intention. Can you review/commit that changes, then I can publish this.
Thanks for your work in explaining this.
Mark

jds-mendix:
Nice suggestion @MarkvanMents. Can we merge it?

MarkvanMents:
Merging.
Thanks for all your work.

PR #4318, opened 2022-03-18
Update trends-v4.md

RCatersels:
5.7 Database Node Disk Throughput does not currently state the unit of measurement. Assumption is (since all the other are in Bytes) that the unit of measurement is Bytes here. But this should be checked and adjusted if it is not in Bytes.
Regards,
Ronald


jeohist (approved):

MarkvanMents:
Thanks for another useful submission, Ronald (@RCatersels )
The developers have confirmed that this is, indeed, the value in Bytes. I will merge this change immediately.
Mark

PR #4330, opened 2022-03-21
Add list of metrics

hugomendix:
Documenting all Mendix metrics explicitly.

### DIFF
--- a/content/en/docs/refguide/runtime/metrics.md
+++ b/content/en/docs/refguide/runtime/metrics.md
@@ -382,6 +382,100 @@ Below are deprecated usages which will be removed in the future releases,
 
 Metering-related log messages are sent to the `Metering` log node. If a registry is enabled, they will be reported with severity `debug`.
 
-## 7 Read More
+## 7 List of Metrics
+
+The Runtime Server produces various metrics. Some of these metrics are controlled by Mendix: these are prefixed with `mx`.
+Other metrics are produced by Micrometer, the library that is used for metrics. This library will output metrics for other
+libraries that it recognizes, such as the Jetty server that is embedded in the Runtime Server. 
+
+{{% alert color="info" %}}
+The additional Micrometer metrics are not under our control and might change unannounced over time.
+{{% /alert %}}
+
+### 7.1 Runtime Server Metrics
+
+The Runtime Server produces the following metrics out-of-the-box:
+
+| Name | Type | Tags | Description |
+| --- | --- | --- | --- |
+| `mx.runtime.stats.handler_requests` | counter | `XASId`, `name` | The total number of requests on a request handler (`name`) that was received by a node (`XASId`) since it was started. |
+| `mx.runtime.stats.requests{path}` | counter | | The total number of requests on a request handler (`path`) that was received by a node; deprecated, use the previous one instead. |
+| `mx.runtime.stats.sessions.named_users` | gauge | | The current number of active, named users in the database. |
+| `mx.runtime.stats.sessions.named_user_sessions` | gauge | | The current number of sessions in the database for named users. |
+| `mx.runtime.stats.sessions.anonymous_sessions` | gauge | | The current number of sessions in the database for anonymous users. |
+| `mx.runtime.stats.connectionbus.transactions` | counter | `XASId` | The total number of transactions on the database that were created by a node (`XASId`) since it was started. |
+| `mx.runtime.stats.connectionbus.selects` | counter | `XASId` | The total number of `SELECT` statements that were executed on the database by a node (`XASId`) since it was started. |
+| `mx.runtime.stats.connectionbus.inserts` | counter | `XASId` | The total number of `INSERT` statements that were executed on the database by a node (`XASId`) since it was started. |
+| `mx.runtime.stats.connectionbus.updates` | counter | `XASId` | The total number of `UPDATE` statements that were executed on the database by a node (`XASId`) since it was started. |
+| `mx.runtime.stats.connectionbus.deletes` | counter | `XASId` | The total number of `DELETE` statements that were executed on the database by a node (`XASId`) since it was started. |
+| `mx.odata.retrieve` | counter | `entity` | The total number of OData requests that were sent for a particular type of object (`entity`). |
+| `mx.odata.publish.objects` | counter | `entity` | The total number of OData objects that were served for a particular type of object (`entity`). |
+
+{{% alert color="info" %}}
+Note that the actual name may vary slightly per back-end (e.g. Prometheus replaces dots by underscores).
+Additionally, a suffix may be added for the unit of the metric (e.g. `_bytes`).
+{{% /alert %}}
+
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/runtime/metrics.md
```diff
@@ -382,6 +382,100 @@ Below are deprecated usages which will be removed in the future releases,
...
+Additionally, a suffix may be added for the unit of the metric (e.g. `_bytes`).
+{{% /alert %}}
+
+### 7.2 Additional Metrics
+
+#### 7.2.1 System and Process Metrics 
+
+| Name | Type | Description |
+| --- | --- | --- |
+| `system.cpu.count` | gauge | The number of processors available to the Java Virtual Machine. |
```
MarkvanMents, line 426: Real processors or virtual processors?
Does it make a difference?

content/en/docs/refguide/runtime/metrics.md
```diff
@@ -382,6 +382,100 @@ Below are deprecated usages which will be removed in the future releases,
...
+| `system.cpu.usage` | gauge | The recent CPU usage of the system the Runtime Server is running on (in the range [0..1]). |
+| `system.load.average.1m` | gauge | The average number of runnable (queued and running) threads on the available processors. |
+| `process.cpu.usage` | gauge | The recent CPU usage for the Java Virtual Machine process (in the range [0..1]). |
+
+#### 7.2.2 Java Virtual Machine Metrics 
+
+| Name | Type | Tags | Description |
+| --- | --- | --- | --- |
+| `jvm_info` | gauge | `vendor`, `runtime`, `version` | Details about the Java Virtual Machine (represented in the tags, the value is always 1). |
```
MarkvanMents, line 435: The value of what is always `1`? Not sure what this means.

content/en/docs/refguide/runtime/metrics.md
```diff
@@ -382,6 +382,100 @@ Below are deprecated usages which will be removed in the future releases,
...
+| `jvm.classes.loaded` | gauge | | The number of classes that are currently loaded in the Java Virtual Machine. |
+| `jvm.classes.unloaded` | counter | | The total number of classes unloaded since the Java Virtual Machine has started execution. |
+| `jvm.buffer.count` | gauge | `id` | An estimate of the number of buffers in a Java Virtual Machine memory pool (`id`). |
```
MarkvanMents, line 438: memory pool or **buffer** pool?
I assume you mean that you can identify which buffer using `id`?

content/en/docs/refguide/runtime/metrics.md
```diff
@@ -382,6 +382,100 @@ Below are deprecated usages which will be removed in the future releases,
...
+| `jvm.buffer.memory_used` | gauge | `id` | An estimate of the memory (in bytes) that the Java Virtual Machine is using for a buffer pool (`id`). |
+| `jvm.buffer.total_capacity` | gauge | `id` | An estimate of the total capacity (in bytes) of the buffers in a buffer pool (`id`). |
+| `jvm.memory.used` | gauge | `id`, `area` | The amount of memory (in bytes) used for a certain area (`area`): `heap` or `nonheap`. |
```
MarkvanMents, line 441: What is `id` here?

content/en/docs/refguide/runtime/metrics.md
```diff
@@ -382,6 +382,101 @@ Below are deprecated usages which will be removed in the future releases,
...
+| `mx.runtime.stats.connectionbus.deletes` | counter | `XASId` | The total number of `DELETE` statements that were executed on the database by a node (`XASId`) since it was started. |
+| `mx.odata.retrieve` | counter | `entity` | The total number of OData requests that were sent for a particular type of object (`entity`). |
+| `mx.odata.publish.objects` | counter | `entity` | The total number of OData objects that were served for a particular type of object (`entity`). |
+
+{{% alert color="info" %}}
+Note that the actual name may vary slightly depending on the back-end (for example, Prometheus replaces dots by underscores).
+Additionally, a suffix may be added for the unit of the metric (for example, `_bytes`).
+{{% /alert %}}
+
+### 7.2 Additional Metrics
```
MarkvanMents, line 416: Are any of these especially useful to the customer, or do we just leave it to the customer to decide what they want to look at?

MarkvanMents (approved):
Hi Hugo (@hugomendix)
Looks like some great information here.
I have asked a couple of questions, but the PR is ready to go if you don't think anything needs changing.

Let me know if I can publish.

Thanks for the contribution.

Mark

content/en/docs/refguide/runtime/metrics.md
hugomendix: It doesn't make a difference, I think. It's the number of processors on the system that is hosting the JVM. If that's a virtual machine then it's virtual processors and if it's bare metal then it's real processor (cores).

content/en/docs/refguide/runtime/metrics.md
hugomendix: That this is a gauge that does not change. It's numeric value is irrelevant and always 1. The relevant information (which also doesn't change during execution) is in the tags of the gauge. In actual fact, these are just some string constants identifying the JVM.

content/en/docs/refguide/runtime/metrics.md
hugomendix: I copied this verbatim from the Micrometer documentation. The code effectively calls some Java MX bean: https://docs.oracle.com/en/java/javase/11/docs/api/java.management/java/lang/management/BufferPoolMXBean.html#getCount()

Unfortunately, the documentation for that isn't much clearer:
> Returns an estimate of the number of buffers in the pool.

hugomendix: The `id` tag is the name of the buffer pool. From the same page:

> Returns the name representing this buffer pool.

content/en/docs/refguide/runtime/metrics.md
hugomendix: The name of the buffer pool, as used by the JVM.

content/en/docs/refguide/runtime/metrics.md
hugomendix: No, I don't think they're very useful to the customer. They're simply implementation details that leak out through the usage of the Micrometer library.

hugomendix:
AFAIC you can publish this, unless you want to make some change in response to my asnwers.

content/en/docs/refguide/runtime/metrics.md
MarkvanMents: OK - makes sense.

MarkvanMents:
> AFAIC you can publish this, unless you want to make some change in response to my asnwers.

Thanks for the answers. I don't think I need to change anything. If the customer is looking at this, we have to assume a reasonable amount of knowledge (i.e. more than I have). I'll publish this now.

PR #4342, opened 2022-03-22
Multi-language support in Studio

MariaShaposhnikova:

PR #4343, opened 2022-03-22
Copy information on Studio multi-language support

MariaShaposhnikova:
This is a copy of all changes made in https://github.com/mendix/docs/pull/4342.


content/en/docs/studio/general/language-support.md
```diff
@@ -0,0 +1,53 @@
...
+
+{{% /alert %}}
+
+## 2 Switching the Currently Selected Language
+
+You can switch the selected language in the top menu bar:
+
+![Switching Languages](attachments/language-support/selected-language.png)
+
+## 3 Translating Texts
```
MariaShaposhnikova, line 26: Comments from Arjan:

- Enumeration texts can also be translated

- I think we should add a screenshot of the translation dialog + explanation of what is on top (currently selected language in top bar) and how to toggle through the others.

- I think it make sense to have a screenshot with Default language on top + alternative below + third one in carroussel. And then explain what the default means, and what the <some text> means. And what you will see in the Preview / Published app. And that when you change the default, all other languages will fall back to that text, unless specifically specified

- Maybe we should also explain that newly added widgets like buttons will have a default caption populated by us in English as place holder and the developer is responsible to  provide the real meaningful text + translations if necesary

MariaShaposhnikova: @arjan-h , I incorporated some of your feedback:

- Enumeration texts can also be translated
  Added
- I think we should add a screenshot of the translation dialog + explanation of what is on top (currently selected language in top bar) and how to toggle through the others.  I think it make sense to have a screenshot with Default language on top + alternative below + third one in carroussel. 
   Added a small diagram to one of the steps
-  And that when you change the default, all other languages will fall back to that text, unless specifically specified
   Added
- And then explain what the default means, and what the means. 
  I explain the default language at the beggining of section 3. Could you please explain what I should add to it?
- And what you will see in the Preview / Published app. 
  Could you provide a use case you have in mind? 
- Maybe we should also explain that newly added widgets like buttons will have a default caption populated by us in English as place holder and the developer is responsible to provide the real meaningful text + translations if necesary
 Does not this explanation cover this point? "When working in a language which is not the default, you can identify texts that have not been translated yet. They show the text in the default language between angle brackets. For example, <Name>. You can replace the text with the appropriate translation for the currently selected language."

PR #4354, opened 2022-03-24
Updated OpenShift versions supported by Mx4PC (no release date)

zlogic:
This is a small change not linked to any release, and can be published at any moment.

A few customers are asking if/when newer Kubernetes and OpenShift versions are supported.
We now have confirmation that Kubernetes 1.23 and OpenShift 4.10 are supported out of the box, with no changes required in Mx4PC components.


MarkvanMents (approved):
Thanks Dmitrii - I'll merge this immediately.

PR #4367, opened 2022-03-29
Clean up file comments and update to Docsy v0.1.0

NatasaKralj:
I changed comment formatting in html partials from html comments to go template comments.
I adjusted comments and contents in SCSS files to match Docsy v0.1.0.
I added a fix for the height of headers, to allow for selection of items above them.
I adjusted layouts files to match Docsy v0.1.0 templates.

### DIFF
--- a/layouts/docs/baseof.print.html
+++ b/layouts/docs/baseof.print.html
@@ -4,11 +4,11 @@
     {{ partial "head.html" . }}
   </head>
   <body class="td-{{ .Kind }}">
-<!-- MvM - deactivation of original navbar - as in baseof.html
+{{/* MvM - deactivation of original navbar - as in baseof.html
     <header>
       {{ partial "navbar.html" . }}
     </header>
--->
+*/}}
     <div class="container-fluid td-outer">
       <div class="td-main">
         <div class="row flex-xl-nowrap stretch-main">
--- a/layouts/docs/baseof.print.html
+++ b/layouts/docs/baseof.print.html
@@ -3,7 +3,7 @@
   <head>
     {{ partial "head.html" . }}
   </head>
-  <body class="td-{{ .Kind }}">
+  <body class="td-{{ .Kind }}{{ with .Page.Params.body_class }} {{ . }}{{ end }}">
 {{/* MvM - deactivation of original navbar - as in baseof.html
     <header>
       {{ partial "navbar.html" . }}
@@ -12,10 +12,6 @@
     <div class="container-fluid td-outer">
       <div class="td-main">
         <div class="row flex-xl-nowrap stretch-main">
-          <div class="col-12 col-md-3 col-xl-2 td-sidebar d-print-none">
-          </div>
-          <div class="d-none d-xl-block col-xl-2 td-toc d-print-none">
-          </div>
           <main class="col-12 col-md-9 col-xl-8 pl-md-5" role="main">
             {{ block "main" . }}{{ end }}
             {{ partial "print/print-footer.html" . }}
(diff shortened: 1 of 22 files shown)


layouts/partials/navbar.html
```diff
@@ -1,7 +1,7 @@
-<!-- {{ $cover := .HasShortcode "blocks/cover" }}
+{{/* {{ $cover := and (.HasShortcode "blocks/cover") (not .Site.Params.ui.navbar_translucent_over_cover_disable) }}
```
MarkvanMents, line 4: Should we have a comment here?
Do we actually need this, or have we commented out every call to navbar and therefore we don't have to overwrite the Docsy version at all?

layouts/partials/print/render.html
```diff
@@ -54,18 +54,18 @@
 
 {{ partial "print/page-heading.html" . }}
 
-<!-- MvM: Remove ToC from the printed section pages
+{{/* MvM: Remove ToC from the printed section pages
```
MarkvanMents, line 57: **Not related to comment style!**
~This doesn't seem to have worked - need to rethink which comments need to be in HTML comments and which in GoTemplate comments.~
image

MarkvanMents (dismissed):
Seems to be a different issue. Unrelated to comment style. Ignore comment below.

~Need to review whether we are doing comments right. See comments below, but one of the changes certainly hasn't worked.~
~I think that the {{ /* comments work in documents, but in partials/shortcodes the templater doesn't recognize them? Or perhaps having {{ inside {{ messes things up.~

~We should probably go back to using <!-- comments in most places, and then review what gets built to see if we need to change anything?~

layouts/partials/print/render.html
MarkvanMents: Oh no - wasn't that? Perhaps I'm misremembering. It is the other way around on the current site 😅 so that wasn't quite right either.
But doesn't look like it is a comment issue, so I take that back.
image

layouts/docs/baseof.print.html
```diff
@@ -3,19 +3,15 @@
...
+{{/* MvM - deactivation of original navbar - as in baseof.html
     <header>
       {{ partial "navbar.html" . }}
     </header>
--->
+*/}}
     <div class="container-fluid td-outer">
       <div class="td-main">
         <div class="row flex-xl-nowrap stretch-main">
-          <div class="col-12 col-md-3 col-xl-2 td-sidebar d-print-none">
```
MarkvanMents, line 15: These are the lines which were making the print gutter appear to the left rather than the right!

layouts/docs/baseof.print.html
```diff
@@ -3,19 +3,15 @@
...
-          </div>
-          <div class="d-none d-xl-block col-xl-2 td-toc d-print-none">
-          </div>
           <main class="col-12 col-md-9 col-xl-8 pl-md-5" role="main">
```
MarkvanMents: This displays the print version on the whole screen without empty columns on either side
```suggestion
          <main class="col-12 col-md-12 col-xl-12 pl-md-5" role="main">
```
Can you try this out and see if it looks OK for you.

MarkvanMents (commented):
Found the strange behaviour with the print partial

layouts/docs/baseof.print.html
NatasaKralj: Fixed with `<main class="col-12 pl-md-5" role="main">`

layouts/partials/navbar.html
NatasaKralj: If we don't comment it out, we have to add 3 more files from Docsy into layouts to comment out the call to navbar.

NatasaKralj: Added a comment to the top of the file for more information.

NatasaKralj:
I fixed the last 2 partials to be updated to Docsy v0.1.0.
I also changed the Edit button property to take the user to the file blob on GH, but the edit page directly.

NatasaKralj:
Approved via slack, merging

PR #4380, opened 2022-03-30
[WC-1106] Add note to strict CSP incompatible widgets or modules 

Keraito:
Per discussed with @ConnorLand on Slack


Keraito:
This should be released after or together with the 9.12 CSP docs

ConnorLand:
> This should be released after or together with the 9.12 CSP docs

And the documentation on how to fix these issues will come later this week, right?

Keraito:
> And the documentation on how to fix these issues will come later this week, right?

For the ones that we know how to, yes

ConnorLand (approved):

ConnorLand (approved):

PR #4399, opened 2022-04-01
Add menu scrolling

NatasaKralj:
The changes in bootstrap class and SCSS styling add a vertical scroll bar to the left and right menus. If the content fits in the menu section, the scroll bar is not visible.


NatasaKralj:
Merging, collaborated and tested with Mark.

PR #4412, opened 2022-04-05
Removed the project ID explanation

Maismaus:
This is confusing because Project ID and App ID are used interchangeably in the API documentations. In the Deploy API, the Project ID is actually what is here referred to as the App ID, but the App ID is something completely different. Added an additional explanation on the first use of the term to clarify.

Perhaps it would be good to align those terms between the two APIs.


Adam-Dupaski (approved):

Adam-Dupaski:
Thanks for the help @Maismaus

PR #4420, opened 2022-04-07
Headings overlap with MxDock

NatasaKralj:
Fixes overlap of anchor heading with MxDock.
Also fixes not being able to use horizontal scroll on tables, when there's overlap with heading height.
The last commit is just a bit of cleanup needed in shortcode (used wrong formatting).

### DIFF
--- a/assets/scss/_tables.scss
+++ b/assets/scss/_tables.scss
@@ -7,6 +7,8 @@
   margin-bottom: $spacer;
   color: $table-color;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
+  position: relative; // NK - added to allow movement of horizontal scroll bar, when overlaping with header below
+  z-index: 2; // NK - added to allow movement of horizontal scroll bar, when overlaping with header below
 
   th,
   td {
--- a/assets/scss/main.scss
+++ b/assets/scss/main.scss
@@ -27,8 +27,8 @@
     h5[id]:before {
         display: block;
         content: " ";
-        margin-top: -48px; // NK - set back to default to prevent MxDock from going on top of headings
-        height: 48px; // NK - set back to default to prevent MxDock from going on top of headings
+        margin-top: -48px; // NK - set to height of MxDock, to prevent dock from going on top of headings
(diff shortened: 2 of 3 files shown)


assets/scss/main.scss
```diff
@@ -27,8 +27,8 @@
     h5[id]:before {
         display: block;
         content: " ";
-        margin-top: 0; // NK - override for default values, to allow selection of items right above headers
-        height: 0; // NK - override for default values, to allow selection of items right above headers
+        margin-top: -5; // NK - set back to default to prevent MxDock from going on top of headings
+        height: 5; // NK - set back to default to prevent MxDock from going on top of headings
```
MarkvanMents: Did we decide to make this 48px to match the height of MxDock?
```suggestion
        margin-top: -48px; // NK - set back to default to prevent MxDock from going on top of headings
        height: 48px; // NK - set back to default to prevent MxDock from going on top of headings
```

assets/scss/_tables.scss
```diff
@@ -7,6 +7,8 @@
   margin-bottom: $spacer;
   color: $table-color;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
+  position: relative; // NK - added to allow movement of horizontal scroll bar, when overlaping with header below
+  z-index: 2; // NK - added to allow movement of horizontal scroll bar, when overlaping with header below
```
MarkvanMents, line 11: Seems to work well.
I was worried about scrolling code blocks, but they seem to work OK.
We can use the same solution if we find anything else which is obscured by headings.

MarkvanMents (commented):
Looks good.
Can you just have a look and see if you agree with my comments?

assets/scss/main.scss
NatasaKralj: Yes, we did agree on this, slipped my mind. I will adjust and test.

assets/scss/_tables.scss
NatasaKralj: The change also helps code block horizontal scrolling, as they also use tables to render.

PR #4424, opened 2022-04-07
Document comparison with empty values

jds-mendix:


content/en/docs/refguide/modeling/application-logic/expressions/relational-expressions.md
```diff
@@ -201,6 +210,9 @@ Possible values are described in the table below:
 {{% alert color="info" %}}
 The two values should be in the same category (for example, if one is numeric, the other should be numeric too).
 {{% /alert %}}
+{{% alert color="info" %}}
+A comparison with an empty value evaluates as `true` if and only if the other value is `empty`, is an empty list, or is and empty reference set.
```
hugomendix: I don't think 'reference set' is a term that is used in Studio Pro. It's merely an implementation details for the runtime, where we distinguish between lists of objects (`IMendixObject`) and lists of references (`IMendixIdentifier`). From a Studio Pro developer's point of view there's no distinction: it's just a `List` of a certain entity type.
And there's a typo: 'and empty` -> 'an empty` but doesn't matter after anymore after removing that part ;-)

jds-mendix: Screenshot 2022-04-07 at 16 27 57

Reference set seems to be used in studio pro. If it's useful or not to say this here I don't know. Do you have an opinion @MarkvanMents ?

I've fixed the typo

MarkvanMents: Could we change this to "an empty association"? Or is there an association there, but it has a zero-length list of objects?
What would an empty association imply, that EntityA is not associated with (any) EntityB?

jdsalchow: We're talking about a 'zero-length list of objects' here. I'm not sure empty association would make people think of that.

MarkvanMents: OK - so an empty reference set is just an empty list of objects. In which case, I think we can ignore reference sets and just say:
```suggestion
A comparison with an empty value evaluates as `true` if and only if the other value is `empty` or is an empty list.
```
After all it isn't the "reference set" itself which is empty, that is just a type of association, it is the fact that there aren't any objects using the association which is key.

Does that look OK to you @jds-mendix and @hugomendix ?

jdsalchow: Great :)

MarkvanMents: Committed. Are you OK to publish @hugomendix ?

hugomendix: Yes. Go ahead...

PR #4428, opened 2022-04-08
Adjust code block line height and button location

NatasaKralj:
The changes cover the following:

- move the language and copy buttons to the upper left of the code block
- adjust line height so no text gets cut off

I could not get highlighting to work on the entire row, still needs looking into.


MarkvanMents (approved):
Looks great. Thanks for doing that.

PR #4436, opened 2022-04-11
Add warning about cookies for SAML module

feddovanede:
Same warning as for Deeplink module. Also see https://mendixsupport.zendesk.com/agent/tickets/146659


MarkvanMents:
@Luyao-Zhang-1 - I've given this to you as it is a Marketplace module, but please let me know if you have any questions.
The text seems to be a copy of text which is already published in https://docs.mendix.com/appstore/modules/deep-link/#36-configuring-constants.

feddovanede:
@MarkvanMents @Luyao-Zhang-1 Yes, it is indeed a copy of the text in the Deep Link module documentation.

PR #4437, opened 2022-04-11
Correct bad external links

MarkvanMents:
And ignore ones that we know are going to return a bad response


NatasaKralj (approved):
Everything looks good

PR #4440, opened 2022-04-11
Removed NotRunning text

Maismaus:
The original text is incorrect. You can transport a deployment package while an environment is running.


MarkvanMents:
Waiting for confirmation from the development team that this is correct.

MarkvanMents (approved):
@Maismaus 
Hi Marius,
Thanks for picking this up. I've had confirmation that this is correct, and added a bit more information to help users understand what the API call does and what steps they may want to take next.
Thanks for making the Pull Request, it is great to have help from customers to make Mendix documentation better.
Mark.

Maismaus:
Glad to help @MarkvanMents! 🙏

PR #4444, opened 2022-04-11
Introducing the deletable capability

joostverhoog:
Studio Pro 9.13.0 introduces the Deletable capability for published OData resources. This is the last of the 4 CRUD capabilities (Create, Read, Update, Delete)

### DIFF
--- a/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
+++ b/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
@@ -225,7 +225,7 @@ The *inserting objects* functionality was introduced in Studio Pro [9.12.0](/rel
 
 ## 11 Deleting Objects {#deleting-objects}
 
-When a published resource has the [Deletable](/refguide/published-odata-resource/#capabilities) capability, clients can update its attributes and associations by sending a `DELETE` request to the URL of the object (for example, `PATCH /odata/myservice/v1/Exployees(8444249301330581)`).
+When a published resource has the [Deletable](/refguide/published-odata-resource/#capabilities) capability, clients can delete an object by sending a `DELETE` request to the URL of the object (for example, `PATCH /odata/myservice/v1/Exployees(8444249301330581)`).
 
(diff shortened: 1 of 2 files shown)


johannahemminger:
Hey @joostverhoog , thanks for the submission. In published-odata-resource.md, I made a few small corrections and integrated the "Call a Microflow Instead of Write to Database" section with the other sections. Let me know what you think!

content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
```diff
@@ -221,3 +221,12 @@ Clients can only set values for an association from the entity that is the [owne
 
 {{% alert type="info" %}}
 The *inserting objects* functionality was introduced in Studio Pro [9.12.0](/releasenotes/studio-pro/9.12/).
+{{% /alert %}}
+
+## 11 Deleting Objects {#deleting-objects}
+
+When a published resource has the [Deletable](/refguide/published-odata-resource/#capabilities) capability, clients can update its attributes and associations by sending a `DELETE` request to the URL of the object (for example, `PATCH /odata/myservice/v1/Exployees(8444249301330581)`).
```
johannahemminger: @joostverhoog this line, "clients can update its attributes and associations by sending a `DELETE` request". Should it be "clients can _delete_ its attributes and associations"? Or is that not exactly correct?

joostverhoog: No neither of those options would be correct. Instead of "clients can update its attributes and associations" it should say "clients can delete an object".

johannahemminger: OK, will update to that!

johannahemminger:
Hey Joost, I see what you mean. I think in this case it's a small amount of text in the same doc, so will leave it as is, but if it gets wordier in the future I'll try to cut it down.

johannahemminger (approved):

PR #4448, opened 2022-04-12
Remote entities are now called external entities

basdebakker:
What we used to call remote entities are now called external entities


paulmendix:
Looks good to me! I approve

PR #4465, opened 2022-04-14
Update intelligent-document-service.md

mahendra-mahajan-mendix:
Removed the image size 50mb statement.

### DIFF
--- a/content/en/docs/appstore/app-services/intelligent-document-service.md
+++ b/content/en/docs/appstore/app-services/intelligent-document-service.md
@@ -19,8 +19,7 @@ The [Intelligent Document Service](https://marketplace.mendix.com/link/component
 
 ### 1.2 Limitation
 
-* Currently only supports images in JPG and JPEG formats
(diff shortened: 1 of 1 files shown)


content/en/docs/appstore/app-services/intelligent-document-service.md
```diff
@@ -79,7 +78,18 @@ To use the Intelligent Document Service, first [train a model](#document-model-t
 
 ### 4.1 Training a Document Model {#document-model-training}
 
+#### 4.1.1 Quick Tips {#quick-tips}
```
Luyao-Zhang-1, line 81: Could we put the info of these tips right below the related steps? It would be easier for new users to follow.

Luyao-Zhang-1: @mahendra-mahajan-mendix Could we put the info of these tips right below the related steps? It would be easier for new users to follow.

PR #4323, opened 2022-03-18
LM-134: Add File Storage metrics

MarkvanMents:
Replacing #3997  in new Docs Stack structure

PR #4485, opened 2022-04-21
Release notes for LM-134, LM-752 & LM-449

ajinkya-takle:
Release notes for changes that were made as of part of LM-134, LM-752 & LM-449

### DIFF
--- a/content/en/docs/releasenotes/deployment/mendix-cloud.md
+++ b/content/en/docs/releasenotes/deployment/mendix-cloud.md
@@ -14,15 +14,14 @@ For information on the current status of deployment to Mendix Cloud and any plan
 ## 2022
 
 ### April 22nd, 2022
-#### New feauture:
-* We have added two new graphs on the Metrics page for Cloud V4:
-   * Number of files in the storage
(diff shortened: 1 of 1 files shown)


MarkvanMents (changes requested):
Hi @ajinkya-takle ,
1. I don't see any updates to the documentation for https://docs.mendix.com/developerportal/operate/trends-v4/. I think that if there are new graphs, we should have them in the documentation and tell customers what they mean. For examples, are there any limits to the numbers or size of files which they can have on storage that they need to be aware of.
2. Can you sign the CLA - it isn't so important for Mendix employees, but means that it is clearer on our dashboard what needs to be done.

Can you let me know what is being done about updating the documentation?

ajinkya-takle:
Hi @MarkvanMents,
Thank you for reviewing & fine tuning the releasing notes for us.
For the new graphs that we want to add, here is the MR https://github.com/mendix/docs/pull/4323 that is already open. I see that you need the images for the new graphs. I will add the images on your comments shortly.

MarkvanMents (approved):

PR #4486, opened 2022-04-21
Version control faq update

grootjans:
Update of the FAQ, adding a notice about the Git storage format interacting with the Mendix model, and added the difference between Git and Subversion commit representations.

### DIFF
--- a/content/en/docs/refguide/version-control/version-control-faq.md
+++ b/content/en/docs/refguide/version-control/version-control-faq.md
@@ -54,7 +54,7 @@ Currently, there is no out-of-the-box migration service available that keeps his
 However, you can migrate manually by creating a new app based on Team Server Git, exporting the Team Server SVN app, and copying that to the new app repository. This does **not** preserve your app history.
 
 
-## 7 Can I Use Third-Party Tools to Connect to the Team Server?
+## 8 Can I Use Third-Party Tools to Connect to the Team Server?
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/version-control/version-control-faq.md
```diff
@@ -39,8 +43,12 @@ Team Server Git has the following advantages over Team Server SVN:
...
 * You can use modern third-party Git tooling for advanced version control cases like handling file changes.
 
-## 5 My Retrieve and Commit + Push actions are getting slower
-This may be caused by the way the storage format of Git interacts with the way the Mendix Model storage changes. This can accumulate substantial disk space over time. This can be mitigated by performing a `git gc` from the command line. In `git gc` a number of housekeeping operations are formed, but primarily the creation of pack files, which store files relatively to their difference, allowing for a much more optimal storage format. We currently already automatically do these housekeeping actions on the Teamserver, so alternatively you can also check out a fresh copy. We are working to improve this in Studio Pro in an upcoming version
+## 5 Retrieve and Commit + Push Actions Are Getting Slower
+This may be caused by the way the Git storage format interacts with the Mendix model storage changes, which can accumulate substantial disk space over time.
```
grootjans: `This may be caused by the way the Git storage format interacts with the Mendix model storage changes`

This changes the meaning, as it is not about the model storage changes, but the way we represent those changes. I also see that my initial sentence was incorrect. I should have written:
`This may be caused by the way the storage format of Git interacts with the way the Mendix Model **stores** changes`

PR #4487, opened 2022-04-21
[NC-675] add onchange behavior to text area native widget

Ahmad-Elsayed:
Feature is not released yet

### DIFF
--- a/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
+++ b/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
@@ -76,7 +76,7 @@ This means that there is no way for users to trigger an on change event while ty
 
 ##### 2.5.2.2 While User is Entering Data {#user-enters-data}
 
-This option makes it so users trigger an on change event while they are typing. Text area will save changes when the value is not the same as the previously saved value in the database and if the last change made took place after the configured **Apply after (ms)** length of time.
+This option allows users to trigger an on change event while they are typing. Text area will save changes when the value is not the same as the previously saved value in the database and if the last change made took place after the configured **Apply after (ms)** length of time.
 
 With **While user is entering data**, users can adjust one more property called **Apply after (ms)** (described above). This will reduce the amount of the calls made for an on change event, thus improving app performance.
 
--- a/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
+++ b/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
@@ -79,6 +79,7 @@ This means that there is no way for users to trigger an on change event while ty
 This option allows users to trigger an on change event while they are typing. 
 
 Text area will save changes if both these conditions are met:
+
 * The value is different than the previously saved value in the database 
 * The last change occured after the configured **Apply after (ms)** length of time
 
--- a/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
+++ b/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
@@ -63,8 +63,8 @@ The on change property specifies an action that will be executed when leaving th
 
 The on change behavior property lets users select how **on change** is handled in Studio Pro via the following options: 
 
-* When user leaves input field (Default)
-* While user is entering data
+* **When user leaves input field (Default)**
+* **While user is entering data**
 
 ##### 2.5.2.1 When User Leaves Input Field (Default)
 
--- a/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
+++ b/content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
@@ -68,9 +68,8 @@ The On Change Behaviour property lets users select how **on change** is handled 
 
 ##### 2.5.2.1 When user leaves input field (Default)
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
```diff
@@ -59,13 +59,37 @@ The on-change property specifies an action that will be executed when leaving th
...
+#### 2.5.2 On Change Behaviour
+
+The On Change Behaviour property lets users select how **on change** is handled via the following options Studio Pro: 
+
+* When user leaves input field (Default)
+* While user is entering data
+
+##### 2.5.2.1 When user leaves input field (Default)
+
+This option will work as in previous versions of Studio Pro. Text Area will apply changes when a value is not the same as previously saved value in the database and one of the following conditions is met:
```
pratikgoyal9: Will this option work with previous versions of Studio Pro? No, right?

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
```diff
@@ -59,13 +59,37 @@ The on-change property specifies an action that will be executed when leaving th
...
+
+* On enter key pressed: This will trigger on change and on enter key press events
```
pratikgoyal9: On press of the enter key, no `onChangeBehavior` would be triggered in a text area. Instead, the user should be navigated to the next line in the same text area. So, user does not leave the current text area in this context. 

According to me, the user can only leave the textarea by pressing the tab key(in case of the web) or by tapping any other part of the screen except in the current text area. So, this event should be removed.

@jordanch @cjfhodges Would you like to add your two cents?

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
```diff
@@ -59,13 +59,37 @@ The on-change property specifies an action that will be executed when leaving th
...
+* Blurred: This will trigger on change and on leave events
```
pratikgoyal9: I've no idea how this would work, where can I find and test the `Blurred` feature?

pratikgoyal9 (changes requested):

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
jordanch: Makes sense. Maybe missed after copying from [text box](https://github.com/mendix/docs/blob/4487/content/en/docs/refguide/modeling/pages/input-widgets/text-box.md).

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
Ahmad-Elsayed: when you touch screen outside the input and the keyboard, the input shouldn't be active, keyboard should be dismissed and then the `onChange` event should be called if `apply changes` property is set to `when user leaves input field`

pratikgoyal9 (approved):

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
pratikgoyal9: Ah okay! Good to know that it's called a `Blurred` condition.

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
```diff
@@ -59,13 +59,36 @@ The on-change property specifies an action that will be executed when leaving th
...
+#### 2.5.2 On Change Behaviour
+
+The On Change Behaviour property lets users select how **on change** is handled via the following options Studio Pro: 
+
+* When user leaves input field (Default)
+* While user is entering data
+
+##### 2.5.2.1 When user leaves input field (Default)
+
+Text Area will apply changes when a value is not the same as previously saved value in the database and one of the following condition is met:
```
pratikgoyal9: Since we removed the other condition, it should be like the below:

`Text Area will apply changes when a value is not the same as previously saved value in the database and the following condition is met:`

pratikgoyal9 (changes requested):

pratikgoyal9 (approved):
Good work @Ahmad-Elsayed 👍

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
pratikgoyal9: Fixed!

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
pratikgoyal9: Fixed, should be resolved!

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
pratikgoyal9: Fixed, should be resolved!

content/en/docs/refguide/modeling/pages/input-widgets/text-area.md
pratikgoyal9: Fixed, should be resolved!

ConnorLand (approved):

PR #4494, opened 2022-04-22
Reformat tables to encourage narrower columns

MarkvanMents:
Try to remove horizontal scrolling where possible by giving more places where the browser can wrap the paragraphs.
Mainly using `<wbr>` as advised in this blog https://www.cjcid.com/articles/wrapping-long-words-css-html/
Also removed bold, to make text narrower
Sometimes have added spaces where no other alternative seemed to work.
Could look at:
Narrower font
Smaller font
Other options?


NatasaKralj (approved):
The addition of `<wbr>` definitely makes the long entries break in a nice fashion. I can't see any instances where spaces were added. 
The webpage you linked is a great resource!
I'm not sure I would have removed the bold styling, especially if those are UI elements. The first column entries look a bit too plain without the bold, but it's just a personal preference.
Entries such as `ConnectionPoolingSoftMinEvictableIdleTimeMillis `don't have a good place to break, but maybe we should put `<wbr>` into long values as well?

MarkvanMents:
How strongly do you feel about the bold styling. I took it away as it makes each letter slightly wider, but perhaps the difference isn't worth the visual change?

Not sure about breaking something that doesn't have an obvious break like a point. But the entries are ridiculously long, even with changing "millisecond" to "millis".

NatasaKralj:
I don't think taking away the bold is worth the space it saves. The tables can already be hard to read.

If we put in breaks into entries with no obvious breaks, at least we control where it happens. It would be better if we could just have automatic word breaking, but applying that style doesn't work well in a table.

MarkvanMents:
I'll put the bold back, but I think I'll leave the long words for now - breaking them could lead to confusion.

MarkvanMents:
@NatasaKralj - I've put the bold back. OK to merge?

On the very long names, I could break them all after `ConnectionPooling` in that table, since they **all** begin with that? It's the only table that still scrolls when I have my window not set to maximum width. But I worry that it will be confusing for customers to have the word broken.

I would prefer to leave it for now and see if we get any feedback?

The changes so far make things much better.

NatasaKralj:
Merge away, looks good. 

That's a good idea about the long entries. Leave them, if we get feedback we can take another look.

PR #4500, opened 2022-04-26
Landing page styling

NatasaKralj:
The PR does the following:

- removes horizontal scroll bar from front page
- fixed height of 300px for most popular cards
- removes indent of active ToC text
- implements last round of design feedback
     - title font color change
     - fix doc button location
     - more spacing on top of page
     - card title closer to icon


MarkvanMents (changes requested):
Hi Natasa,
ToC indent solution seems to have unintended consequences for subsections ☹ - haven't got time to investigate now, so perhaps revert it and add it to the improvements doc?
image

The top three section still has this issue with Developer Portal. Would be better if "Developer" was next to the icon (or even better if it all fitted on one line - perhaps just make the font smaller for this one card, since it is only on this one page), but this isn't new, so we can do it later if necessary.
image

All the rest looks great.

NatasaKralj:
I reverted the line for margin-left, to keep the top headings from being pushed in ToC. I couldn't find a good/quick solution to the children getting dragged to the left.

I fixed the 'Developer Portal' card text by adding a bit of styling to keep it from wrapping. It works well when I resize my window, could you try it out as well?

MarkvanMents:
Looks good.
Is there any way to match the top of the icon with the top of the text. Because it is now wrapped, the "Developer Portal" looks like it is much lower than the other text in relation to the icon (optical illusion, I think).
image
But we could add this to the list for later - already much better.

NatasaKralj:
I added the vertical centering of the icon to our Ideas and Improvements doc, as I could not find a quick solution to the issue.

NatasaKralj:
@MarkvanMents I can't see a way of resolving the requested changes on my end.
Is the PR OK to merge?

MarkvanMents (approved):
Yes - go ahead and merge.
Thanks.

PR #4504, opened 2022-04-27
Update information about metrics

jeohist:
File storage usage is now visible in the metrics, and alerts are being sent for application CPU usage

cc @ajinkya-takle


MarkvanMents (approved):
Can this be published immediately?

PR #4514, opened 2022-04-29
Removed AWS Code commit from supported systems

grootjans:
Code commit does not support pushing multiple refs, which is required for Studio Pro to push metadata along with commits. Without this capability, we cannot support this system.


Adam-Dupaski (approved):

MarkvanMents (approved):

MarkvanMents:
I'll merge this immediately - looks like it has no dependencies.
Let me know if it shouldn't have been merged yet.

PR #4518, opened 2022-05-02
Mention runtime image library dependencies

lvanengelen:
Recently some issues came up regarding SVG image support in the runtime, which were traced to the image libraries we depend on. To make our limitations more clear, this PR:
- adds a reference to the image library we directly depend on ([TwelveMonkeys ImageIO](https://haraldk.github.io/TwelveMonkeys/))
- adds a specific reference to the SVG library we transitively depend on ([Batik](https://xmlgraphics.apache.org/batik/))

As the runtime only processes images when an `Image` object is _uploaded_, I chose to add this information to the place where this would be most likely encountered: the **Image Uploader** widget. Do note that this behavior is triggered whenever an `Image` (which is a specialization of `FileDocument`) is uploaded. E.g. I think it is possible to do that with a **File manager** widget as well, but that would be a very non-obvious place to put this content.


ConnorLand:
Tom Knippenberg is going to have a quick look at this and then I will merge.

ConnorLand (approved):

PR #4527, opened 2022-05-02
Improve substring function entry

NatasaKralj:
The edits below were based on two different user feedback items for the `substring` function. 
One user asked for a better example when using all function parameters. The old example would give the same result without the optional parameter just because the end of the string was reached.
Another user asked us to include an example to check string `length`, before calling `substring`, to avoid errors. I opted to list the possible cases when an error might occur, this seems more in line with the rest of the document.

This PR has two goals:
- give a better example for `substring` when all function parameters are used
- list situations when errors will occur

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
@@ -138,7 +138,7 @@ The output is described in the table below:
 
 {{% alert color="warning" %}}
 The function will output an error for the following:
-* When the start position of the substring is larger than the position of the last character
+* When the start position of the substring is after the last character in the string
 * When the desired length of the result is larger than the length of the substring
 {{% /alert %}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
```diff
@@ -136,6 +136,12 @@ The output is described in the table below:
 | ------------------------------------------------------------ | ------ |
 | A part of the original string, starting at the start position with a length equal to the desired length. If no desired length is specified, will return a substring starting at the start position and ending at the end of the string. | String |
 
+{{% alert color="warning" %}}
+The function will output an error for the following:
+* When the start position of the substring is larger than the position of the last character
```
MarkvanMents: ?
```suggestion
* When the start position of the substring is after the last character in the string
```

content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
```diff
@@ -136,6 +136,12 @@ The output is described in the table below:
...
+* When the desired length of the result is larger than the length of the substring
```
MarkvanMents:
```suggestion
* When the desired length of the result is longer than the substring
```

MarkvanMents (commented):
I think I'd prefer the length to pad with blanks than error, seems more low-code if you are asking for a specific length. But we have to document what is there.

Have suggested alternative ways of expressing the same thing - but yours are also fine.

content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
NatasaKralj: Thank you for making it simpler @MarkvanMents

PR #4528, opened 2022-05-02
Restored the all-uppercase abbreviations to CamelCase

TimvanSteenbergen:
Restored the all-uppercase abbreviations to camelcase abbreviations, as that is more informative and therefor easier to read.


Adam-Dupaski:
Thanks for the suggestion @TimvanSteenbergen but I've verified with Academy that they are teaching what is currently described, so we are not going to change this now.
Cheers,
Adam

PR #4557, opened 2022-05-10
Adjust vertical spacing and list indents

NatasaKralj:
The PR does the following:

- removes extra vertical padding after a paragraph
- evens out spacing between different heading types
- decreases the indentation of sub-items in lists 

Please see the images below for examples.
image
image

### DIFF
--- a/assets/scss/_content.scss
+++ b/assets/scss/_content.scss
@@ -10,6 +10,14 @@
         font-weight: $font-weight-body-text;
     }
 
+    // NK - custom additions for margin styling in lists
+    li p:first-child {
+        margin-bottom: 0.25rem;
+    }
+    li p:not(:first-child), li figure, li .alert {
+        margin-bottom: 0.5rem;
+    }
+
     > h1 {
         font-weight: $font-weight-light; // NK - changed from bold
         margin-bottom: 1rem;
@@ -38,7 +46,8 @@
         @extend .img-fluid;
     }
 
-    > table {
+    // NK - added li table to have stying applied to tables in lists
+    > table, li table {
         @extend .table-striped;
 
         @extend .table-responsive;
--- a/assets/scss/_content.scss
+++ b/assets/scss/_content.scss
@@ -8,7 +8,6 @@
 
     p, li, td {
         font-weight: $font-weight-body-text;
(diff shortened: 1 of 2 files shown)


MarkvanMents:
Great improvements. I can only see two things which (I think) could be better.
Seem to have lost a gap between a paragraph and the _first_ item in a list. I find that this looks cramped in the example below.
image

A paragraph between two headings (except h2 and h3) is closer to the following section than the section it belongs to. I think it should be the other way around, so perhaps the spacing _above_ a paragraph is a bit too big?
image

Let me know if you want me to push a new version of indentation-spacing-test with these test cases in it?

MarkvanMents (commented):
Have made a couple of comments about the layout in specific scenarios, but overall it looks good.
Can you see if you agree with me?

MarkvanMents:
>perhaps the spacing above a paragraph is a bit too big?
Or perhaps there is a big margin _under_ the headings?
I think this is a better solution than putting back the space under the paragraph, the gap between paragraph and heading could do with being less, I think?

And we have also lost the spacing between paragraphs - they just look like new lines:
image
produces
image
which doesn't show that the lorum ipsem is one paragraph surrounded by others.

NatasaKralj:
I made the margin-top of headings larger, to avoid the inconsistency with text spacing above and under a heading.

I rolled back too broad of a change to paragraph elements, it was the cause to both 
> Seem to have lost a gap between a paragraph and the first item in a list
and
> And we have also lost the spacing between paragraphs

The main issue that was supposed to deal with was the gap spacing between list elements, where sometimes list elements get wrapped in `<p>` elements. This is due to how items are indented under the list item. Indenting an item under a list with an empty line between causes the `<p>` wrapping of list elements. See the new addition to the test file:
image

MarkvanMents:
This still doesn't work for me.
1. There is still no distinction between line breaks and paragraph breaks.
image
2. I think that the space above a heading should be a bit bigger than that above a paragraph to make it clear which heading the paragraph belongs to.

Perhaps we could go through the current formatting and look together at what we think it should look like and how we could achieve that? Either in the office on Friday, or online if you want to look at it before?

NatasaKralj:
Styling changes updated and discussed in meeting. Merging.

PR #4567, opened 2022-05-11
[UFC-497] HTML entities in dynamic label

Robinfr:
Added explanation about using HTML entities when renderXHTML is set to `true` for the dynamic label widget.


DmytroKost:
Looks good to me

ConnorLand (approved):

ConnorLand (approved):

PR #4570, opened 2022-05-12
Adjust formatting of no_list results

NatasaKralj:
The PR adjusts the list generated for immediate children of a page, when `no_list` metadata parameter is set to `false`.
I removed the horizontal line above the list, and made each child page appear as a list item with a description (if available).
The max-width changes were needed due to no_list elements taking up more width than the rest of content.

image

### DIFF
--- a/layouts/partials/section-index.html
+++ b/layouts/partials/section-index.html
@@ -1,0 +1,31 @@
+<div class="section-index">
+    {{ $parent := .Page }}
+    {{ $pages := (where .Site.Pages "Section" .Section).ByWeight }}
+    {{ $pages = (where $pages "Type" "!=" "search") }}
+    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) }}
+    {{ $pages = (where $pages ".Parent" "!=" nil) }}
+    {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) }}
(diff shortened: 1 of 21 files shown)


layouts/partials/section-index.html
```diff
@@ -0,0 +1,31 @@
+<div class="section-index">
+    {{ $parent := .Page }}
+    {{ $pages := (where .Site.Pages "Section" .Section).ByWeight }}
+    {{ $pages = (where $pages "Type" "!=" "search") }}
+    {{ $pages = (where $pages ".Params.hide_summary" "!=" true) }}
+    {{ $pages = (where $pages ".Parent" "!=" nil) }}
+    {{ $pages = (where $pages "Parent.File.UniqueID" "==" $parent.File.UniqueID) }}
+    {{ if or $parent.Params.no_list (eq (len $pages) 0) }}
+    {{/* If no_list is true or we don't have subpages we don't show a list of subpages */}}
+    {{ else if $parent.Params.simple_list }}
```
MarkvanMents, line 10: This would have given us what we wanted in the first place - except that we wouldn't have had the description? Would it be better to modify this?
image
image

(Don't know why this isn't documented better in Docsy/Hugo)

NatasaKralj: I added `description_list` as a bulleted list with descriptions. Now we have the two default options and a third one in between.

NatasaKralj:
Reviewed and approved via meeting

PR #4575, opened 2022-05-13
Add link into every figure shortcode

NatasaKralj:
The shortcode change adds the link parameter to every figure shortcode that doesn't contain it. This makes clicking on the image open it in a new tab.

### DIFF
--- a/layouts/shortcodes/figure.html
+++ b/layouts/shortcodes/figure.html
@@ -5,7 +5,7 @@
         <a href="{{ .Get "link" }}"
     {{- else -}}
         <a href="{{ .Get "src" }}"
-    {{- end }} target="{{ with .Get "target" }}{{ . }}{{ else }}_blank{{ end }}"{{ with .Get "rel" }} rel="{{ . }}"{{ end }}></a>
+    {{- end }} target="{{ with .Get "target" }}{{ . }}{{ else }}_blank{{ end }}"{{ with .Get "rel" }} rel="{{ . }}"{{ end }}>
     <img src="{{ if $src }}{{ $src.RelPermalink }}{{ else }}{{ .Get "src" }}{{ end }}" 
          {{- if or (.Get "alt") (.Get "caption") }}
          alt="{{ with .Get "alt" }}{{ . }}{{ else }}{{ .Get "caption" | markdownify| plainify }}{{ end }}"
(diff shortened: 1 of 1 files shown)


layouts/shortcodes/figure.html
```diff
@@ -0,0 +1,32 @@
+{{ $src := (.Page.Resources.GetMatch (printf "**%s*" (.Get "src"))) }}
+<figure{{ with .Get "class" }} class="{{ . }}"{{ end }}>
+    {{- if .Get "link" -}}
+        <a href="{{ .Get "link" }}" target="_blank"{{ with .Get "rel" }} rel="{{ . }}"{{ end }}>
+    <!-- NK - added automatic insertion of link parameter so image can be opened in new tab on click -->
+    {{- else -}}
```
MarkvanMents, line 6: Seems to be a bug in the way you are implementing the {{ else }}.
image
image

Also, I think we should respect the `target` if someone has used it in the [figure shortcode](https://gohugo.io/content-management/shortcodes/#figure). Perhaps something like:
`target="{{ with .Get "target" }}{{ . }}{{ else }}_blank{{ end }}"`
Can we discuss?

MarkvanMents (changes requested):

layouts/shortcodes/figure.html
```diff
@@ -0,0 +1,32 @@
+{{ $src := (.Page.Resources.GetMatch (printf "**%s*" (.Get "src"))) }}
+<figure{{ with .Get "class" }} class="{{ . }}"{{ end }}>
+    <!-- NK - adds on-click link to images, default target is in new tab/window -->
+    {{- if .Get "link" -}}
+        <a href="{{ .Get "link" }}"
+    {{- else -}}
+        <a href="{{ .Get "src" }}"
+    {{- end }} target="{{ with .Get "target" }}{{ . }}{{ else }}_blank{{ end }}"{{ with .Get "rel" }} rel="{{ . }}"{{ end }}></a>
```
MarkvanMents: `</a>` is in the wrong place. Image needs to be within the tags. `<a></a>`
But looking much better!
If you want to work on this before Friday, then do so.

NatasaKralj: Fixed, removed extra </a> tag.

PR #4584, opened 2022-05-18
Update invoice-processing.md

mahendra-mahajan-mendix:
Updates to the Invoice Processing document as we are expanding our support to bmp, png and pdf invoice files. I will be sharing the new screenshot for sample-extraction-dialog-box.png via email as the download JSON dialog is changed.


Adam-Dupaski (approved):

Adam-Dupaski:
CC @Luyao-Zhang-1 for visibility upon return

PR #4587, opened 2022-05-18
Additional SAML Documentation

MarkvanMents:
A bit of extra information about the key store and `/index3.html` files.


content/en/docs/appstore/modules/saml.md
```diff
@@ -75,14 +75,26 @@ There are different versions of the module, depending on which version of Mendix
...
-	* **SSOLandingPage** – You can specify a different landing page here (for example, redirecting each user accessing the application URL to the SAML login). This requires you to change the *index.html* page by adding `<meta http-equiv="refresh" content="0;URL=/SSO/" />` (so the end-user does not end up on *index.html* again after a login attempt).  By changing this constant to `/index3.html`, the end-user will land on *index3.html* instead of *index.html*.  In this case, you will of course need to add an *index3.html* page to your theme (and you can copy the original *index.html* into *index3.html*).
+	* **DefaultLoginPage** – You can specify a different login page here (for example, configuring the `index.html` page to redirect to `/SSO/`). This constant is only used when the login process fails. When the end-user cannot be authenticated in Mendix, they are presented with a page. If this constant is specified, a button will appear, and by clicking this button, you will be redirected to the default login page (for example, `index.html` or `login.html`).
```
MarkvanMents: I find this reference to redirecting very confusing.
I think I would try to make it simpler:
```suggestion
	* **DefaultLoginPage** – You can specify a different login page here for when the login process fails. When the end-user cannot be authenticated in Mendix, a button will appear, and by clicking this button, you will be redirected to the specified login page. If this is left blank, an unauthenticated user will be redirected to `/login.html`. 
```

feddovanede: Simpler, but not entirely correct:
> When the end-user cannot be authenticated in Mendix

should be

> When the end-user cannot be authenticated in the external Identity Provider

The idea is that when an end-user fails to sign in with the IdP, they are redirected to the normal login page, so they can try to sign in there, directly in the Mendix app.

Also, I think
>you will be redirected to the specified login page.

should be

>they will be redirected to the specified login page.

?

content/en/docs/appstore/modules/saml.md
```diff
@@ -92,31 +104,50 @@ You can choose what you want to enter for the entity ID, organization, and conta
...
 {{% alert color="warning" %}}
-If you have multiple IdPs, please make sure each IdP has a unique **Entity descriptor**. If you add multiple IdPs with the same entity descriptor, you might experience unexepected behavior where a different SSO configuration is selected than the alias provided.
+Resetting the key store or uploading another key store will require all the IdPs to import the new metadata file. 
 {{% /alert %}}
 
-When creating a new IdP configuration, you are guided through a workflow to help you configure everything required for the IdP configuration. Each option in the workflow is explained below.
+Click **Upload** to upload a key store file. Use the **Entity Id** as the the alias of the key store.
 
-Upon completing these steps, you only need to send the metadata file to the IdP and have them configure the authentication on their end. 
+You can create your own key store file, which you can use when configuring other SAML SPs by clicking **Download**.
```
feddovanede: Not sure if this is confusing. Clicking the button doesn't create the file: it is already there. It will just download the existing file.

content/en/docs/appstore/modules/saml.md
```diff
@@ -92,31 +104,50 @@ You can choose what you want to enter for the entity ID, organization, and conta
...
-#### 3.2.1 General
+#### 3.2.1 Creating a New IdP Configuration
+
+{{% alert color="warning" %}}
+If you have multiple IdPs, please make sure each IdP has a unique **Entity descriptor**. If you add multiple IdPs with the same entity descriptor, you might experience unexepected behavior where a different SSO configuration is selected than the alias provided.
+{{% /alert %}}
+
+When creating a new IdP configuration, you are guided through a workflow to help you configure everything required for the IdP configuration. Each option in the workflow is explained below, and can be changed by editing an existing IdP Configuration.
+
+Upon completing these steps, you only need to send the metadata file to the IdP and have them configure the authentication on their end.
```
feddovanede: the _SP_ metadata file

content/en/docs/appstore/modules/saml.md
```diff
@@ -125,7 +156,7 @@ Each IdP (entity descriptor) should have its own configuration set. Every IdP ca
 If you want to automatically synchronize the IdP metadata, go to **\_USE ME** > **Scheduled Events** in the **App Explorer**, double-click **SE_SynchronizeIdPMetadata** and select **Yes** for **Enabled** to enable the automatic synchronization of the IdP metadata.
```
feddovanede: This looks weird: changing the _Enabled_ setting of an scheduled event only applies to running locally, or on a sandbox environment. It doesn't enable the scheduled event when running the app in in the cloud. Maybe we should make this more generic?

>If you want to automatically synchronize the IdP metadata, make sure the **SAML20.SE_SynchronizeIdPMetadata** [scheduled event](/refguide/scheduled-events/) is enabled.

feddovanede (commented):
Left a few comments

content/en/docs/appstore/modules/saml.md
MarkvanMents: Thanks Feddo - taking your suggestions into account, does this look OK?
```suggestion
	* **DefaultLoginPage** – You can specify a different login page here for when the login process fails. When the end-user cannot be authenticated in the external Identity Provider, a button will appear, and by clicking this button, they will be redirected to the specified login page. If this is left blank, an unauthenticated user will be redirected to `/login.html`.
```

content/en/docs/appstore/modules/saml.md
MarkvanMents:
```suggestion
You can download the key store file and use it when configuring other SAML SPs by clicking **Download**.
```

content/en/docs/appstore/modules/saml.md
MarkvanMents:
```suggestion
Upon completing these steps, you only need to send the SP metadata file to the IdP and have them configure the authentication on their end. 
```

content/en/docs/appstore/modules/saml.md
MarkvanMents: I've taken your suggestion and just pointed the user to where to find it.
Should we have a suggestion here on how often to run this, or just leave it to the customer?

```suggestion
If you want to automatically synchronize the IdP metadata, make sure the **SE_SynchronizeIdPMetadata** [scheduled event](/refguide/scheduled-events/) is enabled. This is in the **\_USE ME** > **Scheduled Events** folder of the SAML module.
```

MarkvanMents (commented):
@feddovanede 
Thanks for the useful comments. I've put them all as suggested changes to make it easier to see how I've responded. Can you let me know if the new wording looks OK. (If you can actually commit the changes you like, even better, but I don't know if you can do that?)
Mark

content/en/docs/appstore/modules/saml.md
feddovanede: Yes, looks ok!

content/en/docs/appstore/modules/saml.md
feddovanede: 👍

content/en/docs/appstore/modules/saml.md
feddovanede: 👍

content/en/docs/appstore/modules/saml.md
feddovanede: 👍

MarkvanMents:
Publishing after discussion with Feddo.

PR #4597, opened 2022-05-20
9.13.1 RN

Adam-Dupaski:


atennapel:
Could you add the release note: "* We fixed an issue where scheduled events did not execute in a horizontally scaled environment. (Ticket 146867)" to Fixes for 9.13.1? (this is for RTI-629)

PR #4610, opened 2022-05-24
add Mendix Native Mobile Builder to index page

ThijsSeppen:
This seemed to be missing from this page (though it's available in the tree to the left). 

Also, just as an idea, perhaps it makes sense to rename "Native Builder" and "Mendix Native Mobile Builder" to something more distinct (as one is the CLI tool, while the other is the UI)?


ConnorLand (approved):

ConnorLand:
Thanks Thijs. I share your thoughts on the naming. I'll see what I can do.

PR #4612, opened 2022-05-25
Fixed font configuration

jdiehl:
Font-weights on Noto currently do not work, because they are not included in the GoogleFonts API call. This PR fixes that.


jdiehl:
Before:
before

After:
after

jdiehl:
Fixes #4613

NatasaKralj:
Thanks again @jdiehl for your help. Closing this, picking up changes in https://github.com/mendix/docs/pull/4636

PR #4622, opened 2022-05-27
OIDC SSO Marketplace Module

MarkvanMents:
Work in progress
Documentation currently with module in Marketplace


content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+title: "OIDC SSO"
+url: /appstore/modules/oidc/
+category: "Modules"
+description: "Describes the configuration and usage of the OIDC SSO module, which is available in the Mendix Marketplace."
+tags: ["marketplace", "marketplace component", "oidc", "idp", "identity provider", "platform support"]
+#If moving or renaming this doc file, implement a temporary redirect and let the respective team know they should update the URL in the product. See Mapping to Products for more details.
+# Linked from https://marketplace.mendix.com/link/component/120371
+---
+
+{{% todo %}}Remove temporary images - held locally for GitHub review{{% /todo %}}
```
MarkvanMents: I have put temporary images into the document so that you can view them on GitHub. These will be replaced with images in the proper location when it is published.

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+* [MendixSSO](https://marketplace.mendix.com/link/component/111349) – if your app is targeted at end-users that have signed up to the Mendix platform
+
+### 1.1 Typical Usage Scenarios
+
+* **B2C-apps:** Your app is aimed at consumers who have an identity at a 'social IdP' which uses OIDC, such as Google. In this case your app will only delegate the authentication to the IdP, no further user information is available to the app.
+* **B2E-app:** Your app is aimed at your company's employees and you want these employees to sign in to your app using corporate credentials hosted by your identity provider (IdP) that supports the OIDC protocol. In this case your app may have its own logic to assign user roles or you may use authorization information from your IdP as provided to your app using an access token.
+* **API-consumption.** If your app makes calls to 'downstream' APIs on behalf of your end-user, you can use the access token obtained via the “OIDC SSO” module. This scenario is not supported when using SAML SSO. This makes the OIDC SSO module suitable for Mendix customers using Mendix Data Hub.
```
MarkvanMents: What do you mean by 'downstream'?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+## 2 Dependencies
+
+The OIDC module requires your app to be using Mendix 8.18+ or later. (Mendix 9 compatible).
+
+It requires the following Marketplace modules to be included in your app:
+
+* [Encryption](https://marketplace.mendix.com/link/component/1011) – see [Encryption](/appstore/modules/encryption/) documentation
+* [Community Commons](https://marketplace.mendix.com/link/component/170) – see [Community Commons](/appstore/modules/community-commons-function-library/) documentation
+* [Nanoflow Commons](https://marketplace.mendix.com/link/component/109515) – see [Nanoflow Commons](/appstore/modules/nanoflow-commons/) documentation
+* [Native Mobile Resources](https://marketplace.mendix.com/link/component/109513) – see [Native Mobile Resources](/appstore/modules/native-mobile-resources/) documentation
```
MarkvanMents, line 70: Why do we need Native Mobile Resources if we have said that we don't support mobile apps?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+
+Make a single call from `CUSTOM_UserProvisioning` to your own module where you implement the provisioning flow you need. This way, it will be easy to install new versions of the OIDC SSO module over time without overwriting your custom provisioning.
+
+The OIDC SSO module supports multiple identity providers. Since each provider can provide user data in a different format, you may want to use multiple provisioning flows. See the microflow `UserProvisioning_Sample` for an example and details on how to do this.
+
+## 7 Optional Features{#optional}
+
+### 7.1 Performing API Calls on Behalf of an Authenticated User
+
+You might want to make API calls to other apps/services on behalf of the end-user. As you have used the OIDC module to authenticate the end-user to your app, your app also has an access token for this end-user. You can use this access token to propagate the end-user's identity to the API so the API does not need to have a user identifier in the payload.
```
MarkvanMents: Should we mention that this will only work if the API supports OAuth?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+* **Request:**  – a string containing the content of the DELETE request (most likely a formatted JSON)
+* **Token:**  – the `OIDC.Token` object that should be used for authentication, typically retrieved via the `Token_Account` association (to find the token of the current user/session)
+
+The microflow returns an object of type `System.HttpResponse`. This could indicate an error.
+
+#### 7.1.2 GET
+
+Takes as input:
+
+* **Request:**  – a string containing the URL you want to GET data from
```
MarkvanMents, line 260: Why is this **Request**?
In every other call we pass the URL as **Location**. The customer is going to make mistakes if we aren't consistent.

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+5. Configure the user roles in your app to match the roles returned by SAM. End-users will be given the matching role when they sign into the app. If the role in the SAM token is not found in the Mendix app the end-user will be given the role `User`.
+6. Save the configuration.
+
+#### 7.3.2 Parsing Access Tokens Using a Custom Microflow
+
+If you choose to implement your own microflow to parse an access token, the microflow name must start with `CustomATP`, for example `CustomATP_MyTokenParser`. The custom microflow has the access token as the parameter.
+
+Once you have created the microflow (for example `CustomATP_xxx`), you must do the following:
+
+1. Refresh the module containing your microflow as described in [Installing MxModelReflection](#mxmodelreflection).
```
MarkvanMents, line 339: I've added this. I assume we have to do this, otherwise the new microflow won't be found by the app? But I can remove it if it isn't necessary.

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+{{% todo %}}Remove temporary images - held locally for GitHub review{{% /todo %}}
+
+## 1 Introduction
+
+The [OpenID Connect (OIDC) SSO module](https://marketplace.mendix.com/link/component/120371) allows end-users of your Mendix app to login via Single Sign-on (SSO) using the OIDC protocol.  Besides delegating end-user authentication (OIDC), your app can also delegate authorization (OAuth).
+
+OIDC is an extension of OAuth2 that propagates the end-user's identity to your application.
+
+{{% alert color="warning" %}}
+This OIDC SSO module works with Mendix version 9.0 and above. If you are using a previous version of Mendix, you can use the *community-supported* module [OpenIDConnect Single Sign-on (OIDC, OAuth2, SSO)](https://marketplace.mendix.com/link/component/117529).
```
MarkvanMents, line 18: Does it work with 9.0? Later on we say it has to be 9.6.1 if we are upgrading from Mendix 8.

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+* [Mx Model reflection](https://marketplace.mendix.com/link/component/69) – see [Mx Model Reflection](/appstore/modules/model-reflection/) documentation
+
+## 3 Installation
+
+If you are migrating from the community edition of the module ([OpenIDConnect Single Sign-on (OIDC, OAuth2, SSO)](https://marketplace.mendix.com/link/component/117529)), please refer to the [migration documentation](#migration) below.
+
+1. [Add the “OIDC SSO” module into your app](/appstore/general/app-store-content/).
+2. Add the necessary dependencies (as listed in the previous section) from the Marketplace, if they are not already included in your app.
+3. Add the snippet **Snip_Configuration** in the **USE_ME** > **1. Configuration** folder of the OICD SSO module to a page that  is accessible to admin end-users of your app.
+4. Replace all the layouts that end in `_REPLACEME` used in pages in this module with layouts from your own project. The layouts are in the **Implementation** > **Layouts** folder of the module. Use the [Find Usages](/refguide/find-and-find-advanced/#find-usages) command to find where they are used.
```
MarkvanMents, line 80: The Paper document mentions that there is a batch command to do this.
I can't immediately see how to do this, perhaps we should give more explicit instructions?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+### 5.1 OIDC Provider Configuration
+
+1. In your identity provider, provision a new OpenID client application. You will receive a ClientID and Client Secret.
+2. You will also need the OIDC configuration endpoint (for example: [https://accounts.google.com/.well-known/openid-configuration](https://accounts.google.com/.well-known/openid-configuration))
+3. Register the following callback URLs:
+    * `http://<your-app-url>/oauth/v2/callback`
+    * `makeitnative://<your-app-url>/oauth/callback`
+
+{{% alert color="info" %}}
+If your IdP does not support a custom URL scheme as a callback URL, change the setting **Provider supports custom URL schemes** to `false`.
```
MarkvanMents: Where is this setting?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+
+## 5 Configuration of OIDC Provider{#oidc-configuration}
+
+### 5.1 OIDC Provider Configuration
+
+1. In your identity provider, provision a new OpenID client application. You will receive a ClientID and Client Secret.
+2. You will also need the OIDC configuration endpoint (for example: [https://accounts.google.com/.well-known/openid-configuration](https://accounts.google.com/.well-known/openid-configuration))
+3. Register the following callback URLs:
+    * `http://<your-app-url>/oauth/v2/callback`
+    * `makeitnative://<your-app-url>/oauth/callback`
```
MarkvanMents, line 174: I thought that we weren't supporting native mobile?

If we are supporting it, is it correct that this callback doesn't have a `v2` as part of the URL?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+Follow the instructions to [set an encryption key in the Encryption module](/appstore/modules/encryption/#configuration). The constant to set is called `Encryption.EncryptionKey` and should be a random value 32 characters long. This key will be used to encrypt and decrypt values.
+
+## 5 Configuration of OIDC Provider{#oidc-configuration}
+
+### 5.1 OIDC Provider Configuration
+
+1. In your identity provider, provision a new OpenID client application. You will receive a ClientID and Client Secret.
+2. You will also need the OIDC configuration endpoint (for example: [https://accounts.google.com/.well-known/openid-configuration](https://accounts.google.com/.well-known/openid-configuration))
+3. Register the following callback URLs:
+    * `http://<your-app-url>/oauth/v2/callback`
```
MarkvanMents: `http` or `https`?

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+
+To parse of SAM access tokens you need to have done the following:
+
+1. Set-up the connectivity between your app and SAM as described [above](#oidc-configuration).
+2. Install the **MxModelReflection** module.
+3. Check **Enable Access Token Parsing** and select *Default SAM Token processing* as the **custom AccessToken processing microflow**.
+    {{< figure src="/attachments/appstore/modules/oidc/enable-sam-parsing.png" >}}
+    ![](temp-oidc-images/enable-sam-parsing.png)
+4. Add the scopes `sam_account`, `samauth.role`, `samauth.tier`, and `samauth.ten` to the **Selected Scopes** in the [Client Configuration](#client-configuration).
+5. Configure the user roles in your app to match the roles returned by SAM. End-users will be given the matching role when they sign into the app. If the role in the SAM token is not found in the Mendix app the end-user will be given the role `User`.
```
MarkvanMents, line 330: What happens if a role is changed on the SAM IdP? When is the change reflected in the app user role?
* Never (it is never updated after the user is created in the app)
* When the current token needs to be refreshed
* When the Mendix session ends and the customer next logs in

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,403 @@
...
+
+* **Location:** – a string containing the URL you want to do the PUT on
+* **Request:**  – a string containing the content of the PUT request (most likely a formatted JSON)
+* **Token:**  – the `OIDC.Token` object that should be used for authentication, typically retrieved via the `Token_Account` association (to find the token of the current user/session)
+
+The microflow returns an object of type `System.HttpResponse`. This could indicate an error.
+
+### 7.2 PKCE Configuration{#pkce}
+
+PKCE (pronounced as “pixie”) adds enhanced security to the original OAuth protocol. It is generally recommended to use this feature for better protection against hackers who try to get access to your app. You need to check if your IdP supports PKCE. When it does, it will have “S256” as the value for `code_challenge_methods_supported` on its well-known endpoint.
```
MarkvanMents: How does the customer see the value of `code_challenge_methods_supported`? Is it in the app, or will they need to make their own call to the well-known endpoint using *Postman*?

MarkvanMents (commented):
Hi,
Can you address my questions while you are reviewing this document?

content/en/docs/appstore/modules/oidc.md
JaapF: As in: a sequence which app is making a call to which app. If app 'A' makes a call to an api in app B, I would consider A being upstream from B's point of view, and B would be downstream from A's point of view.

Maybe better to simply remove the word.

content/en/docs/appstore/modules/oidc.md
JaapF: Naveen can you clarify please?

- is it really a dependency?
- can you provide a short explanation so the developer that's reading the documentation can also understand a bit?

content/en/docs/appstore/modules/oidc.md
JaapF: Yes and since OAuth is a set of specs rathger than a single one, we should aim at being even more specific.

Maybe add sentence:

This mechanism only works of your API is supporting OAuth and/or OIDC. More specifically:
- it needs to be able to accept OAuth bearer token in the HTTP "Authorization" request header field, as per https://datatracker.ietf.org/doc/html/rfc6749#section-7
- it needs to accept Access Tokens from the same IDP where your user was authenticated
- it needs to be able to parse the Access Token as JWT  (as suggested by https://datatracker.ietf.org/doc/html/rfc9068 although your Access Tokens do not neccesarily have to be fully compliant with that RFC) or be able to invoke UserInfo endpoint  (as suggested by OIDC specs https://openid.net/specs/openid-connect-core-1_0.html#UserInfo )

content/en/docs/appstore/modules/oidc.md
JaapF: Job, can you answer this one, please?

content/en/docs/appstore/modules/oidc.md
JaapF: They need to make call themselves.

I have item on backlog for the module to make it smarter, because the module could be looking at the well-known endpoint itself. But that's currently not in place yet.

content/en/docs/appstore/modules/oidc.md
JaapF: "Never" is wrong answer.
"login" is right, I think.
"refresh" - I don't know. This breaks into 2 questions:
- is SAM possibily changing the roles during refresh?
- if so, does our OIDC module handle that change?

Job or Naveen (or ...) can you help?

content/en/docs/appstore/modules/oidc.md
MarkvanMents: How about:
```suggestion
* **API-consumption.** If your app makes calls to APIs of other services on behalf of your end-user, you can use the access token obtained via the “OIDC SSO” module. This scenario is not supported when using SAML SSO. This makes the OIDC SSO module suitable for Mendix customers using Mendix Data Hub.
```

content/en/docs/appstore/modules/oidc.md
MarkvanMents: How about:
```suggestion
You might want to make API calls to other apps/services on behalf of the end-user. As you have used the OIDC module to authenticate the end-user to your app, your app also has an access token for this end-user. 

If the API supports OAuth and/or OIDC, you can use this access token to propagate the end-user's identity to the API so the API does not need to have a user identifier in the payload. To do this, the API needs to:

* accept OAuth bearer tokens in the HTTP `Authorization` request header field, as per [section 7 of RFC 6749](https://datatracker.ietf.org/doc/html/rfc6749#section-7)
* accept Access Tokens from the same IdP where your user was authenticated
* parse the Access Token as JWT (as suggested by [RFC 9068](https://datatracker.ietf.org/doc/html/rfc9068) – although your Access Tokens do not necessarily have to be fully compliant with that RFC) or be able to invoke the UserInfo endpoint (as suggested by the [OIDC specs](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo))
```

content/en/docs/appstore/modules/oidc.md
MarkvanMents: How about:
```suggestion
PKCE (pronounced as “pixie”) adds enhanced security to the original OAuth protocol. It is generally recommended to use this feature for better protection against hackers who try to get access to your app. You need to check if your IdP supports PKCE. When it does, it will have “S256” as the value for `code_challenge_methods_supported` which you can discover by manually querying its well-known endpoint.
```

MarkvanMents (commented):
@JaapF 
I've made some suggestions to clarify a couple of points. Let me know if they look good.
I look forward to hearing from the other reviewers on the remaining questions and the document as a whole.

content/en/docs/appstore/modules/oidc.md
JaapF: Perfect!

content/en/docs/appstore/modules/oidc.md
JaapF: Yes looks good, thanks

content/en/docs/appstore/modules/oidc.md
JaapF: Yes, ok!

content/en/docs/appstore/modules/oidc.md
PhetoJob: https I would say, perhaps for local development use http?

content/en/docs/appstore/modules/oidc.md
JaapF: @MarkvanMents , support for 'Native Mobile SSO' is a hidden feature for which we do not yet provide support, but the code is already there.
Let's not draw too much attention and maybe not add any explanation. Or we could say:
"Native SSO is not yet supported, but some initial implementations are already included in the OIDC SSO module".

content/en/docs/appstore/modules/oidc.md
JaapF: I think your suggesting an improvement to developer experience, for now let's document 'as is'.

content/en/docs/appstore/modules/oidc.md
JaapF: As per previous remark: we're not supporting it, but some code is already there.

content/en/docs/appstore/modules/oidc.md
JaapF: I know that when user does a new login, the latest roles indicated by SAM IDP are made effective; userrole tables in Mendix app are updated.
I'm not sure if the same logic applies during refresh. My best guess is: NO, because the runtime may not support this.

content/en/docs/appstore/modules/oidc.md
SharishmaPodduturi999: Hi Mark,
This setting is not a Boolean, we are providing an input text box so that user's can configure their own custom callback URL. If custom callback URL is empty, then the default custom callback URL will be "AppURL + '/oauth/v2/callback'" which is set by OIDC SSO

image

content/en/docs/appstore/modules/oidc.md
SharishmaPodduturi999: Hi Jaap & Mark,
I confirm that the above statements mentioned by Jaap are correct. 
Roles get updated only in a new login. 
Roles will not be updated during refresh.

content/en/docs/appstore/modules/oidc.md
MarkvanMents: I'll just leave this for now and hope we don't get any support calls.

content/en/docs/appstore/modules/oidc.md
MarkvanMents: Won't local development use `localhost:`?
```suggestion
    * `https://<your-app-url>/oauth/v2/callback`
```

content/en/docs/appstore/modules/oidc.md
MarkvanMents: I'll delete this box completely, then as it is the default behaviour.

content/en/docs/appstore/modules/oidc.md
```diff
@@ -0,0 +1,409 @@
...
+1. In your identity provider, provision a new OpenID client application. You will receive a ClientID and Client Secret.
+2. You will also need the OIDC configuration endpoint (for example: [https://accounts.google.com/.well-known/openid-configuration](https://accounts.google.com/.well-known/openid-configuration))
+3. Register the following callback URLs:
+    * `http://<your-app-url>/oauth/v2/callback`
+    * `makeitnative://<your-app-url>/oauth/callback`
+
+{{% alert color="info" %}}
+If your IdP does not support a custom URL scheme as a callback URL, change the setting **Provider supports custom URL schemes** to `false`.
+{{% /alert %}}
+
```
MarkvanMents: Remove this information as it is the default behaviour.
```suggestion
```

content/en/docs/appstore/modules/oidc.md
MarkvanMents: I'll leave this as written then as it is the closest I can get to it without going into a long discussion which is probably irrelevant to 99% of readers.

MarkvanMents (commented):
@JaapF, @SharishmaPodduturi999, @PhetoJob
Thanks for the review.
I think my suggested changes cover everything. Please let me know if I can publish this - we can always improve it later.
Mark

JaapF:
Hi @MarkvanMents , yes I think your remarks make clear it's not perfect.
Still I think it's good to have OIDC SSO documented on docs.mendix.com.

Let's publish and do further iterations to improve.

MarkvanMents:
All issues resolved - publishing.

PR #4625, opened 2022-05-27
Update the Apps Section in Control Center

Luyao-Zhang-1:


beerscheffers:
Our feedback regarding section 3:
-------------------------------------
**General remark:** Not sure if the structure makes sense with having a separate section for active apps and deactivated apps. Now it seems that you only have a “Profile Page” for active apps, and that you can only delete an active app. Also first showing all the actions and possibilities (3.1), and then later explaining what this tab is actually showing (3.1.1) feels like a strange order to me. general remark about the Active/Inactive seems a bit of a strange distinction especially because you're explaining the functionalities for Active and then for Inactive you almost have the same kind of functionalities.

**Section 3.1:** For the first table about showing options for different types of deployment:

- We have both a red cross and “not supported”. What is the difference? -> Kindly requested to change the `Not Applicable `and `Not supported` into red crosses.
- For the Private Cloud Environment, the option to “Delete a free Mendix Could environment” is marked as “not supported”, but that is confusing to me. Technically, an app that is on private cloud can still have a free environment in the Mendix Cloud (in CP), and you could delete that from CC. Besides that, it might be confusing for the user to talk about free Mendix Cloud environment deletion for a private cloud app.
- Is it not worth mentioning SAP cloud as well, maybe there is a reference to private cloud docs that can answer these flavors?, what about IBM?
- Is the user aware of a difference between Free Mendix Cloud Environment and No cloud resource? Or are those situations considered the same for the user?

**Section 3.1: For the second table showing the consequences:**

- AFAIK, the team members can also no longer access the code repository when deactivating an app. Should this be mentioned as well? (There is a difference in something being persistent and being accessible. We show the difference for the app project, but not for the repository.)

**Section 3.1.1:**

- Below the image, there is a text saying: "… with options for exporting app details to an .xlsx file, deactivating apps, and deleting apps" .`Deleting apps` should not be plural, since you can only delete one app at a time.

**Section 3.1.2**

- Maybe move the screenshot of 3.1.2 a bit up after `.. it opens with the app profile.` Otherwise, you read the explanation and then get the context.
- 3.1.2 > There are a lot of screenshots, but then for adding yourself to a project there isn't one. I can imagine this is also explained earlier, but then when to show a screenshot a not is a bit random.
- The three dots will also show for private/SAP cloud apps, only not for licensed apps in the Mendix (Public) Cloud since we base this on the info we get from CP.
- The bullet point “Members” mentions that this tab shows active members, but it also shows deactivated members here.
- The bullet point “Environments” mentions that this tab shows all environments, but it is only the Mendix (Public) Cloud environments.

PR #4631, opened 2022-05-31
Add TODO message to Travis Log

MarkvanMents:
Add the actual text of the TODO to the Hugo warnings so you can see why there is a TODO in the build log and don't have to look at the actual file.


NatasaKralj (approved):
Looks good to me, tried a few combinations, it even lets you do multi-line to-do comments.

PR #4654, opened 2022-06-03
Visibility switch does not control running of microflows.

MarkvanMents:


MarkvanMents:
Approved by Marco in Slack frontend channel

PR #4660, opened 2022-06-07
Refer to system requirements for list of supported databases

basdebakker:
The list of supported databases was missing MariaDB. Instead of adding it, I thought it would be more robust to use a single source of truth and refer to that.


MarkvanMents (approved):
Thanks for the PR bas @basdebakker 
Great idea to reduce the number of places where we document supported databases 💯 
I'll merge immediately.

PR #4684, opened 2022-06-14
Reduce size of PNGs and GIFs

NatasaKralj:
Lots of images and GIFs were unnecessarily large, reducing the size will improve overall performance of docs site.
All images above 0.5 MB were resized to under that size.
All GIFs above 0.5 MB were optimized. Some got compressed more than others, two were unable to be optimized.
I deleted the GIFs in RNs, they seemed to showcase features and were not used in other places.


ConnorLand:
Nice! I wonder what the total size in bytes of these aggregate savings is.

PR #4694, opened 2022-06-16
Update versions of SAP Fiori themes for Mendix 9

MarkvanMents:


MarkvanMents:
Release confirmed over Slack [MCDEP-988]

PR #4700, opened 2022-06-17
Update query-over.md

mkumpf:
Added a link to a YouTube video demonstrating how to use the reversed() keyword.


MarkvanMents:
Hi Mike (@mkumpf)
Thanks for the submission. Just to let you know the status of it:
We don't have many videos at present, and the ones we do have are made by our internal video team. However, I know that many of our customers would like to see more instructional videos, so it is great that you have produced one.
Because we don't currently have a policy about linking to external videos, I need to talk to our head of video content and come up with criteria for approving and linking to video content before I approve this link.

I have a meeting with him later this week and will get back to you after that.
Let me know if you have any further questions before then.
Mark van Ments

MarkvanMents (approved):
Hi Mike, (@mkumpf )
I've now discussed this and have updated the PR to embed the video in the page. I have also credited you with the video.

Thanks for making the video. We get feedback from research that some customers find that short videos are very useful. We don't have the resources to make all these ourselves so it is great that you have done this in response to a forum question.

Mark

PR #4714, opened 2022-06-21
Changes after back porting the module to Mx 8.x

rover886:
Hi @johannahemminger, As discussed I have made some changes w.r.t source module and MX 8.x . Please have a look.

### DIFF
--- a/content/en/docs/appstore/connectors/email-connector.md
+++ b/content/en/docs/appstore/connectors/email-connector.md
@@ -202,7 +202,7 @@ When duplicating this microflow, do not change input parameters’ name and data
 Before subscribing to incoming email, it is recommended to attempt to unsubscribe from incoming email so that application will not end up having duplicate subscription for a single email account. The complete flow of subscription is shown in the microflow **ACT_SubscribeForEmailNotification**.
 {{% /alert %}}
 
-Please note that this subscription for new emails will only work if email account is configured with IMAP/S protocol and if email server supports notification. Also subscription will end if anyhow app is stopped. To subscribe again in between app restarts the microflow **IVK_SubscribeForEmailNotification_AfterStartup** should be registered in **After Startup** option. 
+Please note that this subscription for new emails will only work if email account is configured with IMAP/S protocol and if email server supports notification. Also subscription will end if anyhow app is stopped. To subscribe again in between app restarts the microflow **Sample_ASU_SubscribeForEmailNotification** should be registered in **After Startup** option. 
 
 ### 4.6 Unsubscribing from Incoming Email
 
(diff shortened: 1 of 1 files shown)


content/en/docs/appstore/connectors/email-connector.md
```diff
@@ -110,12 +110,12 @@ When the module is running, click **New** to compose and send new emails.
 
 When modeling your app in Studio Pro, use the **SendEmail** Java action to send emails. The input parameters are the following:
 
-* **Email Account** Email account consisting of incoming email configuration and/or outgoing email configuration.
+* **Email Account** Email account consisting of outgoing email configuration.
 * **Email Message** Email Message object to be sent
 
-The **Return type** is a boolean value. The Java action would connect to email server using the provided details and send an email, and would return **True** if successful.
+The **Return type** is a boolean value. The Java action would connect to email server using the provided details and send an email, and would return **True** if successful otherwise the exception along with the cause.
```
johannahemminger: @rover886 this is the only line I'm confused about, can you explain what "otherwise the exception along with the cause" means?

content/en/docs/appstore/connectors/email-connector.md
```diff
@@ -182,8 +182,8 @@ When modeling your app in Studio Pro, call the **SubscribeToIncomingEmail** Java
 
 The input parameters are the following:
 
-* **Email account** – email account consisting of incoming email configuration and/or outgoing email configuration.
-* **onNewEmailReceivedMicroflow** – a microflow that will be triggered when new email (List) is received from the server, You can process the list per your need. Make sure you have list of **Email_Connector.EmailMessage** as a parameter to this microflow. Refer to the sample microflow **Sample_OCH_EmailFetchMicroflow**
+* **Email account** – email account consisting of incoming email configuration.
+* **onNewEmailReceivedMicroflow** – a callback microflow that will be triggered when new email (List) is received from the server, You can process the list per your need. Make sure you have list of **Email_Connector.EmailMessage** as a parameter to this microflow. Refer to the sample microflow **Sample_OCH_EmailFetchMicroflow**
```
johannahemminger: @rover886 what exactly is a callback microflow?

content/en/docs/appstore/connectors/email-connector.md
rover886: While sending email if something goes wrong then java action will throw error object and that error object will also have the cause of the error. For ex. if mail server is not reachable while sending email then underlying library will throw error with cause something like "server host is unreachable"

content/en/docs/appstore/connectors/email-connector.md
rover886: The word "callback" defines that this microflow should not be called directly as a normal microflow (on a button click) rather this microflow should only be used by some other microflow/java action to trigger some events. In this case that event will be new email received.

content/en/docs/appstore/connectors/email-connector.md
johannahemminger: Ahhhhh understand

PR #4721, opened 2022-06-22
Document mx check and mx show-version

mrseanryan:
Document these mx commands which we released already:

- mx check
- mx show-version

note: this is mxassist-client reference AA-3747

SP version: 9.10+

### DIFF
--- a/content/en/docs/refguide/general/mx-command-line-tool.md
+++ b/content/en/docs/refguide/general/mx-command-line-tool.md
@@ -126,7 +126,7 @@ Return codes are described in the table below:
 | 2 | There is something wrong with the command-line options. |
 | 3 | Converting failed. |
 
-### 3.3 mx create-project Command [version 9+]
+### 3.3 mx create-project Command
 
 The `mx create-project` command creates a new app in the Studio Pro. The app version depends on the version the tool was bundled with. For example, if you are using the mx tool for Studio Pro version 8.1.0.58215,  `mx create project` will create a new app in that version. 
 
--- a/content/en/docs/refguide/general/mx-command-line-tool.md
+++ b/content/en/docs/refguide/general/mx-command-line-tool.md
@@ -140,7 +140,7 @@ The `OPTIONS` are described in the table below:
 | --- | --- | --- |
 | `app-name` | App | Assigns the specified app name to the app. |
 | `output-dir` | Current directory | The directory in which to create the app. |
-| `language-code` | Optional | The default language of the app. | 
+| `language-code` | en_US | The default language of the app. | 
 | `sprintr-app-id` | Optional | Associates the app [feedback features](/developerportal/collaborate/feedback/) with the provided [Developer Portal app](/developerportal/#my-apps). The value is a GUID. When accessing the app portal (on sprintr or on Team Server) it can be seen in the browser's URL - for example `1a428ea7-b00e-4166-9b23-20b7be88a40e`. |
 
 `TEMPLATE-MPK-FILE` is an optional path to a Mendix app package (*.mpk*) file. If this argument is omitted, the app is created with a default empty project template.
--- a/content/en/docs/refguide/general/mx-command-line-tool.md
+++ b/content/en/docs/refguide/general/mx-command-line-tool.md
@@ -21,7 +21,7 @@ The mx tool performs the commands described below.
 
 ### 3.1 mx check Command [version 9.10+]
 
-The `mx check` command checks the app MPR file for issues such as Errors, Warnings, Deprecations or Performance recommendations.
+The `mx check` command checks the app MPR file for issues such as Errors, Warnings or Deprecations.
 
 {{% alert color="info" %}}
 The MPR file must be the same version as mx.
@@ -40,7 +40,6 @@ The `OPTIONS` are described in the table below:
 | `--help` | `-h` | Displays the help text and exits. |
 | `--warnings` | `-w` | Include warnings in the output |
 | `--deprecations` | `-d` | Include deprecations in the output |
-| `--performance` | `-p` | Include performance checks in the output (performance recommendations are only output if there are no errors) |
 
 {{% alert color="info" %}}
 Errors in the MPR are always reported.
@@ -56,9 +55,7 @@ Examples of commands are described in the table below:
 | --- | --- |
 | `mx check --help` | Displays the help text for the check command. |
 | `mx check C:\MxProjects\App-main\App-main.mpr` | Checks the app at `C:\MxProjects\App-main\App-main.mpr` for errors. |
-| `mx check C:\MxProjects\App-main\App-main.mpr -p` | Checks the app at `C:\MxProjects\App-main\App-main.mpr` for errors and performance recommendations. |
-| `mx check C:\MxProjects\App-main\App-main.mpr --warnings --deprecations` | Checks the app at `C:\MxProjects\App-main\App-main.mpr` for errors, warnings, and deprecations. |
-| `mx check C:\MxProjects\App-main\App-main.mpr -w -d -p` | Checks the app at `C:\MxProjects\App-main\App-main.mpr` for errors, warnings, deprecations, and performance recommendations. |
+| `mx check C:\MxProjects\App-main\App-main.mpr --warnings --deprecations` | Checks the app at `C:\MxProjects\App-main\App-main.mpr` for errors, warnings and deprecations. |
 
 #### 3.1.3 Return Codes
 
@@ -70,9 +67,8 @@ Return codes are described in the table below:
 | 1 | Errors were found. |
 | 2 | Warnings were found. |
 | 4 | Deprecations were found. |
-| 8 | Performance recommendations were found. |
 
-Those values are logically OR combined to indicate when there are a mix of errors, warnings, deprecations, and performance recommendations.
+Those values are logically OR combined to indicate when there are a mix of errors, warnings, deprecations.
 
 For example:
 - 3 if errors and warnings found
--- a/content/en/docs/refguide/general/mx-command-line-tool.md
+++ b/content/en/docs/refguide/general/mx-command-line-tool.md
@@ -9,21 +9,19 @@ tags: ["mx", "command-line", "tool", "mx", "studio pro", "windows", "linux"]
 
 ## 1 Introduction
 
-The **mx tool** is a Windows and Linux command-line tool that can be used to do useful things with your Mendix app.
+The **mx tool** is a Windows and Linux command-line tool that can be used to perform various actions on a Mendix app.
 
 ## 2 Location
 
-Mendix Studio Pro comes with the mx command-line tool. The executable `mx.exe` file can be found in the same folder that contains `studiopro.exe` (for example, *C:\Program Files\Mendix\8.1.0.58215\modeler\mx.exe*).
+Mendix Studio Pro comes with the mx command-line tool. The executable `mx.exe` file can be found in the same folder that contains `studiopro.exe` (for example, *C:\Program Files\Mendix\9.12.2.44241\modeler\mx.exe*).
 
 ## 3 mx Tool Options
 
-The mx tool enables the options described below.
+The mx tool performs the commands described below.
 
 ### 3.1 mx check Command
 
-The `mx check` command checks the app for issues such as Errors, Warnings, Deprecations or Performance recommendations.
-
-The input is an MPR file.
+The `mx check` command checks the app MPR file for issues such as Errors, Warnings, Deprecations or Performance recommendations.
 
 #### 3.1.1 Usage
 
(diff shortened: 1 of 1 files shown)


mrseanryan:
@MarkvanMents ready for review

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -19,7 +19,60 @@ Mendix Studio Pro comes with the mx command-line tool. The executable `mx.exe` f
...
+
+The `mx check` command checks the app for issues such as Errors, Warnings, Deprecations or Performance recommendations.
+
+The input is an MPR file.
+
+#### 3.1.1 Usage
+
+Use the following command pattern for `mx check`:
+
+`mx convert [OPTIONS] INPUT`
```
MarkvanMents: mx check, surely.
```suggestion
`mx check [OPTIONS] INPUT`
```

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -71,11 +124,11 @@ Return codes are described in the table below:
 | 2 | There is something wrong with the command-line options. |
 | 3 | Converting failed. |
 
-### 3.2 mx create-project Command
+### 3.3 mx create-project Command
```
MarkvanMents, line 129: Should this be deprecated and replaced with
`mx create-app`
?

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -100,7 +153,7 @@ Examples of commands are described in the table below:
 | `mx create-project --app-name "MyFirstApp" --output-dir "C:/Projects/MyFirstApp"` | Creates an app named `MyFirstApp` in the *C:/Projects/MyFirstApp* folder using all the default parameters. |
```
MarkvanMents: The options here are quoted. In `mx convert` they are not quoted. Should we be consistent and always quote the parameters?

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -9,27 +9,82 @@ tags: ["mx", "command-line", "tool", "mx", "studio pro", "windows", "linux"]
...
+
+Those values are logically OR combined to indicate when there are a mix of errors, warnings, deprecations, and performance recommendations.
+
+For example:
+- 3 if errors and warnings found
+- 7 if errors, warnings, and deprecations found
+
+### 3.2 mx convert Command
+
+The `mx convert` command converts the app(s) MPK file(s) to a specific Studio Pro version. For example, if you are using the mx command-line tool for Mendix version 9.12.2.44241, then `mx convert` will convert the app to that version.
```
MarkvanMents, line 79: Why does this work on the MPK file when all the others work on the MPR file?
I think this could cause confusion and errors.
Studio Pro also works on the MPR file when it does a conversion.

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -9,27 +9,82 @@ tags: ["mx", "command-line", "tool", "mx", "studio pro", "windows", "linux"]
...
 ## 3 mx Tool Options
 
-The mx tool enables the options described below.
+The mx tool performs the commands described below.
 
-### 3.1 mx convert Command
+### 3.1 mx check Command
 
-The `mx convert` command converts the app(s) to a specific Studio Pro version. For example, if you are using the mx command-line tool for Mendix version 8.1.0.58215, then `mx convert` will convert the app to that version. 
+The `mx check` command checks the app MPR file for issues such as Errors, Warnings, Deprecations or Performance recommendations.
```
MarkvanMents: I assume this only works if the MPR file is the same version as mx.exe?
If so, I think we should say so.
And does this work for all versions of Mx9, or only from a particular release?

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -85,12 +140,12 @@ The `OPTIONS` are described in the table below:
...
 | `app-name` | App | Assigns the specified app name to the app. |
 | `output-dir` | Current directory | The directory in which to create the app. |
-| `language-code` | Optional | The default language of the app. |
+| `language-code` | Optional | The default language of the app. | 
 | `sprintr-app-id` | Optional | Associates the app [feedback features](/developerportal/collaborate/feedback/) with the provided [Developer Portal app](/developerportal/#my-apps). |
 
 `TEMPLATE-MPK-FILE` is an optional path to a Mendix app package (*.mpk*) file. If this argument is omitted, the app is created with a default empty project template.
 
-#### 3.2.2 Examples
+#### 3.3.2 Examples
```
MarkvanMents, line 148: Would be nice to have an example with a `sprintr-app-id` so customers can see the format of this.

We also don't use Sprintr in customer facing documentation, so can you think of a better name for this option? I assume that it isn't going to be the same as a Team Server app id?

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -85,12 +140,12 @@ The `OPTIONS` are described in the table below:
 | --- | --- | --- |
 | `app-name` | App | Assigns the specified app name to the app. |
 | `output-dir` | Current directory | The directory in which to create the app. |
-| `language-code` | Optional | The default language of the app. |
+| `language-code` | Optional | The default language of the app. |
```
MarkvanMents: Is this really completely optional, or does it default to en-US?
What happens in Studio Pro if there is no default language set?

MarkvanMents (changes requested):
Hi Sean (@mrseanryan )
I've reviewed the whole document, rather than just the changes.
If there is information which is only for Mx 9.16, then we can't release it until then. Should we take that out and put it into a separate PR so we can put this out now.
We should also be clear which commands work in which versions, so customers in 9.1 don't try something that only works in 9.12 and above.

Let me know what we can do here to get the information out to customers as soon as possible.
Mark

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: hm my understanding is that `show-version` and `check` are run on apps that are already in use on the user's machine (and any conversion has already been done). So they are in the 'extracted' state which has an MPR file.

Whereas `convert` is for some app that is not yet in use, and needs to be converted.
`convert` takes a MPK file, or else a folder.

If we want to change how this behaves, that would be some future work 🤔  not related to this PR.

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: Possibly - that would be for SOLP team

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: I will update.

notes:
- mx check == 9.10
- mx show-version == 9.4
- mx convert = (has always been in SP9)

future:
- mx check --performance ==  9.16

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: I believe `sprintr-app-id` is a Team Server app ID.

I will check  ...

mrseanryan: yes they are the same.

The value is a GUID. 
When accessing the app portal (on sprintr OR on Team Server) it can be seen in the browser's URL

sprintr example:
https://sprintr.home.mendix.com/link/buzz/1a428ea7-b00e-4166-9b23-20b7bf88a70d

team server example:

https://teamserver.sprintr.com/1a428ea7-b00e-4166-9b23-20b7bf88a70d

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: In the source code there is a hard-coded default `en_US`

mrseanryan: So, yes it is really optional.   

default = `en_US`

mrseanryan:
> Hi Sean (@mrseanryan ) I've reviewed the whole document, rather than just the changes. If there is information which is only for Mx 9.16, then we can't release it until then. Should we take that out and put it into a separate PR so we can put this out now. We should also be clear which commands work in which versions, so customers in 9.1 don't try something that only works in 9.12 and above.
> 
> Let me know what we can do here to get the information out to customers as soon as possible. Mark

Great!

(done) I will split out the 9.16 specific items to a separate PR.

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: hm normally the quoting is not needed.

so maybe it makes sense to remove the quotes ?

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: (unless I can do in some future crafting ... )

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: related: most of the commands use `Windows` folder separators like `\`
but `mx create-project` uses Unix style.

suggest: `mx create-project` text be changed to be consistent with the other commands

what do you think?

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: done

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: done

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: done

content/en/docs/refguide/general/mx-command-line-tool.md
MarkvanMents: Thanks for the explanation. I've changed the text to highlight more that it is MPKs and not MPRs. If we get lots of support calls I can add a big banner, but I'll leave it for now.

content/en/docs/refguide/general/mx-command-line-tool.md
MarkvanMents: I think putting en_US is more accurate. The first example shows that you don't need to put any options.
```suggestion
| `language-code` | en_US | The default language of the app. | 
```

content/en/docs/refguide/general/mx-command-line-tool.md
MarkvanMents: Great catch.
Yes, I think we should be consistent and use either all Windows style or all Unix style.
And perhaps remove the quotes - but I assume it will work if you quote directory paths with spaces. I think we can assume that customers know this.

MarkvanMents (changes requested):
Hi Sean,
Thanks for your responses.
Can you just check my suggestion and update the quoted strings and directory paths. Then I can publish this.

Thanks for all your work.

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -9,27 +9,82 @@ tags: ["mx", "command-line", "tool", "mx", "studio pro", "windows", "linux"]
...
+| 2 | Warnings were found. |
+| 4 | Deprecations were found. |
+
+Those values are logically OR combined to indicate when there are a mix of errors, warnings, deprecations.
+
+For example:
+- 3 if errors and warnings found
+- 7 if errors, warnings, and deprecations found
+
+### 3.2 mx convert Command [version 9+]
```
MarkvanMents: I think we can remove the version information as it is in all Mx 9 versions?
```suggestion
### 3.2 mx convert Command
```

content/en/docs/refguide/general/mx-command-line-tool.md
```diff
@@ -50,32 +105,32 @@ For `OUTPUT`, enter the output location for the converted results. Mind the foll
...
-| Exit Code | Description |
+| Return Code | Description |
 | --- | --- |
 | 0 | The conversion was successful. |
 | 1 | An internal error occurred. |
 | 2 | There is something wrong with the command-line options. |
 | 3 | Converting failed. |
 
-### 3.2 mx create-project Command
+### 3.3 mx create-project Command [version 9+]
```
MarkvanMents:
```suggestion
### 3.3 mx create-project Command
```

MarkvanMents (changes requested):
Hi - sorry another couple of suggestions.
I don't think we need to put versions on the functions which are available for all Mendix 9 versions, so i've suggested removing them.
Then, with the other suggestions, I think we can publish.

content/en/docs/refguide/general/mx-command-line-tool.md
mrseanryan: done

mrseanryan:
> Hi - sorry another couple of suggestions. I don't think we need to put versions on the functions which are available for all Mendix 9 versions, so i've suggested removing them. Then, with the other suggestions, I think we can publish.

done!

I think it is ready now, let me know if I missed anything.

MarkvanMents (approved):
Thanks Sean, @mrseanryan 
I'll publish this and then review the other one to see if it needs rebasing or anything.
Thanks for working with me.
Mark

PR #4729, opened 2022-06-24
Revert "Revert "Fix statement about changed date""

MarkvanMents:
Need to discuss with the developers.
This behaviour is not consistent for all versions of Mendix and behaviour in 9.5 and before needs to be mentioned.
Reverts mendix/docs#4728


MarkvanMents:
See internal Slack discussion here for reason.
https://mendix.slack.com/archives/C3T260XGA/p1655445857546739

jdsalchow:
Remarking that there was a change here is a really good idea, thanks @MarkvanMents!

PR #4732, opened 2022-06-27
[UIA-118] Use adapter microflows from Administration module

allard-mx:
The Administration module now contains 'adapter' microflows that can be used to set-up Mendix SSO v3.0.0 or higher. The use of the Administration module is recommended. This PR contains an update to the MendixSSO documentation, to point users towards these adapter microflows when configuring Mendix SSO.

 Users can still choose to use the microflows in the **MOVE_THIS** folder. This is described in a separate info alert.


content/en/docs/developerportal/deploy/mendix-cloud-deploy/mendix-sso.md
```diff
@@ -126,21 +126,25 @@ To enable Mendix SSO in your app, follow these steps:
 
 1. Import the [Mendix SSO module](https://marketplace.mendix.com/link/component/111349/) from the Mendix Marketplace.
 
-2. Add the microflow **MendixSSO_AfterStartup** to the **After startup** microflow by performing the following steps:
+2. Optionally, import the [Administration module](https://marketplace.mendix.com/link/component/23513) from the Mendix Marketplace (recommended).
```
MarkvanMents, line 128: Isn't the Administration module included in all starter apps? Customers would have to remove it deliberately for it not to be there?
Or are there some starter apps which don't have it?

content/en/docs/developerportal/deploy/mendix-cloud-deploy/mendix-sso.md
```diff
@@ -126,21 +126,25 @@ To enable Mendix SSO in your app, follow these steps:
...
+
+3. Add the microflow **MendixSSO_AfterStartup** to the **After startup** microflow by performing the following steps:
     1. Open **App Settings** from the **App Explorer**.
     2. Click the **Runtime** tab.
     3. Click **Select…** for the **After startup** microflow.
-    4. Choose the microflow **Marketplace modules** > **MendixSSO** > **MOVE_THIS** > **CustomizableMendixSSOMicroflows** > **MendixSSO_AfterStartup** (you can use the filter to find it quickly) and click **Select**.
+    4. Choose the microflow **Marketplace modules** > **Administration** > **Mendix SSO** >  **MendixSSO_AfterStartup** (you can use the filter to find it quickly) and click **Select**.
```
MarkvanMents, line 134: I believe these instructions were left deliberately neutral as to how customers want to administer users.
Would it be better to leave this document alone and add these instructions in the documentation of the [Administration Module](https://marketplace.mendix.com/link/component/23513)

MarkvanMents (commented):
Hi @allard-mx 
I'm not sure whether this is the right place to add this documentation. I think it might be better in the documentation of the Administration module so we can leave this document neutral as to how the customer does user administration. Particularly as most of the administration doesn't make sense for users who are using SSO.

I have asked for a review in the id-services Slack channel.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/mendix-sso.md
allard-mx: That's right, the module is by default included in all starter apps. This was meant to inform users that the Administration module is optional, and by no means required for Mendix SSO. I agree that the phrasing can be improved here.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/mendix-sso.md
allard-mx: We understood that the Mendix SSO + Administration combination is now the default and recommended way to use Mendix SSO in new apps. That is also the reason why both Mendix SSO and these adapter microflows are now by default included in the blank starter apps. You can still use Mendix SSO without the Administration module, but Administration contains a preconfigured set-up for using Mendix SSO. We would like to somehow point users to this option when they are trying to set-up Mendix SSO.

MarkvanMents: Thanks for the explanation.

I'd just like to get the SSO (identity services) team involved just to make sure that we get a good balance between how they think this is used and pointing the customers in the direction of the .
Unfortunately, they are doing planning this week - but I hope they can look at this in the next few days.
I'll pick this up as soon as I get a response from them.
Hope that is OK for timing.

zaipie:
Hi @allard-mx ,

What team are you in and who gave you this assignment?
I think something very wrong is happening here. I find it very strange that our team was not consulted on updating the documentation for our module, and that somebody we have never heard of is now telling us what the new 'default and recommended' way to use our module is.

@MarkvanMents is completely correct that we created the module and this documentation to be as simplistic and neutral as possible, so it can be used without any dependencies on or complications with other modules. I agree with Marks suggestion to add these instructions to the Administration module documentation.

If it really turns out to be very important to the new starter apps preconfigured set-up to be addressed on the MendixSSO documentation page, I suggest to add an extra section to chapter "5. Customizing Mendix SSO". (As from the atomic MendixSSO module's perspective, the addition of the Administration module is in fact a customization.)

"There are two ways you can modify the Mendix SSO module. You can use snippets from the Marketplace module Mendix SSO in your pages, or you can modify the Mendix SSO module in any way you like to support your end-user administration requirements."

This introduction already clearly mentions the possibility to modify the module in any way you like, this could support a section that explains the Mendix SSO + Administration combination being used in the starter apps.

And please, if this is necessary, come and talk to us before creating a merge request.

Kind regards,
Marijn
Identity Services team

PS - apologies if this comes over too strong, but I have lost count of the number of times the Studio and Workflow teams have taken over the narrative of our module without even communicating afterwards.

allard-mx:
Had a meeting with Marijn & Ronny yesterday in which we discussed this PR and how to move forward. We agreed to:
- Update the docs for the Administration module on how to use the Administration module with Mendix SSO
- Add an extra section to chapter "5. Customizing Mendix SSO" of the Mendix SSO docs that mentions the available adapter microflows in the Administration module, as suggested by Marijn in the comment above

I will close this PR and create a new one for these changes.

PR #4740, opened 2022-06-28
Notice about module deprecation

johannahemminger:


johannahemminger:
Hey @Luyao-Zhang-1 ! The email connector module that is currently being released will eventually replace these two modules (IMAP/POP3 and Email with Templates). It has not been decided when it will happen, but sometime in the next months. What do you think about adding this notice to these files?

Luyao-Zhang-1:
> 

@johannahemminger Looking good!

PR #4743, opened 2022-06-28
New Category for Lead Generation

rajesh-rakshe-mendix:
Added `App Services` category in list which support lead generation. Also corrected existing Category names with Capital case.


Adam-Dupaski (approved):

Adam-Dupaski:
@rajesh-rakshe-mendix category types do not get capital case when cited generically

PR #4758, opened 2022-06-30
security recommendation email template

Andries-Smit:
To prevent man in the middle attacks it would be good to advices developers to preferer SSL and combination with checking server identity.


Luyao-Zhang-1:
@Andries-Smit Thanks for raising the PR. 
May I ask you something? Perhaps it is very obvious for experienced users – but where can I find the **Server configuration** settings where I can set the **Use SSL option** in combination with **Use SSL check server identity**?

Andries-Smit:
Where it can be found in the running application depends on how the developer incorporates the **Administration** snippet
image
Or when setting up the module via the wizard
image

Luyao-Zhang-1:
@Andries-Smit Thank you for your input! I included these details in the doc. 
I will merge this PR tomorrow afternoon. If you have any comments, please let me know before that. Cheers!

PR #4766, opened 2022-07-01
Update text-box.md

RCatersels:
Under Accesibility you have currently two more options in Mx9: Screen reader caption and Aria required. Now the first one I can guess but no clue what that Aria required does. Could these options be explained by somebody from Mendix?


ConnorLand:
Hi @RCatersels . Thank you for your PR. 

Screen reader caption is actually already documented [here](https://docs.mendix.com/refguide/common-widget-properties/#21-screen-reader-caption), because it is a common page editor property. However you are correct that **Aria required** was not documented. I will add the correct documentation and then merge your PR.

Thank you for contributing to the documentation.

ConnorLand (approved):

PR #4797, opened 2022-07-11
Remove (page).URL from our layouts as it is deprecated

MarkvanMents:
Files are slightly different lengths, but cannot identify any differences.
Output compared in VS Code - no differences found
Formatted JSON in VS Code and compared first 150000 characters (cannot format entire file) - no differences found

### DIFF
--- a/layouts/_default/list.mxdocsalgolia.json
+++ b/layouts/_default/list.mxdocsalgolia.json
@@ -148,7 +148,7 @@ Define Global variables
                             {{- end -}}
                         {{- end -}}
                         {{- $scratch.SetInMap $uniqueHit "weight.tag_name" $tagName -}}
-                        {{- /* .URL takes into account any changes by using the url: tag in the front matter */ -}}
+                        {{- /* .RelPermalink gives the page address relative to the baseURL. It reflects any changes from the url: tag in the front matter */ -}}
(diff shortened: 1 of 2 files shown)


MarkvanMents:
Also removed the else in the navbar.html as this would also trigger a deprecation - cannot identify any issues.

layouts/_default/list.mxdocsalgolia.json
```diff
@@ -149,7 +149,7 @@ Define Global variables
                         {{- end -}}
                         {{- $scratch.SetInMap $uniqueHit "weight.tag_name" $tagName -}}
                         {{- /* .URL takes into account any changes by using the url: tag in the front matter */ -}}
```
NatasaKralj: I would only change the comment on line 151 to reflect the new parameter being used. While `.RelPermalink` does give the same result as `.Params.url` (grabs the front matter parameter), it is a parameter that can be configured separately through `permalink`.
"Grabs address relative to baseURL, can be affected by permalink parameter" as the comment perhaps?

NatasaKralj:
The changes look good, work well.

layouts/_default/list.mxdocsalgolia.json
MarkvanMents: Good idea - I'll do that, then merge.

PR #4824, opened 2022-07-18
OAuth 2.0 support for Email with Templates module

johannahemminger:


johannahemminger:
@rover886 I have added basically all the same documentation as we did to the IMAP/POP3 connector. Let me know what differences there are, or if there are additional sections we need to include. Thanks!

rover886:
@johannahemminger This module already has dependency on Encryption module and it's already mentioned in section 1.2.1. So I thought we can remove the section 3.1. What do you say?

rover886:
Also this module is only available in the Mendix 7 and 8 (there is no release for Mendix 9),  so accordingly there should be changes at line `This feature is supported for Studio Pro [7.23)](https://github.com/mendix/docs/blob/jh-emailwithtemplates-oauth2/releasenotes/studio-pro/7.23) and subsequent version of Studio Pro 7, and will soon be available for Studio Pro 8 and Studio Pro 9.`

rover886:
@johannahemminger We are going good, thanks to you :)

rover886:
@johannahemminger one more small change, under section 3.1.2 we need to change the Azure permission image for this module. As this module is only being used for sending emails so accordingly permissions should be as shown below:

image

rover886:
@johannahemminger This looks perfect now, please merge this to master. We will release the module by EOD

PR #4828, opened 2022-07-19
Recommend v1.x of OpenSSL

MarkvanMents:


geoffreyme (approved):

ThijsSeppen:
Seems fine! Perhaps we can add that recommend the latest 1.x (as long as we're recommending things haha)?

MarkvanMents:
> Seems fine! Perhaps we can add that recommend the latest 1.x (as long as we're recommending things haha)?

I'll do that - although you might expect people to use the latest.
Actually, looking at the download page and things that the developer is asked, perhaps it isn't obvious to use the latest version?

Issue #4803, opened 2022-07-12
Issue with the dependencies


Telping:
First dependency 

The Microsoft .NET Desktop Runtime 6.0 executable to windowsdesktop-runtime-6.0.3-win-x64.exe

.....is currently not working.... 

Downloading Microsoft .NET Desktop Runtime 6.0.6 and renaming to to windowsdesktop-runtime-6.0.3-win-x64.exe works

Second dependency 

The Java Development Kit 11 (x64) msi to adoptopenjdk_11_x64.msi

Needs to rename to adoptiumjdk_11_x64

MarkvanMents:
Hi Tarig (@Telping)
Thanks for raising this issue.

It is not clear which document you are talking about. If you use the installer online these dependencies should be installed automatically. If you are doing the [installation offline](https://docs.mendix.com/howto/general/install/#offline) step 5 actually mentions that you need to rename these files.
![image](https://user-images.githubusercontent.com/35492184/178432453-6f6ff0f2-b2a9-4a11-b621-1cfd65e9a6d5.png)

Did you come across this issue in a different scenario, or a different document? Or can you suggest a way of making this clearer?
Yours
Mark

Telping:
Thanks for your response Mark. Just heard from one of my team members, that was using the [installation offline ](https://docs.mendix.com/howto/general/install/#offline) documentation, they weren't able to complete the installation with the steps shown. But they did get it to work by replacing the first step:

The Microsoft .NET Desktop Runtime 6.0 executable to windowsdesktop-runtime-6.0.3-win-x64.exe

with the following

Microsoft .NET Desktop Runtime **6.0.6** and renaming to to windowsdesktop-runtime-6.0.3-win-x64.exe works

and the second step

The Java Development Kit 11 (x64) msi to adoptopenjdk_11_x64.msi

The Java Development Kit 11 (x64) msi to **adoptiumjdk_11_x64**

MarkvanMents:
Thanks for raising this Tarig @Telping 
Thanks for helping with the update which I've merged in #4831. Your help in Slack was invaluable in helping me identify what needed to be change.
Please let us know if you find any other documentation which is wrong or out of date.
Yours
Mark

Issue #4788, opened 2022-07-07
Install Mendix Studio Pro


NicolasBRUNAUD:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [Install Mendix Studio Pro](https://docs.mendix.com/howto/general/install/)

Document link: [install.md](https://github.com/mendix/docs/blob/development/content/en/docs/howto/general/install.md)

## My Issue/Suggestion

Mendix 9.14 has moved from adopt open JDK to eclispse.
Offline installation procedure is obsolete

MarkvanMents:
Thanks for raising this Nicolas @NicolasBRUNAUD 
I've picked up what I believe is the issue and merged it in #4831.
Please let us know if you find any other documentation which is wrong or out of date.
Yours
Mark

PR #4831, opened 2022-07-19
Update offline installation instructions

MarkvanMents:
This addresses issues #4788 and #4803 which identify issues with the offline installation instructions with later versions of Mendix Studio Pro 9


MarkvanMents:
Looks good from R&D PoV. Will merge now and pick up later if there are any further issues.

PR #4841, opened 2022-07-22
WIP: Shorten File Paths

MarkvanMents:
Some file paths are very long which causes issues for local Hugo builds.
This PR reduces the length of some folder names.


MarkvanMents:
Throwing this change away - don't need to change links etc. as we use URL to specify the URL on the site.

PR #4844, opened 2022-07-22
WIP: Shorten File Paths

MarkvanMents:
Some File Paths are very long. This causes issues for local builds where Windows path lengths are exceeded.
Shortening some file paths - but without changing the URL of the resulting pages.


MarkvanMents:
Discussed with @NatasaKralj - will take another approach.
Closing this PR.

PR #4849, opened 2022-07-25
fix link in documentation

ThijsSeppen:
Not sure whether this is the correct way (but looks like it, just copied the way it's handled for conditional visibility just above it). The issue was that the link was not working.


ConnorLand (approved):

ConnorLand:
Thanks for this correction @ThijsSeppen !

PR #4851, opened 2022-07-25
Feat/add preference doc 

DISW-Mendix-3DViewer:
add user guide of new widget: Preference widget


DISW-Mendix-3DViewer:
@Luyao-Zhang-1

PR #4871, opened 2022-07-29
Updated component limitations

jaspervanderhoek:
Added limitations on Security Mode to the known module limitations


johannahemminger:
Hi @jaspervanderhoek , thanks for submitting this PR! Just did a quick style check, but want to ask if this is something that is already released, or is for an upcoming one?

jaspervanderhoek:
This change is already applicable on the original version of the component
we just didn't realize it. The use case for this component is enormous so
finding what it can and can't do is a continued effort. FYI, there might be
more limitations added in the future. If you see those, it's all for this
same reason.

PR #4874, opened 2022-08-02
Remove the note

Luyao-Zhang-1:
Remove the note as the information is already contained in the doc.


Luyao-Zhang-1:
Revert [PR4872](https://github.com/mendix/docs/pull/4872).

PR #4884, opened 2022-08-04
Adjust code fence styling

NatasaKralj:
Ths PR adjusts the padding within a code fence, along with the font size within.

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -7,10 +7,11 @@
 
         margin: 2rem 0;
         padding: 0;
+        font-size: 0.75rem; // NK - added to match code text
 
         pre {
             margin: 0;
-            padding: 1.75rem 1rem 1rem 0.75rem;  // NK - changed padding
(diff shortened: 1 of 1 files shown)


MarkvanMents (commented):
Looks good. Difference is quite subtle but definitely an improvement.
The only thing I wonder is whether we can change any of the spacing around the Prism-formatted code block. There seems to be a lot of space above and below - perhaps we can't remove the spacing above, where the `copy` button appears. But perhaps the space below can be removed. You could try `padding: 1.5rem 0.5rem 0.5rem 0.5rem;` in `.td-content .highlight pre` - but perhaps that will affect other tables (the tables in Indentation and Spacing Test seem to be immune from this change)?

NatasaKralj:
I used your suggestion, along with standardizing the size of the text. Using percentages and/or other units did not give the results I wanted. 
I looked at a few tables, they don't seem to be affected at all.
Inline code text is also larger by default now.

MarkvanMents (approved):
Looks good - thanks for the improvement.

PR #4912, opened 2022-08-12
Add smiley to give an example

johannahemminger:
Just an example for YY


johannahemminger:
Closing this because no smileys are needed.

PR #4945, opened 2022-08-19
[WTF-1126] Add documentation for linked association properties (9.17)

LEGIO-SEXTA-FERRATA:
As WTF we are introducing a new feature in 9.17 with which we are enabling data source linkable association properties for pluggable widgets. Meaning, association properties are now able to have their own data sources similar to attribute properties. This MR is for updating the pages and adding the related new sections in the Pluggable Widgets API part for the Mendix Docs.

### DIFF
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -461,13 +461,13 @@ You can obtain an instance of `EditableValue` by using the `ListAttributeValue` 
 
 #### 3.2.2 Attribute ID, Sortable and Filterable Flags {#listattributevalue-id-sortable-filterable}
 
-`id` field of type `ListAttributeId` represents the unique randomly generated string identifier of an attribute. That identifier could be used when applying sorting and filtering on a linked data source property to identify which attribute should be used for sorting and/or filtering. Check [Sorting](#listvalue-sorting) and [Filtering](#listvalue-filtering) sections for more information.
+`id` field of type `ListAttributeId` represents the unique randomly generated string identifier of an attribute. That identifier could be used when applying sorting and filtering on a linked data source property to identify which attribute should be used for sorting and/or filtering. For more information, see the [Sorting](#listvalue-sorting) and [Filtering](#listvalue-filtering) sections.
 
-Fields `sortable` and `filterable` specify if the attribute could be used for sorting and/or filtering. Those flags have to be checked before a widget applies filtering or sorting on a data source property. An attempt to filter on a non-filterable attribute or sort on a non-sortable attribute would lead to an error during the execution time.
+Fields `sortable` and `filterable` specify if the attribute could be used for sorting and/or filtering. Those flags have to be checked before a widget applies filtering or sorting on a data source property. Any attempt to filter on a non-filterable attribute or sort on a non-sortable attribute leads to an error during the execution time.
 
 #### 3.2.3 Attribute Type
 
-[Attribute](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#attribute) property defines which attribute types could be configured for that property. For example, an attribute property may be configured to allow attributes of type `String` and `Integer` in order to present progress. While this is convenient for users it may require some additional work for a developer by processing different data types.
+The [attribute](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#attribute) property defines which attribute types can be configured for that property. For example, an attribute property may be configured to allow attributes of type `String` and `Integer` in order to present progress. While this is convenient for users, it may require additional work for developers by processing different data types.
 
 It is possible to determine the type of attribute by checking the `type` field on an attribute property. The following code sample shows how to check the attribute type on the property named `myAttributeOnDatasource`:
 
@@ -483,9 +483,9 @@ if (this.props.myAttributeOnDatasource.type === "String") {
 
 #### 3.2.4 Formatter and Universe
 
-`formatter` field represents the default formatter that is going to be used on values obtained by `get` function.
+The `formatter` field represents the default formatter that is going to be used on values obtained by `get` function.
 
-Optional `universe` field represents an array of possible values of an attribute. See `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value) for more information.
+Optional `universe` field represents an array of possible values of an attribute. For more information, see the `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value).
 
 ### 3.3 ListReferenceValue and ListReferenceSetValue {#listassociationvalue}
 
@@ -506,9 +506,9 @@ export interface ListAssociationValue<T extends ObjectItem | ObjectItem[]> {
 
 #### 3.3.1 Obtaining Association Values
 
-In order to work with an object or objects that are associated with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. 
+In order to work with an object or objects with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. 
 
-If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type. See [Association Type](#association-type) section for more information.
+If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type. For more information, see the [Association Type](#association-type) section.
 
 Consult the following example code, which assumes widget properties are configured with the `myAssociationOnDatasource` property allowing association of type `Reference`:
 
@@ -527,19 +527,19 @@ The following code example shows how to get a `DynamicValue<ObjectItem>` that re
 const associationValue = this.props.myAssociationOnDatasource.get(this.props.myDataSource.items[0]);
 ```
 
-This will return an `ObjectItem` representing the associated object, since in this example, the widget is configured to allow only singular associations. If you want to access the individual attribute values of this associated object, you may use an attribute property linked to the selectable objects' data source and pass the associated object to it. For more information, see [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value).
+This will return an `ObjectItem` representing the associated object, because in this example the widget is configured to allow only singular associations. If you want to access the individual attribute values of this associated object, you may use an attribute property linked to the selectable objects' data source and pass the associated object to it. For more information, see [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value).
 
 Please note these code samples omit checks of `myDataSource` status and availability of items for simplicity. See [DynamicValue section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#dynamic-value) for more information on the usages of `DynamicValue`.
 
 #### 3.3.2 Association ID and Filterable Flags {#listassociationvalue-id-filterable}
 
-`id` field of type `ListAssociationId` represents the unique randomly generated string identifier of an association. That identifier could be used when applying filtering on a linked data source property to identify which association should be used for filtering. Check [Filtering](#listvalue-filtering) sections for more information.
+The `id` field of type `ListAssociationId` represents the unique randomly-generated string identifier of an association. That identifier can be used when applying filtering on a linked data source property to identify which association should be used for filtering. For more information, see the [Filtering](#listvalue-filtering) section.
 
-`filterable` field specifies if the association could be used for filtering. This flag has to be checked before a widget applies filtering on a data source property. An attempt to filter on a non-filterable association would lead to an error during the execution time.
+THe `filterable` field specifies if the association can be used for filtering. This flag has to be checked before a widget applies filtering on a data source property. An attempt to filter on a non-filterable association leads to an error during the execution time.
 
 #### 3.3.3 Association Type {#association-type}
 
-[Association](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) property determines which association types could be configured for that property. For example, an association property may be configured to allow associations of type `Reference` and not `ReferenceSet`.
+The [association](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) property determines which association types could be configured for that property. For example, an association property may be configured to allow associations of type `Reference` and not `ReferenceSet`.
 
 It is possible to determine the type of association by checking the `type` field on an association property. This is useful if the property has been configured to allow both references and reference sets. The following code sample shows how to check the association type on the property named `myAssociationOnDatasource`:
 
@@ -554,8 +554,7 @@ if (this.props.myAssociationOnDatasource.type === "Reference") {
 
 ### 3.4 ListWidgetValue {#listwidgetvalue}
 
-`ListWidgetValue` represents a [widget property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#widgets) that is linked to a data source. 
-This allows the client component to render child widgets with items from a `ListValue`.
+`ListWidgetValue` represents a [widget property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#widgets) that is linked to a data source. This allows the client component to render child widgets with items from a `ListValue`.
 `ListWidgetValue` is an object and its definition is as follows:
 
 ```ts
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -269,7 +269,7 @@ if (this.props.myAssociationReference.filterable) {
 }
 ```
 
-Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects data source:
+Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable object's data source:
 
 ```ts
 import { association, literal, notEquals, contains } from "mendix/filters/builders";
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -201,16 +201,19 @@ It is possible to set filtering conditions for items of a datasource. `setFilter
 
 `setFilter` accepts only a specially created object of type `FilterCondition` that describes desired filtering behavior. It is possible to create a filter conditions object using functions provided in `mendix` module under `mendix/filters/builders` path. Those functions we call filter builders. Also `undefined` could be passed to `setFilter` to clear filtering conditions.
 
-Some examples of builder functions are `equals`, `greaterThan`, `lessThanOrEqual` for filtering on `DateTime` or `Decimal` attributes. Functions like `startsWith`, `contains` are useful for filtering on `String` attributes.
+Some examples of builder functions are `equals`, `greaterThan`, `lessThanOrEqual` for filtering on `DateTime` or `Decimal` attributes. Functions like `startsWith`, `contains` are useful for filtering on `String` attributes. Filtering based on associations is also possible. For example, you can use `equals` with references and `contains` with reference sets.
 
-The following code samples show how to use filter builders and apply filtering to a data source property with three linked attributes:
+The following code samples show how to use filter builders and apply filtering to a data source property with three linked attributes and two linked associations:
 
 ```ts
 interface MyListWidgetsProps {
     myDataSource: ListValue;
+    mySelectableObjects: ListValue;
     myAttributeString: ListAttributeValue<string>;
     myAttributeBoolean: ListAttributeValue<boolean>;
     myAttributeNumber: ListAttributeValue<BigJS>;
+    myAssociationReference: ListReferenceValue;
+    myAssociationReferenceSet: ListReferenceSetValue;
 }
 ```
 
@@ -236,7 +239,7 @@ if (this.props.myAttributeString.filterable) {
 }
 ```
 
-First step the code takes is checking for the possibility to use filtering on `myAttributeString` property by checking `filterable` flag. Then `filterCond` filter condition is constructed which specifies that attribute represented by `myAttributeString` should start with character `"B"`. `setFilter` call applies the filter, and on the next re-render the component gets only items where the value of an attribute represented by property `myAttributeString` is starting with `"B"`.
+First step the code takes is checking for the possibility to use filtering on `myAttributeString` property by checking the `filterable` flag. Then `filterCond` filter condition is constructed which specifies that attribute represented by `myAttributeString` should start with character `"B"`. `setFilter` call applies the filter, and on the next re-render the component gets only items where the value of an attribute represented by property `myAttributeString` is starting with `"B"`.
 
 Similarly, to apply a condition where the value on an attribute represented by `myAttributeBoolean` property is set to true:
 
@@ -252,6 +255,38 @@ if (this.props.myAttributeBoolean.filterable) {
 }
 ```
 
+And, to apply a condition to match only objects that are associated with another object:
+
+```ts
+import { association, literal, notEquals, empty } from "mendix/filters/builders";
+
+// in the widget code
+if (this.props.myAssociationReference.filterable) {
+    const filterCond = notEquals(association(this.props.myAssociationReference.id), empty());
+    this.props.myDataSource.setFilter(filterCond);
+} else {
+    console.log("Association is not filterable");
+}
+```
+
+Similarly, to apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects data source:
+
+```ts
+import { association, literal, notEquals, contains } from "mendix/filters/builders";
+
+// in the widget code
+if (this.props.myAssociationReferenceSet.filterable) {
+    // assuming those two objects are available
+    const objectItem1 = this.props.mySelectableObjects.items[0];
+    const objectItem2 = this.props.mySelectableObjects.items[1];
+    
+    const filterCond = contains(association(this.props.myAssociationReferenceSet.id), literal([objectItem1, objectItem2]));
+    this.props.myDataSource.setFilter(filterCond);
+} else {
+    console.log("Association is not filterable");
+}
+```
+
 The following code sample shows how to remove current filtering condition:
 
 ```ts
@@ -260,10 +295,10 @@ this.props.myDataSource.setFilter(undefined);
 
 #### 2.4.2 Advanced Filtering {#advanced-filtering}
 
-In some use cases it is necessary to apply more complex filtering conditions. For example if a use case requires fetching only items where `myAttributeString` starts with `"B"` and `myAttributeBoolean` is set to `true`, or items where `myAttributeNumber` is greater than `10`. In order to construct such condition special filter builders `and` and `or` have to be used. The following code sample shows how to use them. Note that check for `filterable` flags omitted for simplicity. Real widgets should always take `filterable` flag into account.
+In some use cases it is necessary to apply more complex filtering conditions. For example if a use case requires fetching only items where `myAttributeString` starts with `"B"` and `myAttributeBoolean` is set to `true`, or items where `myAttributeNumber` is greater than `10` and `myAssociationReference` is associated with another object while `myAssociationReferenceSet` is not associated with any other objects. In order to construct such a condition, special filter builders `and` and `or` have to be used. The following code sample shows how to use them. Note that checks for the `filterable` flag have been omitted for simplicity. Real widgets should always take the `filterable` flag into account.
 
 ```ts
-import { attribute, literal, startsWith, equals, greaterThan, and, or } from "mendix/filters/builders";
+import { attribute, association, literal, startsWith, equals, notEquals, greaterThan, and, or } from "mendix/filters/builders";
 
 // in the widget code
 if (/* check that all properties are filterable */) {
@@ -272,7 +307,11 @@ if (/* check that all properties are filterable */) {
             startsWith(attribute(this.props.myAttributeString.id), literal("B")),
             equals(attribute(this.props.myAttributeBoolean.id), literal(true))
         ),
-        greaterThan(attribute(his.props.myAttributeNumber.id), literal(10))
+        and(
+            greaterThan(attribute(this.props.myAttributeNumber.id), literal(10)),
+            notEquals(association(this.props.myAssociationReference.id), empty()),
+            equals(association(this.props.myAssociationReferenceSet.id), empty())
+        )
     );
     this.props.myDataSource.setFilter(filterCond);
 } else {
@@ -389,7 +428,7 @@ export interface ListAttributeValue<T extends AttributeValue> {
 }
 ```
 
-#### 3.2.1 Obtaining Attribute Value
+#### 3.2.1 Obtaining Attribute Value {#obtaining-attribute-value}
 
 {{% alert color="warning" %}}
 Due to a technical limitation it is not yet possible to edit attributes obtained via `ListAttributeValue`. `EditableValue`s returned by `ListAttributeValue` are always **readonly**.
@@ -428,9 +467,9 @@ Fields `sortable` and `filterable` specify if the attribute could be used for so
 
 #### 3.2.3 Attribute Type
 
-[Attribute](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#attribute) property determines which attribute types could be configured for that property. For example a property may allow attributes of type `String` and `Integer` in order to present progress. While this is convenient for users it may require some additional work for a developer by processing different data types.
+[Attribute](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#attribute) property defines which attribute types could be configured for that property. For example, an attribute property may be configured to allow attributes of type `String` and `Integer` in order to present progress. While this is convenient for users it may require some additional work for a developer by processing different data types.
 
-It is possible to determine type of attribute by checking `type` field on a property. The following code sample shows how to check the attribute type on the property named `myAttributeOnDatasource`:
+It is possible to determine the type of attribute by checking the `type` field on an attribute property. The following code sample shows how to check the attribute type on the property named `myAttributeOnDatasource`:
 
 ```ts
 if (this.props.myAttributeOnDatasource.type === "String") {
@@ -448,7 +487,72 @@ if (this.props.myAttributeOnDatasource.type === "String") {
 
 Optional `universe` field represents an array of possible values of an attribute. See `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value) for more information.
 
-### 3.3 ListWidgetValue {#listwidgetvalue}
+### 3.3 ListReferenceValue and ListReferenceSetValue {#listassociationvalue}
+
+`ListReferenceValue` and `ListReferenceSetValue` are both used to represent an [association property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) that is linked to a data source.
+This allows the client component to access associated values of individual items from a `ListValue`. `ListReferenceValue` and `ListReferenceSetValue` are both objects and their definitions are as follows:
+
+```ts
+export type ListReferenceValue = ListAssociationValue<ObjectItem> & { type: "Reference" };
+
+export type ListReferenceSetValue = ListAssociationValue<ObjectItem[]> & { type: "ReferenceSet" };
+
+export interface ListAssociationValue<T extends ObjectItem | ObjectItem[]> {
+  get: (item: ObjectItem) => DynamicValue<T>;
+
+  id: ListAssociationId;
+  filterable: boolean;
+}
+```
+
+#### 3.3.1 Obtaining Association Value
+
+In order to work with an object or objects that are associated with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type. See [Association Type](#association-type) section for more information.
+
+Let's take a look at some example. Assuming widget properties are configured as follows with `myAssociationOnDatasource` property allowing association of type `Reference`:
+
+```ts
+interface MyListWidgetsProps {
+    myDataSource: ListValue;
+    mySelectableObjects: ListValue;
+    myAssociationOnDatasource: ListReferenceValue;
+    myAttributeOnSelectableObjects: ListAttributeValue;
+}
+```
+
+The following code sample shows how to get a `DynamicValue<ObjectItem>` that represents a read-only value of an associated object of the first element from the `myDataSource`.
+
+```ts
+const associationValue = this.props.myAssociationOnDatasource.get(this.props.myDataSource.items[0]);
+```
+
+This will return an `ObjectItem` representing the associated object, since in this example, the widget is configured to allow only singular associations.
+If you want to access the individual attribute values of this associated object, you can use an attribute property linked to the selectable objects data source and pass the associated object to it. Refer to [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value) for details on how to do this.
+
+Note: in this code sample checks of status of `myDataSource` and availability of items are omitted for simplicity. See [DynamicValue section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#dynamic-value) for more information about usage of `DynamicValue`.
+
+#### 3.3.2 Association ID and Filterable Flags {#listassociationvalue-id-filterable}
+
+`id` field of type `ListAssociationId` represents the unique randomly generated string identifier of an association. That identifier could be used when applying filtering on a linked data source property to identify which association should be used for filtering. Check [Filtering](#listvalue-filtering) sections for more information.
+
+`filterable` field specifies if the association could be used for filtering. This flag has to be checked before a widget applies filtering on a data source property. An attempt to filter on a non-filterable association would lead to an error during the execution time.
+
+#### 3.3.3 Association Type {#association-type}
+
+[Association](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) property determines which association types could be configured for that property. For example, an association property may be configured to allow associations of type `Reference` and not `ReferenceSet`.
+
+It is possible to determine the type of association by checking the `type` field on an association property. This is useful if the property has been configured to allow both references and reference sets. The following code sample shows how to check the association type on the property named `myAssociationOnDatasource`:
+
+```ts
+if (this.props.myAssociationOnDatasource.type === "Reference") {
+  console.log("Reference association");
+} else {
+  // TypeScript will narrow it down to "ReferenceSet" when the type is not equal to "Reference"
+  console.log("ReferenceSet association");
+}
+```
+
+### 3.4 ListWidgetValue {#listwidgetvalue}
 
 `ListWidgetValue` represents a [widget property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#widgets) that is linked to a data source. 
 This allows the client component to render child widgets with items from a `ListValue`.
@@ -483,7 +587,7 @@ The `get` method was introduced in Mendix Studio Pro v9.0.
 You can obtain an instance of `ReactNode` by using the `ListWidgetValue` as a function and calling it with an item. This is deprecated, will be removed in Mendix 10, and should be replaced by a call to the `get` function.
 {{% /alert %}}
 
-### 3.4 ListExpressionValue {#listexpressionvalue}
+### 3.5 ListExpressionValue {#listexpressionvalue}
 
 `ListExpressionValue` represents an [expression property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#expression) or [text template property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#texttemplate) that is linked to a data source. This allows the client component to access expression or text template values for individual items from a `ListValue`. `ListExpressionValue` is an object and its definition is as follows:
 
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -483,9 +483,9 @@ if (this.props.myAttributeOnDatasource.type === "String") {
 
 #### 3.2.4 Formatter and Universe
 
-The `formatter` field represents the default formatter that is going to be used on values obtained by `get` function.
+The `formatter` field represents the default formatter used on values obtained by the `get` function.
 
-Optional `universe` field represents an array of possible values of an attribute. For more information, see the `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value).
+The optional `universe` field represents an attribute's possible array values. For more information, see the `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value).
 
 ### 3.3 ListReferenceValue and ListReferenceSetValue {#listassociationvalue}
 
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -239,9 +239,9 @@ if (this.props.myAttributeString.filterable) {
 }
 ```
 
-First step the code takes is checking for the possibility to use filtering on `myAttributeString` property by checking the `filterable` flag. Then `filterCond` filter condition is constructed which specifies that attribute represented by `myAttributeString` should start with character `"B"`. `setFilter` call applies the filter, and on the next re-render the component gets only items where the value of an attribute represented by property `myAttributeString` is starting with `"B"`.
+The first step the code takes is checking for the possibility to use filtering on `myAttributeString` property by checking the `filterable` flag. Then the `filterCond` filter condition is constructed, which specifies that attribute represented by `myAttributeString` should start with character `"B"`. The `setFilter` call applies the filter, and on the next re-render the component gets only items where the value of an attribute represented by property `myAttributeString` is starting with `"B"`.
 
-Similarly, to apply a condition where the value on an attribute represented by `myAttributeBoolean` property is set to true:
+Similarly, code like this can apply a condition where the value on an attribute represented by `myAttributeBoolean` property is set to true:
 
 ```ts
 import { attribute, literal, equals } from "mendix/filters/builders";
@@ -255,7 +255,7 @@ if (this.props.myAttributeBoolean.filterable) {
 }
 ```
 
-And, to apply a condition to match only objects that are associated with another object:
+Code like this can apply a condition to match only objects that are associated with another object:
 
 ```ts
 import { association, literal, notEquals, empty } from "mendix/filters/builders";
@@ -269,7 +269,7 @@ if (this.props.myAssociationReference.filterable) {
 }
 ```
 
-Similarly, to apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects data source:
+Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects data source:
 
 ```ts
 import { association, literal, notEquals, contains } from "mendix/filters/builders";
@@ -287,7 +287,7 @@ if (this.props.myAssociationReferenceSet.filterable) {
 }
 ```
 
-The following code sample shows how to remove current filtering condition:
+The following code sample shows how to remove the current filtering condition:
 
 ```ts
 this.props.myDataSource.setFilter(undefined);
@@ -489,8 +489,7 @@ Optional `universe` field represents an array of possible values of an attribute
 
 ### 3.3 ListReferenceValue and ListReferenceSetValue {#listassociationvalue}
 
-`ListReferenceValue` and `ListReferenceSetValue` are both used to represent an [association property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) that is linked to a data source.
-This allows the client component to access associated values of individual items from a `ListValue`. `ListReferenceValue` and `ListReferenceSetValue` are both objects and their definitions are as follows:
+`ListReferenceValue` and `ListReferenceSetValue` are both used to represent an [association property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) that is linked to a data source. This allows the client component to access associated values of individual items from a `ListValue`. `ListReferenceValue` and `ListReferenceSetValue` are both objects and their definitions are as follows:
 
 ```ts
 export type ListReferenceValue = ListAssociationValue<ObjectItem> & { type: "Reference" };
@@ -505,11 +504,13 @@ export interface ListAssociationValue<T extends ObjectItem | ObjectItem[]> {
 }
 ```
 
-#### 3.3.1 Obtaining Association Value
+#### 3.3.1 Obtaining Association Values
 
-In order to work with an object or objects that are associated with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type. See [Association Type](#association-type) section for more information.
+In order to work with an object or objects that are associated with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. 
 
-Let's take a look at some example. Assuming widget properties are configured as follows with `myAssociationOnDatasource` property allowing association of type `Reference`:
+If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type. See [Association Type](#association-type) section for more information.
+
+Consult the following example code, which assumes widget properties are configured with the `myAssociationOnDatasource` property allowing association of type `Reference`:
 
 ```ts
 interface MyListWidgetsProps {
@@ -520,16 +521,15 @@ interface MyListWidgetsProps {
 }
 ```
 
-The following code sample shows how to get a `DynamicValue<ObjectItem>` that represents a read-only value of an associated object of the first element from the `myDataSource`.
+The following code example shows how to get a `DynamicValue<ObjectItem>` that represents a read-only value of an associated object of the first element from the `myDataSource`:
 
 ```ts
 const associationValue = this.props.myAssociationOnDatasource.get(this.props.myDataSource.items[0]);
 ```
 
-This will return an `ObjectItem` representing the associated object, since in this example, the widget is configured to allow only singular associations.
-If you want to access the individual attribute values of this associated object, you can use an attribute property linked to the selectable objects data source and pass the associated object to it. Refer to [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value) for details on how to do this.
+This will return an `ObjectItem` representing the associated object, since in this example, the widget is configured to allow only singular associations. If you want to access the individual attribute values of this associated object, you may use an attribute property linked to the selectable objects' data source and pass the associated object to it. For more information, see [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value).
 
-Note: in this code sample checks of status of `myDataSource` and availability of items are omitted for simplicity. See [DynamicValue section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#dynamic-value) for more information about usage of `DynamicValue`.
+Please note these code samples omit checks of `myDataSource` status and availability of items for simplicity. See [DynamicValue section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#dynamic-value) for more information on the usages of `DynamicValue`.
 
 #### 3.3.2 Association ID and Filterable Flags {#listassociationvalue-id-filterable}
 
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
@@ -269,7 +269,7 @@ if (this.props.myAssociationReference.filterable) {
 }
 ```
 
-Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable object's data source:
+Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects' data source:
 
 ```ts
 import { association, literal, notEquals, contains } from "mendix/filters/builders";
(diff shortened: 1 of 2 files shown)


ConnorLand:
Hi @legio-vi-ferrata . Is this API doc update associated with a future release date?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -201,16 +201,19 @@ It is possible to set filtering conditions for items of a datasource. `setFilter
...
 
 ```ts
 interface MyListWidgetsProps {
     myDataSource: ListValue;
+    mySelectableObjects: ListValue;
     myAttributeString: ListAttributeValue<string>;
     myAttributeBoolean: ListAttributeValue<boolean>;
     myAttributeNumber: ListAttributeValue<BigJS>;
+    myAssociationReference: ListAssociationValue<ObjectItem>;
+    myAssociationReferenceSet: ListAssociationValue<ObjectItem[]>;
```
tamelbrun: This should be `ListReferenceValue` and `ListReferenceSetValue`, the types we actually generate in the pluggable widget tools.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -201,16 +201,19 @@ It is possible to set filtering conditions for items of a datasource. `setFilter
 
 `setFilter` accepts only a specially created object of type `FilterCondition` that describes desired filtering behavior. It is possible to create a filter conditions object using functions provided in `mendix` module under `mendix/filters/builders` path. Those functions we call filter builders. Also `undefined` could be passed to `setFilter` to clear filtering conditions.
 
-Some examples of builder functions are `equals`, `greaterThan`, `lessThanOrEqual` for filtering on `DateTime` or `Decimal` attributes. Functions like `startsWith`, `contains` are useful for filtering on `String` attributes.
+Some examples of builder functions are `equals`, `greaterThan`, `lessThanOrEqual` for filtering on `DateTime` or `Decimal` attributes. Functions like `startsWith`, `contains` are useful for filtering on `String` attributes. Filtering based on associations is also possible. In this case, you can use `equals` and `notEquals` with references,  while `contains` is the only function available for reference sets.
```
tamelbrun: This is incorrect: `equals` and `notEquals` are also available for reference sets, but only for comparison with `empty()`.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -252,6 +255,38 @@ if (this.props.myAttributeBoolean.filterable) {
...
+```ts
+import { association, literal, notEquals, empty } from "mendix/filters/builders";
+
+// in the widget code
+if (this.props.myAssociationReference.filterable) {
+    // assuming those two objects are available
+    const objectItem1 = this.props.mySelectableObjects.items[0];
+    const objectItem2 = this.props.mySelectableObjects.items[1];
+    
+    const filterCond = contains(association(this.props.myAssociationReference.id), literal([objectItem1, objectItem2]));
```
tamelbrun: This should be referring to `this.props.myAssociationReferenceSet`.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -252,6 +255,38 @@ if (this.props.myAttributeBoolean.filterable) {
...
+    this.props.myDataSource.setFilter(filterCond);
+} else {
+    console.log("Association is not filterable");
+}
+```
+
+Similarly, to apply a condition so that only the objects that are associated with at least the first two objects that are brought by the selectable objects datasource will be returned:
+
+```ts
+import { association, literal, notEquals, empty } from "mendix/filters/builders";
```
tamelbrun: `empty()` is not actually used in this example, but it might be a good idea to add that. At the same time, `contains` is missing from this import.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -252,6 +255,38 @@ if (this.props.myAttributeBoolean.filterable) {
...
+// in the widget code
+if (this.props.myAssociationReference.filterable) {
+    const filterCond = notEquals(association(this.props.myAssociationReference.id), empty());
+    this.props.myDataSource.setFilter(filterCond);
+} else {
+    console.log("Association is not filterable");
+}
+```
+
+Similarly, to apply a condition so that only the objects that are associated with at least the first two objects that are brought by the selectable objects datasource will be returned:
```
tamelbrun: This sentence is very complicated to read. I would suggest `that are brought by` -> `from` and maybe also `so that only ... will be returned` -> `to match only ...`. (The latter would also apply to the previous sentence on line 258.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -260,10 +295,10 @@ this.props.myDataSource.setFilter(undefined);
...
 #### 2.4.2 Advanced Filtering {#advanced-filtering}
 
-In some use cases it is necessary to apply more complex filtering conditions. For example if a use case requires fetching only items where `myAttributeString` starts with `"B"` and `myAttributeBoolean` is set to `true`, or items where `myAttributeNumber` is greater than `10`. In order to construct such condition special filter builders `and` and `or` have to be used. The following code sample shows how to use them. Note that check for `filterable` flags omitted for simplicity. Real widgets should always take `filterable` flag into account.
+In some use cases it is necessary to apply more complex filtering conditions. For example if a use case requires fetching only items where `myAttributeString` starts with `"B"` and `myAttributeBoolean` is set to `true`, or items where `myAttributeNumber` is greater than `10` and objects that are associated with other objects. In order to construct such condition special filter builders `and` and `or` have to be used. The following code sample shows how to use them. Note that check for `filterable` flags omitted for simplicity. Real widgets should always take `filterable` flag into account.
```
tamelbrun: Given that the other props are referred to by name, let's do the same for the association property: "...and `myAssociationReference` is associated with another object" or something to that effect.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
 
 Optional `universe` field represents an array of possible values of an attribute. See `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value) for more information.
 
-### 3.3 ListWidgetValue {#listwidgetvalue}
+### 3.3 ListReferenceValue and ListReferenceSetValue {#listassociationvalue}
+
+`ListReferenceValue` and `ListReferenceSetValue` are both used to represent an [association property](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) that is linked to a data source.
+This allows the client component to access associated values of individual items from a `ListValue`. `ListReferenceValue` and `ListReferenceSetValue` are both objects and their definition are as follows:
```
tamelbrun: definition -> definitions

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+
+```ts
+export type ListReferenceValue = ListAssociationValue<ObjectItem> & { type: "Reference" };
+
+export type ListReferenceSetValue = ListAssociationValue<ObjectItem[]> & { type: "ReferenceSet" };
+
+export interface ListAssociationValue<T extends ObjectItem | ObjectItem[]> {
+  get: (item: ObjectItem) => DynamicValue<T>; // NOTE: DynamicValue obtained from ListAssociationValue always readonly
```
tamelbrun: The `NOTE:` is unnecessary as `DynamicValue` - unlike `EditableValue` - is _always_ read-only.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+  id: ListAssociationId;
+  filterable: boolean;
+}
+```
+
+#### 3.3.1 Obtaining Association Value
+
+{{% alert color="warning" %}}
+Due to a technical limitation it is not yet possible to edit associations obtained via `ListReferenceValue` or `ListReferenceSetValue`. `DynamicValue`s returned by `ListReferenceValue` or `ListReferenceSetValue` are always **readonly**.
+{{% /alert %}}
```
tamelbrun: This warning is not necessary, for the same reason: `DynamicValue` is _always_ read-only.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+
+In order to work with the associated object of a particular item of a `ListValue` first an instance of `DynamicValue<T>` should be obtained by calling `ListReferenceValue.get` or `LostReferenceSetValue.get` with the item. The type `<T>` depends on the allowed value types as configured for the association property.
```
tamelbrun: 1. Lost -> List
2. The references to type `<T>` are inaccurate. In the case of associations, the generated type is either `ListReferenceValue` (which always produces `DynamicValue<ObjectItem>`), `ListReferenceSetValue` (which always produces `DynamicValue<ObjectItem[]>`) or `ListReferenceValue | ListReferenceSetValue` (which must be discriminated first using the `type` property).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+```
+
+The following code sample shows how to get a `DynamicValue<ObjectItem>` that represents a read-only value of an associated object of the first element from the `myDataSource`.
+
+```ts
+const associationValue = this.props.myAssociationOnDatasource.get(this.props.myDataSource.items[0]);
+```
+
+This will return an `ObjectItem` representing the associated object, since in this example, the widget is configured to allow only singular associations.
+If you want to access the individual attributes of this associated item, you can use an attribute property. Refer to [Obtaining Attribute Value section](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#obtaining-attribute-value) for details on how to do this.
```
tamelbrun: We should add "...attribute property _linked to the selectable objects data source and pass the associated item to it_" (or something to that effect).

LEGIO-SEXTA-FERRATA:
Hello @ConnorLand, yes, this is the documentation additions and updates for a new feature (Linked Association Properties) to be released with 9.17.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+
+#### 3.3.2 Association ID and Filterable Flags {#listassociationvalue-id-filterable}
+
+`id` field of type `ListAssociationId` represents the unique randomly generated string identifier of an association. That identifier could be used when applying filtering on a linked data source property to identify which association should be used for filtering. Check [Filtering](#listvalue-filtering) sections for more information.
+
+`filterable` field specifies if the association could be used for filtering. This flag has to be checked before a widget applies filtering on a data source property. An attempt to filter on a non-filterable association would lead to an error during the execution time.
+
+#### 3.3.3 Association Type
+
+[Association](/apidocs-mxsdk/apidocs/pluggable-widgets-property-types/#association) property determines which association types could be configured for that property. For example a property may allow associations of type `Reference` and not `ReferenceSet`.
```
tamelbrun: determines -> defines
may allow -> may be configured to allow

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -448,7 +484,75 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+
+It is possible to determine type of association by checking `type` field on a property. The following code sample shows how to check the association type on the property named `myAssociationOnDatasource`:
+
+```ts
+if (this.props.myAssociationOnDatasource.type === "Reference") {
+    console.log("Reference association");
+} else if (this.props.myAssociationOnDatasource.type === "ReferenceSet") {
```
tamelbrun: An `else if` is not a good example, as it will trigger an ESLint warning. (After checking `type === "Reference"`, the type of the `type` property is narrowed to `"ReferenceSet"` and the second check is flagged as always true. We can do three things: convert it to a switch, convert it to two separate ifs, or simply remove the second check (which might be confusing at first glance but compile-accurate).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
```diff
@@ -520,26 +520,32 @@ Then the Studio Pro UI for the property appears like this:
...
+If a `dataSource` attribute is not specified, the client will receive a `ReferenceValue` for references (singular references) and/or `ReferenceSetValue` for reference sets (multiple references) or a union of them, if the widget is configured to allow both. For more information, see the [ModifiableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#modifiable-value) section of *Client APIs Available to Pluggable Widgets*.
+
+When a `dataSource` attribute is specified and configured by the user, it is passed as a [`ListReferenceValue` or `ListReferenceSetValue`](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#listassociationvalue) depending on the configuration of the property. For more information, see the [Datasource](#datasource) section below.
 
 {{% alert color="info" %}}
 This property type was introduced in Mendix [9.13](/releasenotes/studio-pro/9.13/).
+And the ability to link association properties to a data source was introduced in Mendix [9.17](/releasenotes/studio-pro/9.17/).
```
tamelbrun: I would put this in a separate alert box.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
```diff
@@ -520,26 +520,32 @@ Then the Studio Pro UI for the property appears like this:
...
+| Attribute           | Required | Attribute Type | Description                                                                                                         |
+|---------------------| -------- | -------------- |---------------------------------------------------------------------------------------------------------------------|
+| `type`              | Yes      | String         | Must be `association`                                                                                               |
+| `key`               | Yes      | String         | See [key](#key)                                                                                                     |
+| `onChange`          | No       | Property Path  | The path to an Action property that will be executed by the Mendix Platform when the value is changed by the widget |
+| `required`          | No       | Boolean        | Decides if the property must be specified by the user, `true` by default                                            |
+| `selectableObjects` | Yes      | Property Path  | The path to a Datasource property that will provide selectable objects for the association                          |
```
tamelbrun: 🎄 Let's also change this "Datasource property" into a link, while we're at it :).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
```diff
@@ -643,7 +649,7 @@ Then the Studio Pro UI for the property appears like this:
 
 ### 4.8 Datasource {#datasource}
 
-The datasource property allows widgets to work with object lists. The client component will receive value prop of type [`ListValue`](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#listvalue) and may be used with [`action`](#action), [`attribute`](#attribute), [`expression`](#expression), [`text template`](#texttemplate) and [`widgets`](#widgets) properties. See [Data Sources](/refguide/data-sources/#list-widgets) for available data source types.
+The datasource property allows widgets to work with object lists. The client component will receive value prop of type [`ListValue`](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis-list-values/#listvalue) and may be used with [`action`](#action), [`attribute`](#attribute), [`association`](#association), [`expression`](#expression), [`text template`](#texttemplate) and [`widgets`](#widgets) properties. See [Data Sources](/refguide/data-sources/#list-widgets) for available data source types.
```
tamelbrun: As previously suggested, let's add the "The ability to link an association property..." alert box here as well.

tamelbrun (changes requested):
First review pass complete :)

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Do you think I should include that information as well? I think this is just a paragraph that includes some examples, not an exhaustive list of all functions. Therefore; `Filtering based on associations is also possible. For example, you can use equals with references and contains with reference sets.` would be my attempt at fixing this. What do you think?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Complete removal or just the DynamicValue part?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: > If you want to access the individual attribute **values** of this associated ~item~ **object**, you can use an > attribute property linked to the selectable objects data source and pass the associated ~item~ **object** to it using its get function.

What do you think?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Should I fix those for the Attribute part as well?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: I guess the same for Attribute part, as well?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: Mentioning a few examples instead of everything would indeed be a good alternative. My main issue was that the part about `contains` being the only function available for reference sets was not just an example, it was a lie ;).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: Ah, it was adapted from there? Yeah, let's take this opportunity to make that better as well 🎄

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: Complete removal, the entire warning makes no sense here. It was only necessary for attributes because there you were getting an `EditableValue` that was not actually ever editable. _That_ required a warning :).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: I think the reference to the `get()` function is a bit overkill, that's why we're referring to the `[Obtaining Attribute Value section]` for the details. I don't have a strong preference between `object` and `item`,  but the previous sentence uses associated _object_, so that would make it more consistent.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: Attribute properties don't have a `selectableObjects` attribute in the XML, so no ;). Unless you meant the `dataSource` XML attribute, in which case: by all means! (But then we should do the other linkable properties as well, not just attribute.)

LEGIO-SEXTA-FERRATA: Yeah, I meant data sources in general and yeah, it would be even nicer if we did it for all properties not just Attribute. I'll proceed with that.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Partially done. I added contains. More on using empty() in the example later.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done for both parts.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Partially done. I fixed the typo. More on `<T>` later.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done for both parts.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Done and I **also** made a small adjustment for the description of selectableObjects.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: The reference (no pun intended) on line 278 is still wrong :).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: :+1:

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: While we're at it, let's fix the grammar of those last two sentences (both here and in their original location):
_Note that checks for the `filterable` flag have been omitted for simplicity. Real widgets should always take the `filterable` flag into account._

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: :+1:

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: :+1:

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: :+1:

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: Now that it has its own alert box, remove the "And" :)

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: I noticed an existing typo in the description for `dataSource` for the `widgets` property:
```Specifies the path to a [`datasource`](#datasource) property linked to this widget's property```
That apostrophe is wrong, it should be _this widgets property_ (plural). Let's clean that up as well :). (Maybe we should put the `widgets` in code formatting to prevent it from accidentally getting "corrected" again.)

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 13 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: How about, simply `console.log(this.props.myAssociationOnDatasource.type)`?

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 3 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: Done. 🤦

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 4 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: We have an example with empty, in the one immediately above this. Do you mean that I should also include an empty check in this specific example? (This example demonstrates reference sets and it uses contains. Isn't an empty check redundant for this example?)

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 6 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: Done. I also fixed some other `filterable` mentions that lacked the article 'the'. But this section is under ListValue. It does not have association and attribute counterparts. Therefore I am not sure what you mean by 'original location'?

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 14 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: Done. 🤦‍♂️

Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: I noticed this 'and' when I first looked at your comment and pushed without fixing it. 🤦‍♂️

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 15 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: Oh yeah... This is an awesome finding. :) Fixed it.

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 16 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: The same version annotation you mean?

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 4 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: I was considering it, especially because I'm worried about the discoverability of `equals/notEquals(empty())` for reference sets if it is not mentioned anywhere in the examples. At the same time, I think it would make these simple examples too complicated.

Let's add an `equals(empty())` filter for reference sets to the Advanced Filtering section instead, it already has a similar filter for references.

(And remove `empty` from the imports here, then. :))

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 6 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: I think that at first glance the references for `filterable` looked more similar to me than they actually were :P. But yeah, my intent was to fix the articles everywhere 👍🏻.

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 13 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: That would completely defeat the purpose of showing the discriminated union (and subsequent type narrowing) at work.

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 14 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: 👍🏻

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 15 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: Did you also put the code formatting around it? I don't see it in the current diff.

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md (hunk 16 above)
Comment (tamelbrun) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md]: Yes, the one that no longer starts with And ;)

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 10 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: I pushed a "slightly more liberal" new version for describing this.

File: content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md (hunk 13 above)
Comment (LEGIO-SEXTA-FERRATA) [reply content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md]: Then, I am inclined towards simply removing the second check and perhaps indicating the obvious (narrowing down of type) in a comment ~perhaps~? (Below is an example.) What do you say?

```ts
if (this.props.myAssociationOnDatasource.type === "Reference") {
    console.log("Reference association");
} else {
    // TypeScript will narrow it down to "ReferenceSet" when the type property is NOT "Reference"
    console.log("ReferenceSet association");
}
```

switch and two separate ifs seems sub-optimal to me.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Fixed: I was inclined to put it right after the heading at first (similar to Attribute Property) but there is another annotation before the end of the section and I placed our addition next to that.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Oh, sorry. When I first read your comment, I thought by `code formatting` you meant adding it to a spell check dictionary or something. 🤦‍♂️ (Hinted by automatic correction.) I added the code formatting now. (Assuming I understood correctly this time: You just meant to make it like this: `code formatting`.)

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: I would avoid the word "property" (ambiguous in this context), simply "type" will suffice. But otherwise, I think this is indeed the most elegant solution.

(And maybe the ALL CAPS on NOT is a bit overkill. :))

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: In general, I think it would better to place it right after the paragraph that mentions the version-specific feature (`association` in this case), so that people won't accidentally overlook this. On the other hand, this is a relatively small section. @ConnorLand, what do you think?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: That is `indeed` what I `meant` :).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: The current sentence has become a long-winding sequence of commas and parentheses. I would propose the following simplified version:

> ```In order to work with an object or objects that are associated with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item. If the association property has been configured to allow both types of associations, the type of the property is defined as `ListReferenceValue | ListReferenceSetValue` and a check on its `type` should be done to narrow down the type.```

We could also link to the *Association Type* section, if we give it an anchor.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: I just spotted this sentence:
_"It is possible to determine type of association by checking `type` field on a property."_
-> _"It is possible to determine the type of association by checking the `type` field on an association property."_

(Edit: maybe we should also add something like "This is useful if the property has been configured to allow both references and reference sets.")

tamelbrun: And I would add a comma after "For example" here:
_"For example a property may allow associations of type `Reference` and not `ReferenceSet`."_

LEGIO-SEXTA-FERRATA: > I would avoid the word "property" (ambiguous in this context), simply "type" will suffice. But otherwise, I think this is indeed the most elegant solution.
> 
> (And maybe the ALL CAPS on NOT is a bit overkill. :))

My new proposal: TypeScript will narrow it down to "ReferenceSet" when the type ~property~ _field_ is not _equal to_ "Reference"

LEGIO-SEXTA-FERRATA: > I just spotted this sentence: _"It is possible to determine type of association by checking `type` field on a property."_ -> _"It is possible to determine the type of association by checking the `type` field on an association property."_
> 
> (Edit: maybe we should also add something like "This is useful if the property has been configured to allow both references and reference sets.")

**Summary of Proposals:** Some small additions and also doing the same thing for the same part in Attribute Property.

**Proposal 1:** 
"For example a property may be configured to allow attributes of type `String` and `Integer` in order to present progress."
->
"For example, _an attribute_ property may be configured to allow attributes of type `String` and `Integer` in order to present progress."

**Proposal 2:**
"It is possible to determine type of attribute by checking `type` field on a property."
->
"It is possible to determine _the_ type of attribute by checking _the_ `type` field on _an attribute_ property."

**Proposal 3:**
"For example a property may allow associations of type `Reference` and not `ReferenceSet`."
->
"For example, _an association_ property may _be configured to_ allow associations of type `Reference` and not `ReferenceSet`."

**Proposal 4:**
"It is possible to determine type of association by checking `type` field on a property."
->
"It is possible to determine _the_ type of association by checking _the_ `type` field on _an association_ property. This is useful if the property has been configured to allow both references and reference sets."

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
ConnorLand: I think putting it right after the paragraph that mentions the feature is a good practice. Sometimes it leads to duplicated alert boxes, but as long as we keep the alert boxes short it is ok (and a best practice).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: We can do it. Which word or words should have the anchor, or should I add something like "See Association Type for more information."?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Would you like me to relocate the existing alert box to the top of the section as well? (Right after the heading.)

ConnorLand: Yes, I think all alert boxes that just pertain generically to a property (like "This prop was introduced in version X" or "the ability to do x in y was introduced in version Y") should go right after the heading at the top of the section.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Resolved this conversation.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Resolved this conversation.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Resolved this conversation.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Resolved this conversation.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
ConnorLand: Resolved this conversation.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: > My new proposal: TypeScript will narrow it down to "ReferenceSet" when the type ~property~ _field_ is not _equal to_ "Reference"

"Field" is not really JS/TS terminology. Why not drop it altogether? "Type" by itself is perfectly clear in this context.

tamelbrun: Proposals 1-4 seem fine to me.
On a side note: you need to be very careful, it looks like you're force-pushing away commits added by Connor. Did you use force-with-lease?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: I meant that the _Association Type_ section currently does not have an anchor/hashtag that we can link to. I think a separate sentence is more clear; if I recall correctly, we do this in other places as well.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: The existing alert box is not about a version, actually:

> Only list datasources are supported, therefore specifying `isList="true"` is required.

To me, this alert box is more about a platform limitation, not generically applying to all properties. But, having one alert box at the top of the section and one below seems sub-optimal to me.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Ah... that's what I was afraid of... :(

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
ConnorLand: Right, so please just put one (perhaps at the top).

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: **I propose:** In some use cases it is necessary to apply more complex filtering conditions. For example if a use case requires fetching only items where `myAttributeString` starts with `"B"` and `myAttributeBoolean` is set to `true`, or items where ~`myAttributeNumber` is greater than `10` and~ `myAssociationReference` is associated with another object _while `myAssociationReferenceSet` is not associated with other objects; meaning objects that has at least one singular association but no multiple associations._ In order to construct such a condition, special filter builders `and` and `or` have to be used. The following code sample shows how to use them. Note that checks for the `filterable` flag have been omitted for simplicity. Real widgets should always take the `filterable` flag into account.

```ts
import { attribute, association, literal, startsWith, equals, notEquals, greaterThan, and, or } from "mendix/filters/builders";

// in the widget code
if (/* check that all properties are filterable */) {
    const filterCond = or(
        and(
            startsWith(attribute(this.props.myAttributeString.id), literal("B")),
            equals(attribute(this.props.myAttributeBoolean.id), literal(true))
        ),
        and(
            // REMOVING THIS: greaterThan(attribute(this.props.myAttributeNumber.id), literal(10)),
            notEquals(association(this.props.myAssociationReference.id), empty()),
            // ADDING THIS:
            equals(association(this.props.myAssociationReferenceSet.id), empty())
        )
    );
    this.props.myDataSource.setFilter(filterCond);
} else {
    console.log("Some attribute is not filterable");
}
```

LEGIO-SEXTA-FERRATA: A complex filter made of two attribute components, two association components? That's why I removed the greaterThan filter but we can carry it over to the attribute group as well?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: We're dealing with two different types of alert boxes here: one is a warning about a limitation of data source properties in general, the other is information specifically about the version in which `association` was added (which would go right after the paragraph which mentions `association`). Since the limitation is enforced by the XML validation and therefore impossible to overlook, I think its existing location at the bottom of the section is preferable.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: I don't see a reason to remove anything from the filter conditions, it's an advanced filter after all. And I think that _"; meaning objects... etc."_ only introduces clutter and confusion, I would remove it.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: All done. (@ConnorLand, sorry again; I promise to be more careful next time.)

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
LEGIO-SEXTA-FERRATA: No reason programmatically but we are describing the filter in the paragraph above it. The sentence becomes weirder to me as we add stuff to it. But I can easily roll that back, of course.

LEGIO-SEXTA-FERRATA: Done.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: The info box about `association` is now _before_ the paragraph that mentions it, which feels really counter-intuitive to me.

LEGIO-SEXTA-FERRATA: You meant, placing the box after the paragraph that mentions linking with other attributes.
If so, it should be done now.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: Alright, I think I can live with it 👍🏻.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
tamelbrun: 👍🏻

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
tamelbrun: 👍🏻

ConnorLand:
Is this commit of mine correct? It's the one selectable object's data source right? https://github.com/mendix/docs/pull/4945/commits/f9c1bfbddd69d0086499b359edc9ca3ad798a713

tamelbrun:
> Is this commit of mine correct? It's the one selectable object's data source right? [f9c1bfb](https://github.com/mendix/docs/commit/f9c1bfbddd69d0086499b359edc9ca3ad798a713)

The plural was correct: the selectable objects data source is defined using the `selectableObjects` attribute of the association property.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -444,14 +483,78 @@ if (this.props.myAttributeOnDatasource.type === "String") {
 
 #### 3.2.4 Formatter and Universe
 
-`formatter` field represents the default formatter that is going to be used on values obtained by `get` function.
+The `formatter` field represents the default formatter used on values obtained by the `get` function.
+
+The optional `universe` field represents an attribute's possible array values. For more information, see the `universe` field of [EditableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#editable-value).
```
tamelbrun: The previous version, "an array of possible values of an attribute", was correct. An attribute does not have array values :). One small improvement on this would be "an array of possible values **for** an attribute".

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -444,14 +483,78 @@ if (this.props.myAttributeOnDatasource.type === "String") {
...
+  get: (item: ObjectItem) => DynamicValue<T>;
+
+  id: ListAssociationId;
+  filterable: boolean;
+}
+```
+
+#### 3.3.1 Obtaining Association Values
+
+In order to work with an object or objects with a particular item returned by `ListValue`, first an instance of `DynamicValue<ObjectItem>` (for `ListReferenceValue`) or `DynamicValue<ObjectItem[]>` (for `ListReferenceSetValue`) should be obtained by calling `get` with the item.
```
tamelbrun: The removed phrase "that are associated with" is necessary for this statement to be correct, as the association property deals with associations between objects.

tamelbrun (changes requested):

ConnorLand (approved):

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Done

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
ConnorLand: Corrected

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
```diff
@@ -269,7 +269,7 @@ if (this.props.myAssociationReference.filterable) {
 }
 ```
 
-Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable object's data source:
+Similarly, code like this can apply a condition to match only the objects that are associated with at least the first two objects from the selectable objects' data source:
```
LEGIO-SEXTA-FERRATA: I think this term is more correct as "selectable objects data source" or even simply `selectableObjects`. Technically, this "selectable objects data source" is defined in the XML, just like any other `datasource` (meaning there is no separate property `type` like `selectableObjectsDatasource`, just `datasource`) and while, it is correct in a sense that this `datasource` belongs to the `selectableObjects`; in the rest of our documentation where this is being referred to, we just use the nominal form (not possessive) "selectable objects data source". And I think the `datasource` itself is technically owned by the widget, not `selectableObjects` attribute that it is linked to.

Also of note, the difficulty regarding one of our most frequently used terms: `datasource`. Codebase-wise it is currently being referred to as `datasource` (depending on where in code it is involved with, either as `datasource` or as `dataSource`)  but actually as a technical concept in Mendix platform, in English, we refer to it as data source. (It should be written as two words.)

@alihcsumer @mlaponder @tamelbrun, would you agree?

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
```diff
@@ -520,7 +520,7 @@ Then the Studio Pro UI for the property appears like this:
 
 The association property type allows a widget to work directly with both reading and writing associations between entities. Depending on the widget's purposes, a widget should define association types it supports.
 
-If a `dataSource` attribute is not specified, the client will receive a `ReferenceValue` for references (singular references) and/or `ReferenceSetValue` for reference sets (multiple references) or a union of them, if the widget is configured to allow both. For more information, see the [ModifiableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#modifiable-value) section of *Client APIs Available to Pluggable Widgets*.
+If a `dataSource` attribute is not specified the client will receive a `ReferenceValue` for references (singular references), possibly a `ReferenceSetValue` for reference sets (multiple references), or a union of them (if the widget is configured to allow both). For more information, see the [ModifiableValue](/apidocs-mxsdk/apidocs/pluggable-widgets-client-apis/#modifiable-value) section of *Client APIs Available to Pluggable Widgets*.
```
LEGIO-SEXTA-FERRATA: I think this should be just or:

> The client will receive, `ReferenceValue` ... or `ReferenceSetValue` ... or a union of them.

That should be the idea and the correct statement. Because the supplied value (depending on how the widget is defined in its XML definition file) is exclusive among those three options. It cannot be `ReferenceValue` and `ReferenceSetValue`, it will be either one of those or a (`TypeScript`) union like this: `ReferenceValue | ReferenceSetValue`

And also, it is not actually client that is receiving the prop (and therefore the mentioned values) but the **client component** of a pluggable widget which is essentially a React component. (The client reads the XML and handles instantiation of the properties per widget and supplies them to the widgets as React props.) We can adjust this, too, I guess, while we are at it, if you guys also agree.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/pluggable-widgets-client-apis-list-values.md
mlaponder: @ConnorLand I agree with @legio-vi-ferrata here, I don't think the apostrophe is needed here. imo if the apostrophe is dropped we are good to go here.

content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-property-types.md
mlaponder: @ConnorLand I agree also with @legio-vi-ferrata that just a `or` would be better. I think we should go with that suggestion.

ConnorLand: I've changed the language per the suggestion

PR #4963, opened 2022-08-23
Use SSOLandingPage

Luyao-Zhang-1:
Update saml.md


elliotclements-mendix (commented):
The documentation looks good for Mendix v9. My only doubt is if the old documentation is still valid for Mendix versions lower than 9. e.g. for lower than v 9 does the index.html page have the document.cookie logic or would they need to manually add the old redirect logic:  
`<meta http-equiv="refresh" content="0;URL=/SSO/" />`

content/en/docs/appstore/modules/saml.md
```diff
@@ -92,9 +92,15 @@ There are different versions of the module, depending on which version of Mendix
...
+You can use SSO to automatically sign users in to your app by redirecting every user accessing `index.html` to the Mendix `/SSO/` endpoint as follows:
 
-**SSOLandingPage** specifies a different landing page so the end-user does not end up on `index.html` again after a login attempt.  We recommend that you change this constant to `/index3.html` and create an `index3.html` page in your `/theme` folder and copy contents of the original `index.html` (without the added redirect) into it. The authenticated end-user will then land on `index3.html` which will then display the content of the app. If the user authentication fails, the user will be directed to the **DefaultLoginPage** instead. 
+If you use Mendix version lower than 9, change the `index.html` file in the `/theme` folder by adding `<meta http-equiv="refresh" content="0;URL=/SSO/" />` to the file.
```
feddovanede: This is not limited to Mendix 8 and lower: you can use this in Mendix 9 as well.

content/en/docs/appstore/modules/saml.md
```diff
@@ -92,9 +92,15 @@ There are different versions of the module, depending on which version of Mendix
...
+
+If you use Mendix version 9 and higher, there is no `index.html` file, so you need to create one first. To do so, run your app in Studio Pro, and then copy the `index.html` file from the `/deployment/web/` folder to the `/theme/web` folder. Mendix then uses this version as the template for `index.html`. Then edit the `index.html` file by changing the URL within the `originURI` to `\SSO\`, for example: `document.cookie = "originURI=/SSO/" + (window.location.protocol === "https:" ? ";SameSite=None;Secure" : "");`. This cookie determines where the Mendix Client will redirect users to when they need to sign in. If you are already signed in, you are not redirected again.
```
feddovanede: This is not only applicable for Mendix 9: you can use this for Mendix 8 and lower as well.

So, basically, you can use both methods in all Mendix versions. Neither of the methods is "better" than the other, both have some pros and cons.

The main reason for using this method over redirecting using a _meta_ tag, is that it allows you to redirect users that have not yet signed in, automatically to `/SSO/` when opening `index.html`, while still allowing users to open `login.html` directly and sign in using a local account, bypassing SSO. You don't always want to allow this.

feddovanede: It says `\SSO'\`, this should be `/SSO/`

feddovanede (commented):
Both methods can be used in all Mendix versions.

content/en/docs/appstore/modules/saml.md
elliotclements-mendix: The point I want to clear up in the documentation is that a dev should either use the originURI method OR via
`<meta http-equiv="refresh" content="0;URL=/SSO/" />` method. 
The guide did not work when I followed it because by default the index.html page contains the originURI redirect. Therefore when I appending the _meta_ tag redirect (as suggested in the documentation) it created issues as it was doing two redirects. 

Would you agree @feddovanede ? I want to ensure this documentation is clear because as a layman following the previous guide word for word it didn't work.

feddovanede: Yes, it makes sense to only use on the methods, not both.

content/en/docs/appstore/modules/saml.md
```diff
@@ -92,9 +92,13 @@ There are different versions of the module, depending on which version of Mendix
...
+If you want to redirect users who have not yet signed in automatically to `/SSO/` when opening `index.html`, while still allowing users to open `login.html` directly and sign in using a local account, bypassing single sign on,  you should edit the `index.html` file by changing the URL within the `originURI` to `/SSO/`, for example: `document.cookie = "originURI=/SSO/" + (window.location.protocol === "https:" ? ";SameSite=None;Secure" : "");`. This cookie determines to which location the Mendix Client will redirect users when they need to sign in. If you have already signed in, you are not redirected again.
+
+If you do this without any other changes, the app will come back to `index.html` which will be redirected again to single sign on.
```
feddovanede: I do not fully understand this line. What does _this_ refer to? The first method? Or the second?

feddovanede (commented):
Left a comment

content/en/docs/appstore/modules/saml.md
Luyao-Zhang-1: Hi @feddovanede thanks for your review. 
The line `If you do this without any other changes, the app will come back to `index.html` which will be redirected again to single sign on.` is from the [original doc](https://docs.mendix.com/appstore/modules/saml/#ssolandingpage). So I assume `this` here should refer to both methods? Could you advise?

feddovanede: @Luyao-Zhang-1 The method with changing the cookie value will only redirect users if they are not signed in yet. If users go through the whole SAML flow and are signed in to the app, they will not be redirected again from _index.html_ if only the cookie value is set.
Only if you use the meta redirect method, you will _always_ be redirected if you go to _index.html_, regardless of whether you are signed in or not.

Luyao-Zhang-1:
Thank you @feddovanede and @elliotclements-mendix for your input and clarification. I will go head and merge the PR.

PR #4973, opened 2022-08-24
Update pluggable-widgets-studio-apis.md

m3kh66:
This MR will replace the `className` preview property with `class`.

Release version: 9.18


ConnorLand (approved):

ConnorLand:
Hi @MehdiGolchin . Looks like a fairly straightforward PR. 

Did you search all docs for other possible instances which might need changes? If not, could you quickly scan them using VSC? I'll do the same.

m3kh66:
Hi @ConnorLand,
Thanks for the review. I've quickly looked at the docs and couldn't find anything except in version 8, which is expected.

An unrelated question: is it better to mention that we've deprecated the `className`, or just replacing that is enough?

PR #4974, opened 2022-08-25
Draft Management Release changes. My Drafts Page added.

rajesh-rakshe-mendix:
Release is planned on 27th Aug Saturday 9AM - 1PM CEST. after that this needs to be released


Adam-Dupaski:
This will be merged on Monday

PR #4975, opened 2022-08-25
Shorten Long File Paths

MarkvanMents:
Long file paths can cause problems with Git if checkout is not to a very short base path
This PR shortens the longest filepath on the author's local build:

- from 253 characters for markdown files
- from 248 characters for static attachments (images)


MarkvanMents:
Discussed with Nataša. Merging.

PR #4976, opened 2022-08-26
File document in REST operation

joostverhoog:
Clarified what it means to return a FileDocument from a published REST operation. We received feedback that many users use a file document with a mapping. That does not give streaming behavior, nor does it give a good client experience (and is not what the REST standard intends). So now it's the first option discussed.

Also addressed a rendering issue in an alert.


johannahemminger:
@joostverhoog thanks for this!

PR #4981, opened 2022-08-29
Update NBUI release notes

diego-antonelli:


diego-antonelli:
@ConnorLand it was not released yet but I've opened already the PR

ConnorLand (approved):

PR #4988, opened 2022-08-30
Docsy feedback widget

NatasaKralj:
This PR turns on the default Docsy feedback widget. The widget captures Yes/No answers to the question "Was this page helpful?" by triggering a Google Analytics event.

To run the PR locally, use the following command (widget only shows up in production environment):
` .\hugo server --environment production --gc --disableFastRender --poll 700ms`
Everything after `production` is optional, but recommended.

Steps taken:
- Added custom feedback partial for styling of widget section
- Moved already present styling for download button to scss file, now download and Yes/No use the same styling
- Added needed parameters to _default/config.toml
    - @MarkvanMents the parameter doesn't work outside of production environment, but should I move the config parameters to the production config?
- Added hide_feedback to pages that are only lists of child links and RNs

### DIFF
--- a/assets/scss/_content.scss
+++ b/assets/scss/_content.scss
@@ -103,3 +103,10 @@
         font-size: 3rem;
     }
 }
+
+.btn-primary {
+    color: #fff;
+    background-color: #007bff;
+    border-color: $primary;
+    border-radius: 0.25rem;
+}
--- a/config/_default/config.toml
+++ b/config/_default/config.toml
@@ -31,7 +31,7 @@ disableKinds = ["taxonomy","taxonomyTerm","term"]
 
 [services]
     [services.googleAnalytics]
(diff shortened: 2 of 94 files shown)


MarkvanMents:
Hi @NatasaKralj 
Looks good. However,

1. I don't think we should be excluding it from so many pages - I think it is useful even on the landing pages.
2. The link to Github should prefill the issue with the same information as clicking the feedback button at the top of the page.
3. The link to Github should open in a new tab.
4. Would be nice if it wasn't all the way at the end of the page.
    * Could put it at the top (next to the date). It would then have to be smaller - perhaps just a thumbs-up/thumbs-down but I think it would be more obvious even so.
    * If this snippet redesign isn't possible, perhaps we should add a heading so that people can go to the feedback at the end of the page using the ToC.

PR #4989, opened 2022-08-30
Clarify microflow list arguments for grid buttons

r0b1n:
Clarify the behavior for ticket 155164.


Adam-Dupaski (approved):

Adam-Dupaski:
@Yiyun333 this is an example PR for logic/microflows

Yiyun333:
> 

Thanks! I wonder whether changing _single-selection_ to _single selction_ will create inconsistent style with _multi-selection_ in this paragraph?

Adam-Dupaski:
@Yiyun333 good question but "multi" is a prefix and "single" is an adjective, so I see them functioning differently here

Yiyun333:
> @Yiyun333 good question but "multi" is a prefix and "single" is an adjective, so I see them functioning differently here

I see! Thanks for the clarification.

PR #4995, opened 2022-09-01
Document Multiple User Session Best Practice (Solves TW-1709)

ConnorLand:
This documentation update should help customers design their apps to end sessions as intended.

https://mendix.atlassian.net/browse/TW-1709


MarkvanMents (approved):
Good comprehensive description of the issue and solution. Thanks.
I might have tried to shorten it a bit, but yours is nice and conversational and easy to read - so that is just a question of personal preference rather than anything you need to change.
Thanks for picking this up and doing it so promptly.
Can you post something to the two Slack threads on the Jira issue once this is merged, so Support can point customers to the documentation if they get this in the future? Thanks.

ConnorLand:
> Good comprehensive description of the issue and solution. Thanks. I might have tried to shorten it a bit, but yours is nice and conversational and easy to read - so that is just a question of personal preference rather than anything you need to change. Thanks for picking this up and doing it so promptly. Can you post something to the two Slack threads on the Jira issue once this is merged, so Support can point customers to the documentation if they get this in the future? Thanks.

Will do.

PR #5003, opened 2022-09-07
Describe database-specific limitations on unlimited string uniqueness

dahfjkg:
Was not sure whether to describe the limitation here or in `case-sensitive-database-behavior`

### DIFF
--- a/content/en/docs/refguide/runtime/data-storage/case-sensitive-database-behavior.md
+++ b/content/en/docs/refguide/runtime/data-storage/case-sensitive-database-behavior.md
@@ -79,10 +79,10 @@ This table presents the default case sensitivity by different database types:
 |------------------:|:--------------:|:-----------:|:--------------------:|
 | HSQLDB            | I              | I           | I                    |
 | POSTGRESQL        | S              | S           | I                    |
-| DB2               | S              | S¹          | I³                    |
+| DB2               | S              | S¹          | I³                   |
 | MARIADB & MYSQL   | C              | C           | C                    |
 | ORACLE            | C¹             | C           | I                    |
-| SAP HANA          | S¹             | S¹          | I²                    |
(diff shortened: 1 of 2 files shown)


MarkvanMents:
Hi Pavel (@dahfjkg),
I think it would be better in this area of the documentation: https://docs.mendix.com/refguide/data-storage/. There is already a link to https://docs.mendix.com/refguide/case-sensitive-database-behavior/, so I think it would be better to have this information there. But then, I think we need to rename the document to something like **Database String Behavior**.

What do you think?
@katarzyna-koltun-mx : do you have any thoughts on this?

In the Oracle documentation, we also mention uniqueness: https://docs.mendix.com/refguide/oracle/#34-uniqueness-constraint - but this seems to be a stricter limitation than the one we are talking about here, so I think it needs to stay specific to the Oracle description.

katarzyna-koltun-mx:
@MarkvanMents That makes sense, but also, we have individual articles on different database types (e.g. https://docs.mendix.com/refguide/mysql/) - maybe this info should be distributed across these articles? I'm thinking a user would typically only need to see the behaviour for their own database type.

But then also the other contents of the article you linked would need to be moved to database-specific articles, so I'm not sure there. If we want to keep it all on one page then the article you linked looks great.

dahfjkg:
@MarkvanMents @katarzyna-koltun-mx I agree. There are more database-specific differences in unlimited string behavior that ability to create a uniqueness constraint. I will create a separate page in data storage and call it `Unlimited String Behavior`

dahfjkg:
@MarkvanMents @katarzyna-koltun-mx I've updated the PR and created a separate page with unlimited string behavior per database. Please check the format of the page

katarzyna-koltun-mx:
@dahfjkg Hi Pavel, I've merged your requested change. It should be published tomorrow.

PR #5016, opened 2022-09-08
Update Mx4PC Grafana instructions (no release date, can be published at any moment)

zlogic:
These changes are not associated with any release and can be reviewed/published at any convenient time.

* Grafana/Loki have updated their Helm chart and the current instructions no longer work (even when using version `2.5.1` which worked before), updated the Helm chart parameters to match Grafana's latest version
* A few small adjustments to the OpenShift SecurityContextConstraints - to reduce permissions granted to Loki/Promtail
* Updated the Grafana dashboards to be compatible with the latest version of Grafana/Loki
* Removed a warning that is no longer relevant


MarkvanMents (approved):
These all look like technical changes.
Nothing I can add. Will merge immediately.

PR #5026, opened 2022-09-09
Hide feedback widget on old content

NatasaKralj:
Added `hide_feedback: true` to cascade through all older content spaces


MarkvanMents (approved):
Looks good.
Interesting that we have an old_content attribute - we should look if we do anything with that (perhaps in Algolia?)

NatasaKralj:
@MarkvanMents `old_content` is a custom label we use to control styling of older versions (graying out in menu). It might indeed be useful in Algolia.

PR #5032, opened 2022-09-13
Update quickstart-part1.md

wiwengweng:
fix the mistakes


wiwengweng:
finally understand why we create app with a mobile template here.

PR #5066, opened 2022-09-19
Update language example

Yiyun333:
@Luyao-Zhang-1 and @Adam-Dupaski I have updated the language example to Chinese using the latest interface. @Luyao-Zhang-1 Would you mind double-checking the Chinese example for me? Thanks!


Luyao-Zhang-1:
@Yiyun333 Looks good to me. Nice work!

PR #5070, opened 2022-09-20
improve description of Save Changes action

basi01:


MariaShaposhnikova:
@basi01 , thank you for your input! I verified and reviewed changes and adjusted the use case to a more general one, information on committing objects is true not only for data views but other widgets as well.
Could I ask you to sign the CLA please? You only need to sign it once. 
Cheers,
Maria

basi01:
> Could I ask you to sign the CLA please

done

PR #5085, opened 2022-09-22
Improve OIDC Documentation

MarkvanMents:


PhetoJob (commented):
Looks good to me, cannot comment on if there are other changes in the paper that need to be applied.

MarkvanMents:
Hi Job (@PhetoJob )
This should have everything in it from the Paper document (as at 23/9/2022).
Let me know if you spot anything I missed.

MarkvanMents:
Agreed with Jaap to publish.
Will review later if we get customer feedback.

PR #5113, opened 2022-09-29
Private Cloud Operator Release Notes

MarkvanMents:


content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
```diff
@@ -13,6 +13,16 @@ For information on the current status of deployment to Mendix for Private Cloud
 
 ## 2022
 
+### September 29th, 2022
+
+#### Mendix Operator v2.8.0{#2.8.0}
+
+* We have now reduced the security permissions required by Mx4PC pods. This means that pods do not need elevated permissions on the Kubernetes cluster.
```
MarkvanMents: I rewrote this as it wasn't entirely clear. Does it still make sense?

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
```diff
@@ -13,6 +13,16 @@ For information on the current status of deployment to Mendix for Private Cloud
...
+* We have limited the Mendix app emptyDir size to 4 MiB.
```
MarkvanMents: Original was 4 Mi. I assume this is 4 MiB, please let me know if I'm wrong.
Will customers know what the emptyDir size is, and why it matters?

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
```diff
@@ -13,6 +13,16 @@ For information on the current status of deployment to Mendix for Private Cloud
...
 ## 2022
 
+### September 29th, 2022
+
+#### Mendix Operator v2.8.0{#2.8.0}
+
+* We have now reduced the security permissions required by Mx4PC pods. This means that pods do not need elevated permissions on the Kubernetes cluster.
+* We have limited the Mendix app emptyDir size to 4 MiB. 
+* To improve the security score of all Mendix for Private Cloud images, we upgraded the dependencies to address CVEs.
+    * This update allows us to address CVE-2022-1996, CVE-2020-27191 and CVE-2022-29526. These CVEs do not affect previously released versions of Mendix for Private Cloud.
```
MarkvanMents: Is this because we don't/didn't actually use any of the code in the dependencies?
If so, perhaps we should say so?

MarkvanMents (commented):
@Nidhi251289 
I've made a few comments. Can you let me know if this is OK to publish when you are ready.
Thanks
Mark

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
Nidhi251289: Yes. Looks fine now

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
Nidhi251289: we can write it as :

We have limited the mendix app emptyDir size to 4 MiB, in order to limit the consumption of memory by Mx4PC pods

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
MarkvanMents: Agreed over Slack.

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
MarkvanMents: Updated.

content/en/docs/releasenotes/deployment/mendix-for-private-cloud.md
Nidhi251289: meaning is .. the cves which arises basically if there are any updates made to any external dependencies

and thus we have to fix those CVE

MarkvanMents: But why does this not affect previous versions? They had different dependencies?

Nidhi251289: yes.. as i mentioned that the dependencies (which are external) in previous version are updated (we dont have control on them) and because of which CVEs arises

MarkvanMents: Sorry, I'm being stupid here.
Suppose I have Operator v2.5.1 which I downloaded in May and is running on an air-gapped cluster.
You are changing the dependencies in Operator 2.8.0 to address some CVEs.
How does this affect my Operator v2.5.1 - it can't see the dependencies you have updated, so if it was affected by a CVE in May, it must be affected now?

MarkvanMents: Or do the dependencies of v2.5.1, 2.6.0, 2.7.0 all predate these CVEs, which were introduced later. In which case, we have never released anything containing those CVEs, haven't actually changed anything for the customer, and we don't need a release note.

MarkvanMents: Removed

PR #5132, opened 2022-10-04
Change most of "drag and drop" instances to "drag"

Yiyun333:
…s have not been reviewed)


content/en/docs/appstore/modules/data-widgets/data-grid-2.md
```diff
@@ -8,7 +8,7 @@ tags: ["marketplace", "marketplace component", "data grid", "platform support"]
...
+[Data Grid 2](https://marketplace.mendix.com/link/component/116540) is the successor to the standard data grid widget for displaying content in a tabular form. It comes with many powerful new features and settings like support for widgets, row and cell coloring, responsive layout, accessibility, and different paging options like virtual scrolling. It offers personalization support so that end-users can show, hide, and re-order columns. Personalizations can be persisted in the database for flexibility and control. The data grid is packaged as a module and uses a flexible approach for filtering. Developers can drag data-grid-specific widgets inside the header of the grid and tailor the behavior of the filters. This module includes filters for text, numbers, dates, and the option to add drop-down filters for single or multiple selections.
```
Luyao-Zhang-1:
```suggestion
[Data Grid 2](https://marketplace.mendix.com/link/component/116540) is the successor to the standard data grid widget for displaying content in a tabular form. It comes with many powerful new features and settings like support for widgets, row and cell coloring, responsive layout, accessibility, and different paging options like virtual scrolling. It offers personalization support so that end-users can show, hide, and re-order columns. Personalizations can be persisted in the database for flexibility and control. The data grid is packaged as a module and uses a flexible approach for filtering. Developers can drag data-grid-specific widgets into the header of the grid and tailor the behavior of the filters. This module includes filters for text, numbers, dates, and the option to add drop-down filters for single or multiple selections.
```

content/en/docs/appstore/widgets/accessibility-helper.md
```diff
@@ -22,9 +22,9 @@ The [Accessibility Helper](https://marketplace.mendix.com/link/component/114803/
 
 The widget acts like a container and alters the elements inside it. In order to change elements, HTML attributes act in ways described in the sections below.
 
-### 2.1 Drag and Drop Widgets Inside Accessibility Helper
+### 2.1 Drag Widgets Inside Accessibility Helper
 
-Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag and drop the widgets they want to change inside Accessibility Helper.
+Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag the widgets they want to change inside Accessibility Helper.
```
Luyao-Zhang-1:
```suggestion
Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag the widgets they want to change into Accessibility Helper.
```

content/en/docs/appstore/widgets/accessibility-helper.md
```diff
@@ -22,9 +22,9 @@ The [Accessibility Helper](https://marketplace.mendix.com/link/component/114803/
 
 The widget acts like a container and alters the elements inside it. In order to change elements, HTML attributes act in ways described in the sections below.
 
-### 2.1 Drag and Drop Widgets Inside Accessibility Helper
+### 2.1 Drag Widgets Inside Accessibility Helper
```
Luyao-Zhang-1, line 25: Dragging Widgets into Accessibility Helper

Luyao-Zhang-1:
```suggestion
### 2.1 Dragging Widgets into Accessibility Helper
```

content/en/docs/appstore/widgets/accessibility-helper.md
```diff
@@ -22,9 +22,9 @@ The [Accessibility Helper](https://marketplace.mendix.com/link/component/114803/
...
 
-Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag and drop the widgets they want to change inside Accessibility Helper.
+Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag the widgets they want to change inside Accessibility Helper.
 
 ### 2.2 Specify the Target Selector
```
Luyao-Zhang-1:
```suggestion
### 2.2 Specifying the Target Selector
```

content/en/docs/appstore/modules/data-widgets/data-grid-2.md
Luyao-Zhang-1:
```suggestion
[Data Grid 2](https://marketplace.mendix.com/link/component/116540) is the successor to the standard data grid widget for displaying content in a tabular form. It comes with many powerful new features and settings like support for widgets, row and cell coloring, responsive layout, accessibility, and different paging options like virtual scrolling. It offers personalization support so that end-users can show, hide, and re-order columns. Personalizations can be persisted in the database for flexibility and control. The data grid is packaged as a module and uses a flexible approach for filtering. Developers can drag data-grid-specific widgets into the header of the grid and tailor the behavior of the filters. This module includes filters for text, numbers, dates, and the option to add drop-down filters for single or multiple selections.
```

Luyao-Zhang-1:
@Yiyun333 I made a few suggestions. Please let me know if you think differently.
I also changed three verbs in the headings in the "Usage" section of the "Accessibility Helper" doc from the infinitive form to their gerund form – sorry that this is not related to the "drag and drop" issue. (I thought it is too much to open a new PR to change only three words.)

content/en/docs/data-hub/share-data/_index.md
```diff
@@ -38,7 +38,7 @@ Follow these steps to create a simple app in Mendix Studio Pro.
 
 1. In Studio Pro, click **New App** to create a new app using the **Blank** app template. Fill in the **App Name** as *{yourname}CustomerServiceApp* and select **Create app**. 
 2. In the App Explorer, double-click the **Domain Model** in **MyFirstModule**. 
-3. In the **Toolbox**, select **Entity** and drag and drop it to the domain model.
+3. In the **Toolbox**, select **Entity** and drag it to the domain model.
```
Adam-Dupaski: Would use "into" instead of just "to"

content/en/docs/howto/logic-business-rules/add-action-to-workflow-toolbox.md
```diff
@@ -44,7 +44,7 @@ To create the calling microflow and add it as the custom action to the workflow
 
 1. Create a new microflow that calls **ACT_SAP_INVENTORY_LAPTOP_CHECK**. Name the new microflow **WFS_MY_SYSTEM_TASK**.
 2. Make sure that the input parameters of the **WFS_MY_SYSTEM_TASK** microflow are mapped to the workflow definition. For the **Employee_Onboarding** workflow, add the EmployeeOnboarding parameter because the **EmployeeOnboarding** entity is set as the **Workflow entity**. 
-3. In **Toolbox**, select **Microflow call** and drag and drop it to the flow:
+3. In **Toolbox**, select **Microflow call** and drag it to the flow:
```
Adam-Dupaski, line 47: Same here - "to" is not clear/strong enough, maybe even too close to "towards" or "around"

content/en/docs/howto/logic-business-rules/add-action-to-workflow-toolbox.md
```diff
@@ -80,7 +80,7 @@ Now that you have exposed the microflow as the custom workflow action, you can u
 
     {{< figure src="/attachments/howto/logic-business-rules/add-action-to-workflow-toolbox/workflow-toolbox.png" alt="Workflow Toolbox" >}}
 
-3. Drag and drop the action to the workflow.
+3. Drag the action to the workflow.
```
Adam-Dupaski: Update "to" throughout

Yiyun333:
> @Yiyun333 I made a few suggestions. Please let me know if you think differently. I also changed three verbs in the headings in the "Usage" section of the "Accessibility Helper" doc from the infinitive form to their gerund form – sorry that this is not related to the "drag and drop" issue. (I thought it is too much to open a new PR to change only three words.)

@Luyao-Zhang-1 Haha, no problem! I've accepted your commit suggestions. Thank you very much for checking the docs.

content/en/docs/howto/logic-business-rules/add-action-to-workflow-toolbox.md
Yiyun333: No problem. Will have a closer look at these insances tomorrow.

content/en/docs/studio-how-to/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p1.md
```diff
@@ -68,8 +68,8 @@ The attribute is created.
 To configure the decision with the attribute or parameter of the enumeration type, follow these steps:
 
 1. [Create a new microflow](/studio/microflows/) and name it, for example, *Show_grade_specific_page*.
-2. In the **Toolbox** tab, select **Decision**, drag and drop it to the microflow.
-3. You need to pass a parameter to configure the decision correctly.  In the **Toolbox**, select **Parameter** and drag and drop it to the microflow.
+2. In the **Toolbox** tab, select **Decision**, drag it onto the microflow.
```
MariaShaposhnikova: I would use "to" or "into", but most certainly not "onto".

content/en/docs/studio-how-to/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p1.md
```diff
@@ -68,8 +68,8 @@ The attribute is created.
...
+3. You need to pass a parameter to configure the decision correctly.  In the **Toolbox**, select **Parameter** and drag it onto the microflow.
```
MariaShaposhnikova: Same here

content/en/docs/studio-how-to/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p1.md
```diff
@@ -159,8 +159,8 @@ The attribute for the **Customer** entity is created.
 To configure the decision with the attribute of the Boolean type, follow these steps:
 
 1. [Create a new microflow](/studio/microflows/) and name it, for example *Customer_status_check*.
-2. In the **Toolbox** tab, select the decision, drag and drop it to the microflow. 
-3. You need to pass a parameter to configure the decision. In the **Toolbox** tab, select **Parameter**, and drag and drop it to the microflow.
+2. In the **Toolbox** tab, select the decision, drag it onto the microflow.
```
MariaShaposhnikova: Same here

content/en/docs/studio-how-to/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p1.md
```diff
@@ -159,8 +159,8 @@ The attribute for the **Customer** entity is created.
 To configure the decision with the attribute of the Boolean type, follow these steps:
 
 1. [Create a new microflow](/studio/microflows/) and name it, for example *Customer_status_check*.
-2. In the **Toolbox** tab, select the decision, drag and drop it to the microflow. 
-3. You need to pass a parameter to configure the decision. In the **Toolbox** tab, select **Parameter**, and drag and drop it to the microflow.
+2. In the **Toolbox** tab, select the decision, drag it onto the microflow. 
+3. You need to pass a parameter to configure the decision. In the **Toolbox** tab, select **Parameter**, and drag it onto the microflow.
```
MariaShaposhnikova: "into" or "to"

content/en/docs/studio-how-to/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p2.md
```diff
@@ -58,9 +58,9 @@ To embed the microflow with a decision (the attribute of the enumeration type) t
 
     The page with customer details is generated.
 
-7. Return to the **Customer_details** page, and in **Toolbox**>**Widgets** >**Data Containers**, find **List View**, drag and drop it to the page.
+7. Return to the **Customer_details** page, and in **Toolbox**>**Widgets** >**Data Containers**, find **List View**, drag it onto the page.
```
MariaShaposhnikova, line 61: I would use "on"

content/en/docs/studio-how-to8/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p1.md
```diff
@@ -65,8 +65,8 @@ The attribute is created.
 To configure the decision with the attribute or parameter of the enumeration type, follow these steps:
 
 1. [Create a new microflow](/studio8/microflows/) and name it, for example, *Show_grade_specific_page*.
-2. In the **Toolbox** tab, select **Decision**, drag and drop it to the microflow.
-3. You need to pass a parameter to configure the decision correctly.  In the **Toolbox**, select **Parameter** and drag and drop it to the microflow.
+2. In the **Toolbox** tab, select **Decision**, drag it onto the microflow.
```
MariaShaposhnikova: For every instance "onto + microflow", it does not seem to be correct. We do not mean "drag it on top of the microflow", we mean "in the editor". "Into" would be the best choice for microflows.

content/en/docs/studio-how-to8/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p2.md
```diff
@@ -34,14 +34,14 @@ To embed the microflow with a decision (the attribute of the enumeration type) t
 
     {{< figure src="/attachments/studio-how-to8/microflows/microflows-how-to-configure-decision/microflows-how-to-configure-decision-p2/data-view.png" >}}
 
-3. Drag and drop **Data View** to the page.
+3. Drag **Data View** onto the page.
```
MariaShaposhnikova, line 37: With all instances of "page", I would use "on". "Onto" reads "on top of" for me.

content/en/docs/studio-how-to8/domain-model-how-to-configure/_index.md
```diff
@@ -63,7 +63,7 @@ To add the product to your domain model, follow the steps below:
 1. Create the **Product** entity. Do the following:
 
     1. Open your [domain model](/studio8/domain-models/).
-    2. Open **Toolbox**, drag and drop **Entity** in your domain model:
+    2. Open **Toolbox**, drag **Entity** into your domain model:
```
MariaShaposhnikova, line 66: Why are we changing instance in Mendix 8? I don't think it's needed.

MariaShaposhnikova:
I see there are a lot variations of using "into", "in", "to", "on", and "onto", here is what I found in the Microsoft Style Guide:
In the example with "drag" they use "to", where we are using "into" in this PR (Drag the Filename file **to** the Foldername folder.), although there are no explicit guidelines for that. https://learn.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui 
There is a definition for "onto" that says: "to indicate moving something to a position on top of something else". https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/onto
CC: @Adam-Dupaski

MariaShaposhnikova:
My suggestion would be:

- Use "into" to emphasize that something is inside something else, for example, "drag a list view into the data view."
- Use "to" or "on" for page(s).
- Use "to" for all other editors, such as microflows and workflows.

Adam-Dupaski:
Nice find @MariaShaposhnikova . I don't find "to" to be illustrative or strong enough as a preposition. I advised against "to" already in earlier comments. I'd personally go for "into" or "onto" for almost all cases:

- "into" - inside something else, like your example, "drag a list view into the data view"; also for editors, since you edit an element in an editor, "drag a microflow into the editor"
- "onto" - for something sitting on top of another element, like a page, "drag a widget onto the page"

MariaShaposhnikova:
@Adam-Dupaski , thanks for the suggestions! Could you please explain for me what your suggestions are based on? I agree with "into", but for me "onto" reads "on top of". We don't drag elements on top of a page, they are inside it, in the layout, but "into a page" does not sound grammaticaly correct. I would suggest to go with "drag on a page". Analyzing Google searches gives me:

- "drag on a page" - 287.000 results
- "drag onto a page" - 1.400 results

content/en/docs/appstore/widgets/accessibility-helper.md
Luyao-Zhang-1:
```suggestion
Since Accessibility Helper is a containment widget it will try to change the widgets inside it. Thus the user must drag the widgets they want to change into Accessibility Helper.
```

Yiyun333:
Thanks for the insights @MariaShaposhnikova @Adam-Dupaski 
According to the Merriam-Webster dictionary, the first definition of "onto" is: **to a position on**. 
"To" here specifies a moving direction, which, I think, suits nicely with the verb "drag". 
So for instance, we can understand "drag a widget onto a page" as "drag a widget to a postion on a page". I think it is more illustrative than "Drag a widget on a page". "Drag a widget on a page" to me sounds like the widget is already on the page and you are dragging it on the page to adjust its position on the page (also makes sense: but this is not what we mean in the docs).

As for "to the microflow/nanoflow", I agree with @Adam-Dupaski that "to" is not illustrative enough becasue it only specifies a moving direction but not the spatial relation between the object and its ending location (e.g., on/in). 
Personally I prefer to use "onto" for microflows and nanoflows. I understand them as the logic (the flow) that is presented in an microflow/nanoflow editor, and visually we can see the flow/the line in the editor. So visually speaking, I find using "drag an activity onto a microflow" is easier to understand than "drag an activity into a microflow". It's only a flow/a line, which does not have an inside area (except for a sub-microflow lol).

"Drag an activity into a microflow editor" is good, but I don't think we always mean an editor when we use "microflow". "Microflow" to me means the logic/the flow of actions. 

Maybe I am too sensitive with these distinctions, haha, but I'd love to hear everyone else's opinion. It might be just personal preferences, but these differences can make a difference in meaning, I think.

Adam-Dupaski:
> @Adam-Dupaski , thanks for the suggestions! Could you please explain for me what your suggestions are based on? I agree with "into", but for me "onto" reads "on top of". We don't drag elements on top of a page, they are inside it, in the layout, but "into a page" does not sound grammaticaly correct. I would suggest to go with "drag on a page". Analyzing Google searches gives me:
> 
>     * "drag on a page" - 287.000 results
> 
>     * "drag onto a page" - 1.400 results

"Drag on a page" is fine but I don't think "onto" is incorrect or awkward, as "on" is still there actively, and "onto" works better than "to."

Adam-Dupaski:
> "Drag a widget on a page" to me sounds like the widget is already on the page and you are dragging it on the page to adjust its position on the page (also makes sense: but this is not what we mean in the docs).

This is a good example and clear explanation. I support "onto." Google numbers are illustrative but not semantically substantial for this kind of decision. I think discussion of the preposition choice with examples is a good way to make a final decent decision.

MarkvanMents:
> As for "to the microflow/nanoflow", I agree with @Adam-Dupaski that "to" is not illustrative enough becasue it only specifies a moving direction but not the spatial relation between the object and its ending location (e.g., on/in).
Personally I prefer to use "onto" for microflows and nanoflows. I understand them as the logic (the flow) that is presented in an microflow/nanoflow editor, and visually we can see the flow/the line in the editor. So visually speaking, I find using "drag an activity onto a microflow" is easier to understand than "drag an activity into a microflow". It's only a flow/a line, which does not have an inside area (except for a sub-microflow lol).

In the microflow editor you can have a free-floating action which you can position anywhere, on top of another action or on top of an existing flow line.
However, it is not then part of the microflow, it is unconnected and you will get an error in the error pane saying that it needs to have a flow going into/out of it.
What you need to do is drag something so that it goes into the flow - the existing flow line splits and reconnects itself to be a flow into and out of the new action you have dragged. It is not enough to have the action within the editor pane.

Therefore, I think that **into** is preferable to **onto**.

This is also the case if you consider _flow_ in normal English use. I would merge into the flow of traffic on a freeway, not onto it.

I agree with Adam that **onto** is better than **on** if you are dragging from the toolbox onto a page. This is because **drag** normally goes with **to** (I drag the chair to the table, for example), However, if you are moving widgets about within a page you _could_ also use **on** because the widget is already on the page - but perhaps consistency becomes an issue here and it would be better to use the same preposition for both.

Yiyun333:
@MarkvanMents After your explanation, I agree that "into" conceptually works better with microflows than "onto". I will adopt the preposition "into" (instead of "onto") for the microflow/nanoflow/workflow instances. 

I will use "onto" for "drag a widget onto a page" for the cases where we mean the "widget" is NOT already on the page.

PR #5139, opened 2022-10-05
Deep Link Continuation Issue

Luyao-Zhang-1:


ThijsSeppen:
Looks good! 

The only thing (which is not necessarily something that should be fixed in this MR) is that I've tried to look for some definitions/docs of a custom login page (which could be nicely linked here) a few times in the last months. Perhaps it's nice to add that some day as well?

Luyao-Zhang-1:
> Looks good!
> 
> The only thing (which is not necessarily something that should be fixed in this MR) is that I've tried to look for some definitions/docs of a custom login page (which could be nicely linked here) a few times in the last months. Perhaps it's nice to add that some day as well?

@ThijsSeppen 
Thanks. We can definitely link some information about custom login page here. Do you have suggestions on what should be linked here?

ThijsSeppen:
@Luyao-Zhang-1, Unfortunately, no I don't. 
With my previous comment, I meant that I failed to locate such docs in recent months. Perhaps there is no specific document (yet)?

PR #5183, opened 2022-10-17
Generic button shortcode

NatasaKralj:
This PR makes different kinds of buttons available to use for links within content.

When building locally, you can view the different kinds available on the indentation-spacing-test page.

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -32,7 +32,7 @@ $light: #D3F3EE !default;
 
 $dark: #0A1325 !default; // MX Midnight 100 (MxDock color)
 $blue: #0CABF9 !default; // MX Blue
-$orange: #BA5A31 !default;
+$orange: #E39600 !default; // MX colors.light.brand.yellow.7 (from PDS Figma colors board)
 $gray-light: #F3F3F4 !default; // NK - added for design proposal
 $gray-100: #F8F8F8 !default; // NK - changed to better align to Figma Docs mockup
 $gray-200: #E7E7E9 !default; // NK - changed to better align to Figma Docs mockup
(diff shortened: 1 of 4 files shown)


layouts/shortcodes/button.html
```diff
@@ -0,0 +1,5 @@
+{{ $color := .Get "color" | default "info" }}
+{{ $link := .Get "href" }}
+{{ $text := .Get "text" | default "Download" }}
+{{ $title := .Get "title" }}
+<a class="btn btn-{{ $color }}" href="{{ $link }}" title="{{ $title }}" target="_blank">{{ $text }}</a>
```
MarkvanMents, line 6: The only issue here is if someone puts in a color which is not one of the standard ones.
For example "color="#FF00FF"
However, the same issue holds for alerts, and it doesn't crash, so let's assume that people will check that it works, and we can link to a list of the valid ones (or list our preferred ones) from our documentation.

MarkvanMents (approved):
See comment - perhaps need to make sure we document it well?

layouts/shortcodes/button.html
NatasaKralj: I agree, it has to be documented well. I'll add a section in the Use the Docs Site doc to cover all options before merging.
I don't think we should run strict tests on what content is submitted in a button shortcode. As you mentioned, the same is possible with alerts. It should be on the writer to double check if it looks the way they want. Bad links will still fail in htmltest.
As far as default colors and styling go, we can always adjust those later.

MarkvanMents: Yes - good plan. That's why I approved it as well as making the comment.

assets/scss/_variables_project.scss
```diff
@@ -46,6 +46,24 @@ $gray-800: #3B4251 !default; // MX Midnight 80
 $gray-900: #0A1325 !default; // MX Midnight 100 
 $black: #000 !default;
 
+// NK - copied from themes\docsy\assets\vendor\bootstrap\scss\_variables.scss to add extra theme color,
```
NatasaKralj: @MarkvanMents I added theme colors into variables, as a way to override the list of default theme colors. This lets us add any color keyword we want.  
This way I just define a new theme color, give it a color code > Docsy/Bootstrap does the rest. It calculates all the color gradients for any action connected to the button. 
This way is cleaner than copying lots of internal Bootstrap styling for buttons, just to define a new button. There are lots of operations done (hover, click, disable effects), that apply color changes based on the initial theme color chosen.

MarkvanMents: Not sure I understand how this works. You'll have to explain it sometime. But it seems OK.

MarkvanMents (approved):
Looks OK - I'd quite like to go through what you achieve in creating $theme-colors, but you don't have to wait before merging.

The only concern I have is with **outline-light**. The text is so light it is difficult to see.
image

Is there any scope to change the text color to black (while we are in light mode - we can look at the impact of dark mode some other time 😀).

assets/scss/_variables_project.scss
NatasaKralj: I ended up removing the theme color expansion. It's nice to know how to do this, if we ever need to, but we don't have much of a use case for it ATM.

MarkvanMents: Great research. I'm sure it will come in useful in doing improvements in other areas of the site.

Issue #5152, opened 2022-10-10
implement-mendix-best-practices-for-development


CoenPijpers1707:
Page link: https://docs.mendix.com/howto/general/dev-best-practices/

Document link: [dev-best-practices.md](https://github.com/mendix/docs/blob/development/content/en/docs/howto/general/dev-best-practices.md)

## My Issue/Suggestion
At the complexity section, provide an example of doing it wrong and doing it right:
The text: Nested IF statements in a single microflow expression are not recommended. If multiple checks depend on one another, this should be represented by multiple decisions in the microflow, so that the complexity is not hidden away in the expressions. You can use AND and OR operators to produce complex expressions if necessary.

Yiyun333:
@CoenPijpers1707 Thank you very much for your suggestion! Examples are now added for this paragraph in the [Best Practices doc]( https://docs.mendix.com/howto/general/dev-best-practices/) (see section 4.2.4).

PR #5196, opened 2022-10-20
Add examples for best practices for section 4.2.4 microflow complexity

Yiyun333:
This is based on GitHub issue #5152


reinouts (commented):
- The caption of the decision should probably reflect that it checks both if `$Parameter`  is `true` and whether `$DeviceType` equals `Phone`.
- To make the difference between the two examples clearer, the picture should probably show both microflows in which the decisions appear and the expressions inside the decisions.

Yiyun333:
> * The caption of the decision should probably reflect that it checks both if `$Parameter`  is `true` and whether `$DeviceType` equals `Phone`.
> * To make the difference between the two examples clearer, the picture should probably show both microflows in which the decisions appear and the expressions inside the decisions.
@reinouts Thank you very much for the comments. I just updated the doc accordingly. Please let me know if there are more things to improve.

reinouts:
Still has to be reviewed by the doc team but from a technical/theory point of view I think this is fine.

Yiyun333:
> Still has to be reviewed by the doc team but from a technical/theory point of view I think this is fine.

@reinouts Great! Thanks a lot!

content/en/docs/howto/general/dev-best-practices.md
```diff
@@ -335,7 +335,11 @@ If you decide to color code the different activities in your app, be sure to ali
...
 
-Nested `IF` statements in a single microflow expression are not recommended. If multiple checks depend on one another, this should be represented by multiple decisions in the microflow, so that the complexity is not hidden away in the expressions. You can use `AND` and `OR` operators to produce complex expressions if necessary.
+Nested `IF` statements in a single microflow expression are not recommended. If multiple checks depend on one another, this should be represented by multiple decisions in the microflow, so that the complexity is not hidden away in the expressions. You can use `AND` and `OR` operators to produce complex expressions if necessary. Examples of what is not recommended and what is recommended are shown below. If necessary, the single microflow expression in the following example can also be written as `if ($currentDeviceType = System.DeviceType.Phone and $Parameter = true) then true else false` using the `AND` operator. However, we still recommend the low-code approach shown below, which makes it easier to understand what is really happening.
+
+{{< figure src="/attachments/howto/general/dev-best-practices/not-recommended-microflow.png" width="450px">}}
```
MariaShaposhnikova: It's not the best practice to place two images one under the other. I would make it explicit in the text. I would also suggest to swap the order which they are presented and start from the recommended approach. 
For example:
In the example below, there is a low-code approach which makes it easier to understand what is happening in the microflow:
Image
The example below shows a possible but not recommended approach: 
Image
I would also remove exclamation marks, a cross mark, and a check mark from the images to tone it down a bit + these icons are not standardized.

content/en/docs/howto/general/dev-best-practices.md
```diff
@@ -335,7 +335,11 @@ If you decide to color code the different activities in your app, be sure to ali
 
 #### 4.2.4 Complexity
 
-Nested `IF` statements in a single microflow expression are not recommended. If multiple checks depend on one another, this should be represented by multiple decisions in the microflow, so that the complexity is not hidden away in the expressions. You can use `AND` and `OR` operators to produce complex expressions if necessary.
+Nested `IF` statements in a single microflow expression are not recommended. If multiple checks depend on one another, this should be represented by multiple decisions in the microflow, so that the complexity is not hidden away in the expressions. You can use `AND` and `OR` operators to produce complex expressions if necessary. Examples of what is not recommended and what is recommended are shown below. If necessary, the single microflow expression in the following example can also be written as `if ($currentDeviceType = System.DeviceType.Phone and $Parameter = true) then true else false` using the `AND` operator. However, we still recommend the low-code approach shown below, which makes it easier to understand what is really happening.
```
MariaShaposhnikova: I would start describing the examples with a new paragraph.

Yiyun333: @MariaShaposhnikova Many thanks for the review and advice. I just updated the doc based on your suggestions. Please let me know if you have any further advice on the updated version.

content/en/docs/howto/general/dev-best-practices.md
```diff
@@ -337,11 +337,11 @@ If you decide to color code the different activities in your app, be sure to ali
...
 
-The example below shows a low-code approach we recommend because it presents a clear picture of what is happening in the microflow:
+The example below shows a low-code approach that we recommend because it presents a clear picture of what is happening in the microflow:
 
 {{< figure src="/attachments/howto/general/dev-best-practices/recommended-microflow.png" width="700px">}}
 
-The example below shows an approach we do not recommend. We can rewrite the microflow expression in this example as `if ($currentDeviceType = System.DeviceType.Phone and $Parameter = true) then true else false` using the `AND` operator. However, it is still not clear enough and the low-code approach shown in the above example is better.
+The example below shows an approach that we do not recommend. We can rewrite the microflow expression in this example as `if ($currentDeviceType = System.DeviceType.Phone and $Parameter = true) then true else false` using the `AND` operator. However, it is still not clear enough and the low-code approach shown in the above example is better.
```
MariaShaposhnikova:
```suggestion
The example below shows an approach that we do not recommend. You can rewrite the microflow expression in this example as `if ($currentDeviceType = System.DeviceType.Phone and $Parameter = true) then true else false` using the `AND` operator. However, it is still not clear enough and the low-code approach shown in the above example is preferable.
```

MariaShaposhnikova:
@Yiyun333 , thanks for implementing all comments! I added one suggestion, the rest looks good!

Yiyun333:
> @Yiyun333 , thanks for implementing all comments! I added one suggestion, the rest looks good!

@MariaShaposhnikova Thanks for the review!

PR #5203, opened 2022-10-21
Clarify Use Docker with Minikube documentation

MarkvanMents:
The current documentation gives an example of using Docker with Minikube. This is intended as a proof of concept explanation rather than a full-blown explanation of using Mendix on Kubernetes. However it appears above Mendix for Private Cloud, the fully supported offering for deploying to Kubernetes, in Google searches such as _Mendix Kubernetes_.

This PR highlights the use case for deploying on Minikube and clarifies the limitations compared to Mendix for Private Cloud.

**Publishing this will not immediately update Google search - we should force Google to re-index this manually so that we can review the impact on search results.**


zlogic:
Looks good to me 👍

PR #5217, opened 2022-10-25
Drop support for Postgres 10 and MS SQL 2017

dahfjkg:
To be merged after both 9.20.0 and 9.6.15 are released (probably 9.6.15 will be released later than 9.20)


Adam-Dupaski:
To be deprecated in RN for 9.20.0, 9.18.2, 9.12.9, 9.6.15

PR #5219, opened 2022-10-25
Dynamo doc

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@trongbtran I've created a draft based on your document. I would like to expand on the Configuration section - will schedule a meeting to discuss.

Issue #5184, opened 2022-10-17
audit-trail


OlavJansen:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [audit-trail](https://docs.mendix.com/appstore/modules/audit-trail/)

Document link: [audit-trail.md](https://github.com/mendix/docs/blob/development/content/en/docs/appstore/modules/audit-trail.md)

## My Issue/Suggestion

_2.1 Associations and Events
For this configuration option, add a reference set association from your entity to the Log entity or an association from the Log entity to the entity that needs to be audited. Add an after-create, before-commit, and before-delete event to your entity that is identical to the events on the AudittrailSuperClass. The Java actions will automatically create the log item and all the required log lines based on the changes._

After create is not needed and also not used on the AudittrailSuperClass entity

Luyao-Zhang-1:
Hi @OlavJansen ,

Thank you for your feedback. I have removed the after-create event from the [paragraph](https://docs.mendix.com/appstore/modules/audit-trail/#21-associations-and-events).  

Please continue to let us know in the future if you see documentation that can be improved.
Thank you!

PR #5226, opened 2022-10-26
Remove "After-create" from the Audit Trail doc

Luyao-Zhang-1:


Luyao-Zhang-1:
For issue[#5184](https://github.com/mendix/docs/issues/5184)

PR #5235, opened 2022-10-27
Development

MichielArts:
Reference to microflow settings added to microflow properties section, because i spent a long time finding the info.


Yiyun333:
@MichielArts Thank you very much for adding this to our doc. It is a nice improvement. I only made some changes to the language according to our style. The rest looks good!

PR #5283, opened 2022-11-09
Kk s3

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@CasperSpronk27 Hey Casper, here's the updated S3 document, can you take a look and let me know if anything is incorrect?

PR #5284, opened 2022-11-10
Bump Hugo to 105 - no changes to Docsy or npm

MarkvanMents:


NatasaKralj:
This PR is not needed, as the Docsy update to 0.6.0 supersedes it.

PR #5296, opened 2022-11-11
A note about mapping user roles

katarzyna-koltun-mx:


content/en/docs/appstore/general/app-store-content.md
```diff
@@ -217,6 +217,8 @@ To update the module in your app to a newer version, follow these steps::
 
 Each module that you download from the Marketplace is different. Some modules can cause errors because they are connected to other modules. For example, after you download the [Database Replication](/appstore/modules/database-replication/) module, you also need to download the [Mx Model Reflection](/appstore/modules/model-reflection/) module to accomplish the configuration. If you run into any problems, always check the **Documentation** tab on the page of this module in the Marketplace for installation guidelines and details on any dependencies.
 
+Some modules you can download may come with its own [user roles](/studio/settings-security/), and you should keep them in mind to ensure the security of your application. If the Security level of your app is set to **Prototype/demo** or **Production**, you need to map the module roles to the applicable user roles in your app.
```
Adam-Dupaski: "Some" > "their own" (instead of "its own")

"and you should keep them" > I'd recommend "which you should keep"

"Security level" > "**Security** level" (verified against UI?)

Is it clear what it means to "keep in mind here," or is that too abstract and there's a better way of writing that?

katarzyna-koltun-mx: Done.

PR #5297, opened 2022-11-11
Update aws-authentication.md

brfreekmx:
Updated the text on Session and Static credentials to explain that Session credentials are the preferred method but if for the given reasons it is more suitable to use Static Credentials it is okay to use this method.


katarzyna-koltun-mx:
@brfreekmx Thank you! I've also added 2 more changes in the same vein. Merging now. When you have a moment, please sign the CLA (https://cla-assistant.io/mendix/docs?pullRequest=5297) - this should prevent the above license error.

PR #5339, opened 2022-11-23
Link to doc explaining trial mode restriction.

emirmx:
It may be good idea to link to the document explaining what are  the trial mode restrictions.


MarkvanMents (approved):
Thanks for the update, Emir. Making a PR makes our life much easier!
Approving and publishing.

PR #5342, opened 2022-11-24
Moving AWS IoT to community supported

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Per agreeement with Maarten de Vries, documentation for this connector will be moved to the marketplace and removed from docs.

PR #5344, opened 2022-11-24
Adding AWS in the Strategic Partners guide

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Reviewed by Maarten during a call.

PR #5355, opened 2022-11-28
Change GB to GiB for Mendix Cloud packages

MarkvanMents:


MarkvanMents:
Waiting for alignment with cloud UX before publishing.
Will be contacted when it needs to be published.

MarkvanMents:
Merging independently of any changes in the UX as these will take longer.

PR #5360, opened 2022-11-29
Add buttons to widget docs

ConnorLand:
I'm adding buttons at the top of widget docs which take users to the component's Marketplace page. The button pops a bit more than a link, and can let our users navigate to Marketplace faster.

If I have any shortcode issues, I'll work them out with @NatasaKralj when she's back from sick leave.


MariaShaposhnikova:
@ConnorLand , is this PR still relevant?

PR #5381, opened 2022-12-02
Update database-connector.md

brfreekmx:
Updated the download link to the postgresl driver download. The .html suffix shouldn't there.


johannahemminger:
Thanks very much @brfreekmx !

PR #5387, opened 2022-12-05
Add Swagger page types to Algolia index

MarkvanMents:
The new API page was not being indexed for Algolia.
This is because it has a different page type ("swagger") so this needed to be added to the Hugo template for building the Algolia index.


NatasaKralj (approved):
Looks good, searching the algolia json file after a local production build confirms the new swagger pages are included.

PR #5393, opened 2022-12-06
How to create a form and show its details in SP

MariaShaposhnikova:


arjan-h (commented):
Maybe change "Configure a form (a data view)" into "Configure a form (using a data view)"

arjan-h (approved):

PR #5397, opened 2022-12-06
Add new OQL String functions

dahfjkg:
Needs an internal review.
To be added in 9.21.0 release.


content/en/docs/refguide/modeling/resources/data-sets/oql/oql-expressions/oql-functions/oql-lower.md
```diff
@@ -0,0 +1,21 @@
...
+
+## 2 Syntax
+
+The syntax is as follows:
+
+```sql
+LOWER ( expression )
+```
+
+`expression` specifies the string to which the transformation is to be applied. The value can be an attribute, a constant, or an OQL parameter.
```
basdebakker: This is not correct. Attributes, constants and parameters are examples of expressions, but the expression can be anything. The value can only be a string.

basdebakker (approved):
Approved by data storage team.

NatasaKralj:
Language reviewed and edited. PR is ready for 9.21.0 release.

NatasaKralj:
Added new functions to list of reserved OQL words

basdebakker:
> Added new functions to list of reserved OQL words

The new functions are not reserved words. We took care of that in the implementation to avoid invalidating older OQL that may use these.

NatasaKralj:
> > Added new functions to list of reserved OQL words
> 
> The new functions are not reserved words. We took care of that in the implementation to avoid invalidating older OQL that may use these.

Oh, my mistake then. Thank you for letting me know. I'll revert the reserved words list.

NatasaKralj (approved):
Ready for 9.21 and approved

PR #5401, opened 2022-12-07
Add Appcues script

NatasaKralj:
Adds a call to Appcues script, as requested from C&C. It will only show up and work on the production version of the site.


MarkvanMents (approved):
Looks good. But we won't know if it works until we try it in production!
I have Slacked you one concern privately, but the technical solution looks fine.

NatasaKralj:
It was decided against using anonymous. Closing PR.

PR #5412, opened 2022-12-08
Add information about internal vs external end-users

MarkvanMents:


MarkvanMents:
Should we also add a definition of "Active Users"?

satyammx (approved):

satyammx (approved):

PR #5419, opened 2022-12-12
ibm cloud deprecate docs

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Confirmed with the data storage team that DB2 isn't deprecated in this version of Mendix, so won't be making that change. Other changes reviewed by Riyad. Ready to merge.

PR #5421, opened 2022-12-13
Remove Expectations for when Garbage Collector runs

MarkvanMents:
The Java garbage collector works differently depending on the version of Java, and many factors influence when it runs and what it does.
This PR removes any expectation of when the GC might run, as it is not possible to predict with any certainty.

### DIFF
--- a/content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
+++ b/content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
@@ -5,6 +5,8 @@ weight: 2
 tags: ["runtime", "java", "memory usage", "memory", "studio pro"]
 ---
 
+## 1 Introduction
+
 The Java memory is divided in different Memory Usage blocks. Each of these blocks are a snapshot of the actual memory usage of that segment. Each of the memory usage blocks can be broken down into four different values
 
 | Memory block | Description |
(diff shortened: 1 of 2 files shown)


content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
```diff
@@ -58,4 +60,4 @@ A healthy Mendix application that consumes a small amount of memory will show a
 
 The graph on the right shows an unhealthy application. As can be seen here, the memory usage steadily increases throughout the span of one week. This can only be caused by a process that keeps consuming memory.
 
-It is acceptable for an application to consume a lot of memory in the tenured generation space, the JVM should run the major garbage collection and reduce the tenured generation to zero.
+It is acceptable for an application to consume a lot of memory in the tenured generation space, the JVM should run the major garbage collection and reduce the tenured generation to zero.
```
MarkvanMents, line 65: This seems to contradict the paragraph above which says that consuming a lot of memory is unhealthy. Here we say it is acceptable. Which is correct?

grootjans: The critical part in the first paragraph here is that the memory usage is steadily increasing. If after a collect in the tenured space the amount held is reduced again, and does not rise with subsequent collects, it does not indicate an unhealthy application.

MarkvanMents: Thanks. I suppose I thought that the tenured space was only being collected where there was a big drop at the beginning of week 42 - but if we are expecting the tenured space to be collected more often, that definitely makes sense.
Thanks for clarifying.
image

content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
```diff
@@ -16,11 +18,11 @@ The Java memory is divided in different Memory Usage blocks. Each of these block
 
 For all Mendix applications the value for init and max start with identical values. Immediately after startup the JVM can execute the garbage collection and correct the memory usage.
 
-## Memory Segments
+## 2 Memory Segments
 
-### **_Perm Gen and Code Cache    _**
+### 2.1 Permanent Generation and Code Cache
```
grootjans: The perm gen has been replaced since Java 8. I suggest we replace this with:

Classes and libraries withing the JVM are allocated in metaspace. This is a separate part of native memory. The allocated memory to the Perm Gen stays fairly static and only increases when new libraries or classes are loaded into the application. The metaspace is not part of the heap, but in native OS memory. In JVM version 7 and before this data was stored in what was called the 'permanent generation'. In contrast to the so-called 'Perm-gen' metaspace is able to collect classes that are no longer used and reduce memory impact.

content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
grootjans: On that part you are right. It is being collected at that point. 
What I would say is that the second paragraph is contradicting the fact that it is unhealthy. What I would say is that it would be unhealthy if at wk 42 the collect would be done, but the tenured space would not be reduced to a small amount. this means that 1. memory has been consumed that survived eden collects, so is not short lived 2. could not be collected after a week. 
This together indicates a high chance of a memory leak, which is a definite sign of an unhealthy application.

What you could say of the first graph that it is definitely more healthy. Memory that is only used in a transient fashion e.g. only for a small moment in time, is directly collected and not promoted to the tenured space. This application can likely work with much less memory assigned. The fact that the permanent generation is also not collected sooner is also interesting, but a definite story about when the JVM decides to GC is not something I can confidently share. It likely does collects when it has enough 'time' and an application under less stress will likely do a collection sooner. But with modern GCs I cannot say what that threshold is without more research.

content/en/docs/refguide/runtime/runtime-java/java-memory-usage.md
MarkvanMents: Updated with your suggestion.

grootjans (approved):
Looks good to me!

PR #5425, opened 2022-12-15
Update quickstart-part1.md

TimvanSteenbergen:
Function trim never returns empty. Always an empty string, even if the Picture is null


ConnorLand (approved):

ConnorLand:
Danny verified the change, so I'll merge. There's also an added link to the Trim section now in case our Quickstart Guide users want to practice exploring other areas of the documentation.

PR #5429, opened 2022-12-16
OIDC and SAML incompatible with Mx 9.20+

MarkvanMents:


feddovanede (commented):
1. The same issue applies to the MendixSSO and XSUAA modules as well.
2. According to the latest information, the change was made in 9.20, not 9.19.

MarkvanMents:
Approval received over Slack

PR #5453, opened 2022-12-21
Add linter GH action

NatasaKralj:
This PR adds the yaml config for the linting GH action.
It also adds the tmp.log file used in the action to gitignore.

### DIFF
--- a/.github/workflows/lint-action.yml
+++ b/.github/workflows/lint-action.yml
@@ -1,0 +1,55 @@
+# This is a basic workflow to help you get started with Actions
+name: Auto lint content
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow at a specific time
+  schedule:
+    - cron: '0 20 1 * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "lint-pr"
+  lint-pr:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout repo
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-action.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Run markdownlint-cli2 on docs
+          title: '[Auto] Lint docs'
+          body: |
+            ${{ env.VER }}
+          branch: lint-docs
+          assignees: NatasaKralj
+          reviewers: NatasaKralj
+          labels: Internal WIP
```
MarkvanMents: Should we delete `./tmp.log` before we create the PR. The content should be in the **VER** variable so we could delete it in an extra step. Perhaps unnecessary, but makes things a bit clearer.
Having it in .gitignore is also useful.

MarkvanMents: Also need a blank line at the end of the file? GitHub likes complaining if it isn't there.

MarkvanMents (approved):
Hi Natasa,
Just a couple of comments - mainly about good programming practice (although my programming days were long ago).
Looks great - is a really good addition to the site. Amazing to have got it working just as you designed it.
I'll assume that the actual yaml works, since you did a demo last week.
Let me know if you have any questions, but there is nothing which has to be changed so I'll approve it.

.gitignore
```diff
@@ -6,6 +6,8 @@
 hugo.exe
 /htmltest.exe
 /.hugo_build.lock
+# Extra file from GH Actions
+tmp.log
```
MarkvanMents: I put a comment here - but it's got lost.

I think we shouldn't ignore every tmp.log on the site. I think we should specify which one we are ignoring - just in case we need to debug something else in a year's time.

.github/workflows/lint-action.yml
NatasaKralj: As discussed, I added the path for our content files to the PR action. This way we don't have to have tmp.log in the gitignore or delete it prior to creating the PR, it just never gets added to the PR.

Issue #5438, opened 2022-12-17
parse-and-format-date-function-calls


rondv5:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [parse-and-format-date-function-calls](https://docs.mendix.com/refguide/parse-and-format-date-function-calls/)

Document link: [parse-and-format-date-function-calls.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide/modeling/application-logic/expressions/parse-and-format-date-function-calls.md)

## My Issue/Suggestion
When I use the format examples for formatDate in my code I don't get warnings but I do get crashes in execution. Looks like the examples are not valid. Another observation, where is mm representing month listed in the table in the page? I don't see it?
The examples where yyyy/mm/dd are listed, do you mean yyyy/ll/dd (or the caps version thereof)?
I'm confused, this page does not help resolve how I should be using the functions. Can it be adjusted and improved, or can you explain why yyyy-mm-dd does work as 2020-12-16 when the table seems to indicate that it would be minutes for mm?

Yiyun333:
Hi @rondv5:

Thank you very much for your feedback!

In the example yyyy/mm/dd, mm should be MM, and capitalization does make a big difference. We will change this example in the doc. Moreover, if you run into problems with MM, LL (uppercase) is a valid alternative. We are currently working on updating this document to make things more clear.

If you continue to have problems with MM or LL, please raise a support ticket here (https://support.mendix.com/hc/en-us).

Regards,
Yiyun

Yiyun333:
Hi @rondv5:

We've further updated the doc (see PR: https://github.com/mendix/docs/pull/5473). Currently, both the pattern letters M and L work completely as expected in microflows. However, in nanoflows, there are limitations, especially with languages with the genitive case (the month name changes depending on the context). Please see the detailed description in the PR.

Thank you for your patience. We hope that these limitations will be fixed in the future! I am closing this issue. Please raise a new issue if you have further questions.

Best regards,
Yiyun

PR #5457, opened 2022-12-21
Update doc Parse and Format Date Function Calls

Yiyun333:
Add M to the table and add a note about the inconsistent behaviors.
Related to GitHub issue: https://github.com/mendix/docs/issues/5438

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/expressions/parse-and-format-date-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/parse-and-format-date-function-calls.md
@@ -36,7 +36,7 @@ The following pattern letters can be used to parse and format Date and time valu
 | S      | Millisecond                               | 201                    |
 
 {{% alert color="info" %}}
-At this moment, neither the pattern letter M nor L works consistently in distinguishing between context and standalone in nanoflows and when used for text that has three or more characters (for instance, MMM or MMMM).
+In nanoflows, neither the pattern letter M nor L works consistently in distinguishing between context and standalone currently when the pattern has three or more characters (for instance, MMM or MMMM).
 {{% /alert %}}
 
 The following pattern letters are only available for microflows:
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/application-logic/expressions/parse-and-format-date-function-calls.md
```diff
@@ -34,6 +35,10 @@ The following pattern letters can be used to parse and format Date and time valu
 | s      | Second in minute                          | 50                     |
 | S      | Millisecond                               | 201                    |
 
+{{% alert color="info" %}}
+At this moment, neither the pattern letter M nor L works consistently in distinguishing between context and standalone in nanoflows and when used for text that has three or more characters (for instance, MMM or MMMM).
```
MarkvanMents: I think it would be better to introduce that this is about nanoflows at the beginning, rather than the end? Perhaps just swapping "in nanoflows" with "At this moment" (or "currently" or "at present" to make it a bit shorter?).
Also, I don't think you need "and" - the context and standalone distinctions are only (I think) for the text versions of months - it doesn't affect MM or M which are interpreted as the month number.
```suggestion
At this moment, neither the pattern letter M nor L works consistently in distinguishing between context and standalone in nanoflows when used for text that has three or more characters (for instance, MMM or MMMM).
```
Also, it isn't really the "text" which has three or more characters, it is the "pattern" - "MMMM" can produce "May" or "December".

MarkvanMents (changes requested):
Hi Yiyun
Here are my comments (again - don't know what happened to the ones I did yesterday).
Nice response to the customer in the original issue.
I've made some suggestions to make it easier for the customer to decide more quickly whether it is important for them.
The commit suggestion is more to point out which "and" I meant than to resolve all my comments.
Let me know if you disagree or have any questions.
Mark

Yiyun333:
> Hi Yiyun Here are my comments (again - don't know what happened to the ones I did yesterday). Nice response to the customer in the original issue. I've made some suggestions to make it easier for the customer to decide more quickly whether it is important for them. The commit suggestion is more to point out which "and" I meant than to resolve all my comments. Let me know if you disagree or have any questions. Mark

Hi Mark,
I just updated the note based on your comments (see commit 1b195db6eb910607c66753b8ccfcafeda6931a06). Thanks for the nice comments. 
Let me know if there is more that needs to be improved.
Yiyun

MarkvanMents (approved):
Great.

PR #5467, opened 2022-12-27
Add top/skip info to OData doc

johannahemminger:
https://mendix.atlassian.net/browse/TW-1942

### DIFF
--- a/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
+++ b/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
@@ -136,7 +136,7 @@ You can select which attributes and associations to return by specifying the `$s
 
 ## 7 Paging {#paging}
 
-Paging, or pagination, allows you to load data incrementally. See [Pagination](https://learn.microsoft.com/en-us/odata/client/pagination) in Microsoft's OData documentation for more information.
+Paging, or pagination, allows you to load data incrementally to better handle large amounts of data. See [Pagination](https://learn.microsoft.com/en-us/odata/client/pagination) in Microsoft's OData documentation for more information.
 
 ### 7.1 Top (Limit)
 
--- a/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
+++ b/content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
@@ -136,6 +136,8 @@ You can select which attributes and associations to return by specifying the `$s
 
 ## 7 Paging {#paging}
 
+Paging, or pagination, helps you allows you to load data incrementally. See [Pagination](https://learn.microsoft.com/en-us/odata/client/pagination) in Microsoft's OData documentation for more information.
+
 ### 7.1 Top (Limit)
 
 You can limit the number of returned objects using the `$top` query option, where the limit is a positive integer. For example: `?$top=100`.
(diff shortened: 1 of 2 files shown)


content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
```diff
@@ -136,6 +136,8 @@ You can select which attributes and associations to return by specifying the `$s
 
 ## 7 Paging {#paging}
 
+Paging, or pagination, allows you to load data incrementally to better handle large amounts of data. See [Pagination](https://learn.microsoft.com/en-us/odata/client/pagination) in Microsoft's OData documentation for more information.
```
joostverhoog: I think we should not use this link https://learn.microsoft.com/en-us/odata/client/pagination
First it's about microsoft's OData client, and there will be very few customers that use that client.
And second OData has moved from being a Microsoft standard to being an Oasis standard, so links to Microsoft in this context may suggest that we still treat it as a non-open standard.

content/en/docs/refguide/modeling/integration/published-odata-services/published-odata-resource.md
```diff
@@ -191,6 +193,12 @@ The *Deletable* capability was introduced in Studio Pro [9.13.0](/releasenotes/s
 
 ## 10 Query Options {#query-options}
 
-Select the options that you would like to include in each OData capability.
+Select the options to include for the **Readable** OData capability.
 
 * **Countable** – required for getting the total number of records
+* **Top supported** – required when **Use paging** is selected.  This option supports the $top query, which requests the number of items in the queried collection to be included in the result.
```
joostverhoog: Can we put `required when **Use paging** is selected` at the end of this line? It's not as important as the rest of this line, so having that come first would improve readability.

joostverhoog: "This option supports the $top query" -> "This option indicates whether clients can specify that they want to retrieve only a limited number of items."

Saying that the option supports something is not correct, my suggestion tries to rewrite it.

content/en/docs/refguide/modeling/integration/published-odata-services/published-odata-resource.md
```diff
@@ -191,6 +193,12 @@ The *Deletable* capability was introduced in Studio Pro [9.13.0](/releasenotes/s
...
 
-Select the options that you would like to include in each OData capability.
+Select the options to include for the **Readable** OData capability.
 
 * **Countable** – required for getting the total number of records
+* **Top supported** – required when **Use paging** is selected.  This option supports the $top query, which requests the number of items in the queried collection to be included in the result. 
+* **Skip supported** – required when **Use paging** is selected. This option supports the $skip query, which requests the number of items in the queried collection that are to be skipped and not included in the result.
+
+For more information, see the following:
+* [Pagination](https://learn.microsoft.com/en-us/odata/client/pagination) in Microsoft's OData documentation
```
joostverhoog: I would like to have this link removed. The next one is fine.

content/en/docs/refguide/modeling/integration/published-odata-services/odata-query-options.md
johannahemminger: Makes sense, thanks.

PR #5499, opened 2023-01-05
DD_LOG_LEVEL clarification

denis-shuvalov:
Added a note which clarifies how to set correct logging level of mx app.


MarkvanMents (approved):
Hi Denis,
This information was already in section 5.1 of this document - but I'm happy to add it here too if customers are missing it.
image

I've proofread it - let me know if I can publish.

PR #5503, opened 2023-01-09
Add icon collection consistency errors

hendrik-jacobs-mendix:
This pull request adds information on 2 consistency errors that might occur due to new capabilities added to the Icon Collections features.
This feature is still not yet publicly available and should be released as part of version 9.23.0

### DIFF
--- a/content/en/docs/refguide/modeling/menus/view-menu/errors-pane/consistency-errors/consistency-errors-pages.md
+++ b/content/en/docs/refguide/modeling/menus/view-menu/errors-pane/consistency-errors/consistency-errors-pages.md
@@ -290,7 +290,7 @@ To fix the consistency errors, finish configuring the on click event (for exampl
 
 ## 10 Icon Collection Consistency Errors
 
-Icon collections have configurable class names, icon prefixes, and icon names. These properties are used to generate the CSS code for our pages, so to prevent generating invalid CSS, you can come accross the following consistency errors:
+Icon collections have configurable class names, icon prefixes, and icon names. These properties are used to generate the CSS code for your pages. You can come accross the following consistency errors that prevent you from generating invalid CSS:
 
 | Error code | Message in the Error Pane | Cause of the Error | Way to Fix |
 | ---------- | ------------------------- | ------------------ | ---------- |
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/menus/view-menu/errors-pane/consistency-errors/consistency-errors-pages.md
```diff
@@ -288,6 +288,15 @@ The most common consistency errors are connected with not configuring the on cli
...
+
+Icon collections have configurable class names, icon prefixes, and icon names. These properties are used to generate the CSS code for our pages, so to prevent generating invalid CSS, you can come accross the following consistency errors:
+
+| Error code | Message in the Error Pane | Cause of the Error | Way to Fix |
+| ---------- | ------------------------- | ------------------ | ---------- |
+|   CE1616   | You cannot duplicate the icon collection class name {icon collection class name}. | There is more than one icon collection configured with the same class name. | Change the class name for one of the icon collections. |
+|   CE5696   | Duplicate CSS class name {icon class name} for icons, resulting from the combination of the icon collection prefix and the icon name. | The combination of the icon collection prefix and the icon name has automatically generated the same class name. | Do one of the following: <ul><li>Ensure that the icons have unique names that are not used in any of the icon collection prefixes.</li><li>Ensure that all the icon collections related to the icons have unique prefixes that will not be used for icon names.</li></ul> |
```
MariaShaposhnikova: @hendrik-jacobs-mendix , there is something wrong with the error text itself for CE5696. It basically tells a user to duplicate CSS class name, while I think we want to say: "Duplicated CSS class name {icon class name}, resulting from a combination of the icon collection prefix and the icon name."

hendrik-jacobs-mendix: There seems to be a standard within the errors when it comes to duplicates where the error would just say Duplicate and not duplicates, a couple of examples are as follows:

- Duplicate enumeration values.
- Duplicate indexes.

In this case I also want to specify that this duplicate is for icons and not for the icon collection.
I also want to say that the css class names are a result of combining two other fields and therefore it can be a change on either the icon collection prefix or the icon name causing the error.

Given the above should I still change it, or would something like "Duplicate CSS class names for icons, resulting from a combination of the icon collection prefix and the icon name" be acceptable?
The key being that I pluralized _names_

hendrik-jacobs-mendix:
Please be aware that we are expecting some more changes on this documentation following some user feedback so the messages will need to change as well, and the release version needs to be bumped to 9.24

MariaShaposhnikova (approved):

PR #5527, opened 2023-01-12
Mendix Cloud replaces custom HTTP Reason Codes

MarkvanMents:


feddovanede (commented):
Looks good to me. Should we also add it to https://docs.mendix.com/developerportal/deploy/mxcloudv4/#other-considerations, where we list considerations/limitations/behavior for Mendix Cloud?

feddovanede (approved):
Looks good to me!

PR #5529, opened 2023-01-13
Remove git clone and git submodule output from log.

MarkvanMents:
New Log Scan feature in Travis is flagging the commit numbers reported by the git process as potential security leaks.
Suppressing this output so we don't get false positives.


MarkvanMents:
Adding `quiet:true` to the `git` section does not work as it doesn't get applied to the `git submodules` command.
Discussing with Travis support.

MarkvanMents:
Agreed to merge this - will follow up with Travis on alternative solutions.

PR #5531, opened 2023-01-13
Remove deprecation from radio button list

HedwigAR:
Only a part of this component was deprecated in the past, the warning for this was updated a few months ago include the entire thing. This isn't correct, so instead we'd like to remove the warning all together and add it again when we're starting the work on the new replacement widget.

### DIFF
--- a/content/en/docs/appstore/widgets/radio-button-list.md
+++ b/content/en/docs/appstore/widgets/radio-button-list.md
@@ -7,10 +7,6 @@ tags: ["marketplace", "marketplace component", "widget", "radio button list", "p
 #If moving or renaming this doc file, implement a temporary redirect and let the respective team know they should update the URL in the product. See Mapping to Products for more details.
 ---
 
-{{% alert color="warning" %}}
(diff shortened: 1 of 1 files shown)


HedwigAR:
Just to note: I have already accepted the license agreement, but for some reason it never works for me
Screenshot 2023-01-16 at 10 20 54

ConnorLand (approved):

ConnorLand (commented):
Let me check on this with you @HedwigAR . If a part of a widget or module is deprecated, the entire widget or module is deprecated.

ConnorLand:
Because Attribute Radio Button list is still deprecated, this warning will not be deleted. I'll confirm with my team, but I'm fairly sure  someone needs to add a label to the Marketplace component indicating the widget's deprecated status.

HedwigAR:
With the latest Radio Button List release, the deprecation has been removed from the name of it. PM and SDM have agreed to revert the deprecation for this widget until we can point people to a replacement.

ConnorLand (approved):

PR #5557, opened 2023-01-20
Update URL names in Mendix Cloud

MarkvanMents:


Sakshi-Dhakad:
Looks good, thanks for the quick turnaround on this!

PR #5559, opened 2023-01-22
Update _index.md

RCatersels:
For a university program I was trying out a scratch build to help the students along. But I noticed when pressing F1 that the documentation did not give me any help on what I wanted to do namely creating a date time attribuut from a string. Strange that the formatDateTime is not mentioned at all on this page while the FormatDecimal is. So hence I added it to the Date creation part. Or you should add a complete new section and also point to the formatDate and formatTime function calls because they also seem to be missing here.

### DIFF
--- a/content/en/docs/studio/expressions/_index.md
+++ b/content/en/docs/studio/expressions/_index.md
@@ -152,6 +152,7 @@ The list of expressions that are used in Studio the most is represented below. F
 ### 3.8 Date Creation
 
 * [`dateTime`](/refguide/date-creation/) – creating a date value using the server's calendar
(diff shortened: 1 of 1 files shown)


MariaShaposhnikova:
Hi @RCatersels , thank you for your PR, that's a good addition to the document!
I've added a new section with Format Date Function Calls and included the functions you mentioned in your comment. Please note that the list in the Studio guide is not the full list. The Studio list includes the most used expressions, while Studio Pro guide contains the full list (it is mentioned in section 3 **Expression Types** of the document). 

Best regards,
Maria

PR #5575, opened 2023-01-25
Release notes for Mx4PC version 2.10.1 (patch update, planned for January 26th)

zlogic:
We've prepared a patch version of Mx4PC - mostly focusing on CVE fixes. We're planning to release this on January 26th or 27th, depending on the test results.

There's also a new feature to use built-in AWS authentication to connect to the database instead of passwords. This feature was tested to work correctly, but depends on an upcoming Mendix Runtime feature to work correctly. We'll provide additional documentation later, after we have confirmation that everything works correctly.


Adam-Dupaski:
@katarzyna-koltun-mx @MarkvanMents may be worth checking with Security team (Olga) about listing these CVEs here - if that should be done, if/when they will be published by Siemens, and whether https://docs.mendix.com/releasenotes/security-advisories/ should be updated.

MarkvanMents:
@katarzyna-koltun-mx let me know if you need help with this.

PR #5580, opened 2023-01-26
Improved documentation for published OData resource key

joostverhoog:
Addressed a couple of things for choosing a good key for a published OData resource

wrap-services-odata
 * This is about Studio Pro 9.17 already, so that does not need to be mentioned in the key selection section -> deleted
 * This is about non-persistable entities, so talking about the Mendix ID and the database is not correct -> deleted
 * The documentation about what is a good key should not be in this guide, but rather in the documentation of the resource itself.  -> moved; The link for more information is now the other way around.

published-odata-resource
 * There were two sections about the key, with almost the same information. Removed one of them


johannahemminger:
Thanks @joostverhoog !

PR #5637, opened 2023-02-14
typescript is being depricated, right?

jasonteunissen:
typescript is being depricated, right?


Adam-Dupaski:
Thanks @jasonteunissen After discussion with a PM, I'm removing that line entirely, as it is not a necessary recommendation.

PR #5647, opened 2023-02-15
clarification on how to use svg in native

jasonteunissen:
clarification on how to use svg in native

### DIFF
--- a/content/en/docs/refguide8/mobile/native-mobile/native-svg.md
+++ b/content/en/docs/refguide8/mobile/native-mobile/native-svg.md
@@ -33,16 +33,30 @@ We suggest manually removing these elements from your SVGs, or using the tools m
 
 ## 4 Styling SVGs
 
-You might want to change certain colors in your SVG, for example when adding an image. Mendix allows you to do this by setting the `fill` and `stroke` properties in image's styling. These properties will then be applied to *all* the elements inside the SVG that do not have these properties.
+You might want to change certain colors in your SVG, for example when adding an icon. Mendix allows you to do this by setting the `fill` and `stroke` properties in image's styling of the .js. These properties will then be applied to *all* the elements inside the SVG that do not have these properties.
 
 Take the following SVG as an example:
+
+```svg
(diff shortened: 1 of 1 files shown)


ConnorLand:
Hi Jason. Can you give me a little context for your rationale behind these changes? I'm especially curious about the longer code snippet you introduce and then pare back.

jasonteunissen:
When trying to build this out, I had to troubleshoot what worked and what not because there was no clear "before and after" code that was exported from figma, to code that worked in reactnative.

I hope that showing a before and after will help people understand what elements can be removed from the code.

also having the stroke class still in there in the example confused me.

jdiehl:
Thanks for the suggestion! I like that you want to make the example more realistic by using a Figma export. Maybe we can extend that and use something more interesting than just a rectangle that actually makes good use of preserving fill/stroke somewhere?

Also could you explain the other changes you made? Why change the viewBox? Why remove the other svg attributes?

content/en/docs/refguide8/mobile/native-mobile/native-svg.md
```diff
@@ -33,16 +33,30 @@ We suggest manually removing these elements from your SVGs, or using the tools m
 
 ## 4 Styling SVGs
 
-You might want to change certain colors in your SVG, for example when adding an image. Mendix allows you to do this by setting the `fill` and `stroke` properties in image's styling. These properties will then be applied to *all* the elements inside the SVG that do not have these properties.
+You might want to change certain colors in your SVG, for example when adding an icon. Mendix allows you to do this by setting the `fill` and `stroke` properties in image's styling of the .js. These properties will then be applied to *all* the elements inside the SVG that do not have these properties.
```
jdiehl, line 36: SVGs can also be used for other things (not only icon) so we should not change this here.

jdiehl: Also styling of the .js seems wrong - we are not styling javascript.

PR #5664, opened 2023-02-20
Add Solution tab description

mikita-haurykau:


content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
```diff
@@ -319,20 +319,30 @@ A microflow selected for this setting will start every time a user task changes
...
-## 9 Miscellaneous Tab {#miscellaneous}
+## 9 Solution Tab {#solution}
+
+Settings on the **Solution** tab allow you to configure application distribution as an [adaptive solution](/appstore/creating-content/sol-solutions-guide/). 
+
+If you want to distribute your app as an adaptive solution package and allow upgrading it from the [implementer's side](/appstore/creating-content/sol-solutions-impl/), you need to **Enable solution adaptation** on this tab. The title of your app in the App Explorer will change to *Solution* and the solution version will be displayed after the app name.
+
+A distributable app must have a **Solution version** that you can set on this tab.
+
+If you are implementing a solution, *Based on* setting will show the version of the solution package your app is currently based on.
```
MariaShaposhnikova: @mikita-haurykau , implementing a solution is done by someone else, correct? Is the **Based on** setting something that can be set by the creator of the solution, not the implementor?

Adam-Dupaski:
@MariaShaposhnikova @mikita-haurykau added cross-reference: https://github.com/mendix/docs/pull/5664/commits/4bf3b1ec15429a3366b84d834b66f8399d765838

content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
mikita-haurykau: Based On version cannot be set manually. The process is as follows:
1. When a Publisher creates an app and Enables distribution for it they set the Version of the Solution.
2. When they export the App as an mxsolution package the value from Version field is moved to BasedOn Version field in the package.
3. When Consumer implements a solution (imports mxsolution package) the resulting app will have BasedOnVersion set in step 2 and empty Version. Reasoning for this is that an app initialized from an mxsolution is an App. it becomes a solution only when Distribution is enabled. Only then a Solution Version makes sense.
4. Then if a consumer decides to RE-distribute the app (cascading distribution) they have to Enable distribution again, set their own version and then continue as described starting from Step 2. Based On Version of the mxsolution package exported by Consumer (RE-distributing) will be overwritten by the version set bu consumer.
e.g.

**Mendix** creates warehouse management App and decides to distribute it.
Mendix Enables distribution as a solution for this app,
  Sets version to 1.5.8
  Exports the app as `warehousing.mxsolution` package. (in this package the version is empty and Based On Verison is 1.5.8)
**Albert Heijn** buys this warehousing solution.
  Initializes their app from `warehousing.mxsolution` (Based On version shows the version of the mendix solution the app is based on = 1.5.8)
  Adds Promotions module to the app
  Decides to re-distribute it with the new module
  Enables distribution
  Sets their own version (5.6.7)
  Exports as new mxsolution package.
**Jumbo** buys solution from Albert Heijn
  initializes the app ((Based On version shows the version of the mendix solution the app is based on = 5.6.7))

MariaShaposhnikova (approved):

PR #5668, opened 2023-02-21
Update Custom Domain Certificate image

MarkvanMents:


MarkvanMents:
@feddovanede 
I have a feeling that someone forgot to tell me that this change had been released.
Are the buttons now square, and should I publish this small change 😀

MarkvanMents:
OK, I've checked and it seems it doesn't look like either of them any more. So I'll upload the new one and merge this PR.
image

feddovanede:
This is probably related to the migration of Cloud Portal from Mx7 to Mx8. This did include some (minor) styling updates. The buttons are still all in the same place, and they still do the same, they might look slightly different. That means that all images on https://docs.mendix.com/developerportal/deploy/environments-details/ are also outdated.

MarkvanMents:
> This is probably related to the migration of Cloud Portal from Mx7 to Mx8. This did include some (minor) styling updates. The buttons are still all in the same place, and they still do the same, they might look slightly different. That means that all images on https://docs.mendix.com/developerportal/deploy/environments-details/ are also outdated.

@Feddo - I have a task to review a whole load of images, so I'll add all the "Cloud Portal" ones as well.

PR #5677, opened 2023-02-22
Remove beta alert

mikita-haurykau:
Update image

Private Git is becoming GA.
We remove the setting and make it always ON


MariaShaposhnikova (approved):

MariaShaposhnikova:
Due to a Sourcetree issue, I had to create another PR with more changes on the topic that seem to contain changes in this PR too. Just a note that merging only the second PR should be enough, but I will keep this one open for now: https://github.com/mendix/docs/pull/5778

MariaShaposhnikova:
UPD: The Private Git setting will stay in **Preferences** which means this PR is more up-to-date. I will close my other PR: https://github.com/mendix/docs/pull/5778

PR #5689, opened 2023-02-24
Add release notes for Windows Service 4.7.1 and update requirements

jveldhuizen:
Windows Server 2008 is not supported anymore by Microsoft, so we now require Windows Service 2012 with IIS 8.

### DIFF
--- a/content/en/docs/releasenotes/studio-pro/windows-service.md
+++ b/content/en/docs/releasenotes/studio-pro/windows-service.md
@@ -19,7 +19,7 @@ To download the Windows Service, go to the [Get Studio Pro](https://marketplace.
 * Add possibility to add extra logsubscribers in the Settings.yaml file. See https://github.com/mendix/m2ee-tools/blob/v7.2.3/examples/full-documented-m2ee.yaml#L392 for more information. Compared to that information, the 'Logging' property has to start with an uppercase letter, so 'Logging:' instead of 'logging:'. The loglevel setting as described is not supported for the Windows Service.
 * We replaced the modified YAML parser YamlSerializer 0.9.0.2 by YamlDotNet 12.3.1.
 * We updated the Newtonsoft Json.NET library from version 13.0.1 to 13.0.2.
-* Add build number information to the version information of the executables and the subject field of the MSI installer file.
+* We have added build number information to the version information of the executables and the subject field of the MSI installer file.
 
 ### 4.7.0
 
--- a/content/en/docs/releasenotes/studio-pro/windows-service.md
+++ b/content/en/docs/releasenotes/studio-pro/windows-service.md
@@ -12,6 +12,17 @@ To download the Windows Service, go to the [Get Studio Pro](https://marketplace.
 
 ## 4.7
 
+### 4.7.1
+
+**Release date: February 24th, 2023**
+
+* Add possibility to add extra logsubscribers in the Settings.yaml file. See https://github.com/mendix/m2ee-tools/blob/v7.2.3/examples/full-documented-m2ee.yaml#L392 for more information. The 'Logging' property has to start with an uppercase letter, so 'Logging:' instead of 'logging:'. The loglevel node is not supported.
+* We replaced the modified YAML parser YamlSerializer 0.9.0.2 by YamlDotNet 12.3.1.
+* We updated the Newtonsoft Json.NET library from version 13.0.1 to 13.0.2.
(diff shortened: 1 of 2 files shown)


Adam-Dupaski:
@MarkvanMents is this one for you with the `/deploy/on-premises-design/ms-windows/_index.md` dependencies?

MarkvanMents:
I'll pick this up. Thanks Adam.

content/en/docs/releasenotes/studio-pro/windows-service.md
```diff
@@ -12,6 +12,17 @@ To download the Windows Service, go to the [Get Studio Pro](https://marketplace.
 
 ## 4.7
 
+### 4.7.1
+
+**Release date: February 24th, 2023**
+
+* Add possibility to add extra logsubscribers in the Settings.yaml file. See https://github.com/mendix/m2ee-tools/blob/v7.2.3/examples/full-documented-m2ee.yaml#L392 for more information. The 'Logging' property has to start with an uppercase letter, so 'Logging:' instead of 'logging:'. The loglevel node is not supported.
```
MarkvanMents: I find this a bit confusing.
Does this mean that they can set logging as described in the m2ee yaml, but they have to use `Logging:` instead of `logging:` which is shown on line 392?  

I assume that none of the other settings need to use upper case at the beginning. Would it help to tell customers why they need to do this?

MarkvanMents: And all the settings under `logging` are supported except `loglevel`? Would it be better to perhaps call this a setting, rather than a "node"?

content/en/docs/releasenotes/studio-pro/windows-service.md
```diff
@@ -12,6 +12,17 @@ To download the Windows Service, go to the [Get Studio Pro](https://marketplace.
...
+* We replaced the modified YAML parser YamlSerializer 0.9.0.2 by YamlDotNet 12.3.1.
+* We updated the Newtonsoft Json.NET library from version 13.0.1 to 13.0.2.
+* Add build number information to the version information of the executables and the subject field of the MSI installer file.
```
MarkvanMents:
```suggestion
* We have added build number information to the version information of the executables and the subject field of the MSI installer file.
```

So you now have something like `4.7.1.123456` as the version number?

MarkvanMents (commented):
Hi Jonathan @jveldhuizen,
Thanks for the PR.
I have a couple of questions about the Settings.yaml where it could perhaps be more obvious to the customer what is happening and why?
Can you look at my questions and see if I can make anything clearer?

content/en/docs/releasenotes/studio-pro/windows-service.md
jveldhuizen: Indeed. In fact, that was already the situation before, but that 'four-parts-number' was only visible in Windows if you look-up the version of the installed Windows Service application in the overview of installed programs on your pc. That was not so easy.  And the main executables only had the first three parts of the version number visible in the Properties of the executable in Windows Explorer, so without the build number.
We now extended the version number with the build number on the executables and you now see the whole number when you right-click the installer, choose Properties and then see the Subject field on the Details tab. That's important, because I could not see earlier what the exact version of an installer was before actually installing it. I want that, because sometimes I test versions, make a fix and rebuild the same version again with the fix. Then build numbers are nice information to know before actually installing.

content/en/docs/releasenotes/studio-pro/windows-service.md
jveldhuizen: I think that's not so important for customers, but if you want... it doesn't matter.
The reason for the uppercase in `Logging:` is that it's consistent with the other settings in the Yaml file of the Windows Service.
The subsettings within the Logging part are based on what Runtime offers and requires. Currently, these are all lowercase, like for the types `syslog` and `file`. The Windows Service does nothing with the subsettings within the Logging part. In only sends them unchanged to the Runtime as part of the logging configuration of the Runtime. The description of the specific log subscribers should be somewhere in the Runtime documentation. However, I think we still don't have any documentation on that area.

jveldhuizen: Indeed, I renamed it and added some information in the third commit of this MR.

MarkvanMents (approved):
@jveldhuizen 
Thanks for addressing my questions and updating the PR.
I will merge this immediately, since the release was on Friday.

PR #5724, opened 2023-03-07
[README.md] Added a note how to limit Hugo memory usage.

zlogic:
When running Hugo in a container, or environment like Github Codespaces, Hugo will load all pages into memory by default.

At the moment, this requires 12+GB of available memory; with --renderToDisk, Hugo needs less than 4GB RAM. In my case, I didn't see any performance impact.


MarkvanMents:
Hi @NatasaKralj 
This looks like a good suggestion. I don't see so much difference under Windows - although Hugo's memory use seems to drop slowly and perhaps will stabilize at a lot lower memory use eventually. Can you have a look and approve/merge?
Thanks.

NatasaKralj (approved):

NatasaKralj:
Looks like a fine suggestion. It builds slightly slower with this option, but memory usage is significantly decreased. Approving and merging.

PR #5727, opened 2023-03-07
Documented a workaround for Mx4PC/Istio (no release date)

zlogic:
A few customers had network issues when using an Istio service mesh - added a note that the `holdApplicationUntilProxyStarts` feature should be enabled to avoid network issues.


katarzyna-koltun-mx:
@zlogic Reviewed - can I publish the PR or should it wait?

katarzyna-koltun-mx:
Confirmed on slack, merging now.

PR #5738, opened 2023-03-10
[RUN-3390] Add information about Gradle in offline install docs (for 9.24.0)

atennapel:


MariaShaposhnikova:
@atennapel , in release notes we say: Studio Pro now builds apps using Gradle. You can switch back to Ant in the Studio Pro settings. 
Do we need to mention this part in docs? "You can switch back to Ant in the Studio Pro settings."
Is there a new setting somewhere in SP for switching back to Ant? It needs a doc update too then.

atennapel:
I'll clarify the situation. We want to remove building with Ant in Mx10. So our plan was to:
- Introduce building with Gradle by default in 9.24 so we can collect feedback already for Mx10.
- Still have a fallback option to Ant (in Studio Pro preferences) so customers can switch back if something is wrong
- In Mx10 remove Ant completely including the fallback option.
That's why we didn't include it in the documentation because the Ant fallback setting is only available in 9.24 and will be removed in Mx10. What do you think?

MariaShaposhnikova:
@atennapel , since we are updating the documentation for 9.24 now, the setting description should be in the documentation too. Besides, for Mx10 there will be a separate space, for users on 9.24 the setting will still be available and therefore docs describing Mx9 should reflect it. 
Can you please add the description of the setting to this PR?

atennapel:
@MariaShaposhnikova added the Gradle preference information.

MariaShaposhnikova (approved):

PR #5743, opened 2023-03-10
SAP app deletion

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@mayankmendix I've added a note about moving to public cloud before deletion to the SAP documentation. I'm wondering if the migration process should also be described in detail. Or should I say that the client should open a support request for the migration?

content/en/docs/developerportal/deploy/sap-cloud-platform.md
```diff
@@ -734,13 +734,8 @@ A more detailed description of the reason why the environment creation failed wi
...
 
-Note that if you are the last person to leave a Mendix app you can delete the app. However, this will not delete the app or resources on SAP BTP. You can leave the app by going to the **General Settings** page of the Developer Portal and clicking **Leave app**.
-
-{{< figure src="/attachments/developerportal/deploy/sap-cloud-platform/leave-app.png" >}}
-
-If you are the last member of the app development team, you will be asked if you want to delete the app.
-
-{{< figure src="/attachments/developerportal/deploy/sap-cloud-platform/delete-app.png" >}}
+Note that if you are the last person to leave a Mendix app you can delete the app. However, it is only possible to delete apps hosted in the Mendix Public cloud. For SAP BTP, you must first switch back to Mendix Public
+Cloud before you can delete your app.
```
MarkvanMents, line 738: I wonder if we should describe our favoured process very clearly and add the other ones later? I think we need to take the following paragraphs into account as well in our rewrite as I find it unclear exactly what the process is that we are recommending.

Issue #5706, opened 2023-03-02
oidc-sso anonymous user enabling optional


ghost:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [oidc-sso](https://docs.mendix.com/appstore/modules/oidc/)

Document link: [oidc.md](https://github.com/mendix/docs/blob/development/content/en/docs/appstore/modules/oidc.md)

## My Issue/Suggestion

Hey there!
Why is the enabling of anonymous users part of the OIDC documentation [0]? As for me the OIDC module works just fine without enabling it and it moreof seems like a security issue for me. In my opinion this step should at least be marked as optional. Or is there something I'm missing out on?

[0] https://docs.mendix.com/appstore/modules/oidc/#42-allowing-anonymous-users

MarkvanMents:
Hi,
Thank you for bringing this to our attention. I have been in contact with the developers and they have explained why anonymous users are required and have identified that a piece of configuration was left out of the documentation.
I've updated the documentation to add the missing configuration and explain why anonymous users are required. Please see the Pull Request https://github.com/mendix/docs/pull/5750 for details of the changes.
In summary, my understanding is that the anonymous user access is used in two circumstances:

1. When there are multiple IdPs configured, to allow the end user to choose which IdP to use.
2. When the end user signs out, to show them a page where they can choose to sign back in again if required.

Without this access, the OIDC module will still work to log users in, but you may want to test these specific scenarios and ensure that your app behaves as you want it.

I offered your suggestion of making the step optional, but the developers felt that this would affect how the module is supposed to work.

Thanks you for raising this issue and helping us identify some missing documentation.

Yours

Mark van Ments.

PR #5750, opened 2023-03-14
Anonymous users in OIDC module

MarkvanMents:


content/en/docs/appstore/modules/oidc.md
```diff
@@ -164,6 +164,8 @@ You may have to add the *Anonymous* user role if it does not exist already.
 
 ### 4.2 Allowing Anonymous Users
 
+The OIDC module supports multiple OIDC/OAuth-compatible IdPs. If you want to allow your end users to choose from a number of different IdPs, you will need to give them access to the app to choose their preferred IdP. Because they need to do this before they have signed in to the app, you need to give anonymous users access to your app. If you have only configured a single IdP, you do not need to allow anonymous users.
+
```
SharishmaPodduturi999, line 168: Hi Mark,
I have checked the documentation of OIDC SSO module.The missed part in the docs is [Navigation](https://docs.mendix.com/appstore/modules/oidc/#43-configuring-navigation)
Add Role-based home page(Login_Web_Button page) for Anonymous userroles in the Navigation
Screenshot 2023-03-06 at 6 46 07 PM

To be clear:
If we enable/disable anonymous users, users can configure multiple/single IDPs for their app but they will not be able to see the configured IDPs when they logout if they don't configure the Login_Web_button page in the Navigation
Screenshot 2023-03-10 at 9 51 12 AM

Even though if they configure single IDP, they need to allow anonymous users.

Can you please remove the below line
`If you have only configured a single IdP, you do not need to allow anonymous users.`

and add the missing part in the navigation

MarkvanMents: @SharishmaPodduturi999 
I'm puzzled that this is not the experience of the customer who raised the issue of anonymous users #5706. They say that it works fine for them without enabling anonymous users.
However, I'll make the changes you suggest for now and report back to the customer once they have been made.

MarkvanMents: I've updated this in https://github.com/mendix/docs/pull/5750/commits/87f61f397416e1d56bb77a2e2828657a5f164bc6.
Does it look OK now?

SharishmaPodduturi999: @MarkvanMents It looks good to me

MarkvanMents: Thanks - I'll publish now.

PR #5775, opened 2023-03-20
[RUN-3452] Add limitation of Prometheus

atennapel:
Q: Is "warning" the correct level for this note?

### DIFF
--- a/content/en/docs/refguide/runtime/metrics.md
+++ b/content/en/docs/refguide/runtime/metrics.md
@@ -78,7 +78,7 @@ The following settings can be used, depending on the type of metrics being gener
 We do not support multiple [Prometheus](https://prometheus.io/docs/introduction/overview/). When the Prometheus registry is set, it can be accessed through the `/prometheus` context path over the admin endpoint.
 
 {{% alert color="warning" %}}
-Prometheus has a limitation where metrics with the same name but a different set of tags can be ignored. Make sure all metrics with the same name have the same set of tags.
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/runtime/metrics.md
```diff
@@ -77,6 +77,10 @@ The following settings can be used, depending on the type of metrics being gener
 
 We do not support multiple [Prometheus](https://prometheus.io/docs/introduction/overview/). When the Prometheus registry is set, it can be accessed through the `/prometheus` context path over the admin endpoint.
 
+{{% alert color="warning" %}}
+Prometheus has a limitation where metrics with the same name but a different set of tags can be ignored. Make sure all metrics with the same name have the same set of tags.
+{{% /alert %}}
```
MarkvanMents: @atennapel 
Warning looks good for this one. I think it needs more than an info.
You say "can be" ignored. I think we should either say "are" ignored, or  we might want to change the text around a bit as I read it as the user can ignore them, not Prometheus.
```suggestion
{{% alert color="warning" %}}
Prometheus has a limitation where it sometimes ignores metrics with the same name but a different set of tags. Make sure all metrics with the same name have the same set of tags.
{{% /alert %}}
```

Does this suggestion look OK?
And can this be published immediately?

atennapel: Hey Mark, yes this suggestion is great. This can be published immediately, this limitation has been in Studio Pro since 9.6, we've only just come across it.

PR #5781, opened 2023-03-20
[WTF-1438] Document new source variables (9.24)

jeffersonRibeiro:
Add documentation for source variables in text templates, conditional settings, and association widgets in 9.24.

### DIFF
--- a/content/en/docs/refguide/modeling/pages/data-widgets/_index.md
+++ b/content/en/docs/refguide/modeling/pages/data-widgets/_index.md
@@ -27,7 +27,7 @@ The Data containers category contains the following widgets:
 
     {{< figure src="/attachments/refguide/modeling/pages/data-widgets/list-view-example.png"   width="450"  >}}
 
-A [snippet](/refguide/snippet/) can also function as a data container when it has at least one snippet parameter. Unlike other data containers, a snippet never exposes a `$currentObject` variable as it can represent multiple independent objects.
+A [snippet](/refguide/snippet/) can also function as a data container when it has at least one snippet parameter. Unlike other data containers, a snippet never exposes a `$currentObject` variable because it can represent multiple independent objects.
 
 ## 2 Performing Basic Functions
 
--- a/content/en/docs/refguide/modeling/pages/data-widgets/_index.md
+++ b/content/en/docs/refguide/modeling/pages/data-widgets/_index.md
@@ -2,7 +2,7 @@
 title: "Data Containers"
 url: /refguide/data-widgets/
 weight: 15
-tags: ["studio pro", "data widgets", "data containers", "data view", "list view", "data grid"]
+tags: ["studio pro", "data widgets", "data containers", "data view", "list view", "data grid", "snippet"]
 ---
 
 ## 1 Introduction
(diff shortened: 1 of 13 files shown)


ConnorLand:
@jeffersonRibeiro  Is this the main improvement documented in this PR?

"A [snippet](/refguide/snippet/) can also function as a data container when it has at least one snippet parameter. Unlike other data containers, a snippet never exposes a `$currentObject` variable because it can represent multiple independent objects."

If so, we need to add a bullet about that improvement to the [9.24 release notes](https://github.com/mendix/docs/pull/5759).

jeffersonRibeiro:
> @jeffersonRibeiro Is this the main improvement documented in this PR?
> 
> "A [snippet](/refguide/snippet/) can also function as a data container when it has at least one snippet parameter. Unlike other data containers, a snippet never exposes a `$currentObject` variable because it can represent multiple independent objects."
> 
> If so, we need to add a bullet about that improvement to the [9.24 release notes](https://github.com/mendix/docs/pull/5759).

No, the main improvement was that `attribute-based expressions` for input elements can now select entities from outside the default entity context. For example a grand-parent data container or a snippet parameter.

The same improvement was done for the `reference selector` and `input reference set selector` when selecting an attribute path.

Since the refguide for the affected widgets already mention the `refguide/data-widgets` and the current doc description for selecting attributes doesn't specify the previous limitation, the only change needed for the docs was to add a snippet as a data container when it has at least 1 snippet parameter.

content/en/docs/refguide/modeling/pages/data-widgets/_index.md
```diff
@@ -27,6 +27,8 @@ The Data containers category contains the following widgets:
 
     {{< figure src="/attachments/refguide/modeling/pages/data-widgets/list-view-example.png"   width="450"  >}}
 
+A [snippet](/refguide/snippet/) can also function as a data container when it has at least one snippet parameter. Unlike other data containers, a snippet never exposes a `$currentObject` variable because it can represent multiple independent objects.
```
ConnorLand, line 30: Ah, so this is the main upgrade documented in this PR. Ok thank you, I will point the RN bullet to it.

ConnorLand: @jeffersonRibeiro thanks for your explanation in the comment above.

ConnorLand (approved):

ConnorLand:
Basically customers have more options with how their attribute based expressions can refer to snippet parameters and data entities. The docs that discuss data entities and snippets were mostly already correct, except some of the small changes as made in this PR. 

This is a minor change in the product that we do want to let customers know about, but it does not require substantive new docs. Existing documentation on entities/snippets still holds true. So we will just craft a bullet for the 9.24 RNs, and let that bullet be the main way in docs customers learn about this small improvement.

tamelbrun:
> This is a minor change in the product that we do want to let customers know about, but it does not require substantive new docs. Existing documentation on entities/snippets still holds true. So we will just craft a bullet for the 9.24 RNs, and let that bullet be the main way in docs customers learn about this small improvement.

@ConnorLand The "Selecting Attributes" section under "Improvements" already contains the RNs for the related stores, so this additional bullet point is not necessary.

ConnorLand:
> > This is a minor change in the product that we do want to let customers know about, but it does not require substantive new docs. Existing documentation on entities/snippets still holds true. So we will just craft a bullet for the 9.24 RNs, and let that bullet be the main way in docs customers learn about this small improvement.
> 
> @ConnorLand The "Selecting Attributes" section under "Improvements" already contains the RNs for the related stores, so this additional bullet point is not necessary.

Ah I see, thank you.

PR #5782, opened 2023-03-21
Rename Webhooks and Deployment Webhooks

MarkvanMents:
Since the deployment webhooks will be the default webhooks, reverse this current naming to avoid any confusion for users:

"webhooks" --> webhooks for stories & Sprints
"Mendix deployment webhooks" --> webhooks

### DIFF
--- a/content/en/docs/apidocs-mxsdk/apidocs/build-api.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/build-api.md
@@ -15,7 +15,7 @@ The Build API only works for apps which are deployed to the Mendix Cloud.
 
 The Build API allows you to manage deployment packages and create new deployment packages using our build server. You will need the information from the [Teamserver API](/apidocs-mxsdk/apidocs/team-server-api/) as input for these API calls. You will also need to provide authentication for each call; this is described in [Authentication](/apidocs-mxsdk/apidocs/authentication/).
 
-You can use deployment webhooks to trigger CI/CD pipelines which use this API. These are described in [Mendix Deployment Webhooks](/developerportal/deploy/webhooks/).
+You can use webhooks to trigger CI/CD pipelines which use this API. These are described in [Webhooks](/developerportal/deploy/webhooks/).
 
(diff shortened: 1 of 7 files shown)


Sakshi-Dhakad:
@MarkvanMents - Thanks for incorporating the suggested changes, the edits look good! The only over-arching comment I have is that in some places we refer to the 'new' webhooks as 'deployment' webhooks, however in the future, as more events get added to the list - e.g. webhook event for alerts, backup download, etc. then this becomes a place for all things in the app context. Even stories, epics, sprints can be events here. So if possible, I would suggest not using "deployment webhooks" at all as a terminology, otherwise you will have to change it in the future.

MarkvanMents:
> @MarkvanMents - Thanks for incorporating the suggested changes, the edits look good! The only over-arching comment I have is that in some places we refer to the 'new' webhooks as 'deployment' webhooks, however in the future, as more events get added to the list - e.g. webhook event for alerts, backup download, etc. then this becomes a place for all things in the app context. Even stories, epics, sprints can be events here. So if possible, I would suggest not using "deployment webhooks" at all as a terminology, otherwise you will have to change it in the future.

Thanks for the feedback @Sakshi-Dhakad. I'd like to drop the term "deployment", but I have included that in places where the existing webhooks for stories and Sprints are mentioned to try to make a distinction. Otherwise, I think the customers might get confused. I hope that once we introduce more webhooks, these bits of documentation will be retired as customers move to the new webhooks, so I won't have to make any changes other than removing the documentation.

If you can think of a better description then let me know. I don't really want to use "new webhooks" because in 6 months they aren't new any more.

If you are otherwise happy, I'll merge this tomorrow (23 March).

Sakshi-Dhakad:
@MarkvanMents - I understand your reasoning Mark, let's change it when additional webhook events are introduced.

Sakshi-Dhakad:
Good to be published from my point of view 👍

MarkvanMents:
No other comments.  Publishing.

PR #5834, opened 2023-03-30
Removing Leonardo doc as the connector has been retired

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Original file backed up and removed.

PR #5837, opened 2023-04-03
Update SAML documentation for 3.5

MarkvanMents:
Version 3.5 allows encryption settings to be set for each IdP.


content/en/docs/appstore/modules/saml.md
```diff
@@ -499,6 +465,51 @@ Requesting user attributes at the SAML IdP is only available in the following ve
 * v3.3.0/v3.3.1 and above for Mendix version 9
 * v2.3.0 and above for Mendix version 8
 
+## 4.6 Encryption Settings{#encryption-settings}
+
+* **Enable better security for app** (or *Use encryption*) – This setting controls the encryption and signing of messages being exchanged between your app (as an SP) and the IdP. This is in addition to the encryption provided by using a secure HTTPS connection. For security and privacy reasons it is enabled by default. Limitations in your IdP may be a reason to disable it, but this should typically not be the case.
```
JaapF: the word "better" is actually not to the point - let's just say: "Enable security for app".
Without this enabled, the security is actually insufficient. The basic security (i.e. validating signature on assertion) is not in place.  So actually disabling it should currently not be an option.
I'll create a seperat ticket.

All of tis is based on recent PEN-test finding, for which I'll create a separate ticket.

So for now: remove the word "better".

JaapF: I suggest to add the following:
"When using the POST binding ensure the security/encryption settings is enabled. 
When using the artifact binding on the responses, it could be an option to disable the security/encryption setting but it is not recommended."
@MarkvanMents

MarkvanMents: @JaapF ,
I can't remove "better" it is part of the UX:
image

JaapF: @MarkvanMents understood and agreed about not removing the word 'better'.
Can you add the guidance for POST and artifact binding that I suggested, please?

content/en/docs/appstore/modules/saml.md
JaapF, line 1: Hi @MarkvanMents 

I'd like to add a few changes to the PR (although not a showstopper for the release process):
In the 'Other features' section I'd like to add: 
"The SAML module allows to have a SSO connection with multiple SAML IdPs. For each IdP it is possible to upload a keypair."

In the limitations section I'd like to add:
"When using SSO connections with multiple IdP's, the SAML EntityID for your app will be shared between both IdPs. The SAML module does not allow to configure separate EntityID's for each of your connected IdP's."




In the 'known limitations' section I'd like to cla

MarkvanMents: Added in this commit: https://github.com/mendix/docs/pull/5837/commits/049f35908992ecfc46957b1171b69af5db1d6758

MarkvanMents:
3.5 published on 27 April.

PR #5844, opened 2023-04-05
Add release note about cookies for 9.24

atennapel:


content/en/docs/releasenotes/studio-pro/9/9.24.md
```diff
@@ -167,3 +167,4 @@ In the [Runtime API](/apidocs-mxsdk/apidocs/runtime-api/), we did the following:
 ### Breaking Changes
 
 * We removed support for the `FileID` parameter in the File API.
+* We now use a stricter parser for cookies. Make sure your cookies conform to the RFC-6265 standard.
```
grootjans: I would change the wording to:
Due to a regular update of our Jetty Server component, cookies are now correctly parsed according to the RFC-6265 standard. This can cause incorrectly formatted cookies to no longer be accepted.

atennapel: Yes better. Updated.

PR #5853, opened 2023-04-07
Add deprecation notices to widgets/modules

ConnorLand:
Next, I will check with SMEs for alternatives to link at the top of these docs.


ConnorLand:
I have the info I need and will resolve this soon.

PR #5871, opened 2023-04-12
Align docs for feedback module to be in line with the upgrade to v1.1.0

erolsmsr58:


Adam-Dupaski:
@Luyao-Zhang-1 @ConnorLand ownership of this doc may not be totally clear right now, but because it's a module and it's from C&C, I think it should be with @Luyao-Zhang-1 - please let me know if you have different thoughts here!

Luyao-Zhang-1:
> @Luyao-Zhang-1 @ConnorLand ownership of this doc may not be totally clear right now, but because it's a module and it's from C&C, I think it should be with @Luyao-Zhang-1 - please let me know if you have different thoughts here!

I will handle this update.

content/en/docs/appstore/modules/mendix-feedback.md
```diff
@@ -45,73 +45,85 @@ The Mendix Feedback module is easy to set up and automatically attaches addition
 
 ### 1.3 Limitations
 
-* This module can only be used in combination with Mendix apps
-* Compatible with Studio Pro v.9.12.2 or higher
-* [Atlas Core](https://marketplace.mendix.com/link/component/117187) is required to apply the styling 
-* In native mobile apps, some of the feedback metadata such as username, email address, and document name will be hard-coded, as they cannot be retrieved dynamically (to address this you can use the [Native Feedback widget](/appstore/modules/native-mobile-resources/) instead, located in [Native Mobile Resources](https://marketplace.mendix.com/link/component/109513))
+* This module can only be used in combination with Mendix apps.
+* This module is compatible with Studio Pro v.9.12.2 or higher.
```
erolsmsr58: The minimum compatible version should be v.9.18.5.

PR #5878, opened 2023-04-13
Update the API reference link

Luyao-Zhang-1:


Luyao-Zhang-1:
@MarkvanMents  FYI: the current link to the API Reference is broken. The team provides a new link which works, but the URL (https://d102io3ppw18a6.cloudfront.net/index/v2.1/apidoc/index.html) can be improved. The team is working on it. I will first update the link to make it work. When the team provides a better URL, I will replace it.

MarkvanMents:
Thanks for keeping me in the loop, @Luyao-Zhang-1

PR #5880, opened 2023-04-13
[UIA-433] Remove Mendix SSO section for Administration module

allard-mx:
We're planning to release v1.4.0 and v3.0.0 of the Administration module tomorrow. This update removes the adapter microflows that were previously included for Mendix SSO. 

This PR removes the related dependency + steps in the module docs.

### DIFF
--- a/content/en/docs/appstore/modules/administration.md
+++ b/content/en/docs/appstore/modules/administration.md
@@ -23,18 +23,7 @@ The [Administration](https://marketplace.mendix.com/link/component/23513) module
 ## 1.2 Dependencies
 
 * [Atlas UI Resources](https://marketplace.mendix.com/link/component/104730)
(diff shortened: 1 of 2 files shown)


MarkvanMents:
@allard-mx 
Are previous versions of the administration module still available - and still dependent on the Mendix SSO module (for example the versions in starter apps for Mendix 8)?
In this case, we should mention that there is a dependency for versions below 1.4.0 and 3.0.0, rather than just removing the information entirely?

MxMurshed:
@MarkvanMents Currently starter apps for Mx8 doesn't start with Mendix SSO, in fact, they start with even an older version of the Administration module even before Mendix SSO adapter microflows were introduced.

MxMurshed:
We have released the updated modules to the marketplace here

PR #5891, opened 2023-04-14
Mx Model Reflection

RCatersels:


Luyao-Zhang-1:
Hi @RCatersels thanks for sending us this PR. 

I discussed this with the developer. The page that's referenced in that paragraph (`MxObjects_Overview`) has a button that invokes the model reflection which does the same thing. It's just that if you want to sync on app startup you can use the `ASu_CheckMetamodel`. Because of this reason, we think it's not necessary to add this text here.

RCatersels:
But the problem is now that nowhere in the documentation this microflow is mentioned. So users do not know that it is a best practise to add this microflow to your after startup and we get forum questions on how to trigger this refresh automaticly. Users should be made aware on how to trigger this update without them each time pressing the button.

Luyao-Zhang-1:
@RCatersels Thanks for the input. Then I agree it makes sense to add info about the microflow in the doc. I added a paragraph about this best practice in the **Configuration** section. Could you take a look? Please let me know how you think about it.

Luyao-Zhang-1:
I checked this paragraph with the developer. I will merge the PR. 
@RCatersels If you have more comments, please let me know.

PR #5914, opened 2023-04-19
Update task-queue.md

emirmx:
Remove unsupported use case for `ProcessedQueueTask`.


MarkvanMents (approved):
@emirmx 
Merging immediately as it is documentation change and not changes in functionality.
Mark

PR #5918, opened 2023-04-20
Overwrite refguide and howto with mx10 directories

ConnorLand:


ConnorLand:
We are leaving this branch just in case, but in all probability it can be deleted soon.

PR #5923, opened 2023-04-20
updates to prerequisites and installation

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@MarkvanMents I have some other ideas (regarding the idp configuration) but those will require more investigation, so we can look at them together when you have some time.

katarzyna-koltun-mx:
@MarkvanMents I've reassigned the PR to you but let me know if you'd like me to pick it up.

PR #5927, opened 2023-04-21
Removed message about preferIPv4Stack

joostverhoog:
The Java 7 issue about `java.net.preferIPv4Stack` does not apply to Studio Pro 10, so this change removes the note about that.


MarkvanMents (approved):
Thanks @joostverhoog

MarkvanMents:
Mx 10 released

PR #5946, opened 2023-04-26
Update to accomodate asynchronous invocation of lambda functions

brfreekmx:
Please wait with releasing until we released the connector! :) I update @katarzyna-koltun-mx when this is done.


katarzyna-koltun-mx:
Reviewed and waiting for the go ahead from @brfreekmx.

PR #5959, opened 2023-04-26
Reproduce PR 5915

MariaShaposhnikova:

PR #5961, opened 2023-04-26
Revert "Reproduce PR 5915"

MariaShaposhnikova:
Reverts mendix/docs#5959


MariaShaposhnikova:
Did not see that this was for future release.

PR #5974, opened 2023-05-01
Update screenshots for the new Properties pane in Mx 10 (Microflows a…

Yiyun333:
…nd Nanoflows)


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/workflow-call.md
```diff
@@ -45,7 +45,11 @@ The workflow that is called by this activity.
 
 The object which you want to use as a context. It should be of the entity type set in the [Workflow](#workflow) property.
 
-### 3.3 Variable Name, Object Name, or List Name {#name}
+### 3.3 Use Return Value
```
MariaShaposhnikova, line 48: @timvandervoornmx , Yiyun noticed there is a setting that was missing in the documentation. I added it, can you please verify that description is correct?

timvandervoornmx: Ah yes, good find. The activity can only have one type of return value: a workflow object, so I suggest the following change:

 If **Use return value** is set to *Yes*, you will need to fill in the [variable name](https://docs.mendix.com/refguide/microflow-call/#name) of the workflow object returned by the activity.

PR #5976, opened 2023-05-01
Make Mendix Version Numeric in Algolia

MarkvanMents:


MarkvanMents:
Checked that all existing sorting is retained.

PR #5990, opened 2023-05-02
Validate webhook in Mendix

MarkvanMents:


content/en/docs/developerportal/deploy/mendix-cloud-deploy/webhooks.md
```diff
@@ -114,6 +114,22 @@ When verifying your webhook signature, bear the following in mind:
 * You should use a constant-time string comparison method in order to prevent timing attacks.
 * You should also verify that **webhook-timestamp** is within your tolerance for the current system time, to ensure you are not receiving timestamp attacks where old webhook payloads are being sent to your endpoint.
 
+### 3.1.1 Verifying Your Webhook Using Mendix
+
+You can use Mendix to verify your webhook, using functions available in the [Community Commons](/appstore/modules/community-commons-function-library/) module. To do this, follow the steps below:
+
+1. Ensure that your app contains the Community Commons module.
```
feddovanede, line 121: The numbering is not correct, it seems? Or will that be updated automatically later?

content/en/docs/developerportal/deploy/mendix-cloud-deploy/webhooks.md
```diff
@@ -114,6 +114,22 @@ When verifying your webhook signature, bear the following in mind:
...
+1. Retrieve a list of `System.HttpHeader` objects over the association `$HttpRequest/HttpHeaders` using the HttpRequest from the webhook.
+1. Obtain the `webhook-id`, `webhook-timestamp`, and `webhook-signature` by using a *Find by expression* **List operation** on the list of `System.HttpHeader` objects. You can use an expression like `toLowerCase($currentObject/Key) = toLowerCase("webhook-id")` and extract the `/Value` attribute.
```
feddovanede: Obtain the [...] _`System.HttpHeader` objects_ by [...]? Or _headers_?

feddovanede (commented):
I hope this is clear enough for users :) I would not know how to make it more clear.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/webhooks.md
MarkvanMents: Hugo (markdown?) is clever enough to number things for you if you put all 1's. And Hugo will renumber them anyway if you get the numbering wrong. So I've decided to let the computer do that for me - I have enough other things to worry about.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/webhooks.md
MarkvanMents: Yes, perhaps that would read better?
And I don't have to lowercase a string which is already in lowercase 😀
So:
```suggestion
1. Obtain the `webhook-id`, `webhook-timestamp`, and `webhook-signature` by using a *Find by expression* **List operation** on the list of headers. You can use an expression like `toLowerCase($currentObject/Key) = "webhook-id")` and extract the `/Value` attribute.
```

(Feel free to commit this if you are happy with the suggestion.

MarkvanMents (commented):
Can you commit my suggestion if you like it and let me know it is OK to publish.
Other than giving someone a module containing the code, I don't know what else I can do.
Unfortunately, I can't really link to your Marketplace module as it is not Platform supported and I don't want support calls on it.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/webhooks.md
feddovanede: I meant the following (should have created a suggestion right away).
```suggestion
1. Obtain the `webhook-id`, `webhook-timestamp`, and `webhook-signature` headers by using a *Find by expression* **List operation** on the list of `System.HttpHeader` objects. You can use an expression like `toLowerCase($currentObject/Key) = 'webhook-id'` and extract the `/Value` attribute.
```

PR #5993, opened 2023-05-03
Add Sign In button to MxDock UPDATE BEFORE MERGING

MarkvanMents:


MarkvanMents:
Cannot see the benefit to docs of having a Sign-In button as the experience for the user will not change.
Will close this for now - if there are future MxDock enhancements, we will pick them up and test them.

PR #5995, opened 2023-05-03
Add a new microflow how-to (retrieve the current user)

Yiyun333:


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,52 @@
...
+
+The following are facts about user management in a Mendix app:
+
+* The currently logged-in user is always available in microflow calls and XPath constraints, and it can be retrieved in a microflow for subsequent querying or manipulation.
+* All application users/accounts are a specialization of `System.User`, and very often and by default, it is `Administration.Account`.
+* In Mendix Studio Pro 10, the [Administration](/appstore/modules/administration/) module is added by default . It allows you to manage local accounts and view app statistics.
+
+This document teaches you how to do the following:
+
+* Retrieve the current user in a microflow using the **Administration** module in Mendix
```
cryy-bit: Using the Administration Module? - A little bit weird terminology here. Shouldn't this be "Entity from Administration Module"?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,52 @@
...
+* Install the latest version of Studio Pro
+* Read about the [Administration](/appstore/modules/administration/) module in the *Studio Pro Guide*
+* Understand the basics about [microflows](/refguide/microflows/)
+
+## 3 Retrieving the Current User
+
+To retrieve the current user using the **Administration** module, do the following:
+
+1. Open your microflow editor.
+2. Navigate to the **Toolbox** and search for the **Retrieve** activity.
```
cryy-bit, line 35: I think there are 2 options here. You can either search through the toolbox for retrieve activity or use the topbar of microflow editor to select "Action activity" then from there select action type to be "Retrieve activity". 

image

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,52 @@
...
+3. Drag the **Retrieve** activity into your microflow.
+4. Double-click the **Retrieve** activity to open its properties.
+5. Choose **From database** as the **Source**.
+6. Click **Select...** for **Entity** and do the following:
+    * In the **Select Entity** dialog box, go to **Marketplace modules** > **Administration** > **Account**, and select **Account** as the **Entity**
+7. Choose **First** as the **Range**.
```
cryy-bit: Do you think it's clear to the audience as to why we select "First" as the range? Not sure if it's required to explicitly explain that but because when Id = $CurrentUser there can only be 1 Account object.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,52 @@
...
+tags: ["microflow", "current user", "retrieve"]
+---
+
+## 1 Introduction
+
+In order to implement a certain logic, You often need to identify who the current user using the application is.
+
+The following are facts about user management in a Mendix app:
+
+* The currently logged-in user is always available in microflow calls and XPath constraints, and it can be retrieved in a microflow for subsequent querying or manipulation.
```
cryy-bit: Specifically System.User is always available not to be confused with other specialisation of System.user eg. Account, MendixSSOUser, etc.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: How about "using the **Account** entity from the **Administration** module"?

cryy-bit: I think that sounds better

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Yes, I think it's very useful to the explanation there!

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: The other specializations of System.user are not always available through microflow calls? Like the Account we are talking about in the how-to?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Nice to know the other option! Our docs always follow the first option, so I'll keep the one I used in the doc for now.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Solved!

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Explanation added: 8a370ad0d215d887c0fbb829f5c8b498b754e4e5

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
cryy-bit: No, specialisations themselves do not appear in microflows/xpath automatically. These specialisations need to be retrieved explicitly.

Yiyun333: Oh, I see what you mean here now. Are the first two facts about user managerment useful then in this how-to? We are not retrieving the current user via System_user in this how-to, so I am not sure whether providing these facts here are more confusing than helful. What do you think?

cryy-bit: I think they are useful to know because even though we are not retrieving via System.User we are still using it's ID as an identifier of which account to retrieve. Also, it's very good to know that Account entity takes the specialisation of System.User otherwise it will be confusing as to why Id = $CurrentUser works.

Yiyun333: I see! Thanks for the further explanation. I'll keep them in the doc!

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,53 @@
...
+linktitle: "Retrieving Current User with a Microflow"
+url: /refguide/retrieving-current-user-with-a-microflow/
+weight: 6
+description: "Describes how to retrieve the current user of your application in Studio Pro with a microflow."
+tags: ["microflow", "current user", "retrieve"]
+---
+
+## 1 Introduction
+
+In order to implement a certain logic, You often need to identify who the current user using the application is.
```
OlavJansen: 'You' should be lowercase. I would say you often want to identify the user which is currently logged in. As there is also the possibility the have your app available for anonymous users, those will not have an account and therefore will not be found when doing a retrieve based on current user.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,53 @@
...
+
+The following are facts about user management in a Mendix app:
+
+* The currently logged-in user is always available in [microflow calls](/refguide/microflows/) and [XPath constraints](/refguide/xpath-constraints/), and it can be retrieved in a microflow for subsequent querying or manipulation.
+* All application users/accounts are a specialization of `System.User`, and very often and by default, it is `Administration.Account`.
+* In Mendix Studio Pro 10, the [Administration](/appstore/modules/administration/) module is added by default. It allows you to manage local accounts and view app statistics.
```
OlavJansen: This is not specific for Mx10, as long as I can remember (10 years) the module is added by default.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,53 @@
...
+
+* Install the latest version of Studio Pro
+* Read about the [Administration](/appstore/modules/administration/) module in the *Studio Pro Guide*
+* Understand the basics about [microflows](/refguide/microflows/)
+
+## 3 Retrieving the Current User
+
+To retrieve the current user using the **Account** entity from the **Administration** module, do the following:
+
+1. Open your microflow editor.
```
OlavJansen: I'm not sure everybody is aware you build microflows with the microflow editor. Most developer just see it as Studio pro (eventhough technically there is a page and microflow editor). So I would say something more like; Open an existing microflow or create a new one. Something along these lines.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
```diff
@@ -0,0 +1,53 @@
...
+2. Navigate to the **Toolbox** and search for the **Retrieve** activity.
+3. Drag the **Retrieve** activity into your microflow.
+4. Double-click the **Retrieve** activity to open its properties.
+5. Choose **From database** as the **Source**.
+6. Click **Select...** for **Entity**.
+7. In the **Select Entity** dialog box, go to **Marketplace modules** > **Administration**, and select **Account**.
+8. Choose **First** as the **Range**. This is because when retrieving the current user using `[id = $currentUser]`, only one **Account** object will be returned.
```
OlavJansen: Technically is a bit different. You can still leave the range at 'All' and use [id = $currentUser]. The result of this will be a list with 1 object. This means it needs to be treated as a list in the rest of the microflow. If you use first as range, it will return 1 object. So the xpath constraint doesn't impact the return type (list or object) but the range does. In this case you're always looking for 1 object and there use 'first' as range.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Good reasoning! Thanks for bringing this up.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/retrieving-current-user-with-a-microflow.md
Yiyun333: Good catch! Thanks for correcting the reasoning line here and making it more logical.

Yiyun333:
@OlavJansen Thanks a lot for the comments. They are very clear and with good reasoning behind. I've addressed all your comments in this commit: https://github.com/mendix/docs/pull/5995/commits/6f5d4cc6931b1ac3f46dc86aea5a4f7cd23708e7.

Please let me know if you have any further feedback.

PR #6002, opened 2023-05-03
Kk aws twinmaker

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@CasperSpronk27 It took me longer than expected, but I've finished cleaning the file. I want to make some further updates to the structure, but it should be good to publish as-is in the meantime. Can you take a look and let me know if there are any corrections? If it looks good, I'll go ahead and publish it. After that, I'll do some more updates in a new PR.

PR #6019, opened 2023-05-08
Quickstart updates

Adam-Dupaski:


content/en/docs/refguide/quickstart-guide/_index.md
```diff
@@ -6,13 +6,15 @@ description: "Presents tutorials on how to build an app in Studio Pro quickly an
 tags: ["hello world", "microflows", "widgets", "app", "nanoflow", "app development"]
 ---
 
-## 1 Introduction 
+Welcome to the Mendix quickstart guide for building an app! These tutorials cover the essentials of Mendix development:
```
ConnorLand, line 9: Do we have hard and fast rules for how parent docs should use **Introduction** and **Documents in this Category** headers? 

I recognize that this doc has a low wordcount, but I was just curious if it should have a header or two?

Adam-Dupaski: No hard rules. But with only 1 section, generally no need to have a separate heading, esp if all it will say is "Introduction." With something like this, I think we want to get the user started on the first tutorial as quickly as possible, so we should try to limit the overall text in general.

PR #6043, opened 2023-05-12
[WTF-1508] Add extra information on dynamic classes on grid columns

mlaponder:
Dynamic classes for grid columns are only applied to the col element on not on the individual cells. This doc update is to make that more clear.


ConnorLand:
Ok looks pretty straightforward. Thanks @mlaponder !

ConnorLand (approved):

ConnorLand:
@mlaponder did a support ticket inspire this doc update?

PR #6058, opened 2023-05-17
Add docs for TrackHitCount option

WimJongeneel:


content/en/docs/appstore/modules/deep-link.md
```diff
@@ -99,6 +99,7 @@ On the **Advanced** tab, there are these settings:
 * **Process an argument as an Object** – This is deprecated.
 * **Process an argument as a String** – This is deprecated.
 * **Alternative Index Page** – If selected, the default index location (`index.html`) and the **DeepLink.IndexPage** constant will be overridden by this value. This is useful for theme-related use cases, for example, `index-dark.html`.
+* **Track hit count** - If selected, Deeplink will track the amount of hits on this link in the `HitCount` attribute. This tracking can be disabled if performance issues occur.
```
MarkvanMents:
```suggestion
* **Track hit count** - If selected, Deeplink will track the number of hits on this link in the `HitCount` attribute. This tracking can be disabled if performance issues occur.
```

content/en/docs/appstore/modules/deep-link.md
```diff
@@ -99,6 +99,7 @@ On the **Advanced** tab, there are these settings:
 * **Process an argument as an Object** – This is deprecated.
 * **Process an argument as a String** – This is deprecated.
 * **Alternative Index Page** – If selected, the default index location (`index.html`) and the **DeepLink.IndexPage** constant will be overridden by this value. This is useful for theme-related use cases, for example, `index-dark.html`.
+* **Track hit count** - If selected, Deeplink will track the number of hits on this link in the `HitCount` attribute. This tracking can be disabled if performance issues occur.
```
Robinfr: Also needs to be updated in Mx9 docs.

Robinfr (changes requested):

content/en/docs/appstore/modules/deep-link.md
WimJongeneel: AFAIK `appstore` docs don't have a Mendix 9 version, only the `refguide` docs (I also don't find anything like this document if I do a full-text search in the repro).

Luyao-Zhang-1: @Robinfr , what @WimJongeneel said is correct. We only document the latest version of a Marketplace component. We may add a note about version-specific information if that's really necessary. But for this update, it's not needed to mention that this property is for version 10 and 9.

Robinfr: Oh I was not aware of this. In that case, maybe we should mention this is only an option in DeepLink 9 and 10?

WimJongeneel: I've added a note about it

Luyao-Zhang-1: @WimJongeneel  Thanks. Looks good. I edited the text a bit. Should we merge the PR now?

PR #6066, opened 2023-05-19
Wait for notification

ishan1729:
Needs review from `RnD`.

Targeted for `10-beta-2`.


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
```diff
@@ -47,11 +47,13 @@ The workflow instance that will be notified by this activity.
 
 ### 3.2 Activity
 
-The [WaitForNotification](/refguide/wait-for-notification) workflow activity on which the workflow instance is suspended on and you would like to notify so that the workflow can continue further.
+The [Wait for notification](/refguide/wait-for-notification) workflow activity which the workflow instance is suspended on and you would like to notify so that the workflow can continue further.
+
+### 3.3 Variable Name
```
Yiyun333, line 54: We normally directly explain the more specific properties under the Output section, so the **Variable name** here. I added some extra descriptions, but please verify whether the added descriptions are correct.

Yiyun333:
Hi @ishan1729, I've reviewed the language, style, and the screenshots of the PR. I also left one comment above. My changes are in [this](https://github.com/mendix/docs/pull/6066/commits/b5a4e3e51258129c9c275115e9b82ce98ced9944) commit. 

Please verify if there is anything that is not correct from a technical point of view. Thanks!

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
```diff
@@ -0,0 +1,60 @@
...
+{{% /alert %}}
+
+## 1 Introduction
+
+The **Notify workflow** activity can be used to notify a [workflow](/refguide/workflows/) that is suspended on the [Wait for notification](/refguide/wait-for-notification) workflow activity.
+
+{{< figure src="/attachments/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow/notify-workflow.jpg" width="200" >}}
+
+{{% alert color="warning" %}}
+When you try to notify a workflow which is already `Completed` or `Aborted`, it will result in a Runtime error. For information on how to handle the error, see [Error Handling in Microflows](/refguide/error-handling-in-microflows/).
```
hugomendix, line 22: I know you explain the return value further on, but it seems a bit weird to me to have this sentence and not explain the return value. Maybe you could add the return value text here and refer to it from the variable name section.
Or maybe put everything related to execution in a separate section.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
```diff
@@ -0,0 +1,60 @@
...
+The **Notify workflow** properties pane consists of the following sections:
+
+* [Action](#action)
+* [Common](#common)
+
+## 3 Action Section {#action}
+
+The **Action** section of the properties pane shows the action associated with this activity.
+
+You can open a dialog box to configure this action by clicking the ellipsis (**…**) next to the action.
```
hugomendix, line 42: Is this something we generally explain? Or should that be common knowledge?
IMHO it's better to explain the meaning of the property than how to open it.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
```diff
@@ -0,0 +1,60 @@
...
+
+You can also open the dialog box by double-clicking the activity in the microflow or right-clicking the activity and selecting **Properties**.
+
+### 3.1 Workflow Object {#workflow}
+
+The workflow instance that will be notified by this activity.
+
+### 3.2 Activity
+
+The [Wait for notification](/refguide/wait-for-notification) workflow activity which the workflow instance is suspended on and you would like to notify so that the workflow can continue further.
```
hugomendix: Maybe add here that it is the caption of the Wait for notification activity?

hugomendix (changes requested):

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
ishan1729: Yes, you do have a good point. I basically took inspiration from the https://docs.mendix.com/refguide/unlock-workflow/#action document.

MariaShaposhnikova: @hugomendix , we explain it for all microflow activities since the list of properties in the Properties pane does not match the dialog box. We can delete this sentence once (if) this is fixed in the product.
We also have a general phase about the section: "The **Action** section of the properties pane shows the action associated with this activity." Do you have a suggestion on what we can add to this description?
Please note this is a section description, we describe what properties in this section do below.

hugomendix: OK, I understand your point so let's keep it as is then.

As an answer to your question, I would state something like:

> You can change the type of action by clicking the ellipsis next to the action.

Or possibly:

> You can change the type of action from Notify Workflow to something else by clicking the ellipsis next to the action.

Yiyun333: @MariaShaposhnikova 

> OK, I understand your point so let's keep it as is then.
> 
> As an answer to your question, I would state something like:
> 
> > You can change the type of action by clicking the ellipsis next to the action.
> 
> Or possibly:
> 
> > You can change the type of action from Notify Workflow to something else by clicking the ellipsis next to the action.

@MariaShaposhnikova Regarding @hugomendix 's suggestion here, I think perhaps we need to change this sentence throughout our docs, because indeed, clicking the ellipsis only shows a dialog box that allows you to select another type of action instead of configuring a chosen action now. We can also decide to leave this sentence out if our purpose is just to tell customers how to reach a dialog box that allows for all the configurations. 

What do you think? I can pick this up if you agree.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
ishan1729: Yes, looks good to me.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
Yiyun333: My understanding is that this is not only the caption (if so, you can change the caption there too), but the wait for notification activity you select to get notified.

ishan1729: How about the following text?

```The WaitForNotification workflow activity caption on which the workflow instance is suspended on and you would like to notify so that the workflow can continue further.```

ishan1729: > but the wait for notification activity you select to get notified.

@Yiyun333 I am having second thoughts on my newly proposed text now. You are right about this because for a user it shouldn't matter as all they get to see is a dropdown with a list of wait for notification activities.

hugomendix: That sounds wrong to me: it's now the caption on which the workflow is suspended. I would phrase it as:

> The (caption of the) WaitForNotification workflow activity on which the workflow instance is suspended, which you would like to notify so that the workflow can continue.

ishan1729: > My understanding is that this is not only the caption (if so, you can change the caption there too)

If you change the caption then automatically the newly captioned activity is used in the NotifyWorkflow activity.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
ishan1729: Done.

I felt adding a new section wasn't worth it considering the amount of information.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
ishan1729: Fixed an omission for now in the latest commit.

ishan1729: @Yiyun333 what do you think about @hugomendix proposal?

Yiyun333: @ishan1729 I still think adding "caption" here is a bit misleading because you do not select a caption here but the activity as a whole.  I will phrase the sentence like this:
The **wait for notification** workflow activity that pauses the workflow instance and needs to be notified so that the workflow can continue.

Yiyun333: Or maybe @hugomendix can explain a little bit more about why caption is needed here?

hugomendix: Indeed, you do not select a caption, you select a **wait for notification** activity. However, the text that will show up here is the caption of the selected activity. Since it has also a name (which is not shown), I thought that might be relevant.

Yiyun333: @hugomendix Thanks for the further clarification. Now it makes more sense to me why you want to mention the caption here. I just updated the text in [this](https://github.com/mendix/docs/pull/6066/commits/f68d196a14a76b57393f9e1369a842a6a3001a8d) commit. I added a cross-reference to the caption so that customers can also see some further explanation of it.

@ishan1729 Please let me know if you agree with the update.

ishan1729: > The wait for notification workflow activity that pauses the workflow instance and needs to be notified so that the workflow can continue.

Looks good to me. Only one minor thing is that I would not use the word `pause` instead I will stick to suspend because we also have a pause workflow functionality and it can be confusing to some users.

Yiyun333: @ishan1729 No problem. I'll change the word "pause" to "suspends".

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/notify-workflow.md
MariaShaposhnikova: @hugomendix , I was told before that opening a dialog where you can choose other microflow activities when you click the Action section was a bug to be fixed. It used to open a dialog box with more properties before and we still describe it as an expected behaviour. I will double-check this and get back to you.
CC: @Yiyun333

MariaShaposhnikova: @hugomendix , @Yiyun333 , I double-checked and there is yet no clarity what behavior stays (old or new one). I suggest to leave it as is and potentially update it for a later version, for example, for GA.

Yiyun333: Thanks, @MariaShaposhnikova for the double-check.
@hugomendix I will mark this conversation as resolved for now, but we will keep an eye on whether the current behavior will be fixed, kept, or improved.

Yiyun333 (approved):

Yiyun333:
I've just approved the PR and I will make sure to merge the PR when 10.0.0 Beta 2 is released.

PR #6134, opened 2023-06-06
Filtering capability for runtime metrics to APMs

ekremsekerci:
8 June
Filtering capability for ingestion of runtime metrics to APM vendors are introduced


content/en/docs/developerportal/operate/metrics/monitoring-with-apm/_index.md
```diff
@@ -285,6 +285,35 @@ You pass the configuration to the metrics agent by adding a *Custom Runtime Sett
 | mx.rest.time | The time to handle a REST request |
 | mx.soap.time | The time to handle a SOAP request |
 
-## 6 Read More
+## 6 Filtering Metrics Ingestion
```
MarkvanMents: We call this **Filtering Metrics Ingestion**, but below we say *To filter the **emission** of metrics*.
I think we should be consistent.
I think we are stopping the metrics being produced, not being ingested, so *emission* seems better. Is this the technical term that customers will understand, or could we use something like *production* or *to prevent metrics being sent to APM vendors*?

MarkvanMents (commented):
@ekremsekerci 
Thanks for the PR, I have one comment about consistent use of language.

content/en/docs/developerportal/operate/metrics/monitoring-with-apm/_index.md
```diff
@@ -285,6 +285,35 @@ You pass the configuration to the metrics agent by adding a *Custom Runtime Sett
...
+
+For example, to deny metrics starting with jetty or mx.runtime, the environment variable should be set to `jetty,mx.runtime`.
+
+To deny all metrics, the environment variable should be set without any value (also see note above about the database metrics).
+
+### 6.2 APM_METRICS_FILTER_ALLOW{#app-metrics-filter-allow}
+
+List of prefixes of the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this environment variable, so it makes sense to use this environment variable in combination with the `APM_METRICS_FILTER_DENY`.
+
+For example, to deny all metrics except the session metrics, `APM_METRICS_FILTER_DENY` should be used without any value, and `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`.
```
MarkvanMents: @ekremsekerci 
I've noticed that the examples given in this section are not mentioned in section 5 (Summary of Mendix Metrics). The list there does not include any metrics which begin `mx.runtime.stats` - it says that the core runtime metrics begin `jmx.com.mendix`

Can you check what is in section 5?

Thanks.

content/en/docs/developerportal/operate/metrics/monitoring-with-apm/_index.md
ekremsekerci: We are not stopping producing the metrics, we are just blocking the ingestion to third-party APMs. So ingestion sounds better in this case, I updated accordingly. Thank you for pointing it out 👍

content/en/docs/developerportal/operate/metrics/monitoring-with-apm/_index.md
ekremsekerci: The list in section 5 is very limited. I really don't know why we have this section with a very limited subset of the metrics we provide. And also most of the metrics here are only produced if the customer requests them through `METRICS_AGENT_CONFIG` so it may be a bit confusing to also use them as examples for the filtering functionality now. I just want to use metrics that we provide out-of-the-box in my examples. Luckily, we already have a detailed page explaining all the metrics and it's referenced at the end of section 3.1.

MarkvanMents: I think section 5 was perhaps there before the other document you mention. However, the document [Metrics](https://docs.mendix.com/refguide/metrics/#71-runtime-server-metrics) does not include these particular metrics.
Should I move all of section 5 into the Metrics document to avoid customers having the question that I did. If so, should they be in an existing section of the Metrics, or in a section of their own?

MarkvanMents (approved):
Thanks for the PR, @ekremsekerci .
Have sorted out the missing commit and removed the section as discussed.
Will merge immediately.

PR #6142, opened 2023-06-08
Microflow validation best practices

Robinfr:
Adding new documentation around adding validation for inputs in microflows for all Mx7+ versions.

### DIFF
--- a/content/en/docs/howto7/general/dev-best-practices.md
+++ b/content/en/docs/howto7/general/dev-best-practices.md
@@ -322,7 +322,7 @@ Adding validation checks is vital for ensuring that input data conforms to the e
 
 When building microflows, it is important to account for the potential occurrence of unexpected values. These could be empty values, or values outside the expected range or format. It is also important to ensure that read-only attributes only contain expected values.
 
-{{% alert color="info" %}}
+{{% alert color="warning" %}}
 We strongly recommend adding validation checks to all microflows inputs, including read-only attributes. 
 
 We also recommend avoiding storing intermediary values in attributes (such as, *TotalPrice*). Instead, calculate these values when needed to ensure you have the correct values.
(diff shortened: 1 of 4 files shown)


Yiyun333:
@Robinfr Thanks for sending us the PR. 

I've reviewed the language and style of the PR (sorry about the typo in the commit message. I was trying to correct it afterwards, but it seems to create even more issues, so I  just have to tolerate it :)). 

Please let me know if you agree with my changes in [this](https://github.com/mendix/docs/pull/6142/commits/47d89b167a85a8ebed1d5c9f5b2b5ae831a9f4b5) commit. Then I will go ahead and merge the PR.

Robinfr:
@Yiyun333 I agree with most. I'm only not sure about switching the alert box from info to warning. We do want to strongly recommend it. Do you think an info box is enough?

Yiyun333:
> @Yiyun333 I agree with most. I'm only not sure about switching the alert box from info to warning. We do want to strongly recommend it. Do you think an info box is enough?

@Robinfr Warning messages are usually used to call attention to some conditions that might cause problems if not resolved. Recommendations sound like the opposite of that even. In my opinion, info alert suits the vibe of recommendations better than warnings :) Please let me know if you do not agree.

Robinfr:
@Yiyun333 If you do not follow this recommendation, it can cause problems. I think it would be best to keep this as a warning.

Yiyun333:
> @Yiyun333 If you do not follow this recommendation, it can cause problems. I think it would be best to keep this as a warning.

@Robinfr Also makes sense. Then I'll change the style to warning. After that, if nothing else comes up, I'll go ahead and merge the PR.

Robinfr:
That is the only thing. 😄Op 13 jun. 2023 10:47 schreef Yiyun Liao ***@***.***>:

@Yiyun333 If you do not follow this recommendation, it can cause problems. I think it would be best to keep this as a warning.

@Robinfr Also makes sense. Then I'll change the style to warning. After that, if nothing else comes up, I'll go ahead and merge the PR.

PR #6151, opened 2023-06-09
Epics restructuring + Epics API + Release Notes

Luyao-Zhang-1:

PR #6161, opened 2023-06-12
Epics API

Luyao-Zhang-1:


Luyao-Zhang-1:
Move the content to another [PR](https://github.com/mendix/docs/pull/6151), since these two need to be released at the same time.

PR #6180, opened 2023-06-14
Doc update for setting an input parameter as the return value

Yiyun333:


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
```diff
@@ -10,18 +10,28 @@ tags: ["studio pro", "parameter", "microflow"]
...
-### 2.2 Data Type
+### 3.1 Data Type
 
 The data type of a parameter defines the type of the value that it expects. See [Data Types](/refguide/data-types/) for the possible data types.
 
 Default: *Object*
+
+### 3.2 Name
+
+**Name** refers to the value of the parameter.
```
Yiyun333: Shouldn't it be `refers to the name of the value of the parameter` or `the name of the parameter`?

ghost: It should indeed be the name of the parameter

ghost (commented):
Reference around name is wrong. everything else seems to be correct.

Yiyun333:
> Reference around name is wrong. everything else seems to be correct.

Thanks for the review! The description of the name is corrected now :)

PR #6181, opened 2023-06-15
Webhooks relnote and removal of beta tag

katarzyna-koltun-mx:


katarzyna-koltun-mx:
@Luyao-Zhang-1 Here's the PR that I mentioned.

PR #6186, opened 2023-06-15
Correct description of custom runtime setting SessionKeepAliveUpdatesInterval

tieniber:
per this discussion and linked thread in #mxperts

https://mendix.slack.com/archives/C3T260XGA/p1686322829829359


MarkvanMents (approved):
Thanks for updating this description Eric @tieniber . I see that it has already been checked by the development team so will merge immediately.

PR #6212, opened 2023-06-21
Multi-user task UI & decision methods update

Luyao-Zhang-1:


content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,326 @@ For example, you can assign a review task to multiple users:
...
-##### 2.4.1.2 Fallback Outcome {#fallback-outcome}
+###### 2.5.2.1.2 Fallback Outcome {#fallback-outcome}
 
-This outcome will be the final outcome of the task in the following cases:
+This outcome will be the final outcome of the multi-user task in the following cases:
 
-* If any of the users during consensus completion criteria selects a different outcome from the rest of the users
-* If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Required input** section:
+* If any of the users during consensus decision method selects a different outcome from the rest of the users
+* If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Participant input** section:
```
Yiyun333: Paul: This is not true I think. The WF goes to failed state. Pls check with Hugo and add a not about this runtime check at the Required Input property.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,326 @@ For example, you can assign a review task to multiple users:
...
+| **Most chosen** | When one of the outcomes gets the most votes (relative majority) and additional votes will not result in another outcome getting more votes, the task is competed with this outcome. |
+
+###### 2.5.2.3.2 Fallback outcome {#majority-fallback-outcome}
+
+This outcome will be the final outcome of the multi-user task in the following cases:
+
+* If it is determined that none of the remaining votes will result in a majority.
+* If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Participant input** section:
+
+For example, if the targeted user count is `10` and you set the absolute number to `20` as the **Participant input**, then majority can never be reached and the fallback outcome will be used as the final outcome to complete the task.
```
Yiyun333: Comment from Paul: This is not true I think. The WF goes to failed state. Pls check and add a not about this runtime check.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,326 @@ For example, you can assign a review task to multiple users:
...
+
+###### 2.5.2.4.2 Fallback outcome {#threshold-fallback-outcome}
 
-### 2.6 Events Section {#events}
+This outcome will be the final outcome of the multi-user task in the following cases:
+
+* If it is determined that it is not possible to reach the threshold with the remaining votes.
+* If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Participant input** section:
+
+For example, if the targeted user count is `10` and you set the absolute number to `20` as the **Participant input**, then the threshold can never be reached and the fallback outcome will be used as the final outcome to complete the task.
```
Yiyun333: Comment from Paul: This is not true I think. The WF goes to failed state. Pls check and add a not about this runtime check.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,326 @@ For example, you can assign a review task to multiple users:
...
+* If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Participant input** section:
+
+For example, if the targeted user count is `10` and you set the absolute number to `20` as the **Participant input**, then majority can never be reached and the fallback outcome will be used as the final outcome to complete the task.
+
+{{% alert color="info" %}}
+A fallback outcome can be both a selectable outcome and a fallback outcome at the same time.
+{{% /alert %}}
+
+{{% alert color="info" %}}
+When the **Majority type** is set to **More than half**, the fallback outcome is only required when there are 3 or more outcomes defined.
```
Yiyun333: Comment from Paul:  Is this statement correct? If it is pls add some explanation

timvandervoornmx: Ah I made a mistake, it should be the `fallback outcome is only required when there are 3 or more outcomes defined.`

The reason for that is if you have a single outcome, it's still a valid configuration because the task can and will complete as soon as more than half of the targeted users have voted.

Yiyun333: @timvandervoornmx I don't see a difference between the version you provided in the comment (`fallback outcome is only required when there are 3 or more outcomes defined.`) and the version in the doc though. Did I miss anyting?

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,326 @@ For example, you can assign a review task to multiple users:
...
+
+Possible options of majority type are described in the table below:
+
+| Majority Type | Description |
 | --- | --- |
-| None | No due date is set. |
-| Duration | You can set the deadline for the user task with the **Due in** option, which indicates the number of hours, days, or weeks the task is due in. Possible values of the property are the following ones:<br /><ul><li>Hour(s)</li><li>Day(s)</li><li>Week(s)</li> </ul> |
-| Expression | You can set a due date for the user task writing an expression. For example, to set a due date to tomorrow, you can use `addDays([%CurrentDateTime%], 1)`. |
+| **More than half** | When one of the outcomes reaches more than 50% of all possible votes (absolute majority), the task is competed with this outcome. |
+| **Most chosen** | When one of the outcomes gets the most votes (relative majority) and additional votes will not result in another outcome getting more votes, the task is competed with this outcome. |
```
Yiyun333: I find this sentence a bit contradictory in its own sense. Is it possible that one outcome gets an additional vote but in the end does not get more votes?

> additional votes will not result in another outcome getting more votes

Do we actually need to say anything about additional votes? How about just the following?

> When one of the outcomes gets most of the votes (relative majority), the task is completed with this outcome. 

My understanding is that if one outcomes gets most of the votes already, it already implies that any additional votes (to this outcome or to another outcome) will not change the fact that it is the most-chosen outcome already. Please let me know if my understand is incorrect.

timvandervoornmx: Correct, it's actually not possible to add additional votes at the moment if the task has been completed due to a majority but I added this to explain why the user task completes; additional votes won't change the outcome so we decided (at least for now) to have the workflow engine complete the task and continue the workflow. 

However, we are discussing with the team to allow for adding additional votes (it can be interesting for the user in the context of, for example, politics, where every vote should be in, no matter the outcome). 

So in short: if there is a majority before all the targeted users have voted, the user task will complete with that outcome because any additional votes wouldn't change that fact. 

Does that clear things up?

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: I have checked it and the workflow indeed will fail, so this explanation is not true. Let's remove the bullet points where I make this statement and maybe we should then add the line somewhere higher (because it holds for all decision methods) stating something like:

`If the number of [targeted users](#users) is less than the [absolute number](#absolute-number) specified in the **Participant input** section it is impossible to complete the task and the workflow will fail`

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: Sorry, 2 outcomes 🙈

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
Yiyun333: Yes! Thanks for the clarification. Makes sense. The current version in the doc is still very confusing though. What do you think this version:

> As soon as one of the outcomes gets most of the votes (relative majority), the task is completed with this outcome. Adding additional votes will not change the final outcome. Currently, it is also not possible to add additional votes after the task is completed. 

I still think if we want to mention adding addition votes, it's better to also mention that it's actually not possible to do so yet. Otherwise, it will give a wrong impression that it is actually currently possible.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
Yiyun333: Haha, now it makes more sense :))

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
Yiyun333: Sounds good! I'll add a note to the decision method section.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: Yes, that's a good suggestion.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,315 @@ For example, you can assign a review task to multiple users:
...
-#### 2.3.1 Input from
+Specifies the microflow used to assign the multi-user task. This option is displayed only when the [Target users using](#target-users) is set to **Microflow**.
 
-You can switch between a single-user task and a multi-user task. For more information on a single-user task, see [User Task](/refguide/user-task/).
+### 2.5 Completion Condition Section {#completion-condition}
 
-The following sections only introduce the properties that are shown after you choose a multi-user task. 
+{{% alert color="info" %}}
+The completion condition is evaluated after each user selects an outcome. When a user selects an outcome and this results in the multi-user task to be completed, this multi-user task will be removed from the inbox of the remaining users, as their votes will not change the final outcome of the multi-user task.
+{{% /alert %}}
```
Yiyun333, line 78: @timvandervoornmx This info note seems to express the same (but a bit more general) idea as what's said in other several info notes (for the decision methods **Veto**, **Majority**, and **Threshold**). Only **Consensus** doesn't have its own more specific note. Is there a reason for that?

I tend to delete all the more specifc info notes and only keep this more general one, but please let me know if you don't agree. Or we should add a more specific info note to the **Consensus** decision method too?

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,315 @@ For example, you can assign a review task to multiple users:
...
+* If any of the users during consensus decision method selects a different outcome from the rest of the users.
+* If the multi-user task is completed with a final outcome before all required users have completed their task:
 
-#### 2.4.1 Consensus
+  For example, when 5 users need to review a contract (`Approve`, `Reject`) and the first 2 users are not in agreement (one selects `Approve` and the other one selects `Reject`), this multi-user task will be completed with the **Fallback outcome**. The input from the other users will not change the final outcome anymore. The task will no longer appear in their inbox.
 
-Select this rule when all users should be in agreement with the provided outcome. That is, they should all select the same outcome. There has to be consensus. When there is no consensus, the multi-user task will complete with the selected [fallback outcome](#fallback-outcome).
+{{% alert color="info" %}}
+The fallback outcome should not be an option a user can select on the task page.
+{{% /alert %}}
```
Yiyun333: Does this information apply to fallback outcome in general? If so, we might want to move it to a more general place.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,315 @@ For example, you can assign a review task to multiple users:
...
-{{% /alert %}}
+For a multi-user task to complete with the non-veto outcome, all users defined in property **Participant input** need to complete their task by selecting the non-veto outcome.
 
 {{% alert color="info" %}}
-The completion rule is evaluated after each user selects an outcome. Therefore，it is possible that a multi-user task completes with a final outcome before all required users have completed their task. For example, when 5 users need to review a contract (`Approve`, `Reject`) and the first 2 users are not in agreement (one selects `Approve` and the other one selects `Reject`), this multi-user task will be completed with the **Fallback outcome**. The input from the other users will not change the final outcome anymore. The task will no longer appear in their inbox.
+The completion condition is evaluated after each user selects an outcome. When one user completes the task by selecting the veto outcome, the multi-user task will be completed with the veto outcome. The task will be removed from the inbox of the remaining users, as their votes will not change the final outcome of the multi-user task.
```
Yiyun333: This is one of the more specific notes I meant in the comment above.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,315 @@ For example, you can assign a review task to multiple users:
...
+| Most chosen | As soon as one of the outcomes gets most of the votes (relative majority), the task is completed with this outcome. Adding additional votes will not change the final outcome. Currently, it is also not possible to add additional votes after the task is completed. |
+
+###### 2.5.2.3.2 Fallback Outcome {#majority-fallback-outcome}
+
+This outcome will be the final outcome of the multi-user task if it is determined that none of the remaining votes will result in a majority.
 
-### 2.6 Events Section {#events}
+{{% alert color="info" %}}
+A fallback outcome can be both a selectable outcome and a fallback outcome at the same time.
+{{% /alert %}}
```
Yiyun333: A similar question here. Does this information apply to fallback outcome in general? If so, we might also want to move this information a bit higher to a more general place. What do you think? (I also see the same note placed for the **Threshold** decision method section but not for other decision methods).

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
```diff
@@ -18,218 +18,315 @@ For example, you can assign a review task to multiple users:
...
+A threshold of `100%` is the same as using **Consensus** as the **Decision method**.
+{{% /alert %}}
+
+###### 2.5.2.4.2 Fallback Outcome {#threshold-fallback-outcome}
+
+This outcome will be the final outcome of the multi-user task if it is determined that it is not possible to reach the threshold with the remaining votes.
+
+{{% alert color="info" %}}
+A fallback outcome can be both a selectable outcome and a fallback outcome at the same time.
+{{% /alert %}}
```
Yiyun333: The same info note as in the **Majority** decision method section.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
Yiyun333: The note is added from line 96-line98 under the Decision Method section in [this](https://github.com/mendix/docs/pull/6212/commits/31a8b8661e6da910fae2a84a9721f8c47d87352c) commit.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: Yeah I agree, I think that's why I moved that to the top and formulate it a bit more generically. So yes, let's remove the other notes and keep this one.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: Fallback outcomes can in fact be outcomes that double as a selectable option or just serves as a fallback so let's remove this as it is incorrect.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/multi-user-task.md
timvandervoornmx: Yes, this does apply to the methods where there can be a fallback outcome so I agree, let's move this up.

Yiyun333 (approved):

PR #6221, opened 2023-06-22
Release Note for BYOIDP and PAT for SVN

MarkvanMents:


content/en/docs/releasenotes/developer-portal/_index.md
```diff
@@ -61,6 +61,13 @@ To see the current status of the Mendix Developer Portal and Control Center, see
     * You can now create portfolios with different levels of visibility to people in your company: private or restricted.
 * We have added the [Portfolios](/developerportal/control-center/#portfolios) page to the Control Center, which provides Mendix Admins with governance features for portfolio management in your company.
 
+### May 16th, 2023{#byoidp}
+
+#### New Features
+
+* You can now [set up an SSO (BYOIDP)](/developerportal/control-center/set-up-sso-byoidp/) to allow Mendix users to sign in to Mendix Platform services and apps using Mendix SSO using identity federation between the Mendix Platform and your corporate IdP.
```
JaapF: Although technically correct, I would leave out the "and apps using Mendix SSO" and I would like to add StudioPro!

JaapF: What do you think of the following:

You can [set up an SSO (BYOIDP)](/developerportal/control-center/set-up-sso-byoidp/) to allow Mendix users to sign in to Mendix Platform services and to Studio Pro using identity federation between the Mendix Platform and your corporate IdP. This feature can not only be used with Mendix 10, but also for customers using Mendix 8 or 9 and for customers using TeamServer SVN and/or Teamserver Git.

MarkvanMents: @JaapF,
I don't think this is the right place to talk about Mendix 10, since it also applies to the other versions of Mendix as well - and on May 16 there was only a beta version of Mendix 10 anyway. I think the way it is written, it implies all versions of Mendix. I would suggest highlighting it as part of the Mendix 10 launch, or in a separate blog. But if feels a bit strange to mention Mendix 10 and then to add that it also applies to earlier versions.

For your other comment, I suggest:
```suggestion
* You can now [set up an SSO (BYOIDP)](/developerportal/control-center/set-up-sso-byoidp/) to allow Mendix users to sign in to Mendix Platform services, including Studio Pro, using identity federation between the Mendix Platform and your corporate IdP.
```

PR #6230, opened 2023-06-26
Correct module link to the Mendix Feedback module

Luyao-Zhang-1:


Luyao-Zhang-1:
Hi @ConnorLand  Just FYI, by accident, I see this link opens the deprecated Feedback widget, so I am changing it to the link to the new Feedback module.

PR #6234, opened 2023-06-26
Occurrence and Separated typos fixed.

dennisreep:
- Changed where Occurrence and Separated were written like Occurence and Seperated to the correct spelling
- The css class "listview-seperated" is still misspelled, but it's probably not a good idea to rename as that will break some applications that make use of that class.

### DIFF
--- a/content/en/docs/appstore/partner-solutions/ats/rg-ats/v1/test-dev/std-actions/selenium-actions/rg-one-find/rg-one-find-element-by-css.md
+++ b/content/en/docs/appstore/partner-solutions/ats/rg-ats/v1/test-dev/std-actions/selenium-actions/rg-one-find/rg-one-find-element-by-css.md
@@ -5,11 +5,11 @@ url: /appstore/partner-solutions/ats/rg-one-find-element-by-css/
 
 ## 1 Description
 
-Find a web element by CSS. Occurence lets you specify which element to fetch from the result-list, starting at 1 for the first element (defaults to the first element).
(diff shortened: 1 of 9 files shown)


Adam-Dupaski:
Thanks @dennisreep 

@ConnorLand can you comment on this proposition: `The css class "listview-seperated" is still misspelled, but it's probably not a good idea to rename as that will break some applications that make use of that class.` ?

Is there a way to verify this?

dennisreep:
> Thanks @dennisreep
> 
> @ConnorLand can you comment on this proposition: `The css class "listview-seperated" is still misspelled, but it's probably not a good idea to rename as that will break some applications that make use of that class.` ?
> 
> Is there a way to verify this?

I looked into a random project that I had open. The class listview-seperated is being used by Mendix starting applications, so if we want to fix the spelling we need to change the naming in the CSS code that all the Mendix applications are using. That's not something that needs to be changed in docs, but also outside the docs within the generic CSS.

Adam-Dupaski:
Thanks @dennisreep 

@ConnorLand is there something you can do to address this on the product side?

ConnorLand:
> Thanks @dennisreep
> 
> @ConnorLand is there something you can do to address this on the product side?

I will talk to Front End and see when this spelling error can be corrected in the product.

Adam-Dupaski:
Will merge for now, @ConnorLand please make required changes when product is changed

ConnorLand:
> Will merge for now, @ConnorLand please make required changes when product is changed

Yes I will do that. It's looking like this is a deprecated value that you can't select anymore in Atlas 3. 

So when the value is updated in Mx9 and Mx8 (Atlas 2) I will update docs with Takuma.

PR #6236, opened 2023-06-26
Highlight performance impact of fallback option

MarkvanMents:


content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -287,11 +287,17 @@ For Premium customers using a Premium resource pack, there are additional featur
...
 
 #### 4.3.2 Fallback {#fallback}
 
-The Fallback option ensures that the data in your database is automatically copied to a database in a second AZ. This, together with the high availability option, ensures that all your data is still available to your app if there is an issue with current availability zone and app instances have to be started in the second AZ.
+The Fallback option ensures that the data in your database is automatically copied to a database in a second AZ. This, together with the high availability option, ensures that all your data is still available to your app if there is an issue with the current availability zone and app instances have to be started in the second AZ. This is important for critical production apps which cannot afford to have downtime during an AZ outage in the AWS data center.
+
+{{% alert color="warning" %}}
+Because one of the instances will be running in a different AZ from the primary database, the performance of your app might be slightly affected for some users. You need to balance this with the risk of running your app without the fallback option.
```
MarkvanMents: Proposed in Slack:
```suggestion
Instances which are running in a different AZ from the primary database can have a slight impact on the performance of your app. You need to balance this with the risk of downtime if you run your app without the fallback option.
```

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -287,11 +287,17 @@ For Premium customers using a Premium resource pack, there are additional featur
 
 #### 4.3.1 High Availability
 
-The High Availability option ensures that if there is a problem with the current availability zone, new copies of your app are started in the new availability zone (AZ). All FileDocument entities are replicated in this new zone, but the data still resides in a single database. This means that you may need to restore a backup of your database if it is no longer available to your app.
+The High Availability option ensures that if there is a problem with the current availability zone, new copies of your app are started in the new availability zone (AZ). All FileDocument entities are replicated in this new zone, but the data still resides in a single database. This means that, to avoid downtime, you also need to enable the fallback option. Without fallback you may need to restore a backup of your database if it is no longer available to your app.
```
feddovanede: To me, it is not clear what the last sentence means?
> Without fallback you may need to restore a backup of your database if it is no longer available to your app.

You can't have this HA feature without the fallback feature as well. Both are included with Premium resource packs. With Standard resource packs, neither feature is available? The location of a database instance doesn't change upon restore.
Also, if your database is not available, how are you going to restore a backup to it?

feddovanede: I don't think this is technically correct. I think the following is more accurate
>The High Availability option ensures that if you horizontally scale your app, the copies of your app are spread across multiple availability zones.

feddovanede: I'm also not sure about the term _option_. It is not optional in the sense that customers can enable or disable this for resource packs at will. If you have a Premium resource pack, this feature is always enabled. If you have a Standard resource pack, it is never enabled.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -287,11 +287,17 @@ For Premium customers using a Premium resource pack, there are additional featur
...
 
 #### 4.3.2 Fallback {#fallback}
 
-The Fallback option ensures that the data in your database is automatically copied to a database in a second AZ. This, together with the high availability option, ensures that all your data is still available to your app if there is an issue with current availability zone and app instances have to be started in the second AZ.
+The Fallback option ensures that the data in your database is automatically copied to a database in a second AZ. This, together with the high availability option, ensures that all your data is still available to your app if there is an issue with the current availability zone and app instances have to be started in the second AZ. This is important for critical production apps which cannot afford to have downtime during an AZ outage in the AWS data center.
+
+{{% alert color="warning" %}}
```
feddovanede: Shouldn't we add this under 4.3.1, as it is about app containers? 4.3.1 is about app containers, 4.3.2 is about databases?

feddovanede (commented):
See comments.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: No - the performance impact is related to where the database is, not where the app is. If there is no interaction with the database, there is no impact on the app performance.

feddovanede: I would argue it is related to where the app is? The latency arises because we spread the app containers across multiple AZs (which is related to High Availability), not because we replicate the database?

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: Since HA is just one feature which is either on or off, I don't think we should have two sections. I think we should just described what it does and that you get it automatically by having a premium plan.

Always assuming that all the historic plans out there which customers still have also work in the same way.

feddovanede: HA (for app containers) and fallback (for the database) are always linked: you can't have one without the other. With Premium resource packs both are enabled, with Standard neither is enabled.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -285,19 +285,23 @@ You can have a pack using a fraction of a vCPU as several environments can run,
 
 For Premium customers using a Premium resource pack, there are additional features available for apps deployed to the Mendix Cloud, and additional ways to deploy your Mendix App.
 
-#### 4.3.1 High Availability
+#### 4.3.1 High Availability and Fallback {#fallback}
 
-The High Availability option ensures that if there is a problem with the current availability zone, new copies of your app are started in the new availability zone (AZ). All FileDocument entities are replicated in this new zone, but the data still resides in a single database. This means that, to avoid downtime, you also need to enable the fallback option. Without fallback you may need to restore a backup of your database if it is no longer available to your app.
+Premium plans come with High Availability and Fallback out of the box. This ensures that if an instance of your app crashes your app remains available and if there is a problem with the current availability zone, new copies of your app are started in a new availability zone (AZ).
```
feddovanede: The last sentence seems to indicate copies of your app will only run in another AZ if the primary AZ is down. This is different from what is stated below and how it actually works.

> Premium plans come with High Availability and Fallback out of the box. This ensures that copies of your app are spread across multiple availability zones (AZs). If there is a problem with a specific AZ, copies of your app running on other AZs will remain available.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -285,19 +285,23 @@ You can have a pack using a fraction of a vCPU as several environments can run,
...
 
-#### 4.3.2 Fallback {#fallback}
+With a Premium plan your app can be [horizontally scaled](/developerportal/deploy/scale-environment/). To fully benefit from high availability there should be at least two instances so the app is running in more than one AZ. This is important for critical production apps which cannot afford to have downtime during an AZ outage in the AWS data center.
 
-The Fallback option ensures that the data in your database is automatically copied to a database in a second AZ. This, together with the high availability option, ensures that all your data is still available to your app if there is an issue with the current availability zone and app instances have to be started in the second AZ. This is important for critical production apps which cannot afford to have downtime during an AZ outage in the AWS data center.
+Fallback ensures that the data in your database is automatically copied to a database in a second AZ. This ensures that all your data is still available to your app if there is an issue with the current availability zone and app instances have to be started in the second AZ.
```
feddovanede: I would use _primary_ availability zone instead of current availability zone.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -285,19 +285,23 @@ You can have a pack using a fraction of a vCPU as several environments can run,
...
-{{% /alert %}}
+The implementation of these features means that, although the connections between AZs are low latency, your monitoring may indicate that apps deployed to the Mendix Cloud under a Premium plan suffer an additional latency of a few milliseconds compared with apps deployed using a Standard plan. For a well-designed app, this difference should not be noticeable for end-users. 
+
+#### 4.3.2 Regional Fallback
+
+You can also purchase a Premium Plus plan which provides all the features of the Premium plan, with the addition of Regional Fallback.
+
+With Regional Fallback, a copy of your database and FileDocuments is maintained in a completely separate region. For example, if your app normally runs in *us-east-1* a copy of your data is made in *us-west-2*. If all the AZs in the primary region become unavailable, you can then choose to run your app temporarily in the secondary region, without losing any data. Once the primary region is back online, you can then revert your app to running in the primary region.
```
feddovanede: >without losing any data
We don't guarantee there will be no loss of data at all. We guarantee an RPO of 15 minutes (so max 15 minutes of data loss), although this is most likely less in practice.

feddovanede (commented):
Left some comments.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: I've now rewritten this completely, so I'll resolve this and we can start reviewing the new version

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: I've now rewritten this completely, so I'll resolve this and we can start reviewing the new version

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: 👍 
```suggestion
Fallback ensures that the data in your database is automatically copied to a database in a second AZ. This ensures that all your data is still available to your app if there is an issue with the primary availability zone and app instances have to be started in the second AZ. 
```

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents: OK. What do you mean by an RPO? Are you saying that AWS only copies data to the backup region every 15 minutes? I thought that the backup process is continuous, so you can only lose data from the time that the region crashes - minus a few milliseconds latency between writing to the local DB and copying it to the DB in the other region. Once the region is offline, then you can't update data through the app so the fact that it takes 15 minutes to start up the app in the new region doesn't mean that you've lost any data, only that no one could make any changes to the data because the app was down.

If the data can really be 15 minutes out of date it is difficult to sell but we could try:
```suggestion
With Regional Fallback, a copy of your database and FileDocuments is maintained in a completely separate region. For example, if your app normally runs in *us-east-1* a copy of your data is made in *us-west-2*. If all the AZs in the primary region become unavailable, you can then choose to run your app temporarily in the secondary region, with the data which has been copied to that region. Once the primary region is back online, you can then revert your app to running in the primary region.
```

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
MarkvanMents:
```suggestion
Premium plans come with High Availability and Fallback out of the box. This ensures that copies of your app are spread across multiple availability zones (AZs). If there is a problem with a specific AZ, copies of your app running on other AZs will remain available.
```

MarkvanMents (commented):
Thanks for your review @feddovanede 
I've made some proposals. Let me know if you think they are OK and I'll commit them.

content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
feddovanede: Yes, you are technically correct: data is kept in sync near real time. But technically, the secondary database could crash at the exact same time as the primary database. In that case, we would have to restore the secondary database based on the snapshots that are taken every 15 minutes (as the primary AZ and database are down). So, we can only _guarantee_ a data loss of maximum 15 minutes, but in practice it will be a lot less, as the likelihood of the secondary database crashing at the same time is pretty low.

feddovanede: So, text looks good to me!

feddovanede (approved):
Looks good!

PR #6245, opened 2023-06-28
Update curate.md

ThijsSeppen:
fixed some typos 
also, I removed one or two "loose" "and"s, not sure whether there actually needed to be a second noun. 

P.S. Shouldn't we remove mentions of Data Hub (or is that already happening on a separate commit)?


johannahemminger:
Thanks @ThijsSeppen for the PR, I am removing the remaining references to DH in #6233

ThijsSeppen:
Nice!

PR #6255, opened 2023-06-29
Remove webhooks, the tab has been deprecated

Luyao-Zhang-1:
The tab is only available to customers who are using it


Luyao-Zhang-1:
@MarkvanMents  FYI.

Issue #6087, opened 2023-05-25
branch-line-manager


axelenexis:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [branch-line-manager](https://docs.mendix.com/refguide9/branch-line-manager-dialog/)

Document link: [_index.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide9/modeling/menus/version-control-menu/branch-line-manager-dialog/_index.md)

## My Issue/Suggestion
Could you add information about the button "Switch branch"? I wonder what is does.

Adam-Dupaski:
Hi @axelenexis thanks for the question and feedback. We don't have a **Switch branch** button on the dialog box, so can you please clarify what you're looking for information on? Thanks.

axelenexis:
It's in the screenshot in the mentioned page https://docs.mendix.com/refguide9/branch-line-manager-dialog/ .

Adam-Dupaski:
Oh I see, sorry, I will take a look for how to document

Adam-Dupaski:
Hi @axelenexis this is documented here: https://github.com/mendix/docs/pull/6262

Thanks for the feedback and helping to improve Mendix docs! Let us know if you have any other questions.

Cheers,
Adam

PR #6262, opened 2023-06-30
Resolve feedback

Adam-Dupaski:
Addresses #6087


MarkvanMents (approved):
Thanks for picking this up @Adam-Dupaski

PR #6268, opened 2023-07-03
Update cloud-foundry-deploy.md

NikolaMendix:


MarkvanMents:
@NikolaMendix 
Thanks for the PR - should it also have a deprecation notice added to this document: https://docs.mendix.com/refguide/run-menu/#2-menu-items-overview?

NikolaMendix:
@MarkvanMents indeed, thanks. I forgot about that part. I have pushed commit for that. Here is how it currently looks in SP 
image

MarkvanMents (approved):
I've added the updated image.
OK to publish @NikolaMendix ?

NikolaMendix:
> I've added the updated image. OK to publish @NikolaMendix ?

This deprecation is already in SP 10 so it is OK to publish. Thanks Mark!

PR #6285, opened 2023-07-06
Add information about IDResolution node

Robinfr:
Node was introduced in Mx10 and 9.24.4. Added a little bit of information about it.


MarkvanMents (approved):
Thanks @Robinfr 
Looks great. I'll go ahead and publish.

PR #6295, opened 2023-07-10
Updated Feedback Module README and images

mel-chiu:


content/en/docs/appstore/modules/mendix-feedback.md
```diff
@@ -58,31 +58,24 @@ You should install the latest version of the Mendix Feedback module, as it will
...
 
     {{< figure src="/attachments/appstore/modules/mendix-feedback/Toolbox-feedback.png" width="250px">}} 
 
-3. Drop it into a position in the layout:
+3.  Drop the Feedback widget into a position in the layout:
 
     {{< figure src="/attachments/appstore/modules/mendix-feedback/feedback-on-layout.png" >}} 
 
-4. [Configure the Feedback widget](#configuration).
+4.  If your app allows anonymous users, do as follows:
```
mel-chiu: @Luyao-Zhang-1 Shall we let users know that they need to configure the widget even if they don't continue with the following anonymous users path?

PR #6296, opened 2023-07-10
Update docs about google tag module

r0b1n:
- bump version of the supported SP
- adjust documentation about recent changes in multiple "config" command restriction.


r0b1n:
This change could be merged directly, no need to wait for any releases.

PR #6298, opened 2023-07-11
Kk saml updates

katarzyna-koltun-mx:


content/en/docs/appstore/modules/saml.md
```diff
@@ -110,28 +110,18 @@ The URL for downloading the SP metadata of your app is independent of the value
...
+    * For Mendix version 9, there are odd- and even-numbered patch releases that contain the same changes and require the same Mendix version, but differ based on the version of Atlas UI that your app uses: 
     
-    * For Mendix version 8, look for releases in the **2.x** range. For example, if your Mendix version is 8.18.7, download the 2.2.3 release of the SAML module.
-    * For Mendix version 7, look for releases in the **1.x** range. For example, if your Mendix version is 7.23.21, download the 1.16.7 release of the SAML module.
-
-1. Click on a release to see the minimum Mendix version it supports. For example, release 3.3.13 supports Mendix versions 9.12.5 and above. 
+        * The even-numbered releases (for example, 3.6.2) are intended for apps that use the 2.0 version of Atlas UI, that is, apps which were originally built on a version of Mendix below 9.0.0, and then upgraded to Mendix version 9.
+        * The odd-numbered releases (for example, 3.6.3) are for new apps that are built using Mendix version 9 or 10 and are using the 3.0 version of Atlas UI.
```
JaapF: This is a sub-bullet to the bullet on Mendix 9, so I feel that including a statement on Mendix 10 may cause confusion.
I propose to say "using Mendix version 9".

content/en/docs/appstore/modules/saml.md
```diff
@@ -110,28 +110,18 @@ The URL for downloading the SP metadata of your app is independent of the value
...
+        * The odd-numbered releases (for example, 3.6.3) are for new apps that are built using Mendix version 9 or 10 and are using the 3.0 version of Atlas UI. 
 
-    It is recommended that you use the highest version of the module which supports your Mendix version (for example, if you are using Mendix 9.22.0 you would preferably use at least release 3.4.0 of the module).
-
-    {{< figure src="/attachments/appstore/modules/saml/saml-versions.png" alt="A SAML module release with the supported Mendix version highlighted">}}
-
-1. To download the required release, click the **Download** link by the number of the release.
-1. Follow the instructions in [How to Use Marketplace Content in Studio Pro](/appstore/general/app-store-content/) to import the SAML module into your app.
+    * For Mendix version 10, you should use the latest version of the module that is compatible with the 3.0 version of Atlas UI (that is, the version for apps newly built on Mendix 9 or newer).
```
JaapF: I think we need to refer to the odd-numbered releases for versions of the SAML module that are 'shared' with Mendix 9.
We will however do enhancements on the SAML module only for Mendix 10, so I think that will lead to V4.x.y for the SAML module. For Mendix 10 (SAML versions 4.x.y) we will be supporting only AtlasUI v3 and we will abandon the practice of even/odd versions. See TW-2181.
Although we don't have a 4.0.0 version yet, I propose to clarify to our customers already that we are stopping support for AtlasUIv2 (until now these were the "even" releases for SAML v3.x.y)

content/en/docs/appstore/modules/saml.md
katarzyna-koltun-mx: Done.

content/en/docs/appstore/modules/saml.md
katarzyna-koltun-mx: Done.

PR #6329, opened 2023-07-19
Add Basque/Euskara to Language Settings

ConnorLand:


ConnorLand:
I am confirming what the region code is.

PR #6348, opened 2023-07-26
[WC-1959]: Update DG2 docs related to MultiSelect

samuelreichert:


ConnorLand:
@samuelreichert why was this closed?

samuelreichert:
@ConnorLand We have created another story to update the docs for data widgets, including several topics, so this update will also be part of it.

PR #6358, opened 2023-07-31
Added info on downloading PCLM CLI

katarzyna-koltun-mx:


ilholbea:
@katarzyna-koltun-mx  Looks good to me. In case there are further questions on this topic, maybe we can add a screenshot, but for now, I think it's fine.

PR #6366, opened 2023-08-01
Add documentation on how to fix missing metadata

mikita-haurykau:
This MR adds a documentation page about what to do in case of missing metadata for a revision


MarkvanMents (changes requested):
Hi @mikita-haurykau 
Thanks for your Pull Request.
Would this information fit better in https://docs.mendix.com/refguide/troubleshoot-version-control-issues/ as it appears to be a version control issue. It put similar information in one place, and avoids adding a new document.
I can pick up this text and put it in that document if you agree.

mikita-haurykau:
> Hi @mikita-haurykau
> Thanks for your Pull Request.
> Would this information fit better in https://docs.mendix.com/refguide/troubleshoot-version-control-issues/ as it appears to be a version control issue. It put similar information in one place, and avoids adding a new document.
> I can pick up this text and put it in that document if you agree.

Sure. Put it wherever you think is best. 
My justification to put it there was to make it closed to the View where it happens.
If you move it, could you please preserve the URL 
/refguide/deployment-package-troubleshooting/
I have reference to it from the code

MarkvanMents:
Thanks for getting back to me.
I think it would be best to have all the troubleshooting in one place. I can put a link to the troubleshooting document where this is likely to happen so the customer can find it.
I can put a temporary redirect from the URL which you have put in the code, but I would like this to be changed in the code once this is published as we have an issue with a large number of redirects on the site where it is difficult to keep track of why they are there and whether they are still needed.
I may have missed it, but it would be better to discuss the placing of new information before coding URL links in the product - part of our job is trying to keep a good structure to the documentation and so having an open discussion about where information should be makes that part of the job easier, and ensures that the product can link to the best place in the documentation.

I'll work on this and ask you to review once I have made the changes.

mikita-haurykau:
> Thanks for getting back to me. I think it would be best to have all the troubleshooting in one place. I can put a link to the troubleshooting document where this is likely to happen so the customer can find it. I can put a temporary redirect from the URL which you have put in the code, but I would like this to be changed in the code once this is published as we have an issue with a large number of redirects on the site where it is difficult to keep track of why they are there and whether they are still needed. I may have missed it, but it would be better to discuss the placing of new information before coding URL links in the product - part of our job is trying to keep a good structure to the documentation and so having an open discussion about where information should be makes that part of the job easier, and ensures that the product can link to the best place in the documentation.
> 
> I'll work on this and ask you to review once I have made the changes.

I didn't merge the changes yet -- so if you let me know the correct URL I can change it before merging this in

MarkvanMents:
Hi @mikita-haurykau 
I've moved the information into the document I suggested and added a link to it from the Create Deployment Package documentation.
The link to this section will be https://docs.mendix.com/refguide10/troubleshoot-version-control-issues/#cannot-create-package

Please can review the changes and let me know if there is anything that needs changing.

Can you also confirm that this can be published immediately and doesn't need to wait for your update to the product.

Thanks

MarkvanMents (commented):
Updated after discussion - awaiting a review of my edits.

MarkvanMents:
@mikita-haurykau 
Will you have time to look at my changes and let me know whether I can publish immediately or if I have to wait until 10.2 release?
Thanks.

mikita-haurykau:
You can publish it now
Thanks

MarkvanMents (approved):

PR #6371, opened 2023-08-02
Updated business events screenshot to Integration pane for Mx 10

ezagroba:


MarkvanMents (approved):
Thanks for the PR - looks good (both the PR and the new image).
Approving and merging immediately.

PR #6376, opened 2023-08-03
Add new region in São Paulo

loadpi:


MarkvanMents (changes requested):
Looks great.
There is also a table about backups which we should update
https://docs.mendix.com/developerportal/operate/backups/#21-nightly-backups
If we can get that information as well, we are ready to go!
(We can always add it later, if necessary!)

MarkvanMents (approved):
Ready for publication on Monday

MarkvanMents:
Approved over Slack.

PR #6377, opened 2023-08-04
Update parseInteger doc to only refer to Integer/Long

Yiyun333:


content/en/docs/refguide/modeling/application-logic/expressions/parse-integer.md
```diff
@@ -3,16 +3,16 @@ title: "Parse Integer"
...
 weight: 140
 description: "Describes the function for parsing integer from strings in Mendix."
-tags: ["studio pro", "expressions", "parse", "integer", "casting", "conversion", "type casting", "type conversion"]
+tags: ["studio pro", "expressions", "parse", "integer", "long", "integer/long", "casting", "conversion", "type casting", "type conversion"]
 ---
 
 ## 1 Introduction
 
-This document describes functions that convert a string to an integer.
+This document describes the function that converts a string to a long integer.
```
lvanengelen: It's not a 'long integer': that would mean the same as 'long'. In other places (like [here](https://docs.mendix.com/refguide/data-types/#2-data-types-supported) and [here](https://docs.mendix.com/refguide/attributes/#221-type)) we call it `Integer/Long` (not ideal, but at least less confusing) so I think we should keep it that way here as well. Same goes for the other occurrences of `long integer'.

content/en/docs/refguide/modeling/application-logic/expressions/parse-integer.md
```diff
@@ -3,16 +3,16 @@ title: "Parse Integer"
...
 
 ## 2 parseInteger
 
-Takes a string and parses it to an integer.
+Takes a string and parses it to a long integer.
```
lvanengelen: I propose also documenting the handling of parsing errors (see the examples) and use that to link to a description of the valid range as well. Something like:

> A default value can be specified to handle parsing errors. The range of valid values for Integer/Long can be found in the [table of supported data types](/refguide/data-types/#2-data-types-supported).

content/en/docs/refguide/modeling/application-logic/expressions/parse-integer.md
```diff
@@ -21,15 +21,15 @@ The input parameters are described in the table below:
 | Value                        | Type    |
 | ---------------------------- | ------- |
 | The string to parse          | String  |
-| Default value **(optional)** | Integer |
+| Default value **(optional)** | Integer/Long |
```
lvanengelen, line 24: A default value can be specified to handle parsing errors. The range of valid values for Integer/Long can be found in the [table of supported data types](/refguide/data-types/#2-data-types-supported).

Yiyun333:
@lvanengelen Thanks for reviewing this PR! I've applied your feedback in this commit: https://github.com/mendix/docs/pull/6377/commits/81527c8bd2b41ac8d59cbb5733d07b3ee2801c8e

Please let me know if my changes are correct. If so, I will apply these changes to also Mx 8 and 9 docs and then merge this PR.

content/en/docs/refguide/modeling/application-logic/expressions/parse-integer.md
```diff
@@ -21,15 +21,19 @@ The input parameters are described in the table below:
...
```
lvanengelen, line 24: 🎄 You might want to update the layout of this table.

Yiyun333: @lvanengelen I am sorry :) I tried to spot what needs to be updated still but cannot figure what. Can you be a bit more specific about what needs to be updated? Do you mean the style of the table or the content of this table?

lvanengelen: I mean the style: all other tables in this document are aligned (i.e. enough spaces and `-`'s to align the `|`'s properly). My comment suggests keeping those aligned.

Yiyun333: I see! Thanks for the clarification. I am not sure whether we have a style guide on such table style though, but I will make this table aligned as you suggested here. We do need to be more consistent with the table style though. I will ask with my team about this for future improvement!

lvanengelen: Personally I prefer not aligning the tables to keep the diff's small, but surrounding style trumps personal preference ;-)

Yiyun333: I think I have the same preference :) Not sure how the surrounding style was determined actually. Can be a nice thing to investigate. But for now, I will just stick to the surrounding style too.

PR #6393, opened 2023-08-09
Clarify that PNs are not supported in PWAs

jdiehl:
@ConnorLand


jdiehl:
Please also apply this to Mx9 docs

ConnorLand (approved):

PR #6412, opened 2023-08-14
Add Premium Plus plan

feddovanede:
Add Premium Plus plan

### DIFF
--- a/content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
+++ b/content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
@@ -230,7 +230,7 @@ For more information, see [Pricing](https://www.mendix.com/pricing/).
 
 Mendix environments are sized by reference to cloud resource packs. The table below shows the current cloud resource packs for standard and premium plans. Resources for the Basic package are fixed as described in [Mendix Basic Package](/developerportal/deploy/basic-package/).
 
-The Standard resource packs listed below can be used with Standard, Premium and Premium Plus plans.
+The Standard resource packs listed below can be used with Standard, Premium, and Premium Plus plans.
 
 | Pack | App RAM | App vCPU | DB RAM | DB vCPU | DB Storage | File Storage |
 | --- | --- | --- | --- | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


content/en/docs/developerportal/deploy/mendix-cloud-deploy/_index.md
```diff
@@ -254,6 +255,16 @@ The Premium resource packs listed below can only be purchased with a Premium pla
...
 | XXXXL21 | 128 GiB | 32 | 128 GiB | 16 | 640 GiB | 1280 GiB |
 
+The Premium Plus resource packs listed below can only be purchased with a Premium Plus plan.
+
+| Pack | App RAM | App vCPU | DB RAM | DB vCPU | DB Storage | File Storage |
+| --- | --- | --- | --- | --- | --- | --- |
+| XL21 | 16 GiB | 4 | 16 GiB | 4 | 80 GiB | 160 GiB |
+| XXL21 | 32 GiB | 8 | 32 GiB | 4 | 160 GiB | 320 GiB |
+| XXXL21 | 64 GiB | 16 | 64 GiB | 8 | 320 GiB | 640 GiB |
+| XXXXL21 | 128 GiB | 32 | 128 GiB | 16 | 640 GiB | 1280 GiB |
```
MarkvanMents: @feddovanede 
I'm getting confused at what we are trying to tell the customer here. These packs (and the premium ones above) seem to be exactly the same specs as the "standard" resource packs above. So I don't really understand why we are repeating them, since the customer can look up an XXL21 resource pack and discover that it is the same irrespective of whether they have a standard, premium, or premium plus plan.

The table doesn't indicate any difference between the XXL21 Standard, XXL21 Premium, and the XXL21 Premium Plus. So I can't see why we are repeating them.

What we don't say is that Premium gives you multi-AZ fallback and Premium Plus gives you multi-region fallback. I wonder if there is a better way to present this?

MarkvanMents: Perhaps something like:
image

feddovanede: I agree. This looks way better! 👍 Let's go with this.

MarkvanMents: Done.
Let me know if you are happy with the result and I'll publish it.
Thanks for the PR.

feddovanede: Looks great to me. Let's publish!!

PR #6425, opened 2023-08-18
Updated a couple names that changed in Sprintr

ezagroba:


MarkvanMents (approved):
Thanks for the PR, Elizabeth @ezagroba.
I've changed this to indicate that Access Management is a tab rather than a page, and I think the description now indicates how the user can get to it.
Can you let me know if it is OK and I'll merge.

Thanks for spotting and correcting the inconsistency.

MarkvanMents:
I found a few more and corrected them in this commit: https://github.com/mendix/docs/commit/161b71a3a2db165d335f85b905de17c6dd991df2

Thanks for the trigger!

ezagroba:
That looks great Mark, go ahead and push it!

PR #6430, opened 2023-08-18
Task Queues and Scheduled Events Crash Behavior

MarkvanMents:


content/en/docs/refguide/modeling/resources/scheduled-events.md
```diff
@@ -9,7 +9,7 @@ tags: ["Scheduled Event", "Execution properties", "Timing", "intervals", "schedu
 
 ## 1 Introduction
 
-With scheduled events you can make the runtime execute a microflow repeatedly at specific intervals.
+With scheduled events you can make the runtime execute a microflow repeatedly at specific intervals. The microflow is run using a [task queue](/refguide/task-queue/).
```
grootjans: Since there is no influence on the task queue, it might be more apt to say 'using the task queue system'

content/en/docs/refguide/modeling/resources/scheduled-events.md
```diff
@@ -177,3 +177,7 @@ This limit cannot be overridden.
 ### 5.3 Unsupported Intervals
 
 Hour- and minute-based intervals can only be integer divisors of 24 or 60, respectively. For example you cannot schedule an event to run every seven minutes as this does not divide precisely into sixty minutes. If it is absolutely critical that an unsupported interval is used, you should schedule the event with interval value of one (every hour or every minute) and decide within the microflow whether it should continue executing at that particular time.
+
+### 5.4 Task Queues
+
+Scheduled events are run using [Task Queues](/refguide/task-queue/). See the task queue documentation for additional information.
```
grootjans: Not sure if this is helpful, as people cannot use the information to influence or control scheduled events.

MarkvanMents: @grootjans 
I included this as the original question which triggered this update came from someone using a scheduled event which appeared not to restart.

Since the resolution is documented in the task queue documentation, then we need to point people in that direction. They can, for example, find out the information which is in the `System.ProcessedQueueTask` which is described in the **task queue** documentation.

I thought this was the best place for the link - but perhaps it should be somewhere else in the document. For example, where we say
> This name is recorded in the `System.ProcessedQueueTask` objects at runtime, so that you can identify when this scheduled event has been processed.

grootjans: I think it is more applicable indeed in the section describing `System.ProcessedQueueTask`, but also adding a link in the introduction as you stated.

MarkvanMents: Updated as suggested. See commit c87fe46e4d36d331d3a2cef42f09259a10fd1d3d

MarkvanMents:
@grootjans 
Hi - does this look OK to publish now?
Thanks for your help.

MarkvanMents:
Approved over Slack.

PR #6434, opened 2023-08-21
OData Actions

sbont:


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
```diff
@@ -0,0 +1,63 @@
...
+The **Call external action** activity allows you to call an external action that is defined by an OData service that your app consumes. You can pass parameters to the action and use its result.
+
+## 2 Properties
+
+There are two sets of properties for this activity, those in the properties pane on the right, and those that are displayed when opening the activity dialog.
+An example of call external action properties is represented in the image below:
+
+{{< figure src="/attachments/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action/properties.png" alt="call external action properties" width="400" >}}
+
+The **Action** section of the properties pane shows the toolbox action associated with this activity. You can open a dialog box to select a different toolbox action clicking the ellipsis (**…**) next to **Action**.
```
MarkvanMents, line 24: What do you mean here by **toolbox action**?
I think this could be confusing, as the actions in a microflow toolbox are things that can be added to microflows (like **Create Object**).
https://docs.mendix.com/refguide/view-menu/#toolbox
I think what you mean here are **service actions** - i.e. actions which can be performed by the consumed service?

content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/_index.md
```diff
@@ -8,7 +8,7 @@ tags: ["studio pro"]
 
 ## 1 Introduction
 
-Data can be published from an app for use by other apps through [published OData services](/refguide/published-odata-services/). Consumed OData services can be used to integrate external data sources in apps through the [Integration Pane](/refguide/integration-pane/).
+Data can be published from an app for use by other apps through [published OData services](/refguide/published-odata-services/). Consumed OData services can be used to integrate external data sources and remote operations in apps through the [Integration Pane](/refguide/integration-pane/).
```
MarkvanMents: We are calling these "external actions" aren't we?
```suggestion
Data can be published from an app for use by other apps through [published OData services](/refguide/published-odata-services/). Consumed OData services can be used to integrate external data sources and actions in apps through the [Integration Pane](/refguide/integration-pane/).
```

content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/_index.md
```diff
@@ -17,14 +17,14 @@ For further details on the consumed OData service document and updating consumed
...
 {{% /alert %}}
 
-## 2 OData Services and External Entities {#external-entities}
+## 2 OData Services {#odata-services}
 
-When an external entity is used in an app, the associated dataset for the entity is retrieved through the information in the consumed OData service contract and returned.
+### 2.1 External Entities {#external-entities}
 
-### 2.1 External Entities
+When an external entity is used in an app, the associated dataset for the entity is retrieved through the information in the consumed OData service contract and returned.
```
MarkvanMents: Are we retrieving a dataset, or is it just the data? The dataset is just the description (contract) held in the catalog, isn't it?
We have a definition of a dataset here: https://docs.mendix.com/refguide/data-sets/ but it is an OQL query on an entity in the domain model, not related to external entities!

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -239,3 +239,40 @@ Clients can only set values for an association from the entity that is the [owne
...
+```json
+{
+  "FirstName": "John",
+  "LastName": "Doe",
+  "FirstWorkingDay": "20240101T00:00:00.000Z",
+  "CreatePayrollAccount": true
+}
+```
+
+If a parameter has type _object_ or _list_, the value of the parameter's property is a JSON object or array respectively, similar to what is expected when [inserting new objects](#inserting-objects) for that entity. It is possible to pass an existing object by using the `@id` syntax to reference the existing object. It is also possible to pass both an `@id` reference _and_ attributes of the object combined, which results in an existing object with the additional attributes' values. For example:
```
MarkvanMents: Three questions:

- **additional attributes** what happens if the object already has a value for this attribute - presumably it is overwritten. 
- where is the description of how the `@id` syntax works. I can only find some examples, but I can't see it defined anywhere. It would be nice to have a link to the definition.
- what happens if you use the `@id` syntax and there is no object with that id?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
```diff
@@ -0,0 +1,67 @@
...
+
+## 3 General Tab {#general}
+
+{{< figure src="/attachments/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action/general.png" alt="call external action general tab" width="700" >}}
+
+### 3.1 Action
+
+The **Action** property, allows you to select the external action of the OData service that will be called. Click **Select...** to select an action from an OData service that you have consumed in your app.
+
+If a parameter or the return value has an enumeration type which is not yet consumed in your app, you will be asked if it can be created.
```
MarkvanMents, line 41: From the later text I assume that you also have to confirm that you want the enumerator added?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
```diff
@@ -0,0 +1,67 @@
...
+
+If a parameter or the return value is an object or a list of an external entity type that is not yet part of your domain model, it will ask you to confirm that this external entity can be added to the domain model.
+
+Click **OK** to confirm that you want to add the necessary external entities and enumerations to the module which the consumed OData service document is part of. Click **Cancel** to prevent the entities and enumerations being added – in this case the action will not be selected.
```
MarkvanMents, line 45: Is this the module with the service, or the module containing the microflow?
I would expect it to be the module containing the microflow as this is where you want to consume the data. But either makes sense.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -239,3 +239,40 @@ Clients can only set values for an association from the entity that is the [owne
 ## 12 Deleting Objects {#deleting-objects}
 
 When a published resource has the [Deletable](/refguide/published-odata-resource/#capabilities) capability, clients can delete an object by sending a `DELETE` request to the URL of the object (for example, `PATCH /odata/myservice/v1/Employees(8444249301330581)`).
+
+## 13 Calling Microflows {#actions}
+
+Microflows that are published in your OData service can be called by sending a `POST` request to the action's endpoint URL, which is defined by the base URL of the OData service and the exposed name of the microflow. For example: `POST /odata/myservice/v1/OnboardNewEmployee`. An example URL can be found when opening the [Edit published microflow](/refguide/published-odata-microflow/) dialog; on the bottom you will find the **example of location** property.
```
MarkvanMents, line 245: I don't see any description of the example of location property in the published-odata-microflow document.
Should we describe it there?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
```diff
@@ -0,0 +1,63 @@
...
+The resulting location must be unique. This means that a published microflow cannot have the same exposed name as another published microflow or published entity.
+{{% /alert %}}
+
+In the **Public documentation** tab you can add a short **Summary** and a longer **Description** to describe the behaviour of the action.
+
+### 2.3 Return Type
+
+This shows the return type of the published microflow.
+
+## 3 Editing Published Microflow Parameters{#pub-odata-mflow-parameters}
```
MarkvanMents, line 47: I've put this information here - this document isn't so long and I don't think we need to create another document for it.

MarkvanMents (commented):
Hi Steven, @sbont 
Thanks for the PR.
I've done a proofread, and you can see my changes in these commits.
https://github.com/mendix/docs/pull/6434/files/b876b3e8ae9b62c36670e58887440a315724aec7..6a49961db46e72a77478ce4f8c343d05ed1108ff
I have also added a few questions to make sure that I've understood everything and to see whether I've missed anything useful that we can link to.
I'll ensure I have some time on Monday to pick up anything which still needs work.
Sorry about the delay in reviewing this - we are a bit short staffed at present, but will hopefully have onboarded some new writers who can specialize in integration before the next Mendix release.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
sbont: In this case we do mean "toolbox actions" in contrast to "external actions" or service actions. This selector does allow you to change the activity to Create object, for example!

content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/_index.md
sbont: Fine by me!

content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/_index.md
sbont: Right, changing to "data" might remove the confusion. 
```suggestion
When an external entity is used in an app, the associated data for the entity is retrieved through the information in the consumed OData service contract and returned.
```

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
MarkvanMents: @sbont 
Do you have a link to the document I've deleted in the product? If so, I'll need to redirect to this document.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
MarkvanMents: Great - thanks for the explanation. I understand more now.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
sbont: - It is indeed overwritten - but I think I'd prefer to not use the term "overwritten" as it could confuse users into thinking that this would also mean updated in the database, which it won't (unless the microflow was explicitly modeled by the user to do that).
- I've added a new commit
- Same.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
sbont: How do you mean, exactly?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
sbont: The module with the service! You can use the action(s) in multiple places, in different modules, whereas the entities and enumerations will only be added once.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
sbont: It can be found in section 2.2 of that doc! ([link](https://github.com/sbont/docs/blob/actions/content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md#22-edit-a-published-microflow))

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
sbont: We had thought of that too, but Studio Pro has a link to this part of the documentation... We had previously removed it and re-added it with [this commit ](https://github.com/mendix/docs/pull/6434/commits/b10e6df8e595bf4220a9b909bbbca0d4707d5092)

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
MarkvanMents: So it is! Not looking properly. Thanks!

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
MarkvanMents: I've added a redirect to make sure customers get taken to the right page.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
MarkvanMents: OK, great. Thanks for the confirmation.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/call-external-action.md
MarkvanMents: @sbont  
Your original text said:

> If a parameter or the return value has an enumeration type which is not yet consumed in your app, Studio Pro will create it.

Which implied that the enumeration was always created without the dev having to agree. But later on you said

> When you confirm this by clicking **OK**, the necessary external entities and enumerations will be added to the module which the consumed OData service document is part of.

So I've assumed the second is true and I changed the first to say that the dev will be asked first.

I just wanted to check that was true 😀

MarkvanMents (approved):
Thanks for making the PR and responding to my questions.
This is ready to be published now (unless my assumption about the enumerations is wrong - but that isn't major enough to prevent publication, I don't think).
I'll publish once Mendix 10.2.0 is released.

PR #6440, opened 2023-08-23
Update app-store-overview.md

KseniiaIankina:
we have removed one more content type, which is now part of module, widget or starter template


Adam-Dupaski (approved):

Adam-Dupaski:
Thanks @KseniiaIankina

PR #6449, opened 2023-08-25
Use Share link to update metadata

joostverhoog:
Studio Pro 10.3.0 introduces a feature where users can use the 'Share' link of a catalog data set to update a consumed OData service.


MarkvanMents (approved):

PR #6459, opened 2023-08-29
Add Deploy V4 API for Mendix Cloud

MarkvanMents:
Planned release 31 August.

### DIFF
--- a/static/openapi-spec/cloud-portal-v4.yaml
+++ b/static/openapi-spec/cloud-portal-v4.yaml
@@ -1,0 +1,1297 @@
+openapi: 3.0.0
+x-stoplight:
+  id: f38beb4c10b50
+info:
+  title: Deploy API
+  version: 4.0.0
+  contact:
+    name: Mendix Support
+    url: 'https://support.mendix.com'
+    email: support@mendix.com
+  termsOfService: 'https://www.mendix.com/terms-of-use/'
+  description: The API to manage apps deployed to the Mendix Cloud.
+paths:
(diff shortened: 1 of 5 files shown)


static/openapi-spec/cloud-portal-v4.yaml
```diff
@@ -0,0 +1,1295 @@
...
+        type: string
+        format: uuid
+      description: The ID of an environment.
+    expand:
+      name: expand
+      in: query
+      required: false
+      schema:
+        type: string
+      description: 'The value of this parameter can only be `package`. When you provide this query parameter, the response will only contain the package object. This query parameter is optional.'
```
mbyregowMx: @SwapnilThange this should be changed i think, if the expand parameter is provided then the response also contains the package object in details instead of only package object

SwapnilThange: Hi @MarkvanMents does following statement looks good to you? 

'The value of this parameter can only be `package`. When you provide this query parameter, the response will contain the package object with details. This query parameter is optional.'

MarkvanMents: Text updated in this commit: https://github.com/mendix/docs/pull/6459/commits/00b683b932c9c5dcb56e0e221530935ea9fd6afd

SwapnilThange:
reviewed and it looks good now to publish

MarkvanMents:
Release confirmed over Slack.

PR #6462, opened 2023-08-30
Remove cached docsy subreference

dbreseman:


MarkvanMents (approved):
Tested and stops errors in Sourcetree when changing branches

PR #6469, opened 2023-09-01

NielsKunst:
Checked in Slack here: https://mendix.slack.com/archives/CJZ85RLTA/p1688659440472209 

Removed the ^and $ in the example based on this note https://mendix.atlassian.net/browse/WTF-1552?focusedCommentId=518230. Since they are implicit having them in the example is kinda confusing


MarkvanMents (approved):
Thanks for the PR, Neils.
Approving and merging.

PR #6483, opened 2023-09-05
Added section for Custom font support

MxMurshed:


MarkvanMents:
@MendixMurshed 
Hi Mohammed,
Luyao is not around this week. She'll be back next week. Can this wait until next week, or does someone need to pick it up this week?

MxMurshed:
@MarkvanMents No problem, let's do that when Luyao is back next week.

PR #6485, opened 2023-09-06
Add information about date time functions and tokens in combination with non-localized attributes

jveldhuizen:


MarkvanMents:
Hi @jveldhuizen 
I've allocated @Yiyun333 to do this but she isn't back until next week. I can pick it up before then if it is urgent.
Also, is this attached to any release, or can it be published once Yiyun has reviewed it?

jveldhuizen:
No, this is describing the situation since the beginning of the evaluation of expressions in the client (conditional formatting, nanoflows), so it's not attached to specific releases. It's fine to be published after Yiyun has reviewed it. It's not that urgent.

content/en/docs/refguide/modeling/domain-model/entities/attributes.md
```diff
@@ -124,17 +124,17 @@ A customer can be active or inactive, which is stored in an attribute named **Ac
...
 
-In technical terms, this property indicates whether the client assumes that the date and time are in a local time zone (Yes) or in UTC (No). In the former case, the date is first converted to UTC before being sent to the server and converted from UTC before being displayed.
+In technical terms, this property indicates whether the client assumes that the date and time are in a local time zone (Yes) or in UTC (No). In the former case, the date is first converted to UTC before being sent to the server and converted from UTC before being displayed. To avoid this time zone conversion from happening twice, UTC functions and tokens like [dateTimeUTC](/refguide/date-creation/) should not be used in client-side expressions to assign values to **Date and time** attributes where localization is disabled.
 
-Default: *Yes*
+Default: *enabled*
```
jveldhuizen, line 137: I like 'enabled'. However, the `Localize` option on the attribute form is not a checkbox. It has two options: 'Yes' and 'No', see  
image  
Do you think 'enabled' is more clear in the documentation?

content/en/docs/refguide/modeling/domain-model/entities/attributes.md
```diff
@@ -124,17 +124,17 @@ A customer can be active or inactive, which is stored in an attribute named **Ac
...
 
-In technical terms, this property indicates whether the client assumes that the date and time are in a local time zone (Yes) or in UTC (No). In the former case, the date is first converted to UTC before being sent to the server and converted from UTC before being displayed.
+In technical terms, this property indicates whether the client assumes that the date and time are in a local time zone (Yes) or in UTC (No). In the former case, the date is first converted to UTC before being sent to the server and converted from UTC before being displayed. To avoid this time zone conversion from happening twice, UTC functions and tokens like [dateTimeUTC](/refguide/date-creation/) should not be used in client-side expressions to assign values to **Date and time** attributes where localization is disabled.
```
jveldhuizen: Is the next change needed as well if we change 'yes/no' to 'enabled/disabled'?
```suggestion
In technical terms, this property indicates whether the client assumes that the date and time are in a local time zone (enabled) or in UTC (disabled). In the former case, the date is first converted to UTC before being sent to the server and converted from UTC before being displayed. To avoid this time zone conversion from happening twice, UTC functions and tokens like [dateTimeUTC](/refguide/date-creation/) should not be used in client-side expressions to assign values to **Date and time** attributes where localization is disabled.
```

content/en/docs/refguide/modeling/domain-model/entities/attributes.md
Yiyun333: The UI in Mendix 10.2 now only shows a toggle button for the **Localize** property:
image

Our main guideline is to follow the newest UI. "Yes/No" is not there anymore, so I used "enabled/disabled" to replace that and to represent the toggle button for this property :)

content/en/docs/refguide/modeling/domain-model/entities/attributes.md
Yiyun333: Sounds good!

content/en/docs/refguide/modeling/domain-model/entities/attributes.md
jveldhuizen: Ah nice! I like this change (or a checkbox). It seems to be changed between version 10.1.1 (which I was using) and 10.2.

PR #6495, opened 2023-09-07
Remove custom domain issue as it is fixed in PDF Generation v1.1.1

pratikgoyal9:


MarkvanMents:
@pratikgoyal9 
Hi Praktik, 
Luyao is away until Monday. Can this PR wait until then, or does it need to be published immediately?

content/en/docs/appstore/modules/document-generation.md
```diff
@@ -60,7 +60,6 @@ The PDF document generation service does not store pages or documents at any tim
 
 * Some widgets, such as the [Charts](/appstore/widgets/charts/) widget, might not be fully loaded if they are rendered before all data is available. We check on pending network requests to prevent this, but this is not 100% reliable.
 * If you have the [Application Performance Monitor (APM)](/appstore/partner-solutions/apd/) or [Application Performance Diagnostics (APD)](/appstore/partner-solutions/apd/) add-on enabled in your app, or set the log level of the **Services** log node to *Trace*, the PDF Document Generation module will not be able to generate documents when used in Mendix Cloud. Note: This is only applicable for apps built in Mendix versions below 9.24.5 and Mendix 10.0.0.
-* If you configured the Mendix Runtime custom app setting [ApplicationRootURL](/refguide/custom-settings/#ApplicationRootUrl) to a custom domain, then the PDF document generation service returns a [timeout error](#module-usage-runtime-issues), and the PDF is not generated.
```
MarkvanMents, line 63: I think we should keep this for customers who haven't upgraded yet and indicate that it is fixed in v1.1.1

Same with the other deletion?

pratikgoyal9: I think we should remove it. Else we have to mention it every new issue we would be fixing in future. We would be publising a changelog for v1.1.1 that will mention that the new version should address this issue. Usually, its a good practice to update to the stable released version.

pratikgoyal9:
> 

@MarkvanMents Sure, it can wait until Monday.

pratikgoyal9:
> @pratikgoyal9 Hi Praktik, Luyao is away until Monday. Can this PR wait until then, or does it need to be published immediately?

I checked with my team. We've released a version on marketplace with the fix. It would be nice to release soon to ensure docs are also updated with the known issues as well. It will give right insights to Mx Developers when going through the documentation. If it's possible to release soon, it would be nice else we can wait until Monday 👍

content/en/docs/appstore/modules/document-generation.md
MarkvanMents: OK, since it was a bug which is fixed rather than a change in the way the module worked I'll go with that suggestion.

MarkvanMents (approved):
@pratikgoyal9 
Thanks for the PR.
I agree with your position and will publish this now.

PR #6496, opened 2023-09-07
Added documentation for subpath support

grootjans:
Adapted the documentation to reflect the following:
- Since 10.3 it is possible to use the Runtime on a Subpath, given the right settings
- Since Mx10 the HTTP headers (host, etc) no longer take precedence of ApplicationRootUrl

Added a longer description on the use cases, update the docs stating that subpaths are not supported.


MarkvanMents (approved):
Proofread and changes agreed over Slack.
Approving ready for Mendix v10.3

grootjans:
@MarkvanMents I should have mentioned version 10.2 in my initial draft, everything should also already be in that version.

PR #6509, opened 2023-09-12
Add npm script to start development server and update README.

lmeijvogel:
This PR changes the documentation `package.json` and README to simplify the Hugo development server startup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,7 @@ To clone the repo please keep in mind:
 
 To run a local version of the site: 
 
-1. Run cmd: `./node_modules/.bin/hugo server --environment development`
-
-    The path syntax prefacing `hugo` may be different based on your operating system and terminal you use. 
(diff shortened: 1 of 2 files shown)


MarkvanMents:
@lmeijvogel 
Thanks for this Lennaert. This makes the command much simpler.
I will discuss with the team whether this is the best name for the command to build and serve the site, but great that you have pointed me at an npm function which I wasn't aware of.
I might keep the other instructions as well, in case someone wants to change the hugo parameters (to build production, for example) but this is a great improvement for 99% of the cases of everyday use.

MarkvanMents (approved):
Hi @lmeijvogel 
Thanks for the PR, it makes life much easier!
I've left `npm start` as a script, so you can still use that, but added `npm run build` and documented that as it is more descriptive for users who are not used to npm conventions.
Thanks for improving the user experience for local builds.
Mark

PR #6521, opened 2023-09-15
Document reverse association direction

EnasAbdelrazek:
New feature for 10.3


content/en/docs/refguide/modeling/domain-model/associations/_index.md
```diff
@@ -76,6 +76,12 @@ Your mouse pointer should not activate a white dot. This indicates the creation
 
 {{< figure src="/attachments/refguide/modeling/domain-model/associations/association-move-arrow.png" >}}
 
+### 3.4 Reverse the Association direction
+
+You can reverse the association direction by right-clicking your association and selecting **Reverse direction**.
```
MarkvanMents: @EnasAbdelrazek 
What effect does this have on the association? I think we should clarify this for the customer. My assumption is:

1. A `1-*` becomes a `*-1` and vice-versa.
2. It has no effect on a `*-*` or a `1-1` association.

Can you confirm this so I can add it to the document?

MarkvanMents (changes requested):
@EnasAbdelrazek 
Can you let me know what this actually does, as I think it is unclear from the description.

content/en/docs/refguide/modeling/domain-model/associations/_index.md
```diff
@@ -76,6 +76,12 @@ Your mouse pointer should not activate a white dot. This indicates the creation
...
+
+{{< figure src="/attachments/refguide/modeling/domain-model/associations/reverse-association-direction.png" >}}
+
```
WimJongeneel, line 92: I think we should also document what happens to existing data when using this option, is it migrated or lost? (I think the last, but then we properly should put a big red warning box here.)

WimJongeneel (changes requested):

content/en/docs/refguide/modeling/domain-model/associations/_index.md
EnasAbdelrazek: It will exchange the ownership of the association. Reversing an association changes how you access associated entities. For example, if you initially had a one-to-many association from "Customer" to "Order" (1-*), it means each customer can have multiple orders. When reversed, you can access related customers from an order entity.

In all cases, all retrieval queries should be updated according to the new direction of the association. You should continue using the new association name to access the associated entities

In the case of a (many-many) or a (1-1) association when both entities refer to each other bidirectionally, reversing the association should have no significant effect on the relationship itself.
The only thing you might need to consider is the new generated association name.

content/en/docs/refguide/modeling/domain-model/associations/_index.md
EnasAbdelrazek: It will be new relationship structure. so the references will be deleted but the existing objects will stay.

WimJongeneel (approved):

content/en/docs/refguide/modeling/domain-model/associations/_index.md
MarkvanMents: Thanks for this. I've updated the warning a little to emphasize that the actual objects are not deleted.

content/en/docs/refguide/modeling/domain-model/associations/_index.md
MarkvanMents: @EnasAbdelrazek Thanks for putting all this into an update. I've proofread this, can you just check it still says what you wanted it to say.

MarkvanMents (commented):
@EnasAbdelrazek and @WimJongeneel 
Thanks for your reviews and update of the documentation to include the additional information.
I have proofread this, so could you just check it once more to confirm I haven't changed the meaning.
Then this will be read to be published on

EnasAbdelrazek:
@MarkvanMents Thanks for your update, I have reviewed it and it looks great.

MarkvanMents (approved):
PR approved and ready to be published with Mendix version 10.3

PR #6527, opened 2023-09-18
releasenotes: drop TLSv1.0 + 1.1 for Teamserver SVN

pommi:
Today TLSv1.0 + 1.1 support was dropped for accessing Teamserver SVN.


MarkvanMents (approved):
@pommi
Thanks for the PR.
Sorry I didn't get around to this yesterday.
I've proofread this and changed the words around a bit. I don't think I've changed the meaning but will wait an hour before publishing, just in case you have any comments.

PR #6531, opened 2023-09-19
Catalog OpenAPI Beta functionality doc

BananiMendix:
This is created as part of Cataloog story CAT-2712


MarkvanMents:
@BananiMendix 
Thanks for the PR. Is this for immediate publication (once we've agreed any proofreading changes), or is there a future publication date?

BananiMendix:
> @BananiMendix Thanks for the PR. Is this for immediate publication (once we've agreed any proofreading changes), or is there a future publication date?

@MarkvanMents This is for immediate publication

MarkvanMents:
> > @BananiMendix Thanks for the PR. Is this for immediate publication (once we've agreed any proofreading changes), or is there a future publication date?
> 
> @MarkvanMents This is for immediate publication

OK - I haven't got time now, but I'll pick it up first thing tomorrow.

content/en/docs/catalog/register/open-api-registration.md
```diff
@@ -0,0 +1,43 @@
...
+
+## 1 Introduction
+
+OpenAPI automatic registration registers all your OpenAPI and OData APIs in the Catalog automatically when you app is deployed to the Mendix CLoud. This makes them easily discoverable within your organization.
+
+## 2 Prerequisites
+
+For OpenAPI, your app needs to be made using Mendix Studio Pro version 10.0 or above.
+
+For OData, your app needs to be made using Mendix Studio Pro version 8.14 or above.
```
MarkvanMents: This document is specifically about OpenAPI, so I don't understand why we are talking about OData here.

MarkvanMents (changes requested):
Hi @BananiMendix 
Thanks for the PR,
I've made some substantial changes around the beta status of this product. We don't have specific beta documentation - it would mean replacing this page when the product becomes GA.

Therefore I've renamed the document to be the product as it will be when it is GA and just added the Beta banner. I've kept the sections about feedback, we can then change/remove those sections as the product matures.

I notice there is reference to OData in the document - can you address my question. Thanks.

sjoerdbreur:
Hi Mark,

Responding here on Banani's behalf, since she is currently on vacation:

I think indeed since this document specifically concerns automatic registration of OpenAPI contracts, it makes little sense to also mention OData registration here. (These are already covered by /content/en/docs/catalog/register/automate-registration.md) At most it might make sense to have the first section mention that we automatically register all supported integrations, with OpenAPI currently being in beta. Beyond that I see no need to keep information about OData in the OpenAPI document.

That said I'm also wondering what we should do with the remaining documents:
- automate-registration.md is still OData exclusive, despite this now also being done for OpenAPI
- register-data.md also refers OData only, which is subject to change
- register-non-odata-resources.md to a lesser extent also refers the previously OData-centric nature of the Catalog

Feels like we should not confuse people into thinking we are OData only, except for in that single automatic registration article for OpenAPI. On the other hand, I'm also not sure if having a separate article for OpenAPI registration is that logical, since the process itself is identical to OData. (The differences are in the usage *after* registration though)

Looping in @thijsrohof as I think he authored the original writeup as well. What do you think Thijs?

MarkvanMents:
Thanks for the response, @sjoerdbreur 
This PR was originally raised to have a documentation about OpenAPI being beta. But I changed it to be more permanent documentation as I don't want something that is only there for a time and then is replaced. This is particularly the case since customers don't have a choice as to whether the API is published or not.

It seems to me that the OpenAPI implementation has a wider impact and that we should reflect what is happening in all the relevant documents rather than have a separate document for OpenAPI.

We can change the `register-non-odata-resources` file so that it is more about registering resources which are not registered automatically. Might have to come up with a clear doc title for that so customers can find it.

content/en/docs/catalog/register/open-api-registration.md
thijsrohof: It's fine to remove the OData mention.

thijsrohof:
We've discussed generalizing our documentation so that it's written API agnostic, and adding a section where we elaborate on what API technologies and versions are supported, and what functionalities are enabled for them. This is however out of scope for this particular update and should be done as a stand-alone update.

MarkvanMents (approved):
Thanks everyone.
Approving this - and checking if it can be published now.

MarkvanMents:
Approved over Slack.

PR #6534, opened 2023-09-19
Proofread and update SP Integration docs

dbreseman:


content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
-    
-12. In the **Select schema elements for export mapping** dialog box, select the **Message definition** option, and then select the **MD_Orders** mapping created earlier via the **Select** button:
+    {{% alert color="info" %}}The next steps explain how to ensure that outputs are generated in JSON. You can do this using [Export Mappings](/refguide/export-mappings/) or in a microflow. Steps 11-19 take you through the steps of building an export mapping. Creating an export mapping is not required because published REST services in Mendix support [content negotiation](https://nordicapis.com/content-negotiation/): the ability to select which media type you want to be returned to the server.<br><br>You can follow the next steps in this document to learn how to set up export mapping, but note that you can also return the list of objects from the microflow. The platform will export it in the desired format, as indicated by the **Accept** header. You can then specify if you want to receive XML or JSON.{{% /alert %}}
```
dbreseman: I would have added this optional section under a different heading and restarted the numbering, but I'm not sure if the anchor links are used outside of this page. Is this tracked, or are there protocols around anchor linking?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+2. Enter *PRS_OrderService* for the **Name** of your REST service. Then press **OK** to create and start editing the new REST service.
+3. Add a new resource to your service by clicking **Add** in the **Resources** section. Enter **GetOrderByID** for the **Resource name**, then click **OK** to close the dialog box.
 
-    Click **OK** to close the dialog box.
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/AddRestResource.png" alt="Adding a GetOrderByID resource" >}}
 
 4. Add an operation to your resource by clicking **Add** in the **Operations for resource** section.
-5. In the **Operation** dialog box, enter`{OrderID}` in the **Operation path** field (make sure to include the braces). This will allow the REST service to be invoked with the OrderID in the URL shown in the **Example location** field of this same dialog box.
+5. In the **Operation** dialog box, enter `{OrderID}` in the **Operation path** field, making sure to include the braces (`{}`). This allows the REST service to be invoked with the order ID in the URL shown in the **Example location** field of the dialog box.
```
dbreseman, line 76: In this doc, I used code font instead of italics for user inputs that have characters (like '*}"$) around them and for values. Is there guidance around when to use code font instead of italics?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+8. <a id="eight"></a>Click **OK** to close out the **Operation** dialog box, and then click **Show** to start editing the newly created microflow. Add two parameters: **httpRequest** and **OrderID**.
 
-9. Add an action to the microflow to convert the **OrderID** variable (string) to an integer variable. This is needed to be able to search for the OrderID (autonumber).
+    {{% alert color="info" %}}These parameters might be added automatically, along with an **httpResponse** parameter. If you create an export mapping in steps 11-19 below, you must remove the **httpResponse** parameter to avoid getting errors.{{% /alert %}}
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/ConvertOrderID.png" >}}
+9. Add a **Create variable** activity to the microflow to convert the **OrderID** variable from **String** data type to **Integer/Long**. This makes it possible to search for the **OrderID** (an **AutoNumber** data type).
```
dbreseman: The style guide says it's okay to use bolding for activities in training docs but not normal documentation. Does this count as a training doc? Does that mean all how-to docs?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+    {{% alert color="info" %}}The next steps explain how to ensure that outputs are generated in JSON. You can do this using [Export Mappings](/refguide/export-mappings/) or in a microflow. Steps 11-19 take you through the steps of building an export mapping. Creating an export mapping is not required because published REST services in Mendix support [content negotiation](https://nordicapis.com/content-negotiation/): the ability to select which media type you want to be returned to the server.<br><br>You can follow the next steps in this document to learn how to set up export mapping, but note that you can also return the list of objects from the microflow. The platform will export it in the desired format, as indicated by the **Accept** header. You can then specify if you want to receive XML or JSON.{{% /alert %}}
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/SelectSchemaForExport.png" >}}
+11. <a id="eleven"></a> Right-click the **RESTExample** module on the **App Explorer** and select **Add other** > **Export Mapping** to add a new export mapping named **EM_ExportOrder**.
```
dbreseman: Do you typically italicise anytime the user is giving something a name for the first time, and then bold all mentions thereafter? Or would you leave EM_ExportOrder bolded here because the wording doesn't explicitly direct the user to type the name?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: "OData Query Options"
 url: /refguide/odata-query-options/
+description: "Lists the query options for OData."
+weight: 20
 tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "insert", "create", "update", "delete", "actions"]
```
dbreseman: The template specifies max 5-7 tags. Should I remove some of these?

Adam-Dupaski: If they're all relevant, could leave them. But some will be less relevant, so `studio pro` (too general) at least could be removed.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
-We currently only support the options described here.
+Mendix currently supports only the options described here.
 {{% /alert %}}
 
 ## 2 Retrieving Objects
 
 ### 2.1 Retrieving All Objects
 
-All objects can be retrieved by specifying the URI. For example: `/odata/myservice/v1/Employees`. You can see this if you specify the URI in a browser.
+To retrieve all objects, specify the URI (for example, `/odata/myservice/v1/Employees`). You can see this if you specify the URI in a browser.
```
Adam-Dupaski, line 21: Nice use of parentheses

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 
 ### 2.2 Retrieving a Single Object
 
-A single object can be retrieved by passing the object identifier in the URI. For example: `/odata/myservice/v1/Employees(8444249301330581)`.
+To retrieve a single object, pass the object identifier in the URI (for example, `/odata/myservice/v1/Employees(8444249301330581)`).
```
Adam-Dupaski: Nice re-construction of a passive

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 
 This table describes how to pass values for different attribute types:
 
-| Type | How to Pass |
-| --- | --- |
-| String | Enclosed in single quotes (for example, `'John'`) |
-| Enumeration | The enumeration member name between single quotes, prefixed with the enum type (for example, `DefaultNamespace.PrimaryColor'Red'`). OData v4.01 syntax without the qualified enum type name is not supported |
-| Datetime | For OData 4: a plain value (for example, `2021-12-31`). For OData 3 (deprecated): Preceded with `datetime` and enclosed in single quotes (for example, `datetime'2021-12-31'` or `datetime'<epoch value here>'`) |
-| Other | Plain value (for example, 15) |
+| Attribute Type | How to Pass                                                 |
```
Adam-Dupaski, line 57: Do these extra `-` along with the spaces need to be added? Does it make it more readable when editing?

Did this come from an automated creation of a table in your editor?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
-| --- | --- | --- |
-| eq | equals | `/Employees?$filter=Name eq 'John'` |
-| ne | does not equal | `/Employees?$filter=Name ne 'John'` |
-| gt | greater than | `/Employees?$filter=Age gt 15` |
-| lt | less than | `/Employees?$filter=Age lt 15` |
-| ge | greater than or equal to | `/Employees?$filter=Age ge 15` |
-| le | less than or equal to | `/Employees?$filter=Age le 15` |
+| Operator | Meaning                  | Example                             |
+| -------- | ------------------------ | ----------------------------------- |
+| `eq`     | equals                   | `/Employees?$filter=Name eq 'John'` |
```
Adam-Dupaski: Good adding code formatting to the operators

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
+| `ne`     | does not equal           | `/Employees?$filter=Name ne 'John'` |
```
Adam-Dupaski: Should we capitalized the `Meaning` column entries, as they basically function as descriptions?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
+| ----------- | ---------------------------------------------------------------- |
+| `and`       | `/Employees?$filter=Name eq 'John' and Age gt 65`                |
+| `or`        | `/Employees?$filter=Age gt 65 or Age lt 11`                      |
+| `not`       | `/Employees?$filter=not(Name eq 'John')`                         |
+| `()`        | `/Employees?$filter=Name eq 'John' and (Age gt 65 or Age lt 11)` |
 
 ### 4.5 Filtering by Association
 
-You can filter on attributes of an associated entity. The way you do this depends on whether the association exposes one object or a list of objects.
+You can filter on attributes of an associated entity. The syntax depends on whether the association exposes one object or a list of objects:
```
Adam-Dupaski: "filter on attributes" > "filter on the attributes" ? (used an indefinite article for "associated entity," so an additional one may may help readability?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 ## 5 Sorting
 
-You can sort the result using the `$orderby` query option. For example: `?$orderby=Name` or `?$orderby=BirthPlace/CityName`.
+To sort the result, use the `$orderby` query option. Here are some examples: `?$orderby=Name` and `?$orderby=BirthPlace/CityName`.
 
-The default direction is ascending, and you can make this explicit. For example: `?$orderby=Name asc`.
+The default sort direction is ascending. You can make this explicit as follows: `?$orderby=Name asc`.
 
-You can also order the result in a descending direction. For example: `?$orderby=Name desc`.
+You can also order the result in descending direction. Here is an example: `?$orderby=Name desc`.
```
Adam-Dupaski, line 126: Some lines use `Here is an example:` and some use `(for example, )`. Would it be good to make it more consistent throughout the doc, or are they're qualifications you're looking at for what structure to use and when?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
+To sort on multiple attributes, separate each attribute with a comma. Here is an example: `?$orderby=Name asc,Age desc`.
 
 ## 6 Selecting fields
 
-You can select which attributes and associations to return by specifying the `$select` query option. For example: `?$select=Name,Age`.
+To select which attributes and associations to return, specify the `$select` query option (for example, `?$select=Name,Age`).
 
 ## 7 Paging {#paging}
 
 Paging allows you to load data incrementally to better handle large amounts of data. Paging occurs when the client requests a lot of data, and the server returns a subset and a link to request the rest.
```
Adam-Dupaski: Does paging `allow you to load data` to really do that, as in, the user is going to load the data incrementally? Or would it be better to use `enables loading data' ? I tend to use `enables` + gerund when talking about something the system does or that is automated, as opposed to a real explicit choice the user needs to make. Fine line, sometimes.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 
 ### 7.1 Top (Limit)
 
-You can limit the number of returned objects using the `$top` query option, where the limit is a positive integer. For example: `?$top=100`.
+To limit the number of returned objects, use the `$top` query option. The limit—the number of objects to return—must be a positive integer. For example, `?$top=100` returns the top 100 objects in the list.
```
Adam-Dupaski: `The limit—the number of objects to return—` may be more readable as `The limit (meaning, number of objects to return)` ?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 
 ### 7.2 Skip (Offset)
 
-You can skip a number of objects before retrieving the result using the `$skip` query option, where the offset is a positive integer. For example: `?$skip=100` will return objects starting with the 101 object in the list.
+To skip a number of objects before retrieving the result, use the `$skip` query option. The offset—the number of objects to skip—must be a positive integer. For example, `?$skip=100` returns objects starting with the one-hundred-and-first object in the list.
```
Adam-Dupaski: Same thing here for `The offset—the number of objects to skip`

Adam-Dupaski: `one-hundred-and-first` > `101st`

dbreseman: This was following the MS style guide (https://learn.microsoft.com/en-us/style-guide/numbers#ordinal-numbers). Do the Mendix docs have different guidelines for ordinal numbers?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
dbreseman: I personally prefer aligned columns when I'm editing in Markdown, because I find the readability helps with finding errors and checking for consistency, but if the team uniformly follows a different standard then of course I can match that instead.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
Adam-Dupaski: Oh wasn't aware of that MSG guideline - we don't follow that one. We do it differently, and do not write out ordinal numbers. I'll update our SG.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
Adam-Dupaski: Can be done if that's your preference. I prefer the opposite :) No uniform guideline on this, and the choice can be made to reformat tables like this or not based on ownership of the doc.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
-You can compare values against the `null` literal. For example: `?$filter=Name eq null`.
+You can compare values against the `null` literal. For example, given a string attribute `Name`, you can use `?$filter=Name eq null` to query for the objects that have no value assigned to the `Name` attribute.
 
-In this example, `Name` is a string attribute that can have no assigned value in the database. Note that `null` means *no value* as opposed to `''` (which is an empty string).
+{{% alert color="info" %}}
+Keep in mind that `null` is different from an empty string; a null attribute has no value, but an empty string (`''`) is a string of length 0.
+{{% /alert %}}
 
-When you filter against associations, null literals can be quite useful. For example: `?$filter=Association_A_B ne null`. In this example, you query for objects of entity type `A` that have at least one association set to objects of entity type `B`.
+When you filter against associations, null literals can be quite useful. For example, with `?$filter=Association_A_B ne null`, you can query for objects of entity type `A` that have at least one association set to objects of entity type `B`.
```
Adam-Dupaski: `quite` = probably not a necessary word

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -9,159 +11,161 @@ tags: ["OData", "filter", "count", "sort", "select", "page", "studio pro", "inse
...
 
 ## 9 Passing Query Options in the Request Body
 
-If the OData query is too long to be sent as a `GET` request, clients can send the query as a `POST` request to the `/$query` endpoint. For example, `GET /Products?$select=Name,Price` and `POST /Products/$query` with `$select=Name,Price` in the request body give the same result. These `POST` requests must specify the header `Content-Type: text/plain`. 
+If the OData query is too long to be sent as a `GET` request, you can send the query as a `POST` request to the `/$query` endpoint. For example, `POST /Products/$query` with `$select=Name,Price` in the request body gives the same result as `GET /Products?$select=Name,Price`. These `POST` requests must specify the header `Content-Type: text/plain`. 
 
 {{% alert color="info" %}}
-The body must adhere to *URL encoding* principles. So, for instance, spaces, tabs, and newlines are not allowed.
+The request body must adhere to URL encoding principles. So, for example, spaces, tabs, and new line characters (`\n`) are not allowed.
```
Adam-Dupaski: Are there other examples to include here that may be useful?

Or maybe `So, for example` can be replaced with `That means` ?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -234,17 +241,19 @@ The body of the request may specify attribute and association values just as wit
...
 
 ## 13 Calling Microflows {#actions}
 
-Microflows that are published in your OData service can be called by sending a `POST` request to the action's endpoint URL, which is defined by the base URL of the OData service and the exposed name of the microflow. For example: `POST /odata/myservice/v1/OnboardNewEmployee`. An example URL can be found when opening the [Edit published microflow](/refguide/published-odata-microflow/) dialog; on the bottom you will find the **example of location** property.
+To call microflows that are published in your OData service, send a `POST` request to the action's endpoint URL. Here is an example: `POST /odata/myservice/v1/OnboardNewEmployee`.
 
-The request body is always a JSON object, with a property for each parameter that is defined in the published microflow. For example:
+The endpoint URL is defined by the base URL of the OData service and the exposed name of the microflow. To find an example URL, you can open the [Edit published microflow](/refguide/published-odata-microflow/) dialog and look at the **Example of location** property.
```
Adam-Dupaski: Probably better to add a proper anchor to /refguide/published-odata-microflow/#32-edit-a-parameter and set the cross-reference link to go to that anchor?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -272,9 +283,9 @@ If the referenced object cannot be found, the action will fail.
 If the action returns a value, it will always be contained in a JSON object with a single property named `value`.
 
 {{% alert type="info" %}}
-Note that an object that is passed to a microflow will not be committed automatically. If you want the passed object to be saved, you will have to implement this in the microflow. 
+An object passed to a microflow will not be committed automatically. If you want the passed object to be saved, you must implement this in the microflow.
 {{% /alert %}}
 
 {{% alert type="info" %}}
-The *publish microflows* functionality was introduced in Studio Pro [10.2.0](/releasenotes/studio-pro/10.2/).
+Functionality for publishing microflows was introduced in Studio Pro [10.2.0](/releasenotes/studio-pro/10.2/).
```
Adam-Dupaski: Maybe clearer to write `The functionality for publishing microflows in your OData service` ?

And is there already a link to https://docs.mendix.com/refguide/published-odata-microflow/ in this doc?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -11,9 +11,9 @@ aliases:
 
 ## 1 Introduction
 
-Mendix allows you to publish REST web services natively from Studio Pro. This how-to will show you how to publish a REST service in an example project. This example will demonstrate the GET operation for a published REST service.
+Mendix allows you to publish REST web services natively from Studio Pro. This how-to demonstrates how to publish a REST service in an example project, and it demonstrates the `GET` operation for a published REST service.
```
Adam-Dupaski: Do we need the verb `demonstrate` twice in this sentence?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -11,9 +11,9 @@ aliases:
...
 
-This how-to will teach you how to do the following:
+This how-to teaches you how to do the following:
```
Adam-Dupaski, line 16: The template uses `will teach`, but yes, for the sake of consistency with other verb forms, we should use `teaches` - I will update the template and the docs, nice change

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
 
 ## 3 Setting Up the Example Project
 
-To create the example project you will use in the next sections for publishing your REST service, follow these steps:
+To create the example project that you will use in the next sections for publishing your REST service, follow these steps:
```
Adam-Dupaski: I don't think `project` is the correct term, as you build an app and do everything in that app - so I think it should be `app` ?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
 
 1. Create a new app and rename the **MyFirstModule** module to **RESTExample**.
-2. Open the domain model of the RESTExample module.
-3. Create entities with an association like this:
+2. Open the domain model of the **RESTExample** module.
+3. Create **Order** and **OrderItem** entities with a one-to-many association, like this:
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/domainmodel.png" >}}
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/domainmodel.png" alt="Many-to-one association from OrderItem to Order" >}}
```
Adam-Dupaski, line 34: Step 3 uses `one-to-many` and the alt-text uses `many-to-one` - should we make these consistent?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
 
-4. You need pages for entering order data, so create an overview page for the **Order** entity In the RESTExample module.
-5. Create a **NewEdit** page for the **OrderItem** entity, then add a data grid to the **Order_NewEdit** page that displays the **OrderItems** over an association.
+4. [Generate overview and detail pages](/howto/front-end/create-your-first-two-overview-and-detail-pages/#3-creating-overview-and-detail-pages-automatically) for the **Order** and **OrderItem** entities.
+5. Add a data grid to the **Order_NewEdit** page. Set it to display the **OrderItem** objects from the database over an association, like this:
```
Adam-Dupaski: Maybe worth adding a cross-reference to /refguide/data-grid/ ?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/DataGridSettings.png" alt="Data grid settings for the Order_NewEdit page" >}}
 
-Your page should now look like this:
+Your completed **Order_Overview** page should look like this:
 
-{{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/order_NewEdit_Page.png" >}}
+{{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/order_NewEdit_Page.png" alt="Structure mode view of the Order_NewEdit page with data grid" >}}
 
-Add the overview page to your app navigation and run the application. Create a couple of orders and order lines by filling in the appropriate fields.
+Add the overview page to your app navigation. Then run the application and create a couple of orders and order items.
```
Adam-Dupaski: Probably a comma would be good after `Then`

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
 
 ## 4 Publishing the Service
 
-To be able to use the data from your model in the REST service, you need to create a message definition.
+To use the data from your model in the REST service, you need to create a message definition.
 
 ### 4.1 Creating the Mapping
```
Adam-Dupaski, line 52: We can always add an intro sentence in a section before a numbered list - can be simple, even `To create the mapping, follow these steps:`

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
 
 ### 4.2 Configuring the REST Service
 
-1. In the **App Explorer**, right-click the **RESTExample** module and select **Add** > **Other** > **Published REST Service**.
-2. Enter *PRS_OrderService*  for the **Name** of your REST service. The REST service is now opened.
-3. Add a new resource to your service by clicking **Add**, and enter *GetOrderByID* for the **Resource name**:
-
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/AddRestResource.png" >}}
+1. In the **App Explorer**, right-click the **RESTExample** module and select **Add other** > **Published REST Service**.
+2. Enter *PRS_OrderService* for the **Name** of your REST service. Then press **OK** to create and start editing the new REST service.
```
Adam-Dupaski, line 70: `press` > `click`

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: I understand your thinking, but that would make more sense if the code was being entered in a code editor, rather than in a regular field in a regular dialog box, as is the case here. So I think italics for these user inputs is still the best choice.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -25,185 +25,176 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+8. <a id="eight"></a>Click **OK** to close out the **Operation** dialog box, and then click **Show** to start editing the newly created microflow. Add two parameters: **httpRequest** and **OrderID**.
 
-9. Add an action to the microflow to convert the **OrderID** variable (string) to an integer variable. This is needed to be able to search for the OrderID (autonumber).
+    {{% alert color="info" %}}These parameters might be added automatically, along with an **httpResponse** parameter. If you create an export mapping in steps 11-19 below, you must remove the **httpResponse** parameter to avoid getting errors.{{% /alert %}}
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/ConvertOrderID.png" >}}
+9. Add a **Create variable** activity to the microflow to convert the **OrderID** variable from **String** data type to **Integer/Long**. This makes it possible to search for the **OrderID** (an **AutoNumber** data type).
```
Adam-Dupaski: `from **String** data type` > `from the **String** data type`

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: I wanted to ask about this one—Microsoft says to avoid "click" in favor of "select" (https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/c/click), but I've seen "click" used pretty widely throughout our documentation and I see it's in the in-house style guide too. I suppose we're assuming the Mendix user is using a mouse? Are there any situations where you would avoid "click" or "select" in our documentation? Do you use the MS guidance on "press" (https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/p/press-and-hold)?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: Nice find, but that whole section "Activities" in the SG should be updated. The best way to write an activity is to capitalize the first word and add a link to its doc. So, `[Create variable](/refguide/create-variable/) activity`.

Adam-Dupaski: Otherwise, it would be correct to bold the activity name, since it does appear in the UI, where you are finding and selecting it.

Either way, the whole name of the activity needs to be offset somehow from regular text, but it should not be in all caps.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: You can search the docs to see if the anchor is used: `name of file + #anchor-name`.

Or are the anchors just used here internally in the doc? If so, would be even easier to change/restructure.

Wherever they are used, you can always update them.

Adam-Dupaski: It seems it'd be good to start a new section.

Adam-Dupaski: `in Mendix support [content negotiation](https://nordicapis.com/content-negotiation/): the ability to` the use of the colon doesn't really help readability, maybe better to use `, which is` or put the description in parentheses?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: I'd italicize this first instance, as the goal is to let the reader know they have to input the text. It can be bolded thereafter if it's in a screenshot to which the text/step is referring; otherwise, it can just be regular text.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: We may use "press" for end-user mobile apps. We use "click" and "right-click" widely. We use "Select" for checkboxes.

Adam-Dupaski: We also use "press" for keyboard buttons

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -23,187 +23,186 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/RetrieveOrder.png" alt="Range and XPath constraint settings in the Retrieve Objects dialog box" >}}
 
-    {{% alert color="info" %}}Creating an Export Mapping is not required because published REST services in Mendix support [content negotiation](https://nordicapis.com/content-negotiation/): the ability for clients to select the media type they want returned to the server. Follow the next steps in this document to learn how to set up Export Mapping in this case, but note that you can also return the list of objects from the microflow. The platform will export it in the format desired by the client, as indicated by the **Accept** header. The client can then specify if it wants to receive XML or JSON.{{% /alert %}}
-    
-12. In the **Select schema elements for export mapping** dialog box, select the **Message definition** option, and then select the **MD_Orders** mapping created earlier via the **Select** button:
+#### 4.2.1 Building an export mapping (optional){#export-mapping}
```
Adam-Dupaski: Headings take title case

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -23,187 +23,186 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+
+The next steps explain how to ensure that outputs are generated in JSON. You can do this using [Export Mappings](/refguide/export-mappings/) or in a microflow. Creating an export mapping is not required because published REST services in Mendix support [content negotiation](https://nordicapis.com/content-negotiation/); in other words, the REST services let you select which media type you want to be returned to the server.
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/SelectSchemaForExport.png" >}}
+{{% alert color="info" %}}You can follow the next steps in this document to learn how to set up export mapping, but note that you can also return the list of objects from the microflow. The platform will export it in the desired format, as indicated by the **Accept** header. You can then specify if you want to receive XML or JSON. If you are using a microflow instead of export mappings, skip ahead to [Viewing the App](#viewing).{{% /alert %}}
```
Adam-Dupaski: Should **Accept** take code formatting instead?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -23,187 +23,186 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+5. In the **Mapping** field of the dialog box, select the **EM_ExportOrder** mapping. For the **Parameter** field, select the **Order** object that was retrieved with the **Retrieve** action in the microflow.
+6. Select **JSON** for the result, and store the output in a **String Variable**. Enter *Order_JSON* for the **Variable name**.
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/httpResponse.png" >}}
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/MFExportWithMapping.png" alt="Export With Mapping dialog box settings" >}}
 
-    The **StatusCode** will return "OK" as a 200 message. The content of the message is mapped to the exported JSON from [step 16](#sixteen). And add the **HttpVersion** that you will be using (in this case, `HTTP/1.1`).
+7. Add a **Create object** activity to the microflow to create an object of type **HttpResponse**. Create three new members: a **StatusCode** that returns the value `200` to indicate success, **Content** mapped to the exported JSON from the previous step, and the **HttpVersion** that you will be using (`'HTTP/1.1'` in this case). It should look like this:
```
Adam-Dupaski: With three things to create, would it be worth creating a sub-numbered list or a bulleted list under this step with those items, to improve readability?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -23,187 +23,186 @@ Before starting this how-to, make sure you have completed the following prerequi
...
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/RESTTestResponse.png" >}}
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/RESTTestResponse.png" alt="Result in Response" >}}
 
-Cool! You have published your first REST service from Mendix.
+Congratulations! You have published your first REST service from Mendix.
 
 ## 5 Error Handling
 
-In this new service, no error handling has been implemented. For example, if you now execute your service and enter a string in the **OrderID** parameter (or leave it out), you will see a 500 or a 404 generic error. So, error handling should be implemented to publish a robust service.
+You have not yet implemented error handling in this new service. For example, if you enter text instead of an integer in the **OrderID** parameter (or if you simply leave it blank) before clicking **Execute**, you will see a generic `500` or `404` error. So, to publish a more robust service, you should implement error handling.
```
Adam-Dupaski: "simply" can be removed - generally we should avoid adverbs

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
```diff
@@ -23,187 +23,186 @@ Before starting this how-to, make sure you have completed the following prerequi
...
+7. Test your error handler, as you did with the app in the [Viewing the App](#viewing) section. Enter some text in the **OrderID** parameter, click **Execute**, and observe the request's response:
 
-    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/ParsingErrorRESTResult.png" >}}
+    {{< figure src="/attachments/refguide/modeling/integration/publish-rest-service/ParsingErrorRESTResult.png" alt="Error in server response" >}}
 
 ### 5.2 Additional Error Handling
 
-Now that you covered the error handling of the parameter parsing, it's time to handle empty responses. This is not really an error, but an indication of what happened when nothing was returned is always a good idea.
+Now that you have covered the error handling of the **OrderID** parameter parsing, it's time to handle empty responses. Empty responses are not technically errors, but it is always a good idea to indicate what happened when nothing is returned. To add error handling for those situations when the **OrderID** parameter is filled with a valid integer but no corresponding **Order** result is found in the database, follow these steps:
```
Adam-Dupaski: "it's" > "it is" - we do not use contractions

static/attachments/refguide/modeling/integration/publish-rest-service/DataGridSettings.png
Adam-Dupaski, line 1: Not much in this screenshot, is it necessary? There are a lot of screenshots in this doc, maybe we can remove some?

static/attachments/refguide/modeling/integration/publish-rest-service/RESTTestOverview.png
Adam-Dupaski, line 1: There are a lot of screenshots in this doc, maybe we can remove some - how useful/relevant is this screenshot?

static/attachments/refguide/modeling/integration/publish-rest-service/RetrieveOrder.png
Adam-Dupaski, line 1: Not much in this screenshot, is it necessary? There are a lot of screenshots in this doc, maybe we can remove some?

static/attachments/refguide/modeling/integration/publish-rest-service/order_NewEdit_Page.png
Adam-Dupaski, line 1: Not much in this screenshot, is it necessary? There are a lot of screenshots in this doc, maybe we can remove some?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: I considered that, but from the last proofreading session I thought we were supposed to avoid line breaks in lists (in this case, the "It should look like this" line that would come after the bulleted list).

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: Thanks, good to know. Are "select" and "click" mostly interchangeable in our documentation, except for checkboxes? Or should I reserve "select" for just menu items and checkboxes?

static/attachments/refguide/modeling/integration/publish-rest-service/RESTTestOverview.png
dbreseman: I agree that this one isn't all that important. What's the protocol for removing images? Should I delete the image from the directory, or just remove the image reference in the doc? Most of the other images were pretty useful though as I tried to follow along.
I would like to discuss the team's image approach further at some point, as there's a lot I'm still trying to understand about it: Are there rough guidelines around image quantity/density, or is it all contextual? What level of expertise do we assume for the user (in general or for various components)? Is it okay to have some essential specifications mentioned only in images and not in the text (and if so, does that affect the alt text requirements for those images)? Is the alt text meant to show up as image tooltips on the docs site?

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: This is one that I would find useful to discuss further to make sure I'm clear on the guidelines (and where this falls on the balance between contextual judgement calls vs uniform docs rules). Line 126 in this doc ("Set the member **Key** to `'Content-Type'` and the **Value** to `'application/json'`") also brought this into question for me—I originally changed the user inputs to italics, but I found that was much less clear, especially because of the apostrophes.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: Sub-numbered lists and bulleted lists under a numbered step can be used

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: If it's a button, we use "click"

Adam-Dupaski: "Select" is not reserved for checkboxes, but it is the correct verb to use there

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: Right, my question is about the line after the sub-numbered list—the line that introduces the image.

Adam-Dupaski: A line break for that line would be warranted in that case - that's the only way to do it to insert the screenshot there if you're including a sub-list, so it's not an arbitrary line break

static/attachments/refguide/modeling/integration/publish-rest-service/RESTTestOverview.png
Adam-Dupaski: If you remove an image from a doc, you should also delete it in the directory. If you want to back it up for whatever reason, you can do so locally or in a team Dropbox folder (e.g., for a project, etc.)

Adam-Dupaski: Using screenshots is mostly contextual - if it supports the user's understanding for locating and doing what they need to do. If it's just visualizing what can easily and clearly be written in text, then maybe we don't need the screenshot. We don't have a number of screenshots that's enough or too many for a doc. But a screenshot per step, for example, feels like a lot. So we should always be looking to cut down screenshots where possible (and also crop them tightly), because it can help with our maintenance of the images and the docs.

Adam-Dupaski: >What level of expertise do we assume for the user (in general or for various components)?

Varies per product and doc.

Adam-Dupaski: >Is it okay to have some essential specifications mentioned only in images and not in the text (and if so, does that affect the alt text requirements for those images)?

If it's a huge list of essential specs, maybe. But we prioritize text over image.

Adam-Dupaski: > Is the alt text meant to show up as image tooltips on the docs site?

For screen-reader only, I believe

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -255,7 +264,9 @@ The request body is always a JSON object, with a property for each parameter tha
 }
 ```
 
-If a parameter has type *object* or *list*, the value of the parameter's property is a JSON object or array respectively, similar to what is expected when [inserting objects](#inserting-objects) for that entity. It is possible to pass an existing object by using the [`@id` syntax](https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_EntityReference) to reference the existing object. It is also possible to pass both an `@id` reference *and* attributes of the object combined, which results in the existing object having the additional attributes' values. For example:
+If a parameter has the data type **Object**, the value of the parameter's property is a JSON object. If a parameter's data type is **List**, the value of the parameter's property is a JSON array. This is similar to what is expected when [inserting objects](#inserting-objects) for that entity.
```
Adam-Dupaski: Maybe the data types don't need bolding? Those are generic data types and I don't think you're referencing a screenshot here?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
```diff
@@ -272,9 +283,9 @@ If the referenced object cannot be found, the action will fail.
 If the action returns a value, it will always be contained in a JSON object with a single property named `value`.
 
 {{% alert type="info" %}}
-Note that an object that is passed to a microflow will not be committed automatically. If you want the passed object to be saved, you will have to implement this in the microflow. 
+An object passed to a microflow will not be committed automatically. If you want the passed object to be saved, you must implement this in the microflow.
```
Adam-Dupaski, line 286: Good removing "Note that"

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: It can be exception, of course. If it improves readability, then it can be done. And these are ultimately referring to code, so it makes sense. If you make a case for it and own the doc, then it's your choice.

Still important to stay aware of where the user is entering something (a regular field [low-code] or a code editor), which guides our choices in formatting.

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
```diff
@@ -19,26 +19,26 @@ Open your [published OData service](/refguide/published-odata-services/), or cre
 
 {{< figure src="/attachments/refguide/modeling/integration/published-odata-services/published-odata-microflow/microflows-grid.png" alt="published microflows" width="600" >}}
 
-In the **Microflows** tab of the published OData service, click **Add** button. This will open a selector dialog where you can select the microflow you want to publish. From here you can also click **New** to create a new microflow.
+In the **Microflows** tab of the published OData service, click **Add**. This will open a selector dialog where you can select the microflow you want to publish. From here you can also click **New** to create a new microflow.
```
Adam-Dupaski: "selector dialog" > "selector dialog box"

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
```diff
@@ -19,26 +19,26 @@ Open your [published OData service](/refguide/published-odata-services/), or cre
...
 
-When you publish a microflow that has object, list, or enumeration parameters or return types, these entities and enumerations need to be published as well.
+When you publish a microflow that has object, list, or enumeration parameters or return types, these entities and enumerations must also be published.
 
-### 2.2 Edit a Published Microflow
+### 2.2 Edit a Published Microflow{#edit-microflow}
 
-Either select the microflow and click **Edit** or double-click the microflow to open its properties. In the [Edit published microflow](#pub-odata-mflow-parameters) dialog you can change the **Exposed name** and change the **Microflow**. Below that, in the box **Example of location**, you will find the URI where the action will be located when the app is running.
+Either select the microflow and click **Edit** or double-click the microflow to open its properties. In the [Edit published microflow](#pub-odata-mflow-parameters) dialog box, you can change the **Exposed name** and change the **Microflow**. Below that, in the field **Example of location**, you will find the URI where the action will be located when the app is running.
```
Adam-Dupaski: "in the field **Example of location**," > in the *Example of location** field," (parallel construction with "the Edit published microflow dialog box")

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
```diff
@@ -56,7 +56,7 @@ You can **Add** or **Delete** parameters from the published microflow.
 
 ### 3.2 Edit a Parameter
 
-To edit a published parameter, either select the parameter and click **Edit** or double-click the parameter. The **Edit published microflow parameter** dialog will allow you to edit the **Exposed name** of the published parameter, and you can set a value for **Can be empty**. If **Can be empty** is disabled, calls to the published microflow will fail if the parameter is not provided.
+To edit a published parameter, either select the parameter and click **Edit** or double-click the parameter. The **Edit published microflow parameter** dialog will allow you to edit the **Exposed name** of the published parameter, and you can set a value for **Can be empty**. If **Can be empty** is turned off, calls to the published microflow will fail if the parameter is not provided.
```
Adam-Dupaski: "**Edit published microflow parameter** dialog will allow" > "**Edit published microflow parameter** dialog box allows"

Adam-Dupaski: "and you can set a value for **Can be empty**" can maybe be on its own line/paragraph? It doesn't really need to be joined to the previous part of the sentence about editing the parameter.

I also don't think "setting a value" for a checkbox is the right phrase, unless I'm missing some value you set here when that checkbox is selected?

Could be: "You can select also **Can be empty**." ?

For "If **Can be empty** is turned off" > maybe a better verb phrase than "turned off"? 

Could be: "If the **Can be empty** checkbox is cleared" ?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-microflow.md
dbreseman: This page wasn't actually one of my proofreading tasks—I just went in to add an anchor so I could link to it from a different page, and a few corrections jumped out at me while I was in there. I could definitely go back and give this one a full edit though!

Adam-Dupaski: No need to go through the whole doc, let's just look at these comments

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
dbreseman: I bolded this because it references the UI text in the product, and it's meant to help the user find their way around the Edit published microflow dialog box.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: I updated this text, but I haven't been able to replicate this: "You can search the docs to see if the anchor is used: name of file + #anchor-name". Is `#` a meaningful search operator for docs.mendix.com?

static/attachments/refguide/modeling/integration/publish-rest-service/RESTTestOverview.png
dbreseman: > If it's a huge list of essential specs, maybe. But we prioritize text over image.

For example, should I change things like step 3 in https://docs.mendix.com/howto/integration/publish-rest-service/#3-setting-up-the-example-project?

content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/odata-query-options.md
Adam-Dupaski: Okay, but that paragraph's main focus is on the JSON and the code snippet, so the UI and dialog box are quite far. It seems more conceptual than referential to me, but your call.

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
Adam-Dupaski: Yes, it is, but not on docs.mendix.com. You need to search the Markdown, which means you should use a code editor, like VSC.

What is the anchor you're searching for?

static/attachments/refguide/modeling/integration/publish-rest-service/RESTTestOverview.png
Adam-Dupaski: No, that's fine and a good example

content/en/docs/refguide/modeling/integration/rest-services/published-rest-services/publish-rest-service.md
dbreseman: Ah thanks, makes sense! All good now. I wanted to confirm that I wouldn't be breaking links in other docs when I deleted the #eight anchor here, but more generally I wanted to know the process for finding and updating external references in the future.

PR #6573, opened 2023-09-28
MQTT proofreading

quinntracy:


content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -57,7 +57,7 @@ Enabling SSL will establish a secure connection to the broker. To input your SSL
 
 To use the MQTT Connector, make sure you have [configured the broker connection](#configure-broker), or [create a broker connection directly in a microflow](#broker-microflow). 
 
-Build microflows that use the connection details from the domain model to connect, and set up connection details using the domain model. Include microflows that first use the connection details from the domain model to connect, before the subscribe or publish actions can be performed.
```
quinntracy, line 63: This section is difficult to understand so I have a few clarifying questions: 
- Is this speaking directly to the user about what they can do with the connector? 
- This wording appears repetitive; is "building microflows using the connection details from the domain model" the same action as "setting up connection details using the domain model"? If not, how are they different? 
- How is "include" used in this context? How does a user include this kind of microflow first in comparison to other microflow use?

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -68,14 +68,14 @@ After you have set up the connection with the broker in the [Configuration](#con
...
-        * **Retained** set to **No** – For publishers, this indicates that this message should not be retained by the MQTT server. For subscribers, it indicates this is a normal message, received as a result of it being published to the server.
+    * **Topic** – Define the topic to which the payload is published.
+    * **Payload** – Pick a string message to be published.
+    * [QoS (Quality of Service)](https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/qos.html) – Choosean attribute of each message being published
+    * **Retained** – The retained setting serves two purposes depending on whether the message it is associated with is being published or received:
+        * **Yes** – For published messages, selecting **Yes** indicates  the MQTT server should retain a copy of the message. The message will then be transmitted to new subscribers to a topic that matches the message topic. For subscribers registering a new subscription, if the flag is **true**, this indicates the received message has been retained by the MQTT server.
```
quinntracy: Waiting for screenshots to determine formatting for "true"

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,38 +9,35 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+The MQTT connector consists of microflow actions that enable you to do the following:
 
-* Publishing data to the MQTT broker
-* Subscribing to the MQTT broker 
-* Unsubscribing from the MQTT broker
+* Publish data to the MQTT broker
+* Subscribe to the MQTT broker 
+* Unsubscribe from the MQTT broker
 
-### 1.2 Features
```
quinntracy, line 22: This repeats exactly what is started in 1.1. I think it makes sense to combine--can 1.1 be changed to Features?

Adam-Dupaski: I think we need to differentiate more between Use Cases and Features, perhaps - are there any use case details that can be added to replace this duplication? How should customers use this component in ways that aren't currently described? Otherwise, yes, just keeping `1.1 Features` should be fine.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
 
-Configure your broker connection information by adding the **ConnectionAdministration** page to your app, then running locally and setting up in that page. The **Connection Name**, **Broker host**, and **Broker port** fields point to the MQTT broker. 
+Configure your broker connection by adding the **ConnectionAdministration** page to your app. Input your broker information, then set up and run the app locally.
+
+{{< figure src="/attachments/appstore/connectors/mqtt-connector/connection-administration.png" >}}
+
+The **Connection Name**, **Broker host**, and **Broker port** fields point to the MQTT broker. 
 
-If the broker needs basic authentication for connection, then you can set up a username and password that will be stored in the domain model. See the [Authentication](#auth) section below.
+If the broker needs basic authentication for connection, you can set up a user name and password that is stored in the domain model. For more information, see the [Authentication](#auth) section below.
```
MarkvanMents: Perhaps "to connect" would be better here?

MarkvanMents: I also think that the user name and password is stored in the database, rather than in the domain model. (A quick search finds "stored in the database" is much more common). Let me know if you want to discuss this.

MarkvanMents: And the authentication section is now called "Setting Up Authentication".

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
 
 #### 2.1.1 Creating a Broker Connection in a Microflow {#broker-microflow}
 
-You can set up a broker connection in a microflow directly if you do not want it to be [stored in the domain model](#configuration). To do this, follow these steps:
+You can set up a broker connection directly in a microflow if you do not want it to be stored in the domain model. To do this, follow these steps:
```
MarkvanMents: stored in the database.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
 
-1. In your microflow, use the [Create object](/refguide/create-object/) activity to create the **ConnectonDetail** object by providing the required parameters.
-2. Use the created **ConnectionDetail** object for **Publish**, **Subscribe**, or **Unsubscribe** operations.
+1. In your microflow, use the [Create object](/refguide/create-object/) activity to create the **ConnectionDetail** object and provide the required parameters.
```
MarkvanMents: Create _a_ **ConnectionDetail** object.
Or you could say _the_ if you refer to relate it to this particular broker.
I think it might help the users to say that **ConnectionDetail** is in the MqttConnector module. You could just use the syntax **MqttConnector.ConnectionDetail** the first time you mention it.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
 
-Enabling SSL will establish a secure connection to the broker. To input your SSL certificates, follow these steps:
+Enabling SSL establishes a secure connection to the broker. To input your SSL certificates, follow these steps:
 
 1. Open your certificates file in any text editor. 
-2. Copy all content of the file and paste it to the **CA Certificate** and **Client Certificate** fields. Certificates start with `-----BEGIN CERTIFICATE-----` and end with `-----END CERTIFICATE-----`. Client keys typically start with `-----BEGIN RSA PRIVATE KEY-----` or `-----BEGIN PKCS7-----`.
-3. Add the **Client Certificate Key** and the **Certificate password** if one was set.
-    * Currently supported certificate types are X.509 format (*.crt*, *.cer*)
-    * Currently supported key types are X.509 format (*.pem*, *.pkcs*)
+2. Open the **ConnectionAdministration** page and enable the SSL toggle.
```
MarkvanMents: Good addition.
I wonder if we need to mention the toggle, or whether we can just say "enable SSL"?
What do you think?

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+3. Copy all file content and paste it to the **CA Certificate** and **Client Certificate** fields. 
+    *   Certificates start with `-----BEGIN CERTIFICATE-----` and end with `-----END CERTIFICATE-----`
+    * Client keys typically start with `-----BEGIN RSA PRIVATE KEY-----` or `-----BEGIN PKCS7-----`
```
MarkvanMents: Does this go with the next step - it is talking about keys.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+4. Add the **Client Certificate Key** and the **Certificate password** if one was set.
+    * Currently supported certificate types are *X.509* format (*.crt*, *.cer*)
+    * Currently supported key types are *X.509* format (*.pem*, *.pkcs*)
+
+{{< figure src="/attachments/appstore/connectors/mqtt-connector/ssl-input.png" >}}
 
 ## 3 Usage
 
 To use the MQTT Connector, make sure you have [configured the broker connection](#configure-broker), or [create a broker connection directly in a microflow](#broker-microflow).
```
MarkvanMents: created?

content/en/docs/appstore/connectors/mqtt.md
MarkvanMents: 1. Yes, it is telling the user what they can do.
2. What it is trying to say is that in the microflow you first have to get the connection details from the database, or create them in your microflow before you can then do any other MQTT actions. The MQTT action needs the connection details to know which broker it is connecting to and to provide authentication if required.
3. I don't think "Include" is the right word here. I think it means more "create" or "use". Perhaps something like "Ensure that your microflows provide the broker connection details when you perform the subscribe or publish actions."?

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
-    * **Topic** – topic to which the payload is published
-    * **Payload** – string message to be published
-    * [QoS (Quality of Service)](https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/qos.html) – an attribute of each message being published
-    * **Retained** – the retained setting serves two purposes depending on whether the message it is associated with is being published or received:
-        * **Retained** set to **Yes** – For messages being published, setting Retained to Yes indicates that the MQTT server should retain a copy of the message. The message will then be transmitted to new subscribers to a topic that matches the message topic. For subscribers registering a new subscription, the flag being true indicates that the received message is not a new one, but one that has been retained by the MQTT server.
-        * **Retained** set to **No** – For publishers, this indicates that this message should not be retained by the MQTT server. For subscribers, it indicates this is a normal message, received as a result of it being published to the server.
+    * **Connection detail** - Select the **ConnectionDetail** entity to publish.
```
MarkvanMents: Great - definitely need this.
I would say "**ConnectionDetail** _object_" rather than entity.
The entity is more the structure of the data (the **table** in database terms) and you then have "objects" which have this structure.

https://docs.mendix.com/refguide/object-activities/ tries to make this distinction but could perhaps be clearer.

MarkvanMents: And perhaps not "to publish" as the connection details are more about where to publish rather than what (which is the **payload**.

Something like "specifying where to publish to"? I'm sure you can think of something better.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+    * **Topic** – Define the topic to which the payload is published.
+    * **Payload** – Choose a string message to be published.
+    * [Qo S (Quality of Service)](https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/qos.html) – Choose an attribute for each message that is being published.
```
MarkvanMents: Good change to update "Qo S" to match the UX - even though it looks horrible!
And I think you need to choose a value here, rather than the attribute. QoS attribute is applied to each message, but the user needs to decide what its value is.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+    * **Retained** – The retained setting serves two purposes depending on whether the message it is associated with is being published or received:
+        * **Yes** – For published messages, selecting **Yes** indicates  the MQTT server should retain a copy of the message. The message is transmitted to new subscribers to a topic that matches the message topic. For subscribers registering a new subscription, if the flag is **true**, this indicates the received message has been retained by the MQTT server.
+        *  **No** – For publishers, selecting **No** indicates this message should not be retained by the MQTT server. For subscribers, this indicates it is a normal message that was received as a result of it being published to the server.
```
MarkvanMents: This seems to have information about both publishing messages and subscribing to them.
I think it might be better to move the sentences "For subscribers" into the next section, about subscribing to the MQTT broker.

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
-    * **Topic** – topic to subscribe to
-    * **OnMessageMicroflow** – microflow to be executed when a message is received on subscribed topic
-    * [QoS (Quality of Service)](https://www.eclipse.org/paho/files/mqttdoc/MQTTClient/html/qos.html) – an attribute of each message being published
+    * **Connection detail** - Select the **ConnectionDetail** entity to subscribe to. 
+    * **Topic** – Define the topic to subscribe to.
+    * **On message microflow** – Identify the microflow to be executed when a message is received on a subscribed topic.
+    * Qo S (Quality of Service) – Choose an attribute for each message that is being published.
 
-When subscribing to a topic, define in the subscribe action which microflow is triggered for each message, and that the topic and payload is passed on into that microflow.
+When subscribing to a topic, in the **Subscribe** action, define which microflow is triggered for each message and check that the topic and payload is passed on into that microflow.
```
MarkvanMents: It was not clear to me how the topic and payload is passed into the microflow. There is a sample microflow (MqttConnector.SampleOnMessageMicroflowForSubscribe) in the MqttConnector module, but perhaps we can specify a bit more here that the topic and payload are passed to the microflow as string parameters.
image

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
 
-#### 3.2.3 Unsubscribe
+#### 3.1.3 Unsubscribe
 
 To unsubscribe from the MQTT broker, follow these steps:
 
 1. Search for *Unsubscribe MQTT* in the **Toolbox**.
-2. Drag the *Unsubscribe MQTT* event into your microflow.
+2. Drag the **Unsubscribe MQTT** event into your microflow.
```
MarkvanMents: Unsubscribe _from_
image

content/en/docs/appstore/connectors/mqtt.md
```diff
@@ -9,95 +9,107 @@ tags: ["marketplace", "marketplace component", "mqtt", "iot"]
...
+        *  **No** – For publishers, selecting **No** indicates this message should not be retained by the MQTT server. For subscribers, this indicates it is a normal message that was received as a result of it being published to the server.
 
-#### 3.2.2 Subscribe
+#### 3.1.2 Subscribe
 
 To subscribe to the MQTT broker, follow these steps:
 
 1. Search for *Subscribe MQTT* in the **Toolbox**.
-2. Drag the *Subscribe MQTT* event into your microflow.
+2. Drag the **Subscribe MQTT** event into your microflow.
```
MarkvanMents: Subscribe _to_
image

static/attachments/appstore/connectors/mqtt-connector/authentication-method.png
MarkvanMents, line 1: All the images are rather wide and don't have much information beyond the left-hand section. They would present more clearly in the docs if it focused on the bits with data. You can do this by one of:
1. Resizing the dialog box before you capture the image (if that is possible in the UX)
2. Capturing just the bit containing the data
3. Manipulating it in SnagIt to crop from the right or cut a bit out of the middle (there is a tool for that).

MarkvanMents (commented):
Hi @quinntracy 
This is a great improvement on the original document and the images really help clarify what is going on.

I have made a lot of comments, but very few of them are related to your changes and are much more about the terminology in the original document.

Please have a look at them as there are some concepts there which will come up throughout Mendix documentation. Do ask questions about what I've said, and perhaps we can have a session when I get back from leave to look at Mendix concepts in general.

But, just to reiterate, your improvements are great and most of the comments are about things which will become clearer as you work more with Mendix.

content/en/docs/appstore/connectors/mqtt.md
quinntracy: I added it to indicate that the user needs to look for the toggle, because when I was repeating the steps, I wasn't sure what to look for to find this step. I think either work though, especially since it's an article that is geared towards more experienced users.

content/en/docs/appstore/connectors/mqtt.md
quinntracy: Good call, I'll move it down

content/en/docs/appstore/connectors/mqtt.md
quinntracy: I went back into SP to check this feature and it doesn't look like **Retained** is an option anymore when entering in these details. The user can check a box that executes the action in a Task Queue, and there's a box to enter the Return type. Fair to say we can removed **Retained** and add in these two other features?

content/en/docs/appstore/connectors/mqtt.md
MarkvanMents: "enable SSL using the toggle"?

content/en/docs/appstore/connectors/mqtt.md
MarkvanMents: When I look at MQTT Publish I still see **Retained** (Version 1.5.0).
image

The Task Queue and Return are standard for a Java Action, so I'm not sure you need to say anything here. You could link to Java Actions, or Task Queue documentation.

MarkvanMents (commented):
Here are some responses to your outstanding questions.
Can you look at them and let me know if you want any further assistance.

content/en/docs/appstore/connectors/mqtt.md
MarkvanMents: The **Retained** values are `true` and `false` rather than `yes` and `no`. So that needs to be revised, I think.

content/en/docs/appstore/connectors/mqtt.md
quinntracy: I updated this section, so can you review for accuracy? I want to make sure I didn't change the meaning of the statement while moving pieces around.

PR #6588, opened 2023-10-03
[RUN-3346] Default cost of BCrypt was changed to 12 in Mx10

atennapel:
Default cost of BCrypt was changed to 12 in Mx10. There was still one reference in the docs that mentioned it as 10.


MarkvanMents (approved):
Thanks for the PR. Approving and merging.

PR #6613, opened 2023-10-06
[UIA-700] Document known issue w.r.t. latest Chromium differences and timeout error caused by incorrect roles assignment

MxMurshed:


Adam-Dupaski:
@MendixMurshed if this is a known issue, it should go into the release notes for the module (for the releases that it affects).

content/en/docs/appstore/modules/document-generation.md
```diff
@@ -311,6 +311,8 @@ In general, we recommend you perform the following steps in case of any issues d
 
 In case of issues regarding styling, we recommend you temporarily add the page microflow to your app navigation (See step 2 in the [Module Usage and Runtime Issues](#module-usage-runtime-issues) section). This allows you to preview the page in your browser and inspect the applied styles. We recommend you use Chrome or Chromium and the [Chrome DevTools](https://developer.chrome.com/docs/devtools/css/) for this, since Chromium is the browser that is used by the document generation service.
 
+* When developing locally using Chrome or Chromium version 117, the scaling of your document might be different compared to previous versions and the cloud service, we are currently looking into this issue.
```
KevinVlaanderen: Make 'we are currently looking into this' a separate sentence; so either a . or a ; in front of it, instead of a comma.

Also, what is our suggestion to users? How can they best deal with this problem? Do we suggest to use 112 for now?

arjan-h: I would add a suggestion that we advice to use the Chromium version mention in the Chromium paragraph to guarantee the same result as from our service

MxMurshed: For best match in the generated documents, I think 112 is indeed the best option.

arjan-h: Yep, and we already have that one described + link in the Chromium paragraph. Let's refer to that paragraph

arjan-h:
Review suggestion based on latest changes:
`* When testing the PDF document generation locally using Chrome or Chromium version 117 or higher, the scaling of your PDF document might be different compared to generated document from the PDF document service in the Mendix Public Cloud. We are currently looking into this issue and we advice to use the Chromium version mention in the Chromium paragraph to guarantee locally the same result as from our PDF document generation service. Please refer section [3.1.2 Chromium](#chromium) for more details.`

KevinVlaanderen (approved):

Adam-Dupaski (approved):

PR #6619, opened 2023-10-09
Updated Linux versions in system requirements

tknippenberg:


content/en/docs/refguide/installation/system-requirements.md
```diff
@@ -146,9 +146,10 @@ The Mendix Docker buildpack supports the following Kubernetes versions:
 
 * Microsoft Windows Server 2012 and above
 * The following Unix-like operating systems:
-    * [Debian OldOldStable (LTS)](https://wiki.debian.org/DebianOldOldStable), [Debian OldStable, Debian Stable](https://wiki.debian.org/DebianReleases#Current_Releases.2FRepositories)
-    * Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7, and Red Hat Enterprise Linux 8
-    * CentOS 6, CentOS 7
+    * Debian [OldOldStable (LTS)](https://wiki.debian.org/DebianOldOldStable), Debian OldStable, Debian Stable
```
MarkvanMents: Is there a reason you removed the link? I thought it might be useful for customers to have a direct link to the OldStable and Stable versions for confirmation?

content/en/docs/refguide9/general/system-requirements.md
```diff
@@ -170,9 +170,10 @@ The Mendix Docker buildpack supports the following Kubernetes versions:
 
 * Microsoft Windows Server 2008 SP2 and above
 * The following Unix-like operating systems:
-    * Debian OldOldStable (LTS), Debian OldStable, Debian Stable
-    * Red Hat Enterprise Linux 6, Red Hat Enterprise Linux 7, and Red Hat Enterprise Linux 8
-    * CentOS 6, CentOS 7
+    * Debian [OldOldStable (LTS)](https://wiki.debian.org/DebianOldOldStable), Debian OldStable, Debian Stable
+    * Red Hat Enterprise Linux 6 and later
+    * CentOS 7
```
MarkvanMents, line 175: Is CentOS 6 still supported for earlier 9.24.x versions of Mendix. If so, we should still keep that in for customers who are not fully up-to-date. Our LTS support does not insist that customers are on the latest patch release, [only that they are on 9.24](https://docs.mendix.com/releasenotes/studio-pro/lts-mts/#lts)

MarkvanMents (changes requested):
@tknippenberg 
Thanks for the PR. I have a couple of very quick questions.

content/en/docs/refguide/installation/system-requirements.md
tknippenberg: Sorry no, that wasn't on purpose. The previous text wrt Debian was probably already good!

content/en/docs/refguide9/general/system-requirements.md
tknippenberg: We don't have a formal policy wrt these versions as we do have for database versions. So if a CentOS 6 question comes in (also for Mx 8) it's going to be a big question whether we can even support this.

Knowing that CentOS 6 is end-of-life since November 2020, I suggest to remove it from the Mendix 9 system requirements.

MarkvanMents: OK. Then best to leave CentOS 6 off. Thanks for the explanation.

content/en/docs/refguide/installation/system-requirements.md
MarkvanMents: I'll put the link back then. And then publish the docs immediately.

MarkvanMents (approved):
Thanks for the PR Tom. I'll publish it once I've put the links back 😀

PR #6633, opened 2023-10-11
Update scheduled-events.md

NielsKunst:
I felt that it was not clear that a user should enable ProcessedTaskQueue runtime variables when an application utilizes scheduled events. 

Therefore, adding a section to the documentation of the scheduled event on this makes sense in my opinion.


MarkvanMents (approved):
@nelis37 
Hi Niels,
Thanks for the PR.
I've updated it slightly, but haven't changed anything so I'll approve and publish.

PR #6652, opened 2023-10-16
Update environments-details.md

NielsKunst:
We had several questions on this in support so I feel it should be documented somewhere. Since it is related to contstant value after  (re)deployment this page seems more appropriate then the Constants information page.

Not sure about the warning banner or the location of the text. I will leave that to you.


dbreseman:
Thanks for raising this, @nelis37 (and thanks for reviewing, @katarzyna-koltun-mx). This new alert is now merged to development.

PR #6670, opened 2023-10-19
Support for saving changes to external entities

sbont:


sbont:
@Yiyun333 this story won't make it into 10.4, will close this PR.

Yiyun333:
@sbont Thanks for the update!

PR #6671, opened 2023-10-19
[WTF-1651] Document "Support configuring arguments for delete client action"

jeffersonRibeiro:
Updates the "on click event" "delete" now that it supports configuring arguments and allows it to select any object in scope to be deleted instead of just the nearest context object.


ConnorLand (approved):

ConnorLand:
I checked that this heading anchor change is good. It should not affect any links. If anything breaks, I'll fix it.

Thanks @jeffersonRibeiro for the update!

PR #6672, opened 2023-10-19
Usage of the Administration module

Luyao-Zhang-1:


content/en/docs/appstore/modules/administration.md
```diff
@@ -29,7 +29,43 @@ The [Administration](https://marketplace.mendix.com/link/component/23513) module
...
+6. Find **Account_Overview** using the search box on the top and select it. The **Select web page** dialog box closes.
+7. Click **OK** to save the settings for the new menu item.
+
+If you run the app, you can see the new page in the navigation.
+
+### 3.2 Adding Users as an Administrator
+
+1. Make sure that you [added the account overview page to the navigation](#add-account-overview).
+
+2. Run the app and log in to the app as the [Administrator](/refguide/app-security/#administrator).
```
MarkvanMents: I think it is better to say as a user who has the administrator role in the Administration module. It doesn't have to be the global administrator account.

content/en/docs/appstore/modules/administration.md
```diff
@@ -29,7 +29,43 @@ The [Administration](https://marketplace.mendix.com/link/component/23513) module
...
+
+1. In App Explorer, open **Navigation**.
+2. In **Menu**, click **New item**. The **New Menu Item** dialog box opens.
+3. Enter a caption.
+4. Select an icon.
+5. In the **On click** drop-down list, select **Show a page**. The **Select web page** dialog box opens.
+6. Find **Account_Overview** using the search box on the top and select it. The **Select web page** dialog box closes.
+7. Click **OK** to save the settings for the new menu item.
+
+If you run the app, you can see the new page in the navigation.
```
MarkvanMents: You can only see it if you are an administrator of the Administration module (or if security is turned off for local testing).

MarkvanMents (commented):
Hi Luyao,
Looks good.
I did a quick proofread here: ac898d606cd8a97ee82b5f2ff3b6e0fe58b8cbdf and I think that you need to be a bit more specific on who can access and use the account overview page.

PR #6677, opened 2023-10-23
Fix issue with CSP docs

Robinfr:
The code contained variables that would not be replaced.


Robinfr:
@Yiyun333 Can be merged as-is.

ConnorLand (approved):

PR #6681, opened 2023-10-23
Add note about on delete behaviour of external entities

sbont:
To be included in 10.4.


MarkvanMents (changes requested):
@sbont 
Hi Steven.
I've split the table into sections as originally the footnotes were just examples, but you are now adding additional information to those sections.
I've also added links to where the reader can find the Parent/Child information as I do not find it obvious which end of an association is the child.
Can you review and ensure that it still makes sense.

sbont:
HI @MarkvanMents, looks good to me. The only thing I had small doubts about are the 3 on delete behaviour options, that don't 100% reflect the option labels in Studio Pro anymore. It's close enough for me however, so if you're OK with that, I'm good with these changes!

MarkvanMents (approved):
> HI @MarkvanMents, looks good to me. The only thing I had small doubts about are the 3 on delete behaviour options, that don't 100% reflect the option labels in Studio Pro anymore. It's close enough for me however, so if you're OK with that, I'm good with these changes!

As discussed - the bullet points are summaries, but the headings do actually match the UX. Approving.

PR #6698, opened 2023-10-25
Update SAML TW- 2284

Karuna-Mendix:


Karuna-Mendix:
incorporated in https://github.com/mendix/docs/pull/6700

PR #6700, opened 2023-10-25
saml changes

Karuna-Mendix:
Linked to JIRA issue: https://mendix.atlassian.net/browse/TW-2284

### DIFF
--- a/content/en/docs/appstore/modules/saml.md
+++ b/content/en/docs/appstore/modules/saml.md
@@ -87,11 +87,7 @@ Some SAML services, such as eHerkenning and DigID in the Netherlands, use option
 
 If you need any of these features, contact your Mendix CSM to discuss inclusion of these features on the Mendix roadmap or customization of the SAML SSO module.
 
-{{% alert color="info" %}}
-Versions of the SAML module below 3.5.0 are limited to single keypair.
-{{% /alert %}}
-
-If you are using one of those older versions and you want to connect your app to multiple SAML IdPs, you cannot use different key pairs and certificates for each of the SSO federations. Instead, you must use a single key pair and certificate for all SAML IdPs. The certificate can be either a self-signed certificate or a certificate issued by a certificate authority (CA). For more details, see [Use a Certificate Issued by a Certificate Authority](#use-ca) section below.
+Versions of the SAML module below 3.5.0 are limited to single keypair. If you are using one of those older versions and you want to connect your app to multiple SAML IdPs, you cannot use different key pairs and certificates for each of the SSO federations. Instead, you must use a single key pair and certificate for all SAML IdPs. The certificate can be either a self-signed certificate or a certificate issued by a certificate authority (CA). For more details, see the [Use a Certificate Issued by a Certificate Authority](#use-ca) section below.
 
 If you use both the [OIDC SSO](/appstore/modules/oidc/) module and the SAML module in the same app, each end-user can only authenticate using one IdP.
 
--- a/content/en/docs/appstore/modules/saml.md
+++ b/content/en/docs/appstore/modules/saml.md
@@ -62,6 +62,10 @@ The SAML module keeps a log/audit trail of login attempts. These can be download
 
 The SAML module allows you to have an SSO connection with multiple SAML IdPs. Each IdP can have its own keypair.
 
+{{% alert color="info" %}}
+The version of the SAML module 3.5.0 and above compatible with Mendix (v9.22.0 and above) supports multiple keypair.
+{{% /alert %}}
+
 ### 1.3 Limitations {#limitations}
 
 The Mendix SAML SSO module does not support the following:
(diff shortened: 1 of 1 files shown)


content/en/docs/appstore/modules/saml.md
```diff
@@ -85,6 +89,10 @@ If you need any of these features, contact your Mendix CSM to discuss inclusion
```
JaapF, line 89: I would suggest inserting that new line of text here.
"The version of the SAML module below 3.5.0 is limited to single keypair: ..."
The "If you want [...] more details)." sentences explain that limitation in older versions in more detail.

Karuna-Mendix: Done https://github.com/mendix/docs/pull/6700/commits/8d7d61b8e6f639bba1da4a82d1b1da21b1f5ba74

content/en/docs/appstore/modules/saml.md
```diff
@@ -83,6 +87,10 @@ Some SAML services, such as eHerkenning and DigID in the Netherlands, use option
 
 If you need any of these features, contact your Mendix CSM to discuss inclusion of these features on the Mendix roadmap or customization of the SAML SSO module.
 
+{{% alert color="info" %}}
+The version of the SAML module below 3.5.0 is limited to single keypair.
+{{% /alert %}}
+
 If you want to connect your app to multiple SAML IdPs, you cannot use different key pairs and certificates for each of the SSO federations. Instead, you must use a single key pair and certificate for all SAML IdPs. The certificate can be either a self-signed certificate or a certificate issued by a certificate authority (CA) (see [Use a Certificate Issued by a Certificate Authority](#use-ca) for more details).
```
JaapF: If you're using one of those older versions and you want to connect ....

content/en/docs/appstore/modules/saml.md
```diff
@@ -83,6 +87,10 @@ Some SAML services, such as eHerkenning and DigID in the Netherlands, use option
 
 If you need any of these features, contact your Mendix CSM to discuss inclusion of these features on the Mendix roadmap or customization of the SAML SSO module.
 
+{{% alert color="info" %}}
+The version of the SAML module below 3.5.0 is limited to single keypair.
```
MarkvanMents: I think there are several versions below 3.5.0?
```suggestion
Versions of the SAML module below 3.5.0 are limited to single keypair.
```

content/en/docs/appstore/modules/saml.md
```diff
@@ -62,6 +62,10 @@ The SAML module keeps a log/audit trail of login attempts. These can be download
 
 The SAML module allows you to have an SSO connection with multiple SAML IdPs. Each IdP can have its own keypair.
 
+{{% alert color="info" %}}
+The version of the SAML module 3.5.0 and above compatible with Mendix (v9.22.0 and above) supports multiple keypair.
```
MarkvanMents: I think there will be more versions, so I think we need to say:
```suggestion
SAML module versions 3.5.0 and above (compatible with Mendix version 9.22.0 and above) support multiple keypairs.
```

MarkvanMents (changes requested):
Looking good. We just need to be prepared for multiple versions of the module. The versioning of this module in particular is very complicated as there are different versions of the module for different Mendix versions, which is why the version information is complicated to add.

Let me know if you have any questions.

content/en/docs/appstore/modules/saml.md
Karuna-Mendix: I agree with these changes

content/en/docs/appstore/modules/saml.md
Karuna-Mendix: Done https://github.com/mendix/docs/pull/6700/commits/d471f799f12f42326c32be29efbaae2ed96a7eb4

content/en/docs/appstore/modules/saml.md
Karuna-Mendix: Done https://github.com/mendix/docs/pull/6700/commits/d471f799f12f42326c32be29efbaae2ed96a7eb4

content/en/docs/appstore/modules/saml.md
MarkvanMents: I think that this should just be in the list of what is not supported (before the bit about DigID)? It doesn't need to be an alert as this section is already listing limitations.
In fact, I think that it is related to the paragraph which is already here about the SAML EntityID (highlighted in the attached image) which is, I think, a technical description of what happens in versions below 3.5.0, rather than a customer-friendly description.

Can you discuss with @JaapF ? Thanks.
image

MarkvanMents (changes requested):
@Karuna-Mendix - let me know if you want to discuss this, or pick it up with Jaap if you understand what I'm trying to say.

content/en/docs/appstore/modules/saml.md
Karuna-Mendix: The highlighted paragraph is about all versions of SAML module and not only for versions below 3.5.0. It is not completely related to versions 3.5.0 and below (the info in the alert box). Alert info doesn't go with the bulleted list as well as that list is about what is not supported for all versions. I removed the info alert and added the sentence in the paragraph where it talks about older version of the module and its related behavior.

MarkvanMents (approved):
Looks good. Thanks for resolving my question.
OK to go if Jaap is happy.

PR #6706, opened 2023-10-25
Update _index.md

NielsKunst:
Added information that is also accessible here: https://docs.mendix.com/releasenotes/mobile/nt-studio-pro-9-parent/


ConnorLand:
@NielsKunst closing this, as even more granular information has been added here (vis a vis which NTs go with which SPs) https://docs.mendix.com/releasenotes/mobile/native-template/

PR #6708, opened 2023-10-26
[WC-2182] Add docs about Export to Excel

iobuhov:
A quick data grid docs update where I'm adding a section about data export.

### DIFF
--- a/content/en/docs/appstore/modules/data-widgets/data-grid-2.md
+++ b/content/en/docs/appstore/modules/data-widgets/data-grid-2.md
@@ -430,7 +430,17 @@ If a filter is being used and its type does not match with any selected attribut
 There cannot be more than one filter of the same type in the data grid header. For example, if a data grid header already has a `Number Filter`, adding another `Number Filter` will cause incorrect data filtering.
 {{% /alert %}}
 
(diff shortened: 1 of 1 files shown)


ConnorLand:
Thanks for the PR @iobuhov . Please change all the - bullets back to * the way they were. I think your IDE does this automatically. We prefer *s on the docs site. Also please change the "tags" section back to the way it was.

I'll do language edits on your new section. It looks great. Thanks for the submission.

iobuhov:
Hi @ConnorLand, I'm wondering if it's possible to define these rules in the root of the repo using prettierrc or any other formatter configurations, so next time my IDE will know about your formatting preferences. Just a small suggestion.

ConnorLand:
> Hi @ConnorLand, I'm wondering if it's possible to define these rules in the root of the repo using prettierrc or any other formatter configurations, so next time my IDE will know about your formatting preferences. Just a small suggestion.

Oh, that's a cool idea. Do you know of any repos that currently do this so we can look at them as an example? If we can, I'll try and make it happen.

ConnorLand (approved):

ConnorLand (approved):

PR #6715, opened 2023-10-27
Wpenris2 patch 2

wpenris2:
-This pull request adds an example java action to replace a workflow context entity.

Just making sure everything is now properly pulled, so here is a copy of the code, below (dont want to do this three times ;-) Please let me know if I did it correctly this time.

#### 3.2.2 Context Entity Replaced

When an app developer changes the **Workflow Context** entity of a workflow, existing instances still run on the old **Workflow Context** entity. This entity may not be compatible with the new **Workflow Context** entity.

You can do one of the following: 

* The workflow can be aborted, for example, by using the **DefaultWorkflowAdmin** page in the Workflow Commons.
* The app developer can create a Java action that uses the `void setWorkflowObject(contextObject: IMendixObject)` method on the `com.mendix.workflows.Workflow` interface to manually set the context object of the workflow parameter context for the right entity. For more information on Java actions, see [Extending Your Application with Custom Java](/refguide/extending-your-application-with-custom-java/).
   * Here is an example of an possible Java Action set up within Mendix, which in this case, sets a new $History object as context: Alt text
   * Here is the example code to be used for the Java action:

```java
import com.mendix.workflows.Workflow;   //add this import up top

// BEGIN USER CODE
		try {
			Workflow myWorkflow =  com.mendix.core.Core.workflows().getWorkflow(getContext(), __WorkflowInput);    //get the proper type Workflow object to work with from the input parameter __WorkflowInput
			myWorkflow.setWorkflowContext(__ContextReplacement);                                                   //now we have the proper type, we can call the setWorkflowContext() function, that replaces the context with __ContextReplacement
			return true;
		}catch (Exception e){
			return false;
		}
// END USER CODE
``` 

* The app developer can revert changes to the original **Workflow** **Context** entity and deploy this change.

To prevent this issue, you need to make sure that context objects cannot be deleted while they are still being used in running workflow instances.


content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -81,6 +81,24 @@ You can do one of the following:
 
 * The workflow can be aborted, for example, by using the **DefaultWorkflowAdmin** page in the Workflow Commons.
 * The app developer can create a Java action that uses the `void setWorkflowObject(contextObject: IMendixObject)` method on the `com.mendix.workflows.Workflow` interface to manually set the context object of the workflow parameter context for the right entity. For more information on Java actions, see [Extending Your Application with Custom Java](/refguide/extending-your-application-with-custom-java/).
+   * Here is an example of an possible Java Action set up within Mendix, which in this case, sets a new $History object as context: ![Alt text](/content/en/docs/refguide/modeling/application-logic/workflows/JavaActionWorkflowContextReplacement.jpg?raw=true "An example of the Java Action Set Up Necessary to Replace a Workflow Context Object")
```
DmytroKost: example of **a** possible

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -81,6 +81,24 @@ You can do one of the following:
...
+   * Here is the example code to be used for the Java action:
+
+```java
+import com.mendix.workflows.Workflow;   //add this import up top
```
DmytroKost: I would add a space before each comment, e.g.

```java
import com.mendix.workflows.Workflow;   // add this import statement up top
```

same on lines 92, 93

and I would rephrase the comment to "add this import to the top"

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -81,6 +81,24 @@ You can do one of the following:
...
+
```
DmytroKost: Replace the code block with (better formatting)

```java
// BEGIN USER CODE
try {
  var workflow = com.mendix.core.Core.workflows().getWorkflow(getContext(), __WorkflowInput); // get the Workflow interface object
  workflow.setWorkflowContext(__ContextReplacement); // replace the current context object with the new one
  return true;
} catch (Exception e) {
  return false;
}
// END USER CODE
```

DmytroKost (commented):
Left comments mostly regarding formatting. Otherwise, looks good.

Yiyun333:
> Left comments mostly regarding formatting. Otherwise, looks good.

@DmytroKost  Thanks a lot for the review! I will update the doc based on your comments. 👍

Yiyun333:
@wpenris2 Thank you for sending us the PR and for the initiative to provide the example! I've asked for a technical review from the team. I will proceed with the comments. After that, I will merge the PR.

PR #6727, opened 2023-10-30
Add notes on complex attributes

sbont:


content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/consumed-odata-service-requirements.md
```diff
@@ -72,11 +74,19 @@ The most commonly used attribute types can be used in your app. The types of the
 | String, Guid                   | String |
 | (Other)                        | (Ignored) |
 
+<small><sup>[1]</sup> In Mendix, Booleans cannot be null. If the service returns null, the value will be false in Mendix.<br /><sup>[2]</sup> Decimal values outside of the range of a [Mendix decimal](/refguide/attributes/#type) are currently not supported. If the service returns a value outside of the range, there will be an error.</small>
```
joostverhoog: The recommended way to refer to the product is "Studio Pro", not "Mendix" (multiple times in this PR)

sbont: This particular bit was only moved (see below); in general I'd find it weird to talk about runtime concepts using "Studio Pro". Are you sure about this? I couldn't find helpful pointers in the R&D handbook about this...

MarkvanMents: This is a discussion which doesn't have a resolution yet. Our style guide says that you can apply version to [**Mendix**, **Studio Pro**, and **Mendix Runtime**](https://mendix.atlassian.net/wiki/spaces/RNDHB/pages/2510356519/Terminology#v-for-version). I tend to agree with Steven, that it doesn't make sense to talk about Studio Pro when referring to runtime versions. But we don't have firm advice on this.

(I'm hoping that at some time in the future, the strict 1-1 relationship between Studio Pro version and Runtime might be loosened, so that patch versions of Studio Pro can all use the same runtime, but I don't see that happening in the next 5 years!)

joostverhoog: So my suggestion is "In Studio Pro, Booleans cannot be null. If the service returns null, the app will use the value `false`."

sbont: Changed.

joostverhoog: Looks good to me!

joostverhoog (approved):

content/en/docs/refguide/modeling/domain-model/entities/external-entities.md
```diff
@@ -92,12 +92,14 @@ The **Edit Attribute** dialog can be used to specify a local name and add a loca
...
 * **General**
     * **Name** – a local name for the attribute
-    * **Original Name** – a read-only value that displays the original name of the attribute in the originating app
+    * **Original Name** – this is a read-only value that displays the original name of the attribute as given in the originating app ¹
     * **Summary** – a read-only summary that displays the description for the attribute in the originating app; local descriptions are entered in the [Documentation tab](#documentation)
     * **Type** – read-only values for the **Type**, **Length**, and **Max length** of the attribute as defined in the originating app
     * **Default value** - the default value for this attribute when the app creates an object (visible only when the entity is creatable)
 * **Documentation** – a description for the attribute as defined in the originating app
 
+¹ For attributes that represent a property of a complex attribute, a forward slash (`/`) separates the name of the attribute of the entity and the name of the property of the complex type.
```
MarkvanMents, line 101: @sbont 
Hi Steven - here you say that the property is separated by a slash, but in the other document you say that they are separated by an underscore.

Are these two different things, or should they be the same?

content/en/docs/refguide/modeling/domain-model/entities/external-entities.md
```diff
@@ -92,12 +92,14 @@ The **Edit Attribute** dialog can be used to specify a local name and add a loca
 
 * **General**
     * **Name** – a local name for the attribute
-    * **Original Name** – a read-only value that displays the original name of the attribute in the originating app
+    * **Original Name** – this is a read-only value that displays the original name of the attribute as given in the originating app ¹
```
MarkvanMents:
```suggestion
    * **Original Name** – a read-only value that displays the original name of the attribute in the originating app ¹
```

content/en/docs/refguide/modeling/domain-model/entities/external-entities.md
sbont: Hey, they are different: the external name is separated by an underscore, for the local Mendix name we use a default name that uses an underscore

MarkvanMents: Thanks for the explanation. I'll resolve this.

sbont:
@MarkvanMents this feature didn't make the freeze and will be postponed to 10.6.

MarkvanMents:
@sbont 
Thanks, Steven. I've updated the label so it doesn't get merged accidentally.
Can you let Quinn and me know if this is now OK from your side so we can approve it and forget about it.
Or is there a possibility that the features will change between now and middle of December?

sbont:
Hi @MarkvanMents and @quinntracy, I've only changed the note on the version in which this feature was introduced. Otherwise it's fine from our side - we've just finished up this feature so we're good for 10.6.

PR #6737, opened 2023-10-31
Added advise for choosing version control system in Mendix 9

tknippenberg:


content/en/docs/refguide9/version-control/svn-git-differences.md
```diff
@@ -46,7 +48,13 @@ In SVN, commits are done to a central server which enforces the commit order. Th
 
 In Git, committing is done locally at first. Then commits are sent to other repositories and they should be uniquely identifiable. Therefore, commits in Git are represented with a SHA-1/SHA-256 hash (e.g. f0e165, bb2327, 76d34e, c31247), as they can be generated in a distributed setting and still be the same on different clients with identical changes.
 
-## 4 Proxy Support {#proxy-support}
+## 4 Local copy of entire repository
+
+Git's local copy of the entire history is a core feature of its distributed version control system. Every user who clones a Git repository essentially creates a full copy of the entire project, including its entire history. 
+
+This local history allows for complete independence, enabling users to work offline, commit changes, and explore the project's history seamlessly without the need for a constant connection to a central server. A downside of this is that a fresh checkout (called 'clone' in Git), will taker longer compared to SVN. Where a checkout in SVN meant downloading the last revision (for example 100MB), doing a clone in Git means downloading the entire repository (for example 1.2GB). 
+
```
tknippenberg, line 58: Docs team, is it sensible to add something along the lines of "Mendix is working on shrinking the MPR and/or repository to mitigate this issue"? Or can we leave it open-ended for now?

content/en/docs/refguide9/version-control/svn-git-differences.md
```diff
@@ -46,7 +48,13 @@ In SVN, commits are done to a central server which enforces the commit order. Th
 
 In Git, committing is done locally at first. Then commits are sent to other repositories and they should be uniquely identifiable. Therefore, commits in Git are represented with a SHA-1/SHA-256 hash (e.g. f0e165, bb2327, 76d34e, c31247), as they can be generated in a distributed setting and still be the same on different clients with identical changes.
 
-## 4 Proxy Support {#proxy-support}
+## 4 Local copy of entire repository
```
MarkvanMents: I don't really like this as a section heading. It doesn't indicate why the reader should read this section. I would suggest something like:
* Support for Fully Distributed Version Control

content/en/docs/refguide9/version-control/svn-git-differences.md
MarkvanMents: I've rewritten the above section a bit to focus a bit more on the "why" rather than the "what", but I'm happy to go back if you prefer the original.

I don't think we should mention work on the MPR or repository here as:
* it is an aspiration and we can't say anything definite
* it is not going to shrink an existing Git repo as the history will still be there
* any changes in the repository or MPR structure are unlikely to be backported to Mx9 as it would cause issues for upgrading to existing Mx10 versions.

MarkvanMents (dismissed):
Hi @tknippenberg 
Thanks for the PR, Tom. I've made a couple of comments and have also changed the new section a bit. Can you have a look and let me know if you are happy with the (proposed) changes.
Thanks.

content/en/docs/refguide9/version-control/svn-git-differences.md
tknippenberg: Fine with me!

content/en/docs/refguide9/version-control/svn-git-differences.md
tknippenberg: The rewritten looks good to me.

I'll elaborate a bit more on my considerations for mentioning 'planned MPR format changes':

- Currently the MPR size is really a problem and we don't have a solution or workaround for it yet. 
- Certain Git vendors (specifically Github, maybe others as well) can't be used if the MPR is larger than 100MB which is a significant issue for organizations. 
- A large MPR (and the fact that we use binaries at all) leads to rapid repository growth, which results in issues everywhere (Studio Pro, server side, Developer Portal, ...).

This makes it an extremely painful situation, which is why I thought it could be worthwhile mentioning that something is on the roadmap.

Your other bullets are relevant as well however.
- For 'shrinking existing Git repo' we are also planning work, but that's not in-progress yet. I also wouldn't know how to put that in this document with the limited info we have today...
- Changes to the MPR are unlikely to be backported to 9 LTS: yes, normally it would have been. However, this is such a problematic issue for customers that have already migrated to Git, while planning to remain on 9 LTS for a long time, that we're actively investigating exceptions. An exception could be changing the format in 9 LTS if a certain setting is enabled, which would break migration to 10.0-10.5, as the new format is obviously not supported there yet.


All in all, I think offering some kind of perspective on "we're working on this major issue" could be useful to add in this document. On the other hand a lot is uncertain (how/when we'll solve it).

If you have ideas how to deal with this wall of text, I'd be happy to hear your input! :)

tknippenberg: Agreed, let's leave it out.

MarkvanMents (approved):
Approved over Slack.

PR #6743, opened 2023-11-01
Add Retrieve Workflows Activity

DmytroKost:
For 10.5


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/retrieve-workflows.md
```diff
@@ -0,0 +1,49 @@
...
+{{% alert color="warning" %}}
+This activity can only be used in **Microflows**.
+{{% /alert %}}
+
+## 1 Introduction
+
+A **Retrieve workflows** activity can be used to retrieve a list of **System.Workflow** objects.
+
+It needs the [Workflow Context](/refguide/workflow-parameters/) object as its input. A given Workflow Context object can be used to instanciate one or several workflow instances. It will break at Runtime if the passed Workflow Context object is empty.
+
```
PaulJongenMx, line 23: It needs **a** [Workflow Context](/refguide/workflow-parameters/) object as its input.
Pls rephrase these lines (it's too techn ical and not clear why it's mentioned here): "A given Workflow Context object can be used to instanciate one or several workflow instances. It will break at Runtime if the passed Workflow Context object is empty."
@DmytroKost is this true and what does it mean? "It will break at Runtime if the passed Workflow Context object is empty." To my knowledge I can create an object, set no property and start a WF. So the context object may be empty (no attributes has a value), but here has to be an object. @Yiyun333 How to write this properly?

PaulJongenMx, line 23: Add: Use this activity when you want to know if the context object is already used in a workflow, and if so which workflow is used to process the context object

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/retrieve-workflows.md
```diff
@@ -0,0 +1,49 @@
...
+### 3.1 Workflow context
+
+An object of the Workflow Context entity available from the microflow parameters.
+
+### 3.2 Variable Name
+
+The name of the variable returned by the activity.
+
+## 4 Common Section {#common}
+
```
PaulJongenMx, line 54: @Yiyun333 what is a good location in this page the add the following note:
Note: The list of workflow objects this activity returns are all of type System.Workflow but can be related to different/multiple? workflow definitions (System.WorkflowDefinition).

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/retrieve-workflows.md
DmytroKost: By empty object, we mean when it's `== empty`, in runtime it's `null` (e.g. the object could be deleted). We have a similar warning for the Retrieve Workflow Context activity.

> It will break at Runtime if the passed workflow object is not of the right type.

But indeed I think we can omit it here.

Yiyun333: Thanks both for the information. According to the discussions above, how about the following version:

'It needs **a** [Workflow Context](/refguide/workflow-parameters/) object as its input. The **provided** Workflow Context object here can be used to **represent** one or several workflow instances."

@DmytroKost Should we keep the warning for the Retrieve Workflow Context activity?

Yiyun333: I've adjusted the sentence a bit and added the following in [this](https://github.com/mendix/docs/pull/6743/commits/3c88b23787cf083b050e76770ca51f5fa7ba8ca0) commit:

`You can use this activity to check if the Workflow Context object is already used in a workflow, and if so, which workflow is used to process the Workflow Context object.`

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/retrieve-workflows.md
Yiyun333: The note is added at the end of the introduction section. See the same commit as mentioned above.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/workflow-activities/retrieve-workflows.md
DmytroKost: > Should we keep the warning for the Retrieve Workflow Context activity?

I think it's okay to remove it

> 'It needs a [Workflow Context](https://github.com/refguide/workflow-parameters/) object as its input. The provided Workflow Context object here can be used to represent one or several workflow instances."

I would put it "The provided Workflow Context object here can be associated with one or several workflow instances.", @PaulJongenMx What do you think?

PaulJongenMx: > The provided Workflow Context object here can be associated with one or several workflow instances.",

@DmytroKost Yes this is better. Thx.
I prefer to keep the line about the runtime error (not use the "will break" wording), just wondering how the correctly write when the parameter contains no object.

Yiyun333: Okay, I think this is a bit too less information for me to come up with a sentence that makes sense. @DmytroKost do you have time this afternoon and can you please show me a small demo on what the parameter and object actually mean in the current context? :) :)

Yiyun333: Hi @PaulJongenMx, Dmytro and I just had a very nice meeting in which he explained to me all the terms that I found confusing in workflows, including the differences between **WorkflowContext** parameter and **WorkflowContext** object. We've also discussed how to describe the Runtime error better. What do you think of the following?

> A Runtime error occurs if the [WorkflowContext parameter](https://docs.mendix.com/refguide/workflow-parameters/) contains no object.

PaulJongenMx: There could be more than 1 WF for Context object. Can we change to:
"You can use this activity to check if the Workflow Context object is already used in a workflow, and if so, which workflow**s are** used to process the Workflow Context object.

PaulJongenMx: Yes that's good, but it's singular so we should omit the last "s"
"... contains no object."

Yiyun333: I just edited my comment (deleted the last "s") but I guess I was 10 seconds late :)

Yiyun333: Changes are commited here: https://github.com/mendix/docs/pull/6743/commits/2ece0921562ff10f31968fdb192768afa51ba26d

Yiyun333: Changes are commited here too: https://github.com/mendix/docs/pull/6743/commits/2ece0921562ff10f31968fdb192768afa51ba26d

Yiyun333 (approved):

PR #6754, opened 2023-11-01
Clarify technology preview status of protected modules in Mendix 9 and availability in 10

dlhartveld:
Mendix 9 contains an incomplete, unsupported version of protected modules, which was intentionally never released as generally available. Documentation has been clarified to point customers to Mendix 10, which contains a generally available and fully supported version of protected modules.

This has led to some unclarity with customers and this pull request intends to solve this problem.

Suggested text included in this pull request. Completeness should be verified, this PR might miss occurrences of the infobox that needs to be updated.

### DIFF
--- a/content/en/docs/refguide9/modeling/app-explorer/modules/_index.md
+++ b/content/en/docs/refguide9/modeling/app-explorer/modules/_index.md
@@ -27,7 +27,7 @@ React Native modules expose native Java/Objective-C and C++ objects allowing for
 ## 2 Module Types {#module-types}
 
 {{% alert color="info" %}}
-Access to this functionality is currently limited and can be gained through the [Mendix Component Partner Program](/appstore/creating-content/partner-program/) and the [Mendix Commercial Solution Partner Program](https://www.mendix.com/partners/become-a-partner/isv-program/).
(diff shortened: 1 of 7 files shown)


content/en/docs/refguide9/modeling/app-explorer/modules/_index.md
```diff
@@ -27,7 +27,7 @@ React Native modules expose native Java/Objective-C and C++ objects allowing for
 ## 2 Module Types {#module-types}
 
 {{% alert color="info" %}}
-Access to this functionality is currently limited and can be gained through the [Mendix Component Partner Program](/appstore/creating-content/partner-program/) and the [Mendix Commercial Solution Partner Program](https://www.mendix.com/partners/become-a-partner/isv-program/).
+In Mendix 9, Add-on modules and Solution modules were previously available in preview for select customers and partners as part of an early access program. In Mendix 10, this functionality is generally available and ready for use by any customer or partner.
```
Adam-Dupaski: @dlhartveld is it correct here to say "were previously available" (instead of "are available") ? Meaning, these are not available for Mx9 via those programs anymore?

dlhartveld: We can selectively give access but should not advertise that.

Adam-Dupaski (approved):

PR #6764, opened 2023-11-03
Length function changes

Robinfr:
In 10.5 the `length` function will also start accepting list variables. To document this I've moved the function out of the string-expression documentation and into its own section.

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/expressions/length.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/length.md
@@ -11,7 +11,7 @@ Basic function to determine the length of a string or a list.
 
 ## 2 length
 
-Determines the length of a string or a list and outputs the length as an integer.
+Determines the length of a string or a list and outputs the value as an integer.
 
 ### 2.1 Input Parameters
 
@@ -38,4 +38,5 @@ The output is described in the table below:
 | `length('thisismystring')` | 14 |
 | `length($MyString)` and `MyString = 'qwer'` | 4 |
 | `length($MyString)` and `MyString` is empty | 0 |
-| `length($EntityAList)` | 12 |
+| `length($EntityAList)` and `EntityAList = '[MxObject, MxObject, MxObject]'`| 3 |
+| `length($EntityAList)` and `EntityAList = '[]'`| 0 |
--- a/content/en/docs/refguide/modeling/application-logic/expressions/length.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/length.md
@@ -1,0 +1,41 @@
+---
+title: "Length"
+url: /refguide/length/
+weight: 130
+tags: ["studio pro", "length", "expression", "expressions"]
+---
+
+## 1 Introduction
+
+Basic function to determine the length of a string or a list.
+
+## 2 length
+
+Determines the length of a string or a list and outputs the length as an integer.
+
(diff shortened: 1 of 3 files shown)


content/en/docs/refguide/modeling/application-logic/expressions/length.md
```diff
@@ -0,0 +1,41 @@
...
+tags: ["studio pro", "length", "expression", "expressions"]
+---
+
+## 1 Introduction
+
+Basic function to determine the length of a string or a list.
+
+## 2 length
+
+Determines the length of a string or a list and outputs the length as an integer.
```
EnasAbdelrazek: Just one small suggestion: can it be changed to "Determines the length of a string or a list and outputs the value as an integer" or just "Returns the length of a string or list as an integer"?

content/en/docs/refguide/modeling/application-logic/expressions/length.md
```diff
@@ -0,0 +1,41 @@
...
+| Length of the list   | Integer |
+
+### 2.3 Examples
+
+| Input | Output |
+| --- | --- |
+| `length('thisismystring')` | 14 |
+| `length($MyString)` and `MyString = 'qwer'` | 4 |
+| `length($MyString)` and `MyString` is empty | 0 |
+| `length($EntityAList)` | 12 |
```
EnasAbdelrazek: Where is the list value?

Robinfr: Any good ideas for how to write an example for the list variable? I couldn't come up with any good ways of doing so :-/

EnasAbdelrazek: an example with an empty list [] is the easiest one :D

EnasAbdelrazek: Maybe we can add a list of empty objects :/

content/en/docs/refguide/modeling/application-logic/expressions/length.md
Robinfr: Changed.

content/en/docs/refguide/modeling/application-logic/expressions/length.md
Robinfr: Done

EnasAbdelrazek (approved):

content/en/docs/refguide/modeling/application-logic/expressions/length.md
RCatersels: It might be me but I find it very strange that length on a list variable gives me the same as the list operation count list. These are string function calls that now get mixed up to get a count on a list. So I fail to grasp why the length function got extended to lists. Very confusing imho.

Robinfr:
@RCatersels Moving the conversation separately from the review comments from Enas. To answer your question, to make it easier to build loops around list size using the While loop + allowing users to use the size of a list for conditional visibility. 

In my experience, it's made it far easier to quickly build a loop that goes over a certain range.

RCatersels:
True, but it is still strange to now have to different actions for the same result. And I get the conditional visibility. But I would rather see that I could use microflows or naanoflows for conditional visibility because this gives me a lot more flexibility.

Yiyun333 (approved):

PR #6766, opened 2023-11-03
Portunus 5 3 1

Nidhi251289:
Resolved an issue where ephemeral storage requests and limits were always set to 1 GB.
Resolved the issue where user was unable to import a large bundle of licenses at one time.


katarzyna-koltun-mx (approved):

katarzyna-koltun-mx:
Waiting for the go ahead to publish.

PR #6786, opened 2023-11-09
Exposing a resource terminology

joostverhoog:
Studio Pro 10.5.0 clarifies the terminology for publishing an entity in a published OData service.

This used to be "Exposing a resource", now it's "Publishing an entity". The former is not wrong, but outside-in, where we prefer inside-out when modeling. There are some exceptions:

* When comparing OData to REST it is clearer to keep referring to published entities as resources
* Attributes still have "exposed names" (not "published names")

Note that this involves a rename of a page that is references by Studio Pro, so it needs a redirect.

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/entities/external-entities.md
+++ b/content/en/docs/refguide/modeling/domain-model/entities/external-entities.md
@@ -19,7 +19,7 @@ To add an external entity to your app model, follow these steps:
 
 1. In the domain model of your app, use the Integration pane to search for the entity or data source you want to use. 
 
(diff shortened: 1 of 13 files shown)


content/en/docs/refguide/modeling/integration/odata-services/published-odata-services/published-odata-resource.md
MarkvanMents, line 1: I can see that this file is deleted, but you say it should have been renamed. Should there be a new file with this content?

Also, if you rename a file can you do that in a separate commit (with no other changes) as this helps Git identify the rename.

joostverhoog: I don't see it as deleted, but as renamed to published-odata-entity. Are you sure you are looking at the latest version?

quinntracy (approved):

PR #6792, opened 2023-11-10
Update using-the-excel-exporter.md

mahendra-mahajan-mendix:
update excel-exporter how-to document based on user feedback. The distinction between template with and without input object is mentioned along with deleteaftterdownload for the file Object.


Luyao-Zhang-1:
@quinntracy I know a bit more about the context of this issue. Let's work on this PR together :)

PR #6795, opened 2023-11-10
9.24.12 RN

Adam-Dupaski:


KevinVlaanderen:
The following notes were missing from the initial changelog:

```
### Improvements
    
- We updated the bundled JDK to version 11.0.20.1.

### Fixes
    
- We fixed an issue in consumed web services where doing requests with attachments would fail due to a missing `Content-Transfer-Encoding` header. (Ticket 199235)
- We fixed an issue where the LastLogin attribute was not readable when it should be. (Ticket 192759)
- We fixed an issue where the session id was incorrectly `null` in WebSockets. (Ticket 195691)

### Deprecations
    
- We dropped support for PostgreSQL 11, as it is no longer supported by the vendor.

```

PR #6799, opened 2023-11-10
Update _index.md

jdiehl:
Added warning about API level


jdiehl:
@ConnorLand 
I don't think this warning needs to be placed anywhere else.

ConnorLand (approved):

PR #6804, opened 2023-11-13
Reword Catalog to be more generic

quinntracy:


content/en/docs/catalog/register/automate-registration.md
```diff
@@ -11,7 +11,7 @@ aliases:
 
 ## 1 Introduction
 
-The [Catalog](/catalog/) is a metadata hub for managing shared registered assets made available through OData services. Registering data to the Catalog happens automatically when deploying an app with a published OData service to Mendix Cloud.
+The [Catalog](/catalog/) is a metadata hub for managing shared registered assets made available through OData and OpenAPI REST services. Registering data to the Catalog happens automatically when deploying an app with a published service to Mendix Cloud.
```
NammNguyen: We're still struggling to come up with the correct terms here. Sjoerd sent out an email suggesting "Published REST" as the servicetype term, but hasn't gotten a response yet. In this file content/en/docs/catalog/manage/user-roles.md you just used "OpenAPI", I would suggest to keep using the same term

content/en/docs/catalog/manage/user-roles.md
```diff
@@ -11,7 +11,7 @@ aliases:
 
 ## 1 Introduction
 
-You can use the Catalog to search for and consume registered OData services. You can also register new services and curate your own services.
+You can use the Catalog to search for and consume registered OData and OpenAPI services. You can also register new services and curate your own services.
```
quinntracy, line 14: @NammNguyen to match your previous comment, would this be updated to "You can use the Catalog to search for and consume published REST services." ?

PR #6809, opened 2023-11-15
Catalog 2.100 release

NammNguyen:


layouts/partials/landingpage/latest-releases.html
```diff
@@ -16,7 +16,7 @@
                                                 <p class="rn-date">Oct 26, 2023</p>
 
 	<li class="link-list"><a href="/releasenotes/data-hub/">Catalog</a></li>
```
NammNguyen, line 18: Should `/releasenotes/data-hub/` not refer to `/releasenotes/catalog/` now? I wasn't sure, so I didn't make the change

MarkvanMents: We have implemented an automatic forwarding, so this will still work.
However, we could go through and update all the links so we can get rid of the automatic forwarding.
I think this should be a separate exercise as we also need to ensure that all the links in the product have been updated before we remove the forwarding.

quinntracy (approved):

content/en/docs/releasenotes/catalog/_index.md
```diff
@@ -17,7 +17,7 @@ These release notes cover changes made to the [Catalog](/catalog/).
 
 #### Fixes
 
-* We fixed an issue where some applications would not show up in the dropdown of the registration form.
+* We fixed an issue where some applications would not show up in the drop-down of the registration form.
```
Adam-Dupaski: "drop-down menu" via SG @quinntracy :)

"would not" > "did not" via SG - use simple tenses

PR #6831, opened 2023-11-17
Add Fixed Known Issue about extra commits

MarkvanMents:
Known issue was never published, and is now fixed.


MarkvanMents:
Approved over Slack.
Thanks for the review @Adam-Dupaski

PR #6839, opened 2023-11-21
Added clarification of MPR size limit

tknippenberg:


MarkvanMents (approved):
Thanks for the PR Tom.
I'll merge it immediately.

PR #6865, opened 2023-11-24
Update 10.5.md

yunyilmaz:
change whitelist to suggestion list


Adam-Dupaski:
Thanks for picking this up @yunyilmaz - we have using "safe list" as the recommendation in our Style Guide, so I've changed it to that.

PR #6880, opened 2023-11-29
Remove reference to removed license details

feddovanede:
Some license details were removed from the Environment Details page, as they were no longer relevant to customers. Documentation is now in line with the page.


MarkvanMents (approved):
Looks good @feddovanede
Will merge now.
@dbreseman - FYI when you are back.

PR #6886, opened 2023-11-30
Reword Catalog docs 

quinntracy:


quinntracy:
Before I make additional changes, I have a couple questions/thoughts:

1. The -[ Register OData Resources](https://docs.mendix.com/catalog/register/register-data/) article has specific OData instructions and links to the article [Published OData Services](https://docs.mendix.com/refguide/published-odata-services/). They are both OData-specific and reference specific OData terms and processes. Therefore, I think it may make more sense to keep these as OData resources, as updates to these articles would make many, widespread changes to multiple Catalog articles. 
2. Would it make sense to change the [Register Non-OData Resources in the Catalog](https://docs.mendix.com/catalog/register/register-non-odata-resources/) article into a more generic "Register Resources in the Catalog" so it can cover anything that is non-OData, such as OpenAPI and any other technologies that are to be added to the Catalog? If so, we would need to verify which steps are the same between technologies and which need specification.
3. Are the OpenAPI/non-OData registration steps the same or similar to the OData steps? If they're similar enough, the article [Register Data Sources (On-Prem or Private Cloud)](https://docs.mendix.com/catalog/register/data-sources-without-mendix-cloud/) can likely be made generic to cover all registration processes. 
4. The articles [Share Data Between Apps](https://docs.mendix.com/catalog/share-data/) and [Write Data to Another App ](https://docs.mendix.com/catalog/write-data/) also have OData-specific steps documented. Similar to #3, if the steps are the same/similar enough to OpenAPI (and upcoming technologies), they can both be generalized to any topic.

NammNguyen:
> Before I make additional changes, I have a couple questions/thoughts:
> 
> 1. The -[ Register OData Resources](https://docs.mendix.com/catalog/register/register-data/) article has specific OData instructions and links to the article [Published OData Services](https://docs.mendix.com/refguide/published-odata-services/). They are both OData-specific and reference specific OData terms and processes. Therefore, I think it may make more sense to keep these as OData resources, as updates to these articles would make many, widespread changes to multiple Catalog articles.
> 2. Would it make sense to change the [Register Non-OData Resources in the Catalog](https://docs.mendix.com/catalog/register/register-non-odata-resources/) article into a more generic "Register Resources in the Catalog" so it can cover anything that is non-OData, such as OpenAPI and any other technologies that are to be added to the Catalog? If so, we would need to verify which steps are the same between technologies and which need specification.
> 3. Are the OpenAPI/non-OData registration steps the same or similar to the OData steps? If they're similar enough, the article [Register Data Sources (On-Prem or Private Cloud)](https://docs.mendix.com/catalog/register/data-sources-without-mendix-cloud/) can likely be made generic to cover all registration processes.
> 4. The articles [Share Data Between Apps](https://docs.mendix.com/catalog/share-data/) and [Write Data to Another App ](https://docs.mendix.com/catalog/write-data/) also have OData-specific steps documented. Similar to [Search improvements #3](https://github.com/mendix/docs/pull/3), if the steps are the same/similar enough to OpenAPI (and upcoming technologies), they can both be generalized to any topic.

1. Hmm.. considering the way we register other resources will be pretty much the same as OData it might make sense here to spend some time re-writing this... See next point below
2. The current "Register Non-OData sources" only explains how other services could be wrapped in OData and then registered as OData into the Catalog. This does not apply for OpenAPI and other future technologies that we support, since we will support those directly. Registration of these will be very similar to OData : Automatic registration through the cloud, manual registration through the API and through the registration form
3. Yes! They are very similar, for automatic registration they will just need to publish a REST service in Studio Pro instead of OData. In the API the payload is the same except the `ApiTechnology` / `Protocol` field should be changed to correspond the technology registered. (`ApiTechnology` / `Protocol` will soon be changed to ServiceType and ContractType`). On the registration form, you will just click a different button from the homepage, but the registration steps are exactly the same.
4. This one I cannot say yet, we don't know how OpenAPI consumption will be done in Studio Pro. I think it will be a bit different though since metadata details of an OData contract are Entitysets while for OpenAPI those probably would be operations and consuming them would be different

PR #6896, opened 2023-12-01
Add Rebase Release Note in 10.5

MarkvanMents:


content/en/docs/releasenotes/studio-pro/10/10.5.md
```diff
@@ -28,6 +28,14 @@ Some advanced XPath features are not yet supported by the Builder. For those cas
 
 Along with the introduction of the new visual Builder, we have also taken the opportunity to combine the **Database** and **XPath** data sources under the **Database** data source.
 
+#### Version Control Uses Rebase When Combining Changes
+
+We have introduced Rebase as the default action for combining local changes with changes on the server. This is an alternative to making a 'Merge commit'. Using Rebase results in a simpler commit history, as all commits are made sequentially.
```
Adam-Dupaski: Capitalize "Rebase" ? Or needs to be bolded? Cross-reference?

`Merge commit` - capitalization correct? Single quote marks need to be removed. Bold? Cross-reference?

content/en/docs/releasenotes/studio-pro/10/10.5.md
```diff
@@ -28,6 +28,14 @@ Some advanced XPath features are not yet supported by the Builder. For those cas
...
+
+You can choose how to combine your work with the server state every time you interact with the server. Whether you are making a commit or performing a pull or a push, Studio Pro will detect if you are behind the server and let you choose how to proceed. We recommend you to use Rebase in most scenarios, but you can adjust the default with a user preference if needed.
```
Adam-Dupaski: "will detect" > "detects"
"let you" > "lets you"
"We recommend you to use" > "Mendix recommends using"

content/en/docs/releasenotes/studio-pro/10/10.5.md
```diff
@@ -28,6 +28,14 @@ Some advanced XPath features are not yet supported by the Builder. For those cas
...
+
+For more information on the differences between using Rebase and Merge commit, and how they work, please see [Combining Changes and Conflict Resolution](/refguide/merge-algorithm/).
```
Adam-Dupaski: "For more information on the differences between using Rebase and Merge commit, and how they work," > "For more information,"

content/en/docs/releasenotes/studio-pro/10/10.5.md
```diff
@@ -28,6 +28,14 @@ Some advanced XPath features are not yet supported by the Builder. For those cas
 
 Along with the introduction of the new visual Builder, we have also taken the opportunity to combine the **Database** and **XPath** data sources under the **Database** data source.
 
+#### Version Control Uses Rebase When Combining Changes
```
Adam-Dupaski: "Uses" > "Using"

content/en/docs/releasenotes/studio-pro/10/10.5.md
MarkvanMents: @Adam-Dupaski
Thanks for the comprehensive review, Adam. I've made all the changes you suggested apart from the decision about the formatting of Rebase.
Here we are not talking about the UX element but the git concept, so I don't think it needs to be bold, and it could be lower case. I have left it in upper case because it is the name of the feature which is being introduced and it is capitalized where it does appear in the UX.

I like it in upper case, but am prepared to change if you think that is better.

I'm not sure whether we should refer to the git concept in a cross-reference as there are a couple of things which are different because we are doing the diff within the Studio Pro model and not with plain text.

MarkvanMents:
Approved by R&D over Slack. Merging.

PR #6900, opened 2023-12-01
Update Backups Page: Creation and Retention Schedules

stasfilin:
Update docs from SLO document


feddovanede (approved):
Looks good to me!

PR #6908, opened 2023-12-04
Update rich-text.md

leonardomendix:
Add one widget known issue.

### DIFF
--- a/content/en/docs/appstore/widgets/rich-text.md
+++ b/content/en/docs/appstore/widgets/rich-text.md
@@ -186,7 +186,7 @@ Features:
 
     {{% alert color="info" %}}To be fully secure, all user HTML input should be sanitized on the server side too. This could be done with the XSSSanitize action found in the [Community Commons](/appstore/modules/community-commons-function-library/). When the option 'Sanitize content' is set to 'false' server side sanitating is required before showing any HTML content.{{% /alert %}}
 
-## 5 Known issue
+## 5 Limitations
 
-There is a problem with the toolbar buttons that causes incorrect rendering when hovering over them, depending on the screen size or scaling. The problem is caused by a miscalculation of the button's border by the browser, as well as styling issues with CKEditor.
+There is problem with the toolbar buttons which causes incorrect rendering when hovering over them, depending on the screen size or scaling. This problem is caused by a miscalculation of the button's border by the browser, as well as styling issues with CKEditor.
 
--- a/content/en/docs/appstore/widgets/rich-text.md
+++ b/content/en/docs/appstore/widgets/rich-text.md
@@ -188,8 +188,7 @@ Features:
 
 ## 5 Limitations
 
-The following limitations apply to the rich text widget:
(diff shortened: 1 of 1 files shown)


content/en/docs/appstore/widgets/rich-text.md
```diff
@@ -185,3 +185,8 @@ Features:
     * Schemes: `http`, `https`, `ftp`, `mailto`
 
     {{% alert color="info" %}}To be fully secure, all user HTML input should be sanitized on the server side too. This could be done with the XSSSanitize action found in the [Community Commons](/appstore/modules/community-commons-function-library/). When the option 'Sanitize content' is set to 'false' server side sanitating is required before showing any HTML content.{{% /alert %}}
+
+## 5 Known issue
```
Adam-Dupaski: If it's a known issue, it will be fixed, and it should go in release notes and be tracked there for the fix.

If it's a limitation, it will not be fixed, and it can be labeled as a "Limitation" and stay here in the doc.

gjulivan: It is limitation on the browser side that we recently discovered.
Since it is actually an issue on the browser, we can't fixed it.
The wording might need to change a little bit.

maybe:

"There is known incorrect hover style issue in the toolbar that occurs when display size is not set to 100%. This is also known issue on the major browser which unable to calculate correct pixel size on the border in CKEditor toolbar styling"

ConnorLand: I will change the language in this doc to reflect that this is a browser limitation with no fix. I will remove instances of the words "known issue".

ConnorLand: @gjulivan by "major browser" do you mean Chrome? If so we need to specify by name.

Also do you happen to have a link to an official documentation of the style and/or browser issues? It might be nice to include them if they are from the browser or CKEditor sites.

gjulivan: wait, let me rephrase again,
this is actually considered as "not a bug" from a browser perspective. because it is a sensible things to do. pixel scales based on screen density. and this is a correct specification for all browser.

but this is more of a "side effect", which cause the rendering position messed up. There are many "workaround" to fix this, but due to the styling is owned by CKEditor and not us, we prefer not to.

how do you want to phrase this?

ConnorLand: I've added some language to try and reflect what you are saying: https://github.com/mendix/docs/pull/6908/commits/ebcc19489b8cbd3b841942d57c4f9bce786a6494

ConnorLand: @gjulivan @leonardomendix do you think the tech is correctly explained in the new Limitations section now? If there are no further issues I can merge.

gjulivan: sounds good!

gjulivan (approved):

ConnorLand (approved):

PR #6939, opened 2023-12-11
Add limitation to Rich Text

rahmanunver:
Adding a limitation faced by the usage of keyboard input sources which use suggestions from the OS ( called Input Method Editors), like Japanese or Chinese.


ConnorLand:
Addresses this support ticket: https://mendix.atlassian.net/browse/WC-2203

ConnorLand (approved):

PR #6953, opened 2023-12-12
Add a warning note

Yiyun333:


Yiyun333:
Upon request from support: https://mendix.slack.com/archives/C63U54B6K/p1702314795428679

PR #6970, opened 2023-12-14
Update prerequisites.md

jdiehl:
Explain which MiN app to use from 10.6.0 on


jdiehl:
@ConnorLand

ConnorLand (approved):

ConnorLand:
Will merge after deciding how to duplicate these ideas in this doc: https://docs.mendix.com/releasenotes/mobile/make-it-native-parent/

While articulating why we recommend MIN9 into SP10

PR #6975, opened 2023-12-14
Amazon rekognition documentation update for Auth 3.0.0

dennisreep:
@katarzyna-koltun-mx This one should be approved when Amazon Authentication 3.0.0 gets released.


katarzyna-koltun-mx:
@dennisreep Reviewed and ready to go when Auth 3.0.0 goes live. BTW "microflow" and "nanoflow" are not capitalized in our docs, so I changed those back.

PR #6979, opened 2023-12-15
Clarified that the Oracle JDK will not be automatically installed

grootjans:
The current text left some customers the impression we install the Oracle JDK


MarkvanMents (approved):
Looks good, Robbert Jan (@grootjans )
Approving and merging.

PR #6984, opened 2023-12-18
Remove instructions for import wizard.

MarkvanMents:
Import wizard was deprecated, but is removed in 10.6.

Removing whole section rather than labelling with version information.


MarkvanMents:
R&D team agrees to this approach.

PR #7011, opened 2023-12-22
10.6.1 RN

Adam-Dupaski:


atennapel:
@Adam-Dupaski We need to add the following change to the 10.6.1 release notes:
```
---
type: "improvement"
rank: 50
---

- We reverted the fix for the issue where created files are not deleted when an error occurs in a microflow. (Reintroduces Ticket 199712) 
- We reverted updating the `changedBy` attribute when `storeFileDocumentContent` or `storeImageDocumentContent` from the public API is used, or when uploading files. 
```
Could you do this for us? We are not sure about the "Reintroduces Ticket ..." part.

Additionally we would like to add a link to 10.6.0 that mentions that the following two changes have been reverted in 10.6.1:
```
- We fixed a an issue where created files were not deleted when an error occurred in a microflow. (Ticket 199712)
- The changedBy attribute is now updated when storeFileDocumentContent or storeImageDocumentContent from the [Mendix Runtime API](https://github.com/mendix/docs/blob/53c1bcd9e2bbc73928c0384928b0019c90c9c6b8/apidocs-mxsdk/apidocs/runtime-api) is used, or when uploading files. This makes it consistent with other ways of updating file documents or images
```
(also note the error in the first change "a an")

Adam-Dupaski:
@atennapel updated

atennapel:
Thank you for making the changes for us, looks good to me.

PR #7023, opened 2023-12-28
PMP relnotes

katarzyna-koltun-mx:


katarzyna-koltun-mx:
I've received feedback that makes it sound like this needs to be hidden until confirmed.

PR #7046, opened 2024-01-04
Add known issue to 10.6.0

tabakx:
There was a known issue in 10.6.0 that we missed. It has already been fixed in 10.6.1 and 10.7, but we missed it in the release notes for 10.6.0.


Adam-Dupaski:
Added here: https://github.com/mendix/docs/pull/7011/commits/3b1d8845afaf93f8aecbe1cad2760b6aad8ab1dc

PR #7053, opened 2024-01-04
SAP BTP metrics

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Updated per feedback from dev team, waiting for the go ahead to publish.

PR #7059, opened 2024-01-08

edwinvanelk:
Example is incorrect.

Result from
'substring('thisismystring', 0, min(length('thisismystring') - 1, 20))' ->
thisismystrin

instead of thisismystring

P.S. Why is this useful example no longer included in refguide8, 9 and 10?

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
@@ -131,13 +131,13 @@ The output is the following:
 If you use a third parameter that specifies the desired length of the output:
 
 ```java {linenos=false}
-substring('funwithmendixapps', 7,6)
+substring('thisismystring', 6, 2)
 ```
 
 The output is the following:
 
 ```java {linenos=false}
-'mendix'
+'my'
 ```
 
 You can also limit the value of the third parameter to prevent it from specifying a desired length that is out of range:
--- a/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
@@ -140,10 +140,10 @@ The output is the following:
 'my'
 ```
 
-You can also limit the value of the third parameter to prevent it from specifying a desired length that is out of range:
+You can also specify the value of the third parameter using the `length` function:
 
 ```java {linenos=false}
-substring('thisismystring', 0, min(length('thisismystring'), 20))
+substring('thisismystring', 0, length('thisismystring'))
 ```
 
 ## 5 find
--- a/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
@@ -128,7 +128,7 @@ The output is the following:
 'mystring'
 ```
 
-If you use a third parameter that specifies the desired length of the output:
+If you use a third parameter to specify the desired length of the output:
 
 ```java {linenos=false}
 substring('thisismystring', 6, 2)
@@ -140,10 +140,10 @@ The output is the following:
 'my'
 ```
 
-You can also specify the value of the third parameter using the `length` function:
+To prevent the value of the third parameter from getting out of range, you can set a limit to the third parameter, for instance, using the `min` and `length` functions:
 
 ```java {linenos=false}
-substring('thisismystring', 0, length('thisismystring'))
+substring('thisismystring', 0, min(length('thisismystring'), 20))
 ```
 
 ## 5 find
--- a/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
+++ b/content/en/docs/refguide/modeling/application-logic/expressions/string-function-calls.md
@@ -128,7 +128,7 @@ The output is the following:
 'mystring'
 ```
 
-Another example of an input is:
+If you use a third parameter that specifies the desired length of the output:
 
 ```java {linenos=false}
 substring('funwithmendixapps', 7,6)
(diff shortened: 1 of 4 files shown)


Yiyun333:
Hi @edwinvanelk,

Thank you for the feedback! 

However, from our documentation, it states that:

> the first character of a string is located at position 0, and the last character is located at position `length(string)-1`.

So my understanding is that the example is correct. It should be `length('thisismystring') - 1` instead of `length('thisismystring')`. Because to have the output `thisismystring`, we need `substring('thisismystring', 0, 13)` instead of `substring('thisismystring', 0, 14)`.`

Please let me know if my understanding is not correct. I've not tested this example in Studio Pro though, so please let me know if you've tested this in Studio Pro and you still find the example incorrect. 

PS: I am not sure why this example was removed from refguide 8, 9 and 10. It is indeed a useful one. I will add the example back after I've received your further feedback.

edwinvanelk:
Yiyun333,

String starting at position 0 makes sense for the "Start position of the substring". but not for "Desired length of the result". For a string like "Edwin" the "Desired length of the result" should be 5 (characters).

As stated, I tested the example from Studio Pro and it gives result: thisismystrin instead of expected thisismystring.

Yiyun333:
> Yiyun333,
> 
> String starting at position 0 makes sense for the "Start position of the substring". but not for "Desired length of the result". For a string like "Edwin" the "Desired length of the result" should be 5 (characters).
> 
> As stated, I tested the example from Studio Pro and it gives result: thisismystrin instead of expected thisismystring.

@edwinvanelk oh, right! Sorry about my misinterpretation. I had a totally different logic in my head. Indeed, the third parameter should mean the "desired length of the result", instead of "the position of the last character".  Not sure why my thought went that way, but I will add the example back to refguide 8, 9, 10, and approve this PR soon. 

Thanks a lot for your explanation!

Yiyun333:
Hello @edwinvanelk,

I've verified the usefulness of the example with our developer. We both don't see the need to use the `min` function here:
`substring('thisismystring', 0, min(length('thisismystring'), 20))`

Hence, I've simplified the example to the following:
`substring('thisismystring', 0, length('thisismystring') )`

Please let me know if you have different opinions. Thanks!

edwinvanelk:
@Yiyun333 Sorry to say, but the new example is not very usefull.

The original example meant: Limit the string to the first 20 characters (or less if the string is shorter then 20 characters). So a very long string like 'thisisastringwith29characters' would result in: 'thisisastringwith29c'. For a very long string you could simply use substring($verylongstringwithmorethan20characters, 0, 20). For a string shorter then 20 characters (e.g. 5 characters) this would fail: substring($veryshortstringwithlessthan20characters, 0, 20) -> 

```
Caused by: com.mendix.modules.microflowengine.MicroflowException: Failed to evaluate expression, error occurred on line 1, character 1
substring($veryshortstringwithlessthan20characters, 0, 20)
^
	at Test.ACT_substring (SendTextMessage : '{1}')

Advanced stacktrace:
	at com.mendix.languages.mxexpressions.MxExpressionImpl.evaluate(MxExpressionImpl.scala:32)

Caused by: com.mendix.languages.expressions.ExpressionException: begin 0, end 20, length 5
	at com.mendix.languages.expressions.Expr.evaluate(Expr.scala:16)

Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end 20, length 5

```

Therefore the example with the min() function was added to the documentation, so this situation will work always:

for a short string (e.g. 5 characters):
substring($veryshortstringwithlessthan20characters, 0, min(length($veryshortstringwithlessthan20characters), 20))
->
substring($veryshortstringwithlessthan20characters, 0, 5)

and for a long string (e.g. more then 20 characters):
substring($verylongstringwithmorethan20characters, 0, min(length($verylongstringwithmorethan20characters), 20))
->
substring($verylongstringwithmorethan20characters, 0, 20)

You'r new example now always returns the complete string, and no longer just a part of the string as expected.

Yiyun333:
> @Yiyun333 Sorry to say, but the new example is not very usefull.
> 
> The original example meant: Limit the string to the first 20 characters (or less if the string is shorter then 20 characters). So a very long string like 'thisisastringwith29characters' would result in: 'thisisastringwith29c'. For a very long string you could simply use substring($verylongstringwithmorethan20characters, 0, 20). For a string shorter then 20 characters (e.g. 5 characters) this would fail: substring($veryshortstringwithlessthan20characters, 0, 20) ->
> 
> ```
> Caused by: com.mendix.modules.microflowengine.MicroflowException: Failed to evaluate expression, error occurred on line 1, character 1
> substring($veryshortstringwithlessthan20characters, 0, 20)
> ^
> 	at Test.ACT_substring (SendTextMessage : '{1}')
> 
> Advanced stacktrace:
> 	at com.mendix.languages.mxexpressions.MxExpressionImpl.evaluate(MxExpressionImpl.scala:32)
> 
> Caused by: com.mendix.languages.expressions.ExpressionException: begin 0, end 20, length 5
> 	at com.mendix.languages.expressions.Expr.evaluate(Expr.scala:16)
> 
> Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end 20, length 5
> ```
> 
> Therefore the example with the min() function was added to the documentation, so this situation will work always:
> 
> for a short string (e.g. 5 characters): substring($veryshortstringwithlessthan20characters, 0, min(length($veryshortstringwithlessthan20characters), 20)) -> substring($veryshortstringwithlessthan20characters, 0, 5)
> 
> and for a long string (e.g. more then 20 characters): substring($verylongstringwithmorethan20characters, 0, min(length($verylongstringwithmorethan20characters), 20)) -> substring($verylongstringwithmorethan20characters, 0, 20)
> 
> You'r new example now always returns the complete string, and no longer just a part of the string as expected.

@edwinvanelk Thank you very much for the further and very clear explanations! It makes much more sense to me now why we use the `min` function in the example. I've updated the example again as well as its description. Please let me know if you have any further suggestions.

Yiyun333:
@edwinvanelk I am merging this PR. If you have any further suggestions, please send us a new PR. Thank you!

PR #7066, opened 2024-01-08
Add data usage note

Yiyun333:


Yiyun333:
Add the note via request from Amir and Danny Roest via email.

PR #7082, opened 2024-01-11
Fix outdated Gradle information about manual Gradle installation for Studio Pro 9

jveldhuizen:


content/en/docs/refguide9/general/install.md
```diff
@@ -103,7 +103,7 @@ It is possible to prepare the prerequisite installers beforehand, so that the Me
     * The `latest` executable to *mendix_native_mobile_builder.exe*
     * The *Git-{version}-64-bit.exe* executable to *git_for_windows_installer.exe*
     * Do not rename the Microsoft Edge WebView2 Evergreen Runtime installer *MicrosoftEdgeWebview2Setup.exe*, keep it as is
-    * The Gradle zip file should be named *gradle-7.6-bin.zip*
+    * The Gradle zip file should be named *gradle-7.6.3-bin.zip*
```
jveldhuizen: From 9.24.0 up to 9.24.10, it should be `gradle-7.6-bin-zip`.  
From 9.24.11, it should be `gradle-7.6-bin-zip`, like in Mendix 10.  
Should we mention both situations or only the latest 9.24 situation?

MarkvanMents: We should mention both as we don't know which version a customer is using (we still have customers on Mendix 7!

And is it just the name of the file which is important, or does it need to have the correct Gradle version in it?
I would suggest:
```suggestion
    * For Mendix 9 versions 9.24.11 and above, use Gradle version 7.6.3 in a zip file named *gradle-7.6.3-bin.zip*
    * For Mendix versions 9.24.0 to 9.24.10, use Gradle version 7.6 in a zip file named *gradle-7.6-bin.zip*
```

content/en/docs/refguide9/general/install.md
```diff
@@ -78,7 +78,7 @@ The prerequisites are the following:
     | [2.37.1](https://github.com/git-for-windows/git/releases/tag/v2.37.1.windows.1) | [2.41.0](https://github.com/git-for-windows/git/releases/tag/v2.41.0.windows.3) |
 
 * [Microsoft Edge WebView2 Evergreen Runtime (x64 or ARM64) (Evergreen Standalone Installer version)](https://developer.microsoft.com/en-us/microsoft-edge/webview2/)
-* For Mendix Studio Pro 9.24.0 and above – [Gradle 7.6 or above](https://services.gradle.org/distributions/gradle-7.6-bin.zip)
+* For Mendix Studio Pro 9.24.0 and above – [Gradle 7.6 or above](https://gradle.org/install/#manually) with Gradle extracted to the parent directory of the folder where Studio Pro is installed (usually `C:\Program Files\Mendix`) instead of `C:\Gradle`.
```
MarkvanMents:
```suggestion
* Mendix Studio Pro version 9.24.0 and above require the following versions of [Gradle](https://gradle.org/install/#manually) with Gradle extracted to the parent directory of the folder where Studio Pro is installed (usually `C:\Program Files\Mendix`) instead of `C:\Gradle`.
    * Mendix Studio Pro versions 9.24.0 to 9.24.10 require Gradle version 7.6 or above
    * Mendix Studio Pro version 9.24.11 and above require Gradle version 7.6.3 or above
```

MarkvanMents (commented):
Thanks for the PR, @jveldhuizen.
The versioning of Gradle seems very complicated - I've tried to make it more precise, but perhaps we should just recommend 7.6.3 and above and just deal with the renaming if customers are using 9.24.0 t/m 9.24.10

content/en/docs/refguide9/general/install.md
jveldhuizen: I think this is fine too, except the '-' in 'versions of - [Gradle'. It reads better when you omit '-', right?

MarkvanMents: Thanks for spotting that. I'll remove it and make the commit.
Then I'll approve and merge the PR.
Thanks for the contribution and clearly explaining the options.

MarkvanMents (approved):
Thanks for the PR, Jonathan.
Approving and merging.

PR #7090, opened 2024-01-11
adding more details on CSP

gjulivan:
adding more details on content security policy for widgets.

This document will be the start of widget's CSP document, following other widget list on the epic:
https://mendix.atlassian.net/browse/WC-2304

Background:
By default, CSP for "self" is sufficient (as mentioned on [current CSP policy document](https://docs.mendix.com/howto/security/using-mobile-capabilities/csp/)
but some widget need access to 3rd party domain (such as map widget), which heavily dependent on the provider.
this document will give overview to that domain access request as part of CSP compliance requirements.


content/en/docs/appstore/widgets/security/map.md
```diff
@@ -0,0 +1,56 @@
...
+## 1 Introduction
+
+[Map widget](/appstore/widgets/maps/) need to have access to map provider based on your configuration. These are the allowlist domains need to be setup for each available provider.
+
+## 2 Setup
+
+### 2.1 Google Map
+
+Please refer to [google maps' content security policy](https://developers.google.com/maps/documentation/javascript/content-security-policy#sample_content_security_policy) document for more detail.
+You can enable allowlist CSP for google maps by including these domains:
```
hnejra: This only necessary when using legacy API or region-specific domain as mentioned on that page: "Websites which load the Maps JavaScript API from a legacy Google APIs domain (for example maps.google.com) or a region-specific domain (for example maps.google.fr), must also include these domain names in their CSP script-src setting, as shown in the following example"
Otherwise these domains should be enough: https://developers.google.com/maps/domains

hnejra: As mentioned here: https://developers.google.com/maps/documentation/javascript/content-security-policy#allowlist_csp

gjulivan: list updated

gjulivan:
Other widget CSP document (as mentioned in the jira epic) will follow after this one being merged and published.
thus, we want to make sure first that this is a correct structure, place, and format for widget CSP document.
@ConnorLand

ConnorLand:
@grand

> Other widget CSP document (as mentioned in the jira epic) will follow after this one being merged and published. thus, we want to make sure first that this is a correct structure, place, and format for widget CSP document. @ConnorLand

I think this is a good structure for CSP material. Is all the draft material submitted for this round? If so, I can get to work editing things.

gjulivan:
> @grand
> 
> > Other widget CSP document (as mentioned in the jira epic) will follow after this one being merged and published. thus, we want to make sure first that this is a correct structure, place, and format for widget CSP document. @ConnorLand
> 
> I think this is a good structure for CSP material. Is all the draft material submitted for this round? If so, I can get to work editing things.

Yes, please continue

content/en/docs/appstore/widgets/security/map.md
```diff
@@ -0,0 +1,67 @@
...
+
+The [Maps](/appstore/widgets/maps/) widget requires access to a map provider (which provider is based on your configuration) in order to work. Below, you can see the `allowlist` domains you need to be set up each available provider.
+
+## 2 Setup
+
+### 2.1 Google Maps
+
+You can enable `allowlist` CSP for Google Maps by including these domains:
+
+```
```
ConnorLand: @gjulivan should these snippets render in a specific code language?

ConnorLand (approved):

PR #7119, opened 2024-01-17
Add retention schedule

feddovanede:
Add retention schedule


dbreseman:
Hi @feddovanede, this information is actually already mentioned on this page ("Logs on Mendix Cloud are rotated and archived every day... By default, archived logs are sorted by date; the current log is the first one listed. They are stored for six months.") Are you suggesting moving this information into a separate section for emphasis?

PR #7134, opened 2024-01-22
updated the epics api specs yaml file

JulesVanThillo:
Updates the epics api specs yaml file. It contains new endpoints and multi error responses.

### DIFF
--- a/static/openapi-spec/epics.yaml
+++ b/static/openapi-spec/epics.yaml
@@ -15,27 +15,27 @@ paths:
     get:
       tags:
         - statuses
-      summary: Get statuses by app ID
-      description: Returns an array of statuses
+      summary: Get statuses by the app ID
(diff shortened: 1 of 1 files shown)


static/openapi-spec/epics.yaml
```diff
@@ -362,7 +362,7 @@ paths:
                   value:
                     status: 400
                     title: Bad Request
-                    detail: Maximum 200 characters is allowed for Title 
+                    detail: Maximum 200 characters are allowed for Title
```
JulesVanThillo: Should we change this into "The maximum number of characters for Title is 200" or "Maximum of 200 characters is allowed for Title"?

Luyao-Zhang-1: We can use "The maximum number of characters for Title is 200."

PR #7142, opened 2024-01-23
Add Authenticating Users to Mobile

jdiehl:
@ConnorLand


ConnorLand:
We are going to add a page to Designing Mobile User Interfaces called Authenticating Users, and we can explain the default process with anonymous users and link to Native SSO. We'll also rename Designing Mobile User Interfaces to something more generic.

ConnorLand (approved):

PR #7170, opened 2024-01-26
Prevent slicing on icons

dbreseman:


MarkvanMents (approved):
Great update, Dana.
I've done some random reviewing and it looks OK to me.
You can merge if you want. But you'll need to go to production to see Algolia work.
Let me know if you want me to look at anything in Algolia directly.

(or I can merge on Monday if you are worried about doing it at the end of Friday!)

MarkvanMents:
Done a couple more tests:

1. The only paragraphs which have additional tags (class= etc) are for the feedback text, so it doesn't matter if these aren't indexed.
2. The full paragraph is now exported to Algolia, the icon is replaced by a space, so it shouldn't corrupt Algolia indexing.

Merging.

dbreseman:
Thanks a lot, Mark!

PR #7174, opened 2024-01-29
Add new Integration activities to Application Logic

quinntracy:


content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services-beta.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "Consumed REST Services (beta)"
+title: "Consumed REST Services (Beta)"
 url: /refguide/consumed-rest-services-beta/
```
Yiyun333, line 3: I see that this url (`/refguide/consumed-rest-services-beta/`) is linked to several places in Studio Pro already. For now, it's okay since the feature is still in beta. However, when the feature goes in GA, this link needs to be updated both in our doc and in Studio Pro. Please let the team know about this. Actually we normally don't add "beta" in a doc url even for a beta feature. Then we/especially the team don't need to update the url in the product again when the feature status changes :)

Moreover, the **Send REST request** activity actually should be linked to `https://docs.mendix.com/refguide/send-rest-request/` in the product, indtead of` https://docs.mendix.com/refguide/consumed-rest-services-beta/#add-entity-to-microflow`. The team need to update this link in the product :)

Haha, okay, the above two points are not really specific to this PR. I've reviewed the PR and updated the icons as well. The icons needs to show both the icon image and the activity name. I usually drag them onto the canvas and then make a screenshot from there. I am approving this PR. Pls let me know if you have questions!

Yiyun333 (approved):

PR #7194, opened 2024-01-30
User's access to metadata of an offboarded app

Luyao-Zhang-1:


content/en/docs/support/new-app-node-request-template.md
```diff
@@ -60,15 +60,7 @@ Offboarding will remove access to everything related to the node. If you want to
...
-When your app is offboarded you will still have access to the app information which is in the **COLLABORATE** and **DEPLOY** sections. This consists of the following:
-
-* Buzz
-* Stories
-* Planning
-* Team
-* Feedback
-* Documents
-* Team Server (containing all committed versions of your app model)
+When your app is offboarded, you still retain all access to the metadata associated with the app, such as the stories in Epics, the feedback in App Insights, and the repository on the Team Server.
```
MarkvanMents: I would just say **Epics** as that is the title of the feature.
Is App Insights also supposed to be "metadata" or a different sort of thing?
The repo is certainly _not_ metadata.

I wonder whether we should not say "metadata" but perhaps "development data". Something like:
```suggestion
When your app is offboarded, only the deployed app is removed. You still retain all access to the data associated with the development of the app, such as Epics, App Insights, and the repository on the Team Server.
```
What do you think?

MarkvanMents (commented):
I've made a suggestion - what do you think?

content/en/docs/support/new-app-node-request-template.md
Luyao-Zhang-1: Your suggestion looks good! Thank you!

PR #7196, opened 2024-01-30
Update fonts

Luyao-Zhang-1:


MarkvanMents (commented):
Change looks good, except that the Docs landing page has a single instance of Patron, which looks odd.
I think this should probably be Noto Sans as well, all our other doc titles are in Noto Sans.
image

PR #7198, opened 2024-01-31
Added link to developer settings

tknippenberg:


content/en/docs/community-tools/mendix-profile.md
```diff
@@ -378,7 +378,7 @@ The following Mendix services support usage of PATs:
 
 ##### 10.4.2.1 Creating a PAT {#create-pat}
 
-When you click **New Token**, you need to set the following characteristics of the PAT:
+When you click **New Token** in the [Developer Settings](https://user-settings.mendix.com/link/developersettings), you need to set the following characteristics of the PAT:
```
MarkvanMents: I think this might need changes elsewhere in the document. This doesn't seem the right place for this?

It should be a link to the documentation about how to create a PAT, not to the app.

tknippenberg: I don't think I agree with that, as I was unable to find how to create a PAT and it took me over 3 minutes with a colleague to figure it out. 
I think it would be beneficial to add clearer links for this process at the point of the documentation which I'm finding through Google or with CTRL+F on a page, which is this chapter.

MarkvanMents: I agree that we need to help users set up PATs. We have full documentation about that. That documentation is here: https://docs.mendix.com/community-tools/mendix-profile/#pat.

The disadvantage with putting a link to the platform here is that:

1. It doesn't give the full context.
2. When external URLs change and break we don't get any warnings - so we want to limit where we have external links if possible.
3. It has to be put everywhere that PATs are mentioned.

I would suggest

1. Adding a link to the PAT documentation, rather than the platform
2. Making sure that users can easily get to the PAT creation from the PAT documentation.

@Luyao-Zhang-1 are you able to pick this up and check where we mention PATs and point them to the right place - and also make sure it is clear how the reader can get to the right page on the platform from the PAT documentation.

PR #7240, opened 2024-02-06
Add dependency on Combo Box widget for the Administration module

MxMurshed:


content/en/docs/appstore/modules/administration.md
```diff
@@ -22,8 +22,10 @@ The [Administration](https://marketplace.mendix.com/link/component/23513) module
 
 ### 1.2 Dependencies
 
-* [Atlas UI Resources](https://marketplace.mendix.com/link/component/104730)
-* [Mendix SSO](https://marketplace.mendix.com/link/component/111349): required for the Administration module versions 1.3.X (for example 1.3.2) and 2.1.X (for example 2.1.2)
+* [Atlas Core](https://marketplace.mendix.com/link/component/117187)
+* [Combo Box](https://marketplace.mendix.com/link/component/219304)
```
MarkvanMents: Is this for all versions, or only for certain versions?

content/en/docs/appstore/modules/administration.md
```diff
@@ -22,8 +22,10 @@ The [Administration](https://marketplace.mendix.com/link/component/23513) module
...
+
+{{% alert color="info" %}} For versions 1.3.X (for example 1.3.2) and 2.1.X (for example 2.1.2) of the Administration module, [Mendix SSO](https://marketplace.mendix.com/link/component/111349) is also required. {{% /alert %}}
```
MarkvanMents: Why does this need to be in an alert?

MxMurshed: It doesn't need to be. However, we noticed that some people did not notice it therefore we changed it to an alert/info.

PR #7247, opened 2024-02-07
Redesign alert boxes to match PDS

dbreseman:


Yiyun333 (commented):
@dbreseman The changes look good to me! Thank you for the great work!

PR #7273, opened 2024-02-13
[RUN-3891] Clarify standard format in docs about datetime formatting

atennapel:


lvanengelen:
Should we add this caveat also to the `parseDecimal` and `parseInteger` functions? Do we need to verify whether the behavior is different between microflows and nanoflows?

atennapel:
I updated the PR to handle 8 and 9 as well, and also Decimal formatting.

emirmx:
Looks good!

atennapel:
@MarkvanMents it is approved from our side, so it's ready for the docs team :)

MarkvanMents:
> @MarkvanMents it is approved from our side, so it's ready for the docs team :)

Thanks. At least I won't start reviewing this one too early 😂

MarkvanMents (approved):
Thanks for the PR, Albert.
Looks like it can be published immediately, so approving and merging.

atennapel:
Thank you, yes this one can be merged immediately.

PR #7314, opened 2024-02-20
Non-persistable external entities

sbont:
We've added the possibility to import NPEs from a consumed OData service in 10.8. These can be used as external action parameters / return types.

### DIFF
--- a/content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/consumed-odata-service-requirements.md
+++ b/content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/consumed-odata-service-requirements.md
@@ -67,7 +67,7 @@ The key can consist of one or more properties, as long as the following conditio
 The list above for supported key fields does not include `Date` or `DateTime` data types.
 {{% /alert %}}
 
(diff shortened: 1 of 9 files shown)


content/en/docs/refguide/modeling/integration/odata-services/consumed-odata-services/consumed-odata-service-requirements.md
```diff
@@ -68,6 +67,8 @@ Furthermore, an entity can only be used if it is uniquely identifiable with a ke
 The list above for supported key fields does not include `Date` or `DateTime` data types.
 {{% /alert %}}
 
+Entities that are not accessible through an entity set can be imported as a [non-persistable entity](/refguide/consumed-odata-services/#external-non-persistable-entities).
```
joostverhoog, line 70: This sentence mixes plural and singular. "can be imported as non-persistable entities"

joostverhoog, line 70: Is "imported" the right word here? "Used" is more neutral, and doesn't cause confusion with "import mappings"

quinntracy (approved):

PR #7323, opened 2024-02-22
Mendix Forum -> Mendix Community

Luyao-Zhang-1:


Luyao-Zhang-1:
Waiting for the info regarding the release date of the Mendix Community. Need to create a Release Note for that. Also, should still use Mendix Forum for any mentions before that date in the Release Notes.

PR #7343, opened 2024-02-27
[RUN-3911] Update Managed Dependency docs page with information about proxy settings

akkrishnakumar:
Summary doc included in the story


content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -98,13 +98,72 @@ There are some dependencies that are required by Mendix. These need to be added
...
+
+### 7.1 Gradle Proxy Settings
+
+You can create a `gradle.properties` file in the [Gradle User Home](https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home) directory and configure the proxy settings there. Just replace the values in the sample file below with your values.
+See the official [Networking with Gradle](https://docs.gradle.org/current/userguide/networking.html) guide for further details.
+
+Sample `gradle.properties` file:
+
+```txt {linenos=false}
+systemProp.http.proxyHost=proxy-host-ip
```
MarkvanMents, line 115: Why do we specify an ip here - in the Gradle example page they suggest www.somehost.org as an example value?

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -98,13 +98,72 @@ There are some dependencies that are required by Mendix. These need to be added
...
+```
+
+### 7.2 VPN Setup
+
+You may have a VPN or firewall that blocks/restricts access to certain websites. In this case, you need to ask your network/system administrators to allow access to the following sites:
+
+* jcenter.bintray.com
+* plugins-artifacts.gradle.org
+* plugins.gradle.org
+* repo.maven.apache.org
```
MarkvanMents, line 141: I have a new PR (#7436) which implies that if you have a VPN these should always be whitelisted. Is that the case (in which case it should also go where the PR suggests) or are these only needed if customers are actively using Managed Dependencies?

MarkvanMents (commented):
Hi Akhil (@akkrishnakumar )
Thanks for the PR. Looks like a good addition to the documentation.
I have a couple of questions. Can you look at these, and check my proofreading.
But I think it is pretty much ready to be merged once you agree.

content/en/docs/refguide/java-programming/managed-dependencies.md
akkrishnakumar: You can either use an IP or hostname. Gradle will internally build a url out of it. 
Eg: I have seen examples where they use `localhost`. 

Also, from the tickets we worked on, the customers used proxy IPs and it seemed to work for them.

content/en/docs/refguide/java-programming/managed-dependencies.md
akkrishnakumar: I'm not sure where the other PR came from, But anything related to Gradle should be in Managed Dependencies for now. 

And yes, these repos need to be included only when the user has module/project dependencies mentioned in the project settings.

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: OK, I'll leave it as it is and think about updating if we get any feedback.

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: The other PR came from someone in Expert Services who was working with a customer who had the issue.
I'll alter his PR so that it points to this document and say it is only needed when there are module/project dependencies mentioned in the project settings.

MarkvanMents (approved):
Hi @akkrishnakumar 
Thanks for responding to my questions.
I'll approve this now. Can it be published immediately?

akkrishnakumar:
Yes, it can be published

PR #7352, opened 2024-02-28
[RUN-3894] Create Java version migration document

jveldhuizen:
From now on, we will mention all known issues due to Java version migration in this document. It will be a growing document, with Java 21 issues as well when we start supporting that version.


jveldhuizen:
Approval from the Runtime Core team is required.

atennapel:
Looks good to me. I am not sure about the location of the document, but I will let the docs team decide on that.

MarkvanMents (commented):
@jveldhuizen 
Thanks for the PR. I have proofread it, so can you have a look and make sure it is still OK.
Like @atennapel, I'm not sure where is best for this. We have two sections, one about how Java works within Mendix, but one which covers Java more generally - for when customers are writing their own Java actions.
These changes look like they will affect Java actions as well as functions within microflows, so they  may fit better in the [Java programming](https://docs.mendix.com/refguide/java-programming/) section (since it has links from the expression section anyhow).

But I'm also happy if you think it is better here.

jveldhuizen:
@MarkvanMents, your commits are fine. But sorry, I don't know how to choose regarding the location. So feel free to move it to Java programming. Both locations have pro's and cons. Java programming is not always true as the Java version has influence on the microflow expression as well. But under 'Mendix Runtime' -> 'Mendix Runtime and Java' is too deep in the tree and Runtime is too technical for people who only work in Studio Pro.

Maybe Java programming is better as it's higher in the hierarchy. But maybe it could be improved now or later by renaming the node 'Java Programming' to simply 'Java' and write a more general text in the 'Introduction' of that page, about that Mendix apps runs on a JVM and that that JVM has influence on the running app, its memory usage, and some functionality of microflow expressions which is handled by the JVM, but also the big benefit of extending Mendix apps with Java code which is possible because of the JVM.

MarkvanMents:
Hi @jveldhuizen 
I think you make a good case for moving it to the Java Programming section, so I'll do that.
Done in this commit: https://github.com/mendix/docs/pull/7352/commits/a2df27f866fb77f2e7013a2f73b6d8dc3463ed3c
OK to publish now?

jveldhuizen:
@MarkvanMents, thanks. Yes you can publish it. Is its 'weight' correct, which is the lowest now in this directory, 4?

MarkvanMents:
Thanks for the reminder - I had made the change locally but forgotten to commit it.
It is now above Managed Dependencies but below the rest of the Java documentation.
I'll merge this and make a Jira issue to pick up the improvement to the Java Programming section.

MarkvanMents (approved):
Thanks for the contribution @jveldhuizen and working together to get the doc in the right place.
Approving and merging.

PR #7369, opened 2024-03-01
Stories API deprecation note

mxfmg1982:
We are introducing a deprecation warning on Stories API


MarkvanMents (approved):
Looks good. Approving and publishing.

PR #7378, opened 2024-03-04
Add new runtime setting for JSON string used with mapping

sidrock147:
Now user could use field com.mendix.integration.MaxJsonReadingLength in studio pro to provide the required value which would override the default 20 million (20000000) max length. If this field is not used then 20 million (20000000) which is a default would be used as max length

**NB: This also needs to be put into 9.24.17 once the text is agreed.**


sidrock147:
@quinntracy : To be released for Mendix 10.9.0

content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -72,7 +72,7 @@ The following custom settings can be configured:
...
 | <a id="ObjectManagementStrictChangeBehavior" href="#ObjectManagementStrictChangeBehavior">ObjectManagement.<wbr>StrictChangeBehavior</a> | Defines the behavior when changing values of Enums and Calculated attributes.<br/>When set to true, setting an invalid value for an Enum attribute and/or setting a value for a Calculated attribute will result in an InvalidEnumerationValueException and/or ReadOnlyAttributeException respectively.<br/>When set to false, changes to the values of Enums and/or Calculated attributes will be allowed.<br/>We plan to remove this setting in Mendix 11, after which, an exception will always be raised when setting an invalid value. | true |
-
+| <a id="mapping.import.MaxJsonReadingLength" href="#mapping.import.MaxJsonReadingLength">mapping.import.MaxJsonReadingLength</a> | The max length of the input JSON string used with import mapping. The current max limit in the Jackson XML library is 20000000 which is the default value. Use this setting when you want to set it to a higher value than the default. |  |
```
MarkvanMents: @sidrock147 
What do you mean by the "input JSON string" - do you mean the JSON string which is received from the remote, or the size of individual strings within the mapping, or the size of the mapping itself?

Also, if there is a maximum limit in the library, how can setting a higher number within Mendix work - the library will still apply its own maximum.

sidrock147: 1) Incoming JSON string received from the remote
2) The library also provide the provision to override its default value. So When this custom setting is not used it would use its default length. But when user use the custom setting the value provided by the user would be considered and would override the default value.

MarkvanMents: Thanks for the response - I will try to clarify so that customers don't have the same questions I had 😀

MarkvanMents: I've updated this - I used the "default" column in the table and removed references to the specific library as I don't think this is relevant to the customer?
@sidrock147 - does this update look OK to you?
https://github.com/mendix/docs/pull/7378/files#diff-7906bdc7f896dc28d165304656f8ac95747ddf824ab51a0937ab53cc945d9c6bR75

sidrock147: Earlier, I had specified the current default value. The problem is that default value comes from third party library. Today it is 20000000 but in future they could change to something else. In the past this value was 500000 and later they changed to 20000000. User should not have impression that this is the default value which Mendix controls

MarkvanMents: @sidrock147 
But it is a value which we **do** control as we decide which version of the library we include?
If we just point to Jackson, the customer still won't know which version of the library we have included, and different versions of Mendix will be using different versions of the library. So we should just keep this number up-to-date, and indicate if it changes and which Mendix version it changes in.

sidrock147: @MarkvanMents  That would mean that whenever runtime team upgrades jackson core library we have to keep in mind to check the default value. if it is different then we need to update it. That might lead to human error since our team is not involved in upgrading the libraries of runtime and usually runtime team does that. So I see that as gap which would be difficult to keep track on.

MarkvanMents: I'll raise this in the runtime Slack channel.
The problem is that the customer doesn't necessarily know which version of Jackson is related to each version of the runtime.

MarkvanMents: I'd also say that the limit in this case is probably only going to get larger in future, so putting in 20,000,000 here isn't so damaging if the limit later rises to 50,000,000. The main damage would be if the limit was reduced in future.

Also, would it not be possible to mock a test as part of the Studio Pro release to see whether the limit was still there and update the documentation if it changes in some way?

Anyhow - I've asked Runtime to have a look and see if they can come up with any ideas.

content/en/docs/refguide/modeling/integration/mapping-documents/import-mappings.md
```diff
@@ -95,3 +95,15 @@ Import mappings have the additional option to receive an incoming parameter. The
...
+
+If you are importing a very long JSON string, the underlying conversion library may not be able to handle it. In this case you will get a message similar to the following:
+
+`Caused by: com.fasterxml.jackson.databind.JsonMappingException: String length (20051119) exceeds the maximum length (20000000)`
+
+Mendix uses the Jackson Core XML library when performing the import mapping and this [has a limit of 20 million](https://javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.15.1/com/fasterxml/jackson/core/StreamReadConstraints.html#DEFAULT_MAX_STRING_LEN) characters (in earlier versions 5 million) in the JSON string.
+
+In Mendix version 10.9.0 and above, you can override this using the [mapping.import.MaxJsonReadingLength](/refguide/custom-settings/#mapping.import.MaxJsonReadingLength) custom setting.
+
+Mendix recommends that you only add this setting if you will be getting very long JSON strings as it will have an impact on performance.
```
fazzdev: Rephrase to "...may result in more memory usages." I think this would give more information rather than performance.

content/en/docs/refguide/modeling/integration/mapping-documents/import-mappings.md
```diff
@@ -95,3 +95,15 @@ Import mappings have the additional option to receive an incoming parameter. The
 To define a parameter for your mapping, click the parameter box and select the data type. You can also drag an entity into the parameter box. 
 
 You can use a parameter as a key or in a microflow to obtain objects. When you use an entity parameter, you can set associations to it. When you use a primitive parameter (string, integer, etc.), you can write the value to an attribute of an object that is being imported.
+
+## 4 Troubleshooting{#troubleshooting}
+
+If you are importing a very long JSON string, the underlying conversion library may not be able to handle it. In this case you will get a message similar to the following:
+
+`Caused by: com.fasterxml.jackson.databind.JsonMappingException: String length (20051119) exceeds the maximum length (20000000)`
```
fazzdev: This is the actual exception the customer receiving based on the ticket `com.mendix.systemvideinterfaces.MendixRuntimeException: com.fasterxm1.jackson.core.exc.StreamConstraintsException: String value length (20051112) exceeds the maximum allowed (20000000, from 'StreamReadConstraints.getMaxStringLength()')`
Would it help the user better if the exception as close as possible?

MarkvanMents: Yes, I'll put the actual exception in. Thanks.
```suggestion
`com.mendix.systemvideinterfaces.MendixRuntimeException: com.fasterxm1.jackson.core.exc.StreamConstraintsException: String value length (20051112) exceeds the maximum allowed (20000000, from 'StreamReadConstraints.getMaxStringLength()')`
```

content/en/docs/refguide/modeling/integration/mapping-documents/import-mappings.md
MarkvanMents: Thanks for clarifying - I think we should be as accurate as possible.
```suggestion
Mendix recommends that you only add this setting if you will be getting very long JSON strings as it may result in more memory usage.
```

MarkvanMents (approved):
Thanks for the review.
Will commit the suggested changes and approve.
Will also create new PRs for 9.24.17 and 10.6.5.

PR #7388, opened 2024-03-05
Remove App Switcher documentation

MarkvanMents:

PR #7393, opened 2024-03-05
Revert "Remove App Switcher documentation"

MarkvanMents:
Reverts mendix/docs#7388

Module is not available, but service is deprecated. So restoring documentation and adding a deprecation notice.


MarkvanMents:
Approved over Slack.

PR #7399, opened 2024-03-06
Added deprecation notice to developer portal release notes and Feedback v1

danlawfordMX:


MarkvanMents (approved):
@danlawfordMX 
Thanks for the PR. Approving and merging as discussed on Slack.

PR #7423, opened 2024-03-11
Copy anchor to clipboard

dbreseman:


assets/js/anchor.js
```diff
@@ -1,46 +1,64 @@
...
+                placement: 'right',
+                trigger: 'hover focus'
+            });
+
+            // Add on-click behavior
+            a.addEventListener('click', (event) => {
+                event.preventDefault(); // Prevent scrolling
+                const href = a.href;
+                navigator.clipboard.writeText(href); // Copy to clipboard
+                window.history.pushState({}, document.title, href); // Update URL
```
MarkvanMents, line 45: Should we change the URL if we aren't scrolling there?
Perhaps talk to UX about what their preferred behavior is?

MarkvanMents (commented):
Hi Dana (@dbreseman)
This looks really great.
Two things which are not important, but might be nice to address.

1. See note about browser URL, below.

2. We still have this issue that there seems to be a space below headings which are nearly the whole width of the window

image
Where the purpose only becomes clear when you mouse over and the link icon appears
image
I don't know if there is an aesthetic solution to this?

Thanks for picking this up so quickly.

assets/js/anchor.js
dbreseman: Hi, thanks for the feedback.
- Extra spacing: Ah no, I had this resolved but must have reintroduced it when I changed the tooltip type. I'll take a look and see if I can eliminate this.
- URL: I think it does make sense to change the URL—if the user is copying something to clipboard, I think it's nice if they can see what it is. This is also what Confluence does. I'm not deeply attached to it though, and can certainly check in with UX on this—do you know who the right contact person or Slack channel would be?

dbreseman: @MarkvanMents Fixed now! If this looks good to you, I'll go ahead and merge, and then add any adjustments from the UX team once they've had a chance to review as well.

MarkvanMents: Thanks Dana:
* Extra spacing: solved (although I had to go to a private window to stop my cached javascript triggering!). Thanks.
* URL: let's publish it as it is, but talk to UX (Ashley?) about it. Normally what is in the URL reflects what is being displayed on the page. If we change the URL without moving the anchor to the top then that breaks the (my😀) paradigm. But perhaps it isn't a problem for UX and most people?

MarkvanMents (approved):
@dbreseman 
Thanks for this. We can follow up the UX issue with the URL later.

assets/js/anchor.js
dbreseman: Sounds good! I reached out to Ashley, so see what she thinks.

PR #7431, opened 2024-03-12
Auth Token Support Docs in 10.9

EnasAbdelrazek:


EnasAbdelrazek:
[Diagram link](https://www.plantuml.com/plantuml/uml/bPB1Rjim44Jl-efjV7M3n4Lww2687TSd5oZgzbOWeAMnj91Cgikb7AVklnTI65iA3BGv211dXyCt8E-y6j6mhP9tMc0BMbS1kPXzuaksjH6pfRL9ornSiDczgvpGQ5UmecVm-1LWKz3lP9ggauMp6grN7s_ci-b9Nl4JQ7ALJ4NSRkZffAFdUXA5yrap9ndaUJ07wbMvdrK1oP8tMB95VpwQVlXyivYTPhq-_VbN8yefryPg3tKeluRvazIdCLryWStuaUuhXjKBCfxIxlTuni3zBLZbBMW5QI2TtNU_D9fVBoQBUM9IvSOeOinn7Nr9J7z_UnkLJxHqELlQiHOKFAP1Y--kXANkgg0Gyb1IHoe1IJol7z10UBCdBT06o9XIWL5qWeVtxEoOl6b0tEPqUcRKh8t7c7vQCZQZdN3SqaByY13jg9z3gNCbmr-UAoBjg_9AEVGX-X_QS-WST5eWBRleVkcOFwOC5RG5xW1pMCXB9Js22LGO3GQ4mRFDpp1VzYcUwLjaNxksFMinsDDo9uV3eVtUrDLi53A8AemdFDOioFvnUusfE6FTazjqQnEqEBLboygcugTpVLbV-lr_lIJ3g-VQKo9Vr_yEDWlRwYy0)

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,50 @@
...
+
+Mendix now utilizes authentication tokens instead of earlier approaches for long-lived sessions to keep them session active while users are not actively working in the app, improving security and providing a more unified solution for session management across various application types. Consequently, long-lived sessions are now supported for both online and offline applications, whether native or web-based.
+
+### 2 Authentocation Token
+
+### 2.1 Enable Authentication Token
+
+We now support a new core client login API, `login2`, that allows you to pass an extra parameter, `useAuthToken`. You can pass `true` to enable authentication token or `false` to disable it. When the runtime receives this parameter during the login action, it adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated in the later `get_session_data` action.
```
WimJongeneel: It think that "get_session_data action" will be confusing for users because get_session_data isn't an action like we use **action** in other places.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,50 @@
...
+
+### 3 Application Behaviour
+
+### 3.1 Native and Offline PWA Applications
+
+This feature maintains backward compatibility, so we preserve the previous session management approach for existing applications. For upgraded offline applications, whether native or web, that currently use the old login API, `login`, are now supposed to use the new authentication token system by default. You can always disable this behaviour by using the new API, login2, and setting the useAuthToken parameter to false.
+
+### 4.1 Offline PWAs
+
+The key change is that Mendix now use authentication tokens to reinitialize the session  instead of the previous approach of keeping the users logged in for a longer period.
```
WimJongeneel: I think it is a bit strange to talk about "key changes" in the docs, we are only documenting the current behavior of the product

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,50 @@
...
+
+### 5.1 Native
+
+The key change is that Mendix now use HTTPOnly cookie to hold the authentication token instead of storing it on the device. This enhances security, as these cookies are inaccessible to anything other than the server.
```
WimJongeneel: I think it is a bit strange to talk about "key changes" in the docs, we are only documenting the current behavior of the product

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,50 @@
...
+
+With the introduction of this new core login API, You can begin modeling your login page to utilize the authentication token. Additionally, you can provide end users the option to decide whether they want to be remembered or not.
+
+### 2.2 Authentication Token Generation
+
+When the runtime receives the `get_session_data`, it checks the existence of the `useAuthToken` cookies. If present, it adds an additional HttpOnly cookie holding the generated unique authentication token `authtoken`. This token is then used in the subsequent actions to authenticate the user and extend the session upon its expiration.
+
+### 2.3 Authentication Token Expiry
+
+There is a custom runtime setting, `com.mendix.webui.HybridAppLoginTimeOut` that determines the validity of the authentication token. If no value is set, the token remains valid for one year.
```
WimJongeneel: Do we truly use a `Hybrid` setting for this?

EnasAbdelrazek: Yes, Its name is confusing as we don't support Hybrid anymore, but it is already used for the native auth token and we wanted our solution to be backwards compatible so we didn't want to add a new one.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: I mentioned that because we already support the old behaviour for the existing applications.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: what about changing it to `get_session_data` request?

content/en/docs/refguide/runtime/session-management.md
WimJongeneel: Oke :)

content/en/docs/refguide/runtime/session-management.md
WimJongeneel: Yes, or you could mention that it is during the client startup.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Removed.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Removed. They can just go and check the refguid9

WimJongeneel (approved):

content/en/docs/refguide/mobile/building-efficient-mobile-apps/offlinefirst-data/auto-committed-objects.md
```diff
@@ -35,9 +35,8 @@ It is true that an offline-first app *can* create auto-committed objects on the
 
 Auto-committed objects live until the user's session expires. When the user's session expires, the Mendix Runtime deletes any auto-committed objects created in that session. This logic works well for web applications, but is not a good fit for offline-first apps for the following reasons:
 
-* **Long-lived sessions** – offline-first progressive web apps use [long-lived sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) with longer session timeout by default. This causes auto-committed objects to remain on the server database as long as a session is active. Other PWA users can synchronize the auto-committed objects to their local databases. For example, a user changes and attempts to synchronize an auto-committed object with validation problems. In that case the server may fail to synchronize it.
```
MarkvanMents, line 38: Do we need to keep this in for customers who are on older versions? E.g. 10.6 MTS?

content/en/docs/refguide/runtime/mendix-client/_index.md
```diff
@@ -254,8 +256,10 @@ When the Mendix client is running, it sets a number of technical cookies to reco
 | **Profile** | Runtime | Holds the navigation profile that is being accessed within the session | `/` | until user closes their browser (session cookie) |  `true` | 
 | **SessionTimeZoneOffset** | Runtime | Holds the time zone offset for the session | `/` | until user closes their browser (session cookie) |  `true` | 
 | **xasid** | Runtime | Used for multi-instance fail-over | `/` | until user closes their browser (session cookie) |  `true` | 
-| **reloginReason** | Runtime | Used to let the client know that a relogin reason should be displayed on the sign in page | `/` | until user closes their browser (session cookie) |  `true` |
```
MarkvanMents, line 257: Should this still be documented for customers using versions before 10.9? E.g. 10.6 MTS.

content/en/docs/refguide/runtime/mendix-client/_index.md
```diff
@@ -254,8 +256,10 @@ When the Mendix client is running, it sets a number of technical cookies to reco
...
 | **SessionTimeZoneOffset** | Runtime | Holds the time zone offset for the session | `/` | until user closes their browser (session cookie) |  `true` | 
 | **xasid** | Runtime | Used for multi-instance fail-over | `/` | until user closes their browser (session cookie) |  `true` | 
-| **reloginReason** | Runtime | Used to let the client know that a relogin reason should be displayed on the sign in page | `/` | until user closes their browser (session cookie) |  `true` | 
 | **XASSESSIONID**<sup><small>2</small></sup> | Runtime | Holds the ID of the user's session | `/` | 1 year for offline sessions<sup><small>1</small></sup>, otherwise based on the session timeout |  `true` | 
+| **useAuthToken**<sup><small>2</small></sup> | Runtime | Used to let the runtime know whether to generate an authtoken or not | `/` | until user closes their browser (session cookie) |  `true` | 
+| **authtoken**<sup><small>2</small></sup> | Runtime | Used to hold the authentication token | `/` | until the authentication token expires |  `true` |
```
MarkvanMents: Are these new in 10.9? Then we need to indicate when they were introduced.

MarkvanMents: Footnote 2 isn't relevant here.
```suggestion
| **useAuthToken** | Runtime | Used to let the runtime know whether to generate an authtoken or not | `/` | until user closes their browser (session cookie) |  `true` | 
| **authtoken** | Runtime | Used to hold the authentication token | `/` | until the authentication token expires |  `true` | 
```

MarkvanMents: Also `useAuthToken` and `authtoken` have inconsistent camelcase for `token`/`Token`. I think that it should be consistent, otherwise customers will type it wrongly.

content/en/docs/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app.md
```diff
@@ -134,9 +134,9 @@ For more information on the available properties in the web app manifest, read t
 
 ### 5.2 Sessions {#sessions}
 
-Offline-first PWAs use long-lived sessions, which keep users logged in for a longer period even after their apps are closed. By default, users will be logged out after 7 days of inactivity. This can be customized using the *LongLivedSessionTimeout* runtime setting.
+Offline-first PWAs use authentication token by default to keep the session active while users are not actively working in the app.
 
-For more information on sessions and how to customize the timeout, see the [Session Duration](/refguide/tricky-custom-runtime-settings/#session-duration) section of the *Advanced Custom Settings in Mendix Runtime Reference Guide*.
```
MarkvanMents, line 139: The link to information on the `session-duration`  has not been carried forward into the new **Session Management** doc, so it is more difficult to find for customers using earlier versions of Mx10.

content/en/docs/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app.md
```diff
@@ -134,9 +134,9 @@ For more information on the available properties in the web app manifest, read t
 
 ### 5.2 Sessions {#sessions}
 
-Offline-first PWAs use long-lived sessions, which keep users logged in for a longer period even after their apps are closed. By default, users will be logged out after 7 days of inactivity. This can be customized using the *LongLivedSessionTimeout* runtime setting.
+Offline-first PWAs use an authentication token by default to keep the session active while users are not actively working in the app.
```
MarkvanMents, line 139: Do we still need to talk about long-lived sessions for customers using older versions, e.g. 10.6 MTS

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,44 @@
...
+
+Mendix versions 10.9.0 and above, use authentication tokens to keep a session active while users are not actively working in the app. This replaces earlier approaches for long-lived sessions, improving security and providing a more unified solution for session management across various types of application. This means that long-lived sessions are now supported for both online and offline applications, whether native or web-based.
+
+### 2 Authentication Token
+
+### 2.1 Enable Authentication Token
+
+We now support a new core client login API, `login2`, that allows you to pass an extra parameter, `useAuthToken`. You can pass `true` to enable authentication token or `false` to disable it. When the runtime receives this parameter during the login action, it adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup.
```
MarkvanMents: I think an example would be useful here. Or at least a link to the document which describes calling the new API.

And if the user has to call the API directly, perhaps a link to the API spec as well?

MarkvanMents (changes requested):
Hi Enas, @EnasAbdelrazek 

The new doc, **Session Management** looks a good description of what is technically going on, but I'm not sure how to use it as a customer. As a customer, I don't call an API to sign in, I just set security to production and set up some users in my app. Does this new session management apply then? How do I pass this `useAuthToken` parameter in this case?

I've also made a few comments below about supporting customers who are not yet on 10.9.

Probably best to discuss this over Slack, or have a quick meeting if you have any questions.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,44 @@
...
+
+### 3 Application Behavior
+
+### 3.1 Native and Offline PWA Applications
+
+This feature maintains backward compatibility, so we preserve the previous session management approach for existing applications. For upgraded offline applications, whether native or web, that currently use the old login API, `login`, are now supposed to use the new authentication token system by default. You can always disable this behavior by using the new API, login2, and setting the useAuthToken parameter to false.
+
+### 3.2 Client-Runtime Session Management Flow
+
+{{< figure src="/attachments/refguide/runtime/session.png" alt="The Session Management Flow between client and the runtime." >}}
```
MarkvanMents, line 79: Could you add the link to plantuml which creates this image. The it is easier to update it, if necessary? Thanks.

content/en/docs/refguide/mobile/building-efficient-mobile-apps/offlinefirst-data/auto-committed-objects.md
```diff
@@ -35,9 +35,8 @@ It is true that an offline-first app *can* create auto-committed objects on the
...
-    * It is useful to keep in mind that a native session may expire on the server if the end-user uses the app while there is no network condition. When this happens, the Mendix Client attempts to create a new session automatically when the device sends a request to the Mendix Runtime.  
+* **Session expiration** – offline apps use regular sessions that expire according to the [`SessionTimeout` runtime setting](/refguide/tricky-custom-runtime-settings/#general-settings). A session on the server may expire as the user uses the app. The server will remove any auto-committed objects, even if the end-user is still interacting with the app.This can cause unexpected behaviors. For example, suppose the user changes the auto-committed object and attempts to synchronize. The Mendix Runtime will not be able to do that, because the runtime database no longer has this object.
+    * It is useful to keep in mind that a session may expire on the server if the end-user uses the app while there is no network condition. When this happens, the Mendix Client attempts to create a new session automatically when the device sends a request to the Mendix Runtime.
```
MarkvanMents: Should this be 
```suggestion
    * It is useful to keep in mind that a session may expire on the server if the end-user uses the app while there is no network connection. When this happens, the Mendix Client attempts to create a new session automatically when the device sends a request to the Mendix Runtime.
```

content/en/docs/refguide/runtime/mendix-client/_index.md
```diff
@@ -254,8 +256,11 @@ When the Mendix client is running, it sets a number of technical cookies to reco
...
-| **reloginReason** | Runtime | Used to let the client know that a relogin reason should be displayed on the sign in page | `/` | until user closes their browser (session cookie) |  `true` | 
+| **reloginReason** | Runtime | Used to let the client know that a relogin reason should be displayed on the sign in page—*not used by apps created in Mendix version 10.9.0 and above* | `/` | until user closes their browser (session cookie) |  `true` | 
 | **XASSESSIONID**<sup><small>2</small></sup> | Runtime | Holds the ID of the user's session | `/` | 1 year for offline sessions<sup><small>1</small></sup>, otherwise based on the session timeout |  `true` | 
+| **useAuthToken** | Runtime | Lets the runtime know whether to generate an authtoken or not—*introduced for apps created in Mendix version 10.9.0 and above* | `/` | until the end-user closes their browser (session cookie) |  `true` | 
+| **authtoken** | Runtime | Used to hold the authentication token—*introduced for apps created in Mendix version 10.9.0 and above* | `/` | until the authentication token expires |  `true` |
```
MarkvanMents, line 262: **Inconsistent Camel Case**
You have two cookies: `useAuthToken` and `authtoken`. The `token` part has inconsistent case. It should be one of

- `useAuthToken` and `authToken`
- `useAuthtoken` and `authtoken`

Otherwise developers will make mistakes when using them.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,52 @@
...
+
+### 2.2 Authentication Token Generation
+
+When the runtime receives the `get_session_data` request, it checks the existence of the `useAuthToken` cookies. If present, it adds an additional HttpOnly cookie holding the generated unique authentication token `authtoken`. This token is then used in the subsequent actions to authenticate the user and extend the session upon its expiration.
+
+This approach enhances security, as these cookies are inaccessible to anything other than the server.
+
+### 2.3 Authentication Token Expiry
+
+There is a custom runtime setting, [`com.mendix.webui.HybridAppLoginTimeOut`](/refguide/custom-settings/#commendixwebuiHybridAppLoginTimeOut) that determines how long the authentication token is valid for. If no value is set, the token remains valid for one year.
```
MarkvanMents, line 49: According to our documentation, the token never expires if this value isn't set.
https://docs.mendix.com/refguide/custom-settings/#commendixwebuiHybridAppLoginTimeOut

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,52 @@
...
+
+### 3 Application Behavior
+
+### 3.1 Native and Offline PWA Applications
+
+This feature maintains backward compatibility, so we preserve the previous session management approach for existing applications. For upgraded offline applications, whether native or web, that currently use the old login API, `login`, are now supposed to use the new authentication token system by default. You can always disable this behavior by using the new API, login2, and setting the useAuthToken parameter to false.
```
MarkvanMents: I'm afraid I don't understand what we are saying here?
The best I can do is:
* this feature mimics the long-lived sessions for offline first apps
* old apps will continue to use the long-lived sessions mechanism even if they are upgraded to 10.9
* all apps created in 10.9 will have the same behavior as offline-first apps did before
* you can turn off this behavior if you are writing custom java or javascript using login2 by setting the value of the `useAuthToken` parameter to `false`.

I'm not sure I've understood it correctly, so I think it could do with some clarification once I'm sure I understand it.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,52 @@
...
+
+If you are using an app created in a version below 10.9.0, then long-lived sessions will be used for offline-first apps. See the [Sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) section of *Progressive Web App* for more information.
+
+### 2 Authentication Token
+
+If you are using standard Mendix sign in flows in version 10.9.0 or above, the authentication token will be generated automatically. If you have custom JavaScript or Java sign in code, you will have to ensure that it uses the `login2` API if you want to create authentication tokens. 
+
+### 2.1 Enable Authentication Token
+
+{{% todo %}}Need a link to login2 api documentation{{% /todo %}}
```
MarkvanMents: Where is the spec for the login2 API - would be good to link to it.

MarkvanMents (changes requested):
I've now made some changes based on what @WimJongeneel told me over Slack.
Can you have a look at my remaining questions and review [the changes I've made](https://github.com/mendix/docs/pull/7431/files/e575c1a4a82e9822fe01197a25a2dea2c05c5a92..5c93f3ca950486a07133128a9e54b2f1dc288230).
Thanks.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,17 +9,25 @@ weight: 35
...
+Mendix versions 10.9.0 and above, use authentication tokens to keep a session active while users are not actively working in the app. This replaces earlier approaches for long-lived sessions in offline-first apps, improving security. This approach provides a more unified solution for session management, allowing long-lived sessions in  both online and offline applications, whether native or web-based.
+
+If you create a new app in Mendix version 10.9.0 or above, session management will work as described below.
+
+If you are using an app created in a version below 10.9.0, then long-lived sessions will be used for offline-first apps. See the [Sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) section of *Progressive Web App* for more information.
 
 ### 2 Authentication Token
 
+If you are using standard Mendix sign in flows in version 10.9.0 or above, the authentication token will be generated automatically. If you have custom JavaScript or Java sign in code, you will have to ensure that it uses the `login2` API if you want to create authentication tokens.
```
EnasAbdelrazek: Offline applications, whether native or web, that use the old login API, are now supposed to use the new authentication token system by default.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Old offline PWA apps will continue to use the long-lived sessions mechanism, but upgraded apps will use the new system by default if they already use the old API but developers can disable this default system if they use the login2 and set useAuthToken param to false. 

Online apps can use the new login API to enable the authentication token system.

EnasAbdelrazek: Since our solution is backwards compatible, the offline PWA apps (not upgraded to 10.9 or above) will continue to use the previous long-lived sessions mechanism and the existing native apps should continue to work as the previously old system of authentication tokens, in which the authentication token is generated during login and stored on the device.

Upgraded offline apps (native or web) will use the new system by default if they already use the old API; however, developers can disable this default system if they use the login2 and set useAuthToken param to false.

Online apps can use the new login2 API to enable the authentication token system.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: It is part of the client API docs which should be at the release of 10.9

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: [Plantuml link](https://www.plantuml.com/plantuml/uml/bPB1Rjim44Jl-efjV7M3n4Lww2687TSd5oZgzbOWeAMnj91Cgikb7AVklnTI65iA3BGv211dXyCt8E-y6j6mhP9tMc0BMbS1kPXzuaksjH6pfRL9ornSiDczgvpGQ5UmecVm-1LWKz3lP9ggauMp6grN7s_ci-b9Nl4JQ7ALJ4NSRkZffAFdUXA5yrap9ndaUJ07wbMvdrK1oP8tMB95VpwQVlXyivYTPhq-_VbN8yefryPg3tKeluRvazIdCLryWStuaUuhXjKBCfxIxlTuni3zBLZbBMW5QI2TtNU_D9fVBoQBUM9IvSOeOinn7Nr9J7z_UnkLJxHqELlQiHOKFAP1Y--kXANkgg0Gyb1IHoe1IJol7z10UBCdBT06o9XIWL5qWeVtxEoOl6b0tEPqUcRKh8t7c7vQCZQZdN3SqaByY13jg9z3gNCbmr-UAoBjg_9AEVGX-X_QS-WST5eWBRleVkcOFwOC5RG5xW1pMCXB9Js22LGO3GQ4mRFDpp1VzYcUwLjaNxksFMinsDDo9uV3eVtUrDLi53A8AemdFDOioFvnUusfE6FTazjqQnEqEBLboygcugTpVLbV-lr_lIJ3g-VQKo9Vr_yEDWlRwYy0)

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: It should be released at 10.9

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Example with useAuthToken set to true
```

mx.login2("fred", "fred's unguessable password", true, function() {
    console.log("success");
}, function() {
    console.log("wrong password");
});
```

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Now, the authtoken is stored as an HttpOnly cookie, and we set the expiry value for the cookie using this timeout, but if it is not specified, we set a default value of one year.

content/en/docs/refguide/runtime/mendix-client/_index.md
EnasAbdelrazek: What do you mean by using them?

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: Sorry, missed that you had already put it in an earlier comment:
https://github.com/mendix/docs/pull/7431#issuecomment-1991684567

content/en/docs/refguide/runtime/mendix-client/_index.md
MarkvanMents: I was thinking that the customer might have to read/write these cookies themselves if they are writing custom login functionality. But if it is all handled then it is only internal developers who have to make sure they type them correctly!

(Would still be nice if they were consistent 😀)

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: So do I need to change the information in https://docs.mendix.com/refguide/custom-settings/#commendixwebuiHybridAppLoginTimeOut
image

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: OK - do you have a link I could use, or do I need to ask nearer the time?

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: This doesn't seem to be the same as what you say here: https://github.com/mendix/docs/pull/7431#discussion_r1527863770

> existing offline PWA apps will continue to use the previous long-lived sessions mechanism and the existing native apps should continue to work as the previously old system of authentication tokens,

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -0,0 +1,52 @@
...
+### 2 Authentication Token
+
+If you are using standard Mendix sign in flows in version 10.9.0 or above, the authentication token will be generated automatically. If you have custom JavaScript or Java sign in code, you will have to ensure that it uses the `login2` API if you want to create authentication tokens. 
+
+### 2.1 Enable Authentication Token
+
+{{% todo %}}Need a link to login2 api documentation{{% /todo %}}
+
+The `login2` API allows you to pass the parameter `useAuthToken`. The value `true` enables the authentication token, `false` disables it. If the login action passes `true` to the runtime, the runtime adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup.
+
```
MarkvanMents: Does it make it clearer with an example?
```suggestion
The `login2` API allows you to pass the parameter `useAuthToken`. The value `true` enables the authentication token, `false` disables it. If the login action passes `true` to the runtime, the runtime adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup.

For example
```java {linenos=false}
mx.login2("fred", "fred's unguessable password", true, function() {
    console.log("success");
}, function() {
    console.log("wrong password");
});
```
```

File: content/en/docs/refguide/runtime/session-management.md (hunk 10 above)
Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md]: Should we add this example, as I suggest here: https://github.com/mendix/docs/pull/7431/files#r1528426157

Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/session-management.md]: No, It should be part of the client API docs https://apidocs.rnd.mendix.com/10/client/mx.html or https://apidocs.rnd.mendix.com/10/client-react/mx.html depending on the used client.

File: content/en/docs/refguide/runtime/mendix-client/_index.md (hunk 13 above)
Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/mendix-client/_index.md:262]: No, they are only accessible by the runtime.

File: content/en/docs/refguide/runtime/session-management.md (hunk 14 above)
Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/session-management.md:49]: Yes, you can change it.

File: content/en/docs/refguide/runtime/session-management.md (hunk 16 above)
Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/session-management.md]: https://apidocs.rnd.mendix.com/10/client/mx.html or https://apidocs.rnd.mendix.com/10/client-react/mx.html depending on the used client.

File: content/en/docs/refguide/runtime/session-management.md (hunk 17 above)
Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/session-management.md]: I updated the comment. I hope now it is more clear.

File: content/en/docs/refguide/runtime/session-management.md (hunk 18 above)
Comment (EnasAbdelrazek) [reply content/en/docs/refguide/runtime/session-management.md]: This example will be included in the client API docs https://apidocs.rnd.mendix.com/10/client/mx.html or https://apidocs.rnd.mendix.com/10/client-react/mx.html depending on the used client.

Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md]: OK - won't put it in this document as well, but have a link to the API spec.

File: content/en/docs/refguide/runtime/session-management.md (hunk 14 above)
Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md:49]: Added in 665c3a223a0b228da6615c1543f4974082af6797

File: content/en/docs/refguide/runtime/session-management.md (hunk 15 above)
Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md]: Updated description in 48e98aeb43628fb87fe0628d302dfa9a7a21036e

File: content/en/docs/refguide/runtime/session-management.md (hunk 17 above)
Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md]: Updated description in 48e98aeb43628fb87fe0628d302dfa9a7a21036e

File: content/en/docs/refguide/runtime/session-management.md (hunk 16 above)
Comment (MarkvanMents) [reply content/en/docs/refguide/runtime/session-management.md]: Added,

File: content/en/docs/refguide/runtime/session-management.md [hunk 19]
```diff
@@ -9,46 +9,62 @@ weight: 35
...
+This approach provides a more unified solution for session management, allowing long-lived sessions to be used in online, as well as offline applications, whether native or web-based. For example, you could customize the sign in to an online app to provide a "remember me" function so that end-users do not have to provide their credentials again if their session expires.
 
 If you create a new app in Mendix version 10.9.0 or above, session management will work as described below.
 
-If you are using an app created in a version below 10.9.0, then long-lived sessions will be used for offline-first apps. See the [Sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) section of *Progressive Web App* for more information.
+If you are using an app created in a version below 10.9.0, then long-lived sessions will be used for offline-first apps. See the [Sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) section of *Progressive Web App* for more information. Online apps in Mendix versions below 10.9.0 cannot have their sessions extended.
```
EnasAbdelrazek: Online apps below 10.9.0 should work with KeepAlive approach check this https://docs.mendix.com/refguide7/tricky-custom-runtime-settings/#21-web-client-settings

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,46 +9,62 @@ weight: 35
...
 
-Your login page will then utilize the authentication token. You can also provide end-users the option to decide whether they want to be remembered or not.
+### 2.2 Enable Authentication Token
 
-Follow the links below to find the login2 API specifications for the following:
+For offline-first apps using standard Mendix sign in with version 10.9.0 or above, the authentication token will be generated automatically.
 
-* [React](https://apidocs.rnd.mendix.com/10/client-react/mx.html#.login2)
-* [Non-react (Dojo)](https://apidocs.rnd.mendix.com/10/client/mx.html#.login2)  
+If you want to generate authentication tokens in your online app, you will have to use custom JavaScript or Java sign in code using the `login2` API.
```
EnasAbdelrazek: It is not correct to mention java action here. login2 is a client API.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,46 +9,62 @@ weight: 35
...
+When the runtime receives the `get_session_data` request, it checks the existence of the `useAuthToken` cookies. If present, it adds an additional HttpOnly cookie holding the generated unique authentication token `authtoken`. This token is then used in the subsequent actions to authenticate the user and extend the session when it expires.
 
-The `login2` API allows you to pass the parameter `useAuthToken`. The value `true` enables the authentication token, `false` disables it. If the login action passes `true` to the runtime, the runtime adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup.
+This approach enhances security, as these cookies are inaccessible to anything other than the server.
 
-Your login page will then utilize the authentication token. You can also provide end-users the option to decide whether they want to be remembered or not.
+### 2.2 Enable Authentication Token
 
-Follow the links below to find the login2 API specifications for the following:
+For offline-first apps using standard Mendix sign in with version 10.9.0 or above, the authentication token will be generated automatically.
```
EnasAbdelrazek:
```suggestion
For offline-first apps using the client API `login`, with version 10.9.0 or above, the authentication token system will be used by default.
```

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek:
```suggestion
If you want to generate authentication tokens in your online app, you have to use the client API `login2`. 
```

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,46 +9,62 @@ weight: 35
...
 
-* [React](https://apidocs.rnd.mendix.com/10/client-react/mx.html#.login2)
-* [Non-react (Dojo)](https://apidocs.rnd.mendix.com/10/client/mx.html#.login2)  
+If you want to generate authentication tokens in your online app, you will have to use custom JavaScript or Java sign in code using the `login2` API. 
 
-### 2.2 Authentication Token Generation
+The `login2` API allows you to pass the parameter `useAuthToken`. The value `true` enables the authentication token, `false` disables it. If the login action passes `true` to the runtime, the runtime adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup. Your login flow will then utilize the authentication token.
```
EnasAbdelrazek:
```suggestion
The `login2` API allows you to pass the parameter `useAuthToken`. The value `true` enables the authentication token, `false` disables it. If the login action passes `true` to the runtime, the runtime adds an HttpOnly cookie `useAuthToken` indicating that an authentication token should be generated during the client startup. 
```

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,46 +9,62 @@ weight: 35
...
 
-When the runtime receives the `get_session_data` request, it checks the existence of the `useAuthToken` cookies. If present, it adds an additional HttpOnly cookie holding the generated unique authentication token `authtoken`. This token is then used in the subsequent actions to authenticate the user and extend the session upon its expiration.
+You could also provide end-users a "remember me" option so they can decide whether they want to be remembered or not.
```
EnasAbdelrazek:
```suggestion
You could also model your login page to provide end-users with a "remember me" option so they can decide whether they want to be remembered or not.
```

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -9,46 +9,62 @@ weight: 35
...
 
 There is a custom runtime setting, [`com.mendix.webui.HybridAppLoginTimeOut`](/refguide/custom-settings/#commendixwebuiHybridAppLoginTimeOut) that determines how long the authentication token is valid for. If no value is set, the token remains valid for one year.
 
-### 3 Application Behavior
+## 3 Application Behavior
 
 ### 3.1 Native and Offline PWA Applications
 
-This feature maintains backward compatibility, so we preserve the previous session management approach for existing applications. For upgraded offline applications, whether native or web, that currently use the old login API, `login`, are now supposed to use the new authentication token system by default. You can always disable this behavior by using the new API, login2, and setting the useAuthToken parameter to false.
+In Mendix version 10.9.0 and above, login should always support authentication tokens.
```
EnasAbdelrazek:
```suggestion
This feature maintains backward compatibility, If end users haven't updated their apps on their devices, but the runtime is a newer version, it still functions correctly to support the previous session management methods.

Upgraded offline-first applications to version 10.9 or above, that currently use the client API, `login`, will use the new authentication token system by default as we internally call the new API login2 and enable the authentication token system. If you do not want your offline-first app to create an authentication token, you can disable it by using the new API, `login2`, but setting the `useAuthToken` parameter to `false`.

Applications created with version 10.9 or above, whether offline or online, have the flexibility to enable or disable the authentication token system via the client API 'login2' and set the 'useAuthToken' parameter according to their preferences.
```

content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -249,7 +249,7 @@ The settings below influence the behavior of the Mendix web client.
...
 | <a id="TabletUserAgentRegEx" href="#TabletUserAgentRegEx">TabletUserAgentRegEx</a> | Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a tablet. The regular expression is matched against the User-Agent header sent by the client's web browser. | Android, iPad |
-| <a id="commendixwebuiHybridAppLoginTimeOut" href="#commendixwebuiHybridAppLoginTimeOut">com.<wbr>mendix.<wbr>webui.<wbr>HybridAppLoginTimeOut</a> | Determines how many minutes your token will remain valid before re-authenticating using your full credentials. If no value is set the token will never expire. | |
+| <a id="commendixwebuiHybridAppLoginTimeOut" href="#commendixwebuiHybridAppLoginTimeOut">com.<wbr>mendix.<wbr>webui.<wbr>HybridAppLoginTimeOut</a> | Determines how many minutes your token will remain valid before re-authenticating using your full credentials. | One year *(Mendix version 10.9.0 and above)*<br />Never *(Mendix versions below 10.9.0)* |
```
EnasAbdelrazek: I don't think there is a need to change the default value since the one-year is related to the authtoken cookie expiry, not this timeout.

EnasAbdelrazek: If no value is set the token will expire after one year.

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: So should we say 
```suggestion
If you are using an app created in a version below 10.9.0, then long-lived sessions will be used for offline-first apps. See the [Sessions](/refguide/mobile/introduction-to-mobile-technologies/progressive-web-app/#sessions) section of *Progressive Web App* for more information.

Online apps in Mendix versions below 10.9.0 have a different approach, using the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to send keep alive requests from the client.
```

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: This is the bit which I find is confusing.

> we preserve the previous session management approaches

I think it is clear that we are talking about apps in 10.9 and above so I don't think we should talk about "preserving" anything. My understanding is that in 10.9 and above they have to use the new mechanism. If you talk about preserving things it sounds like they don't have to do anything.

> are now supposed to use the new authentication token system by default

This doesn't tell me what I need to know. "supposed to" could mean "you should change it but you can leave it as it is if you want", or it could mean "the runtime assumes that this is the case". Saying it is "by default" suggests that this is implemented automatically and that you have to disable it if you don't want it. But you go on to say that it can be disabled and in the next paragraph you say that it has to be enabled (implemented?) by the developer.

*Can you make this clearer as to whether this is something that has to be implemented by the developer and that the login2 API gives them the ability to do this, or whether this is something that is being implemented "by default" for offline-first applications (i.e. by making an app offline-first this mechanism is automatically applied)*

MarkvanMents (changes requested):
Hi @EnasAbdelrazek 
Thanks for the suggestions, I've committed all but one of them where I have asked some questions about it as I find the language needs to be clearer.

I have also made one further suggestion based on your comments, can you see if that looks OK?

Thanks for working with me on this. I think we will end up with something which clearly states what is going on and how the customer should approach it, so the extra work is going to be worth it in the end.

Mark

content/en/docs/refguide/runtime/custom-settings/_index.md
MarkvanMents: But we need to change something as now it says 

> If no value is set the token will never expire

So should we put it in the description instead?

```suggestion
| <a id="commendixwebuiHybridAppLoginTimeOut" href="#commendixwebuiHybridAppLoginTimeOut">com.<wbr>mendix.<wbr>webui.<wbr>HybridAppLoginTimeOut</a> | Determines how many minutes your token will remain valid before re-authenticating using your full credentials.  If no value is set the token will expire after One year *(Mendix version 10.9.0 and above)* or never expire *(Mendix versions below 10.9.0)*. |  |
```

I prefer this as the default (i.e what will happen if you don't explicitly set `commendixwebuiHybridAppLoginTimeOut`) as it is clearer to me. But I can make the change I've suggested if you think that is more accurate.

content/en/docs/refguide/runtime/session-management.md
EnasAbdelrazek: Here I am referring to apps that have not been upgraded we kept our runtime backwards compatible (can handle both approaches).

EnasAbdelrazek: `New offline-first applications created in version 10.9.0 or above will use the authentication token approach if they enable it`.
 we expect they will use the new login2 client API but if they use login It will also utilize the token system by default as mentioned before as they are offline.

content/en/docs/refguide/runtime/custom-settings/_index.md
EnasAbdelrazek: I don't think `never` is a valid default value, so I prefer your suggestion.

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: Here we are talking about apps which are above 10.9 (either new or upgraded) not ones below, so I don't think we should be talking about preserving anything. If they are below 10.9 they will be using an older runtime version.

And if they are below 10.9 they will be calling the login API, so they will get the behaviour that they had with the login API and I don't think we need to mention that either.

I don't understand how using the old API in an app above 10.9 can also utilize the token system by default - in the past they had to set up long-lived sessions if they were using the login API.

As far as I can see there are three situations we need to describe clearly:
1) If you are using Mx above 10.9 and new login2 this is what happens
2) If you are using Mx below 10.9 and old login this is what happens
3) If you use Mx above 10.9 and old login you should change to using login2 but it will carry on working as 2) if you don't.

I think that all this "maintains backward compatibility" is not relevant as we are mixing up talking about two different APIs and two different Mendix versions.

MarkvanMents:
One other question:
Can users with Mendix versions below 10.9 use the login2 call - or will it only work for Mendix 10.9?
If so, will this be included in the https://apidocs.rnd.mendix.com/10/client-react/mx.html#.login documentation. And will this documentation also tell users which version of login they should be using?

content/en/docs/refguide/runtime/session-management.md
MarkvanMents: and
4. You cannot use login2 with Mendix versions below 10.9

EnasAbdelrazek: - Regarding backward compatibility: If end users haven't updated their apps on their devices, but the runtime is a newer version, it should still function correctly to accommodate both old and new clients.
- Upgraded `offline` apps that utilize login API, they will use the new token system, as we internally call the new API login2 with authToken enabled.
- New apps (whether offline or online) created with version 10.9 or above now have two options: login2 or login. If they opt to use the authentication token system, they are allowed to use login2 and set the useAuthToken parameter to whatever value they prefer. However, as previously mentioned, if they used the login API and are offline, it will act login2 with the useAuthToken parameter set to true.

EnasAbdelrazek: I updated the suggestion.

MarkvanMents: That looks great. I now understand it and the documentation is also clear.
I'll commit your suggestion and now I think we are ready to publish.

MarkvanMents (approved):
@EnasAbdelrazek 
Thanks for that final update. I think everything is much clearer now and I understand what is going on - which means our customers should too.

Sorry it has taken so long to get to this place - hope I wasn't being too stupid in not understanding it earlier - but I'm very happy that we now have documentation which is really clear.

The technical side of the documentation was always very clear - great use of the UML to explain the interaction of the client and the server. I think our customers should be very happy with this update.

Mark

EnasAbdelrazek:
@MarkvanMents No worries, I understand your confusion, it is a bit complicated, but the good news is we were able to come up with clear documentation. Thanks for your feedback as well! :)

PR #7432, opened 2024-03-12
Linaaf/buildrestconnector

strategicalliances:
Changed chapter about building a REST based connector


katarzyna-koltun-mx:
I think this was assigned to @quinntracy in error. Reassigning.

Issue #7435, opened 2024-03-12
Add Gradle Networking Requirements


cdcharlebois:
I believe that the use of managed dependencies with version 10.3+ also requires Studio Pro to be able to access the following URLs (to fetch managed dependencies)

```
repo.maven.apache.org
plugins.gradle.org
jcenter.bintray.com
plugins-artifacts.gradle.org
```

This should be added to the documentation, either under system requirements with the other URLs that need to be whitelisted in system requirements [here](https://docs.mendix.com/refguide/system-requirements/#firewall-settings) or with the managed dependencies docs [here](https://docs.mendix.com/refguide/managed-dependencies/#5-migrating-from-unmanaged-to-managed-dependencies)

This issue has come up with multiple customers recently and should be added to the docs.

MarkvanMents:
Hi Connor,
This information is already in our documentation here: https://docs.mendix.com/refguide/managed-dependencies/#8-troubleshooting
And is being rewritten in #7343
I'll pick this up on your PR #7436

PR #7436, opened 2024-03-12
Update system-requirements.md

cdcharlebois:
#7435 add gradle URLs to firewall settings

### DIFF
--- a/content/en/docs/refguide/installation/system-requirements.md
+++ b/content/en/docs/refguide/installation/system-requirements.md
@@ -78,6 +78,10 @@ Studio Pro needs access to the following URLs in order to work. If your firewall
 * `*.mendix.com`
 * `*.mendixcloud.com`
 * `*.teamserver.sprintr.com`
+* `repo.maven.apache.org`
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi Connor,
Hi Connor,
This information is already in our documentation here: https://docs.mendix.com/refguide/managed-dependencies/#8-troubleshooting
And is being rewritten in #7343

Are these URLs always required, or only when using Managed Dependencies?
If the former, then we should put it here. But if the latter, then customers should be able to pick it up when reading the Managed Dependencies documentation.

MarkvanMents (approved):
@cdcharlebois 
Thanks for the PR, Connor.
I've been told by the developer that the additional firewall settings are [only needed if managed dependencies are being used](https://github.com/mendix/docs/pull/7343/files#r1523415242). So I've updated your note to point to the more detailed information in the Managed Dependencies document.

Please let me know if you are happy with the changes. I'm approving it from my side.

MarkvanMents:
No further response, so merging the updated version.

PR #7441, opened 2024-03-13
Catalog release 2.114.0

BananiMendix:
Catalog release 2.114.0

### DIFF
--- a/content/en/docs/releasenotes/catalog/_index.md
+++ b/content/en/docs/releasenotes/catalog/_index.md
@@ -17,7 +17,7 @@ These release notes cover changes made to the [Catalog](/catalog/).
 
 #### Fixes
 
-* We fixed an issue where endpoints would get re-created once for an application that was converted from Mendix 10.6 or older to 10.7 or newer.
+* We fixed an issue where endpoints would get deleted and re-created when an application was converted from Mendix version 10.6 or below to  version 10.7 or above.
 
 ### February 29, 2024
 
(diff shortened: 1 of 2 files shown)


content/en/docs/releasenotes/catalog/_index.md
```diff
@@ -13,6 +13,12 @@ These release notes cover changes made to the [Catalog](/catalog/).
 
 ## 2024
 
+### March 14, 2024
+
+#### Fixes
+
+* We fixed an issue where endpoints would get re-created once for an application that was converted from Mendix 10.6 or older to 10.7 or newer.
```
MarkvanMents: Hi @BananiMendix 
Thanks for the PR.
What do you mean "would get re-created once"? Do you mean they were duplicated - so you got the same endpoint twice in the same application?

Also - was there a zendesk ticket for this we can refer to, or was it reported internally?

BananiMendix: Hi Mark,
The endpoints are not duplicated , instead after creating new endpoints the old ones were being deleted in the process.
There is no Zendesk ticket for this .

MarkvanMents: Thanks for the update.  Can we say:
```suggestion
* We fixed an issue where endpoints would get deleted and re-created when an application was converted from Mendix version 10.6 or below to  version 10.7 or above.
```

MarkvanMents (approved):

PR #7443, opened 2024-03-13
Portfolio Management Peer Review

dbreseman:


content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -167,8 +167,8 @@ To create a new tag, do as follows:
```
dbreseman, line 167: It looks like this has since been changed to an autosave?

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -504,7 +504,7 @@ The table below shows the permissions of Portfolio Managers, Contributors, and V
 
 | Action | Portfolio Manager | Contributor | Viewer\* |
 | --- | --- | --- | --- |
-| Invite users | {{< figure src="/attachments/developerportal/portfolio-management/check-mark.svg" >}} | {{< figure src="/attachments/developerportal/portfolio-management/cross-mark.svg" >}} | {{< figure src="/attachments/developerportal/portfolio-management/cross-mark.svg" >}} |
+| Invite users | {{< icon name="checkmark-circle-filled" color="green" >}} | {{< icon name="checkmark-circle-filled" color="green" >}} | {{< icon name="remove-circle-filled" color="red" >}} |
```
dbreseman: What do you think of embedding these as icons instead of images? It would take care of the padding issue that the images seem to bring, and it gets rid of the implication that you'll learn more about the topic if you click on the image.

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -277,7 +277,7 @@ To change your view on the **Initiatives Overview** page, click the drop-down li
...
 
-In Kanban view, all the initiatives are represented by initiative cards, and are categorized in different columns. A column indicates the stage that an initiative is currently in. Clicking an initiative card opens a [side pane](#view-initiative) that shows initiative details. To move an initiative to a different stage, hover over the initiative card, and then click **Move Left** ({{% icon name="chevron-left" %}}) or **Move Right** ({{% icon name="chevron-right" %}}) on the left or right side of the card.
+In Kanban view, all the initiatives are represented by initiative cards and are categorized in different columns. A column indicates the stage that an initiative is currently in. Clicking an initiative card opens a [side pane](#view-initiative) that shows initiative details. To move an initiative to a different stage, hover over the initiative card, and then click **Move Left** ({{% icon name="chevron-left" %}}) or **Move Right** ({{% icon name="chevron-right" %}}) on the left or right side of the card.
 
 {{% alert type="info" %}}A Portfolio Manager can set up custom stages for initiatives on the [Portfolio Settings](#portfolio-settings) page.{{% /alert %}}
```
dbreseman, line 283: What would you think about tweaking the design of the labeling system, perhaps to something like https://help.tableau.com/current/pro/desktop/en-us/environment_workspace.htm? It seems to me like the current method is fairly clear, but might benefit from the Design team taking a look—I'm curious what your thoughts are, since it seems like your components involve a lot of interfaces that need labeling.

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -64,7 +64,7 @@ On the **Portfolio Landscape Overview** page, you can create a new portfolio as
 
 2. Enter **Portfolio Name** and **Description** for the new portfolio.
 
-3. For **Privacy Settings**, select **Private**, **Restricted**, or **Open**. For details on privacy settings, see the [Different Privacy Settings of a Portfolio](#privacy-settings) section.
+3. For **Privacy Settings**, select **Private**, **Restricted**, or **Open**. For details on privacy settings, see the [Different Privacy Settings of a Portfolio](#privacy-settings) section, above.
```
dbreseman, line 67: I think flags like this for links to other sections in this page would make it easier to develop a mental model for this page's structure.

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -86,7 +86,7 @@ The **Initiatives Overview** page gives an overview of all the initiatives in th
 
 In the search box on the top, you can search for an initiative by the initiative name.
 
-On the upper-right corner, you can use the drop-down list to [change the view](#change-view). Clicking **Filters** enables you to filter initiatives and the results persist even after you change to a new view. You can [export and import Initiatives](/developerportal/portfolio-management/export-import-initiatives/) by clicking the {{% icon name="office-sheet" %}} icon, and [create a new initiative](#create-new-initiative) by clicking the **Create Initiative** button.
```
dbreseman: I see the UI doesn't have a tooltip for this icon yet, but I'd suggest creating your own name for use in the docs (and advocate for the tooltip in the UI if there's an opportunity). Maybe this could be "clicking **Import or Export** ({{% icon name="office-sheet" %}}), and [create..."

(Also for the other nameless icons, if possible)

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -96,7 +96,7 @@ To create a new initiative and add details to it, do as follows:
```
dbreseman, line 96: I recommend adding a little more contextual information here (even just specifying that all these steps are described below, and the user can read through step by step or click a link to go straight to that step).

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -9,7 +9,7 @@ tags: ["Portfolio Management"]
 
 ## 1 Introduction
 
-The [Portfolio Management](https://portfolio.mendix.com) tool is available to all Mendix customers. It enables staying informed about initiatives and managing them in the different development stages. It provides portfolio managers, business stakeholders, and developers with everything they need to collaborate effectively in one place and bring new initiatives to the Mendix Platform. To start the Portfolio Management app, go to the Developer Portal, open the [Switch to menu](/developerportal/), and select **Portfolio**.
+The [Portfolio Management](https://portfolio.mendix.com) tool is available to all Mendix customers. It enables staying informed about initiatives and managing them in the different development stages. It provides portfolio managers, business stakeholders, and developers with everything they need to collaborate effectively in one place and bring new initiatives to the Mendix Platform. To start the Portfolio Management app, go to the Developer Portal, open the [Global Navigation menu](/developerportal/) ({{% icon name="layout-rounded-1-filled" %}}), and select **Portfolio**.
```
dbreseman, line 13: My impression is that you aren't expecting most users to read through this doc from top to bottom, but rather to jump to relevant sections to figure out how to do specific things (right?). With that in mind, I'd suggest bulking up this introductory section to help users orient themselves to what Portfolio Management is and how it's organized. Could you go into more specifics here about how people use Portfolio Management, and what sections they should skip to for the most important use cases?

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -86,7 +86,7 @@ The **Initiatives Overview** page gives an overview of all the initiatives in th
 
 In the search box on the top, you can search for an initiative by the initiative name.
 
-On the upper-right corner, you can use the drop-down list to [change the view](#change-view). Clicking **Filters** enables you to filter initiatives and the results persist even after you change to a new view. You can [export and import Initiatives](/developerportal/portfolio-management/export-import-initiatives/) by clicking the {{% icon name="office-sheet" %}} icon, and [create a new initiative](#create-new-initiative) by clicking the **Create Initiative** button.
+On the upper-right corner, you can use the drop-down list to [change the view](#change-view). Clicking **Filters** enables you to filter initiatives, and the results persist even after you change to a new view. You can [export and import initiatives](/developerportal/portfolio-management/export-import-initiatives/) by clicking the {{% icon name="office-sheet" %}} icon, and [create a new initiative](#create-new-initiative) by clicking the **Create Initiative** button.
 
 ### 3.1 Creating a New Initiative {#create-new-initiative}
```
dbreseman, line 91: I really like all these clear headings—it's helping a lot with keeping track of the structure and navigation.

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -9,7 +9,7 @@ tags: ["Portfolio Management"]
 
 ## 1 Introduction
```
dbreseman, line 10: Overall, I think this is clear, well structured, and thoughtfully put together. It's consistent and detailed, and it seems like a very thorough reference guide for the Portfolio Management page. I especially like all the clear headings—it helps a lot with developing a mental model for how the content is organized.

It does have a slightly overwhelming amount of information. I think you could consider breaking out Initiatives Overview, Archive, Access Management, and Portfolio Settings into individual child pages, and keep this page as an orientation and concept introduction. (I mostly think that would be useful just to reduce the length of this page).

content/en/docs/developerportal/portfolio-management/_index.md
```diff
@@ -35,7 +35,7 @@ On the **Portfolio Landscape Overview** page, each portfolio is presented in a c
 
 Currently, the privacy settings of a portfolio can be private, restricted, or open. The table below shows their differences:
 
-| Portfolio type | Visible in the **Company Portfolios** section?               | How to access the portfolio?                                 |
+| Portfolio type | Visible in the **Company Portfolios** section?               | How to access the portfolio                                 |
```
dbreseman: I think this might be clearer as a bulleted lit or paragraphs instead of a table. I found it a bit hard to compare between the rows in the Access column. Also, I'd suggest removing some of the duplicate links so that within a section, you're only linking once to each heading. This might become more important once the link underlines get implemented, since the underlines make the links a bit more aggressive.

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Thank you. I think this is a good idea. I will propose this idea to the Product Manager.

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Totally agree with you about the links. I must have missed this by an oversight. I will remove them.

I would like to emphasize the differences among these three types of portfolios. That's why I choose to use a table instead of a list or paragraphs. Maybe I can change the text of the access part in the table to make it clearer? We can discuss more on Monday! : )

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Good one. Thanks!

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: I think it's a nice idea. Also nice to advocate for the tooltip in the UI if there is a chance!
Just a small comment on the style for the icon name here—we probably shouldn't use the bold style, as the name we pick here are not really in the UI.

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Good idea! Thanks!

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: I don't really understand this comment. Could you tell me more about this on Monday?

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Sounds good. I will reach out to the design team for some advice!

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: Yes, those SVG files had been used before we introduced this nice way of working with icons. Using icons now would definitely be a cleaner choice.

content/en/docs/developerportal/portfolio-management/_index.md
Luyao-Zhang-1: We can discuss this more on Monday.

content/en/docs/developerportal/portfolio-management/_index.md
dbreseman: I see where you're coming from. Personally, I'd prefer to bold it anyway, because that seems more consistent to me and carries an implication that soon the tooltip will be added, but I also see how your approach makes sense. Up to you! (Or maybe worth discussing in the larger group? Not sure)

PR #7446, opened 2024-03-13
Update Combobox docs with new source type additions

rahmanunver:
This documentation update is to mention the new Data source type support additions for the Combobox widget.

Categorizing the already existing types under "Context"
And adding "Database" and "Static" types as new source types.


gjulivan:
feature already released and document are good to go

ConnorLand (approved):

PR #7448, opened 2024-03-14
Update build-hybrid-locally.md

ThijsSeppen:
with the latest buildpack, it's `env target` instead of `env.target` (see also: the code above without the dot)


ConnorLand:
I didn't find any other instances of env.target, so this is good to go. Thanks @ThijsSeppen.

ConnorLand (approved):

MarkvanMents:
@ThijsSeppen I assume all customers will be using the latest buildpack so we don't need to document the previous value?

ThijsSeppen:
Yes this should be fine @MarkvanMents !

PR #7450, opened 2024-03-14
Link styling

dbreseman:


assets/scss/_styles_project.scss
```diff
@@ -226,6 +226,28 @@
     }
 }
 
+// Style visited links but exclude buttons and left sidebar
+a:not(.btn):not(.td-sidebar-link):visited {
+    color: $purple;
+}
+
+// Add icon after external links (but exclude buttons and the right sidebar icons)
+a[href*=".com"]:not([href*="github.com/mendix/"]):not(.btn)::after,
```
MarkvanMents: This doesn't work. Not all external links will be to *.com.
For example [BBC News](https://news.bbc.co.uk)

image

This is definitely an issue as a lot of APIs use `.io` and a lot of non-profits (open source) use `.org`.

I think a better way of excluding the sidebars would be to only include anchors which are in the `<main>` section of the html.

I've added some tests to the bottom of our test file to make it easier to test: 2a6cc8c8dcbb1e67a2607849c04af32c002115d3

MarkvanMents (dismissed):
This is looking great but there are a couple of issues still with external links.
Have a look yourself, but we can discuss at our meeting this morning as well.

assets/scss/_styles_project.scss
MarkvanMents: 2a6cc8c8dcbb1e67a2607849c04af32c002115d3

MarkvanMents: and external links will always have `http://` or `https://`

We could even assume that we never have an external link to an `http://` site as the link should always be to a secure site.

PR #7453, opened 2024-03-15
Avoid using the acronym CDN

MXClyde:
The use of the term CDN might lead to the misleading understanding that the static content is somehow served from a globally distributed service for serving out static content (the contemporary use of the word CDN), while this does not have to be the case (and is not the default in our public cloud). Good to remove the acronym from the diagram as well, if possible.


MXClyde:
LGTM!

MarkvanMents (approved):
Thanks for the PR, Clyde.
Approving and merging.

PR #7470, opened 2024-03-19
Update windows-service.md with 4.7.4 release notes

yatindrakulkarni-mendix:


yatindrakulkarni-mendix:
We are duplicating the information in this documentation and marketplace module. So removing this PR.

PR #7492, opened 2024-03-22
"Updated Release Notes, How to and documentation for external database connector for 10.9 release.

SurabhiSaraf:


content/en/docs/howto/integration/use-the-external-database-connector.md
```diff
@@ -102,7 +102,21 @@ SQL Query:
 
 {{% alert color="info" %}} For all DML Queries, changes made to database in the Mendix Design phase are automatically rolled back. {{% /alert %}}
 
-### 6 Use the Query External Database Activity
+## 5 Call Stored Procedure
```
quinntracy, line 105: A couple questions for this section:

1. Can you clarify "If the stored procedure returns result-set"? Does this mean if it returns "a result set"? 
2. Does the user click Use Response and immediately follow with clicking Save Query & Create Entity? If so, we can change it to ""click **Use Response** > **Save Query & Create Entity** to save the query and the newly-created entity in the domain model."
3. Does the "Number of affected rows" response display before or after clicking "Save query"? Where and when does this response occur? Could it be moved to a be a sub-bullet of step 2? 
4. Number 4 seems to be a note rather than a step in the process. Can this be changed to an info box?

SurabhiSaraf: Here are the answers to the questions:
1. Yes, this mean if it returns "a result set".
2. Yes, you are right, we can make change as suggested by you.
3. Number of affected rows is displayed before Save Query,
Users click on 'Run Query' -> Response is shown (number of affected rows) -> if user is okay with response should Save Query.
4. Yes, it is a note rather than a step.

MariaShaposhnikova: @quinntracy , I've implemented your comments except for the second one. Please review the changes when you are back.

PR #7495, opened 2024-03-22
Change erroneous text about task queue table lock

basdebakker:


MarkvanMents (approved):
Thanks Bas for the PR.

PR #7508, opened 2024-03-27
Adjusted for public beta Git cleanup

tknippenberg:
- Should be merged no later than 2 April 2024
- ✅Download link to Git cleanup public beta still needs to be inserted


mrseanryan (commented):
[RESOLVED - thank you]

>> For developers on your team this means they have to delete their local folders.

Deleting local folders is risky - the dev may have changes they did not push OR branches they do want to keep etc.

Recommend instead:

- use the  `sp-reset` tool

OR

- *move* the folder of the app, for example by renaming it to x`.old`

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
```diff
@@ -33,26 +33,32 @@ Other places where you might encounter performance issues or timeouts are the fo
 
 ### 4.1 Preventing a Large Repository Size
 
-The *.mpr* storage format will be changed in H1 2024 to reduce the rapid repository growth. This will reduce the repository size for new apps, but will not reduce the repository size for existing apps. Upgrading to a new storage format will happen automatically and will not affect functionality.
+The *.mpr* storage format will be changed to reduce the rapid repository growth. Switching to the new storage format will be done under the hood and does not result in functional changes.
 
-### 4.2 Mitigating Large Repository Size
+For new apps we aim to introduce the new format in Q2 2024. Existing apps will be automatically converted in a later version, targeted for H2 2024.
```
robertvmx: Is it worth communicating a date on this yet, as there are many reasons why this date might change anyway?

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
```diff
@@ -80,7 +86,7 @@ When planning the cleanup, note the following:
 
 * Changes made by other users after you made your local copy are lost after pushing the cleaned repository to the Git server.
 * All users need to get fresh clones from the server after the cleanup is completed. This means all changes that have not been committed and pushed to the server before you download the repository you will use for cleaning, will be lost.
-* The process can take up several hours, so we recommend to align with your team before you start. First do a test run to get familiar with the tool before planning the actual cleanup.
+* The process can take up to several hours, so we recommend to align with your team before you start. First do a test run to get familiar with the tool before planning the actual cleanup.
```
robertvmx: The process will not take up several hours anymore since we will only supply the extreme mode which will take less then a hour.

static/attachments/refguide/version-control/troubleshoot-version-control-issues/git_fixer_mode.png
robertvmx, line 1: It is unclear what a merged commit means, I think it means work of these commits will not be discarded. Maybe we could clarify this a bit.

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
tknippenberg: Although you're right it can still change, I also don't see how it can really cause harm.

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
tknippenberg: Changed in new commit. "Up to an hour".

static/attachments/refguide/version-control/troubleshoot-version-control-issues/git_fixer_mode.png
tknippenberg: If you have suggestions how to make it clearer I'm happy to change it. But couldn't really figure it out myself without using a lot more words.
I hope the image itself is pretty clear.

tknippenberg: Changed

tknippenberg:
@MariaShaposhnikova , ready for review. Download link still needs to be inserted, should have that tomorrow or right after the weekend.
[Edit] link is inserted, ready for review and merging.

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
```diff
@@ -31,32 +31,48 @@ Other places where you might encounter performance issues or timeouts are the fo
...
 
-### 4.2 Mitigating Large Repository Size
+Mendix aims to introduce the new format for new apps in Q2 2024. Existing apps will be automatically converted in a later version, targeted for H2 2024.
 
-In cases where a repository has already grown to a significant size, Mendix recommends you to consider shrinking the repository by removing parts of the history, in case you are encountering performance issues. 
+### 4.2 Working with a Large Repository Size
 
-Mendix has developed a git-fixer – a cleanup tool to assist you in shrinking your repository. For more information, see the [Cleanup Tool](#cleanup-tool) section below.
+When cloning an app, the default behavior of Git is to download the full history. As Mendix uses different folders on disk for different branches, downloading full history is done for each branch. To mitigate that, Mendix has introduced local cloning for subsequent branch downloads. When cloning a new branch, data from a local branch you already have is used to reduce data that needs to be downloaded.
```
MariaShaposhnikova: "Mendix has introduced local cloning for subsequent branch downloads" - @tknippenberg , do you mean the gix-fixer tool here?

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
```diff
@@ -31,32 +31,48 @@ Other places where you might encounter performance issues or timeouts are the fo
...
+### 4.2 Working with a Large Repository Size
 
-Mendix has developed a git-fixer – a cleanup tool to assist you in shrinking your repository. For more information, see the [Cleanup Tool](#cleanup-tool) section below.
+When cloning an app, the default behavior of Git is to download the full history. As Mendix uses different folders on disk for different branches, downloading full history is done for each branch. To mitigate that, Mendix has introduced local cloning for subsequent branch downloads. When cloning a new branch, data from a local branch you already have is used to reduce data that needs to be downloaded. 
+
+Another improvement that is currently in development is partial cloning, which will instruct Git to only retrieve a part of the history from the server when cloning. In case you do need to access older history, for example, when merging an older branch, the data is retrieved from the server only when needed. Mendix expects to deliver this improvement in the summer of 2024 and intends to backport this to the 10.12 MTS release.
```
MariaShaposhnikova, line 44: We usually do not announce anything that is still in development in our documentation, it feels like something we can communicate in an email or somewhere else. What is the reason to include it to docs? @tknippenberg

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
tknippenberg: No, this is actually about "local cloning" :) 

A normal clone (for the first branch) would:
- Contact the server and download everything

A local clone will more-or-less do this:
- Copy-paste the folder of an existing branch you already have on disk (ie the `main` branch, which has data up to February)
- Download the new data up to [now], so everything between February and [now]
- Switch the branch

That means that there's way less network traffic, which is a lot slower than copy-pasting a folder on-disk. Therefore subsequent branch clones are much faster.

--

Perhaps we shouldn't say "Mendix has introduced local cloning" but "Mendix offers local cloning" or something along those lines? It's not really a new feature anymore.

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
tknippenberg: We're not mentioning it in the email as that's not a timeline we can change afterwards. Software development is unpredictable and I'm afraid text in an email will be perceived more as a commitment than mentioning it here.

The reason why I'd want to mention it on this page, is that this it the main "Large repo troubleshoot page" that Support also refers to when customers come in with issues. This way there is a public statement that we're aiming for the 10.12 MTS release to have things more-or-less solved.

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
MariaShaposhnikova: @tknippenberg , thanks for the explanation! I wrote ""Mendix suggests using".

content/en/docs/refguide/version-control/version-control-troubleshooting/repository-size.md
tknippenberg: I see the text is now "suggests using local cloning", but basically this is an under-the-hood optimization that the user doesn't have control over. It's only an explanation why subsequent branch downloads are faster than the initial download.
From this the user could also deduct that it's not ideal to "throw away all local branches" as that prevents local cloning for future branch downloads. But this is not very important (as it's an edge case) and will be made obsolete in months, so I don't think this needs to be made explicit.

MariaShaposhnikova: I see, I interpreted it differently indeed. If I say "Mendix uses" to indicate that this is not something a user has to choose, will that be correct?

tknippenberg: Yes!

PR #7519, opened 2024-03-28
Storing secrets in Mendix 10.9

MarkvanMents:
In Mendix 10.9 there are alternative places to store secrets.


MarkvanMents:
Approved over Slack.

PR #7537, opened 2024-04-02
Configure sidebar badges

dbreseman:


MarkvanMents (dismissed):
Looks great,
I wonder if we should find somewhere to put the badges in https://docs.mendix.com/releasenotes/studio-pro/lts-mts/ and https://docs.mendix.com/releasenotes/beta-features/?

Can you also ask Maria about her feelings - or perhaps demo it at the meeting tomorrow to prompt discussion? There may be other beta features which could benefit from the beta label.
And there may be other

dbreseman:
Sure, I'll add it to the agenda.

PR #7575, opened 2024-04-09
Update TabContainer widget guide

PaulHryn:
Should target 10.11 release of SP

### DIFF
--- a/content/en/docs/refguide/modeling/pages/structure-widgets/tab-container.md
+++ b/content/en/docs/refguide/modeling/pages/structure-widgets/tab-container.md
@@ -72,18 +72,17 @@ Default: *True*
 This property is not supported on native mobile pages.
 {{% /alert %}}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/pages/structure-widgets/tab-container.md
```diff
@@ -72,7 +72,20 @@ Default: *True*
 This property is not supported on native mobile pages.
 {{% /alert %}}
 
-## 6 Read More
+## 6 Usage Details
+
+### 6.1 Common rules {#usage-common}
```
ConnorLand: Are there any reasons why customers might often put a Tab Container inside another container or wrapper widget? If so, try and address those use-cases directly with our best advice. 

If not, I will ensure this section sends the message "never put Tab Container inside another container".

ConnorLand (approved):

PR #7581, opened 2024-04-11
Update community-best-practices-for-app-performance.md

recktenwaldfabian:
Remove microflow recommendation on 'retrieve by association'. The statement 'Use the retrieve over association if possible. This uses caching, it is more readable, and it uses an index. If business logic requires the database value (because the value over association might be changed), then of course a database retrieve is needed.' is wrong or misleading. There is no such caching mechanism documented anywhere and also the statement about indexes is lacking any explanation,

### DIFF
--- a/content/en/docs/refguide/modeling/community-best-practices-for-app-performance.md
+++ b/content/en/docs/refguide/modeling/community-best-practices-for-app-performance.md
@@ -81,7 +81,6 @@ Indexes is a topic with a long history of best practices from the database world
     * If loops contain decisions, consider if the decision logic can be a query before the loop to minimize iterations.
 * Prevent unnecessary retrieves if objects or lists can be passed as parameters.
 * Know and use the retrieve + aggregate optimization. If you retrieve a list and count the list, Mendix will optimize this to one query. If you need the list later in the microflow, after some decisions, it is wise to retrieve the list again so that you only retrieve the data when needed. This also works in batches where you can retrieve the total count optimized and retrieve chunks in a separate query.
-* Use the retrieve over association if possible. This uses caching, it is more readable, and it uses an index. If business logic requires the database value (because the value over association might be changed), then of course a database retrieve is needed.
 * Commit as late as possible. A commit locks that record (or list of records). This means that any other user/logic that wants to commit the same object has to wait until the first transaction is finished.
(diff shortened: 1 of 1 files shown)


MarkvanMents:
Hi Fabian,
Thanks for your PR on the Mendix documentation site.
I have spoken to the developers who feel that the original advice is still good, but agree that the wording was unclear and not backed up.
I have created an alternative PR, #7598 which describes the reasons for using retrieve over association in preference to retrieve from database, but also explains when it is better not to use it. I hope this new PR clarifies why retrieve over association is a best practice when it can be used.
Thanks again for helping to improve the Mendix documentation.
Yours
Mark van Ments

recktenwaldfabian:
Hi Mark,
you added a reference to client side caching while refering to microflows running server side. That's two different things.
The important difference which is now noted (getting changes made in the client) is a good addition.
However, the client ONLY sends changes to the server - so any other object information is anyway retrieved from DB. Also doing two consecutive retrieve by association will also do 2 database retrieves. Why would these database retrieves be any faster?

Who are 'the developers' you spoke with?

regards Fabian

recktenwaldfabian:
I would still like to get forward with that topic. Are these comments actually still read?

MarkvanMents:
Yes, I am working on a new PR and agreeing with the developers how to respond to your comments.
Thanks for your patience.

MarkvanMents:
Hi Fabian,
I have now updated the documentation to clarify that the main advantage in retrieve over association is in getting the current state of the object. This is in PR #7661 .
The developers have, however, confirmed to me that there are occasions where retrieve from association will not involve a call to the database. This happens where the association is from the owning object and the associated object is also retained by the Mendix Client and shared with the Server. However, Mendix does query the database if not all the information is held by the Client. In particular, if you are retrieving from the object which is not the owner (e.g. retrieving order-lines associated with an order) then a database call is needed.
Since, as you point out, Mendix will often still need to access the database, I have removed the speed advantage from the recommendation.
I will look into updating the [Communication Patterns](https://docs.mendix.com/refguide/communication-patterns/) documentation in the future so that it is clearer how Mendix does this.
Thank you for bringing the misleading impression to my attention and helping to clarify the documentation around this.
Yours
Mark van Ments

PR #7583, opened 2024-04-12
Update marketplace-curation.md

Luyao-Zhang-1:


content/en/docs/control-center/marketplace-curation.md
serdarceranmendix, line 1: looks good from my end.

PR #7590, opened 2024-04-12
Update metrics.md

Infra-Red:
Remove references to the Mendix plans as they keep constantly changing and it's hard to keep them up-to-date.


dbreseman (approved):
Looks good to me!

dbreseman:
I just added a link to the resource packs table on the main Mendix Cloud page, so that we only need to maintain the plan details in one place in the docs

PR #7598, opened 2024-04-15
Update microflow best practice for retrieve over association

MarkvanMents:


MarkvanMents:
PR approved by developers over Slack

recktenwaldfabian:
Hi Mark,
you added a reference to client side caching while refering to microflows running server side. That's two different things.
The important difference which is now noted (getting changes made in the client) is a good addition.
However, the client ONLY sends changes to the server - so any other object information is anyway retrieved from DB. Also doing two consecutive retrieve by association will also do 2 database retrieves. Why would these database retrieves be any faster?

Who are 'the developers' you spoke with?

regards Fabian

PR #7601, opened 2024-04-15
Document the Optimize bundle settings 10.10

EnasAbdelrazek:


content/en/docs/refguide/modeling/menus/edit-menu/preferences-dialog.md
```diff
@@ -67,6 +67,10 @@ The JDK is necessary to run Mendix applications, because the Mendix Runtime is w
 
 Enable this setting to increase the speed at which a running application is updated after changes have been made in Studio Pro. When only pages, layouts or snippets have been changed, an entire restart of the application is skipped to decrease deployment time. Also, no Java compilation is performed when no relevant files have changed.
 
+#### 3.1.3 Optimize bundle
+
+This setting is visible and effective when the React client is enabled. By default, It's already enabled to reduce the size of the generated app bundle and speeds up the bundling process by diabling source maps generation. Source maps are files that map code from its transformed state back to its original form, facilitating debugging during development. If you want to generate source maps for debugging purposes, you can disable this setting.
```
WimJongeneel: diabling -> disabling.

I think we should mention that the bundling speed is relevant for local deployment speed, I don't think that is directly clear for all users. 

"Source maps are files that map code from its transformed state back to its original form" might also be a bit too in-dept from a low-code perspective, I would just mention that source maps are used for debugging the bundled code.

WimJongeneel (changes requested):

WimJongeneel (approved):

ConnorLand (approved):

PR #7603, opened 2024-04-15
Remove docs.mendix.com from internal links

dbreseman:


dbreseman:
The external link icon is inserted after the link unless a relative URL is used.

PR #7616, opened 2024-04-18
Implement default borders

dbreseman:


MarkvanMents (dismissed):
Hi Dana,
Looking good so far.
Two things:
1. You haven't removed the `class="image-border"` from the existing figures which currently need a border. This doesn't have a visual effect, but makes the markdown messy and might have strange results in future.
2. Your image examples in "Shortcodes, Markdown, and HTML Examples" link to real images which might, in future, but changed or deleted. I've made copies in commit 5372cf5a9a2d58b2ae79a6edd99042fb1dc31761 to ensure that our examples don't get corrupted in future - and I've added an image which makes the effect of the border easier to test/see.

PR #7618, opened 2024-04-18
Add PAT authentication

Luyao-Zhang-1:


content/en/docs/apidocs-mxsdk/apidocs/authentication.md
```diff
@@ -2,25 +2,38 @@
 title: "Authentication"
 url: /apidocs-mxsdk/apidocs/authentication/
 category: "API Documentation"
-description: "Describes how to retrieve and use authentication keys for Mendix APIs."
+description: "Describes the two methods of Mendix API authentication: using API keys and using personal access tokens."
```
JaapF, line 5: Maybe good to add a bit more explkanation.
We're also working on Eval Guide improvement. For Eval Guide I recently drafted:
"Mendix provides you with various APIs to access the public Mendix platform. These APIs are secured by either [API-keys or Personal Access Tokens (PAT)](https://docs.mendix.com/community-tools/mendix-profile/user-settings/#profile-api-keys). Both mechanisms allow clients such as a CI/CD pipeline to consume the platform APIs on behalf of the platform user who created the token; the same privilege restrictions apply as for the user who created the API-key or PAT. Benefit of PAT compared to API-keys is that the platform user can restrict the scope of delegated access to specific APIs, by selecting so-called ‘scopes’ during creation of the PAT.  API-keys and PATs can no longer be used if the user who created them has been deactivated."

PR #7654, opened 2024-04-25
Document the different sorting strategies in microflows and nanoflows: 10.10

EnasAbdelrazek:
In 10.10


MariaShaposhnikova (approved):

MariaShaposhnikova:
@Yiyun333 , do we also need to update differences between microflows and nanoflows?

Yiyun333:
> @Yiyun333 , do we also need to update differences between microflows and nanoflows?

@MariaShaposhnikova Thanks for handling this PR! You mean adding the difference to the [Differences from Microflows](https://docs.mendix.com/refguide/nanoflows/#3-differences-from-microflows) section? The difference added in the PR seems to only relate to a list sorting operation though. I was wondering whether this difference applies to a broader context. If yes, we might be able to generalize the statement a bit and then add it to the [Differences from Microflows](https://docs.mendix.com/refguide/nanoflows/#3-differences-from-microflows) section. 

@EnasAbdelrazek What do you think? ;)

PR #7659, opened 2024-04-26
Update GitHub Actions

dbreseman:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### DIFF
--- a/.github/workflows/lint-action.yml
+++ b/.github/workflows/lint-action.yml
@@ -53,6 +53,7 @@ jobs:
           body: |
             ${{ env.VER }}
           branch: lint-docs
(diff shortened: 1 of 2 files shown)


MarkvanMents (approved):
This looks good - hopefully changing the committer will work, I didn't know that was an option (perhaps it is new?).

If this can be merged before c.o.b. 30 April then they should be run automatically on 1 May. You can also run them manually if you want, but I think the changes are small enough (and the impact of them failing so small) that it is up to you whether you do that or not.

Thanks for picking this up.

Mark

PR #7698, opened 2024-05-03
Include three new operations in Pgvector Module

tswarmerdam-mx:


MarkvanMents:
Hi Tom (@tswarmerdam-mx )
Is this for immediate publication, or is it dependent on a release?

tswarmerdam-mx:
Hi @MarkvanMents , we have released to marketplace already earlier this week, so the docs can be updated anytime :) 
CC @Luyao-Zhang-1

PR #7704, opened 2024-05-07
OpenAI: Function Calling

liamsommer-mx:
The OpenAI connector now supports "function calling" that can be used within the ChatCompletions operations. The documentation was updated to guide developers and reflect the technical changes. A new subpage for function calling was added to explain the concept.


content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+
+### 1.1 High-level flow {#high-level}
+
+The basic process steps of function calling contain at least the following:
+1. Invoke the chat completions API with a user prompt and a list of available functions with expected input parameters.
+2. Based on the user prompt and the available functions, the model can intelligently decide whether to call one or more functions. If so, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as desceribed in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
```
rik-bos: "whether to call one or more functions" --> "that it's needed to call one or more functions"

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+---
+
+## 1 Introduction {#introduction}
+
+Function calling enables LLMs (Large Language Models) to connect with external tools to gather information, execute actions, convert natural language in structure data and much more. Thereby, function calling enables the model to intelligently decide when to call one or more predefined functions to gather additional information to include in the assistant's response. OpenAI does not call the function; instead, the model returns a tool call JSON structure that is used to build the input of the function(s) so that they can be executed as part of the chat completions operation.
+
+### 1.1 High-level flow {#high-level}
+
+The basic process steps of function calling contain at least the following:
+1. Invoke the chat completions API with a user prompt and a list of available functions with expected input parameters.
```
rik-bos: functions --> functions (microflows)

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+2. Based on the user prompt and the available functions, the model can intelligently decide whether to call one or more functions. If so, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as desceribed in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
+3. Parse the string into JSON and call the function with its input parameters.
```
rik-bos: function --> function (microflow)

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -349,7 +392,41 @@ This is a wrapper for a list of messages comprising the conversation so far.
...
+
+This is a specialization of the [AbstractTool](#abstracttool) entity and represents a tool the model may call.
+
+##### 4.1.3.9 `FunctionRequest` {#functionrequest}
+
+This is a specialization of the [AbstractFunction](#abstractfunction) entity and represents a function the model may call.
+
+| Attribute          | Description                                                  |
+| ------------------ | ------------------------------------------------------------ |
+| `Description`      | A description of what the function does, used by the model to choose when and how to call the function. This attribute is optional. |
```
MarkvanMents: If the model is using this to choose when to call the function how can it be optional?

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+### 1.3 Use cases {#use-cases}
+
+Function calling can be used for a variety of usecases including
+
+* Create assisants that can answer questions about data from your Mendix database or a knowledge base
+    * e.g., getTicketById(String identifier) or findSimilarTickets(String description)
+* Create assistants that can get information from external APIs
+    * e.g., getCurrentWeather(String location)
+* Extract structured data from natural language
+    * e.g., extractBirthday(String birthday), or getSqlQuery(String query)
```
rik-bos: Let's not including SQL query as an example since that may cause OQL like expectations. Let's not do that for now.

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+* Execute action like creating objects
+    * e.g., createTicket(String subject)
```
rik-bos: Let's include a recommendation that if the function manipulates things (e.g. creating a ticket), user validation to confirm the action is recommended to prevent unexpected behavior

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+
+The basic process steps of function calling contain at least the following:
+1. Invoke the chat completions API with a user prompt and a list of available functions with expected input parameters.
+2. Based on the user prompt and the available functions, the model can intelligently decide whether to call one or more functions. If so, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as desceribed in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
+3. Parse the string into JSON and call the function with its input parameters.
+4. Append the existing list of messages with the function response as a new tool message. Then, invoke the chat completions API again and let the model answer your initial prompt with the new information provided by the function.
+
+For more general information on this topic, see [Function Calling](https://platform.openai.com/docs/guides/function-calling).
+
+### 1.2 Function calling with the OpenAI Connector {#openai-connecor}
```
rik-bos: Can we include that the microflows run in the original user session, and have access to `$currentUser` for applying user specific data / security constraints?

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -637,13 +741,20 @@ For [specific models](https://platform.openai.com/docs/guides/text-generation/js
 | `Configuration`  | [Configuration](#configuration-entity)                | mandatory                     | This is an object that contains endpoint and API key.        |
 | `Model`          | String                                                | only mandatory for **OpenAI** | This is the ID of the model to use; not considered for **Azure OpenAI** configurations. |
 | `ResponseFormat` | [ENUM_ResponseFormat_Chat](#enum-responseformat-chat) | optional                      | This can be used to specify the format that the model must output. |
+| `Temperature`    | Decimal                                               | optional                      | This can be used to control the randomness of the output. The value should be a decimal between 0.0 and 2.0. The default value is 1.0. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic. Note: high values for temperature (>1.7) may give unexpected results and even internal server errors. |
```
MarkvanMents: Since the default value is 1.0 I think it would be better to have a higher value than one as your example of a higher value. Otherwise you are saying that the default is more random than setting it explicitly to be more random.

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -656,13 +767,22 @@ Use the microflow `ChatCompletions_Execute_WithHistory` to execute a chat comple
...
 **Return value**
 
 | Name                    | Type   | Description                                                  |
 | ----------------------- | ------ | ------------------------------------------------------------ |
 | `AssistantResponseText` | String | Assistant message that was generated by the model as a response to a user message. |
 
+The following microflows may be used in order to construct and handle the required inputs: 
+
+* `ChatCompletionsSession_Create` is used to create the session wrapper that must be passed as input parameter. 
+* `ChatCompletionsSession_AddMessage` is used to attach the historical messages to the `ChatCompletionsSession`. The content of such a message corresponds to a system, assistant, or user prompt. In the case of multiple historical messages the order is relevant.
```
MarkvanMents: In what way relevant? Which one takes priority, first or last?

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,65 @@
...
+
+Function calling is supported for all chat completions operations by adding the optional input parameter [FunctionCollection](#functioncollection). Functions in Mendix are essentially microflows that can be registered within the request to the LLM​. The OpenAI connector takes care of handling the tool call response as well as executing the function microflow(s) until the API returns the final assistant's response. Currently, function microflows are limited to one input parameter of type string and must return a string.
```
MarkvanMents: This seems to contradict what we say before. If the connector calls the function with the parameters how can we check that the model hasn't hallucinated parameters before the function is called?

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: It's optional in the OpenAI API, so we made it optional for our connector as well. If you give the function a very descriptive name, it may work just fine without the description
https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

MarkvanMents: Thanks for explaining. Should we put something into the docs to explain this?

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,75 @@
...
+The basic steps of function calling are as follows:
+
+1. Invoke the chat completions API with a user prompt and a list of available functions (microflows) with expected input parameters.
+
+    * The model decides whether to call one or more functions based on the user prompt and the available functions.
+
+        * If it calls a function, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as described in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
+        * Parse the string into JSON and call the function with its input parameters. The microflow runs in the original user's session and `$currentUser` can for example be used to apply security constraints.
+        * Append the existing list of messages with the function response as a new tool message.
+
```
MarkvanMents: I don't find it clear which bits are done by the OpenAI connector and what the customer needs to do. I've tried to clarify it, based on my understanding of the rest of the document.

Here, and in the diagrams below, it seems that the OpenAI connector does everything and the customer just has to make a single call.

However, you also say that "the model may hallucinate parameters, so they should be validated before the function is actually called". But if the connector is calling the function directly you can't validate the parameters before you call the function - you can only validate the parameters inside the function after it has been called.

MarkvanMents (changes requested):
Hi @liamsommer-mx @tswarmerdam-mx @rik-bos and @tinaengel 
I have proofread this (see https://github.com/mendix/docs/pull/7704/files/42ee9160a7271837badebcd25909d4603067b1aa..d329670943e699c9e1e7c052711a6ae0a68c197c) but I have a few questions which I've asked in the comments. Can you look into these and get back as I feel that there a couple of things which need to be clarified.

Thanks
Mark

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: it's a chronological order, from oldest message to most recent message being the last one

liamsommer-mx: It's not really about priority, but rather the chronological order. So your last messages are the most current, while the first ones are "older". The model responds to the last message, but takes into account the older ones (like a normal conversation).

content/en/docs/appstore/modules/openai/function-calling.md
tinaengel-mx: you can build validation inside the fucntion microflow. Even though the function microflow is called by the OpenAI connector, the function microflow is implemented by the developer and could some validation as the first action

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: I would phrase it in a way that we recommend adding a description even though it is optional

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -10,46 +10,56 @@ tags: ["OpenAI", "generative AI", "AI", "connector", "marketplace", "chatgpt", "
...
-3. Parse the string into JSON and call the function with its input parameters.
-4. Append the existing list of messages with the function response as a new tool message. Then, invoke the chat completions API again and let the model answer your initial prompt with the new information provided by the function.
+The basic steps of function calling are as follows:
+
+1. Invoke the chat completions API with a user prompt and a list of available functions (microflows) with expected input parameters.
+
+    * The model decides whether to call one or more functions based on the user prompt and the available functions.
+
+        * If it calls a function, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as described in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
```
MarkvanMents: What is the response if it doesn't call a function?
What is the format of the JSON object? I can't see an example anywhere and so I'm not sure how I know which function was called, what the parameters were, and what the function returned before the AI parses it and sends it back.

The examples below do not return the JSON to the customer, they just return a natural language response ("The status of ticket 123 is in progress") so I'm not sure why the customer needs to know about the JSON or what they can do to get it.

content/en/docs/appstore/modules/openai/function-calling.md
tinaengel-mx: @MarkvanMents can we go back to the structure with the 4 steps, which align with the process mentioned in the official openai docs? https://platform.openai.com/docs/guides/function-calling

content/en/docs/appstore/modules/openai/function-calling.md
tinaengel-mx: the format of the JSON object was provided as part of the request so it depends on the input of your function microflow.
This is just some background information for developers who want to know how function calling works under the hood. Developers might need to handle the json object and the function call themselves if they are using the advanced operations, which we generally don't recommend. If they are using with/without history operations, we handle everything inside the connector

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: Agree, can we just remove the "like 0.8"?

content/en/docs/appstore/modules/openai/function-calling.md
tinaengel-mx: It can be validated inside the function microflow which needs to be implemented by the developer

content/en/docs/appstore/modules/openai/function-calling.md
MarkvanMents: I find the four steps confusing for the Mendix documentation as the developer is only interested in step 1 and step 4 while steps 2 and 3 are hidden by the wrapper. 
So my idea is to have four steps, but make it clear that steps 2 and 3 are done automatically by the connector.

I'm happy to go for another structure, but I think it should be clear what is done by the developer and what is done by the connector.

As it was it just described what the openai API does, which the developer can get from the openai docs.

content/en/docs/appstore/modules/openai/function-calling.md
MarkvanMents: I'm finding this confusing. Are we describing here how function calls work, or just how the OpenAI connector does it for you?

If developers are doing the calls themselves, then we should be able to point them at the OpenAI documentation. I think we should focus here on how the non-advanced calls work, or at least make it clear what the differences are. Here we are mixing the two things up which, as someone who doesn't know the internal workings of the connector, is confusing.

Additionally, we don't anywhere describe what JSON we create for a request, we only describe entities where the information is held, so the developer isn't going to know the structure.

I think that if we mention the JSON objects we should have a bit more information about them?

content/en/docs/appstore/modules/openai/_index.md
MarkvanMents: How about
```suggestion
| `Description`      | A description of what the function does, used by the model to choose when and how to call the function. This attribute is optional, but Mendix suggests using it to help the model choose the correct function. |
```

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -309,14 +339,31 @@ For more information on how to manage tokens for text generation, see [Managing
...
-| `Role`    | This is the role of the message author.<br />For more information, see the [ENUM_Role](#enum-role) section. |
+| `ToolType`| The type of the tool. Currently, only function is supported. <br />For more information, see the [ENUM_ToolType](#enum-tooltype) section. |
+
+##### 4.1.2.4 `AbstractFunction` {#abstractfunction}
+
+This is the abstract entity for `Function` reused in the Chat Completions request and response. Do not use this entity directly. Instead, use one of its specializations. 
+
+| Attribute | Description                                                  |
+| --------- | ------------------------------------------------------------ |
+| `Name`    | The name of the function to call.                            |
```
MarkvanMents, line 362: Would it be a good idea to add something here to support what you answered below:
```suggestion
| `Name`    | The name of the function to call. The model will use this to help identify which function to call where it has a choice. |
```

content/en/docs/appstore/modules/openai/_index.md
MarkvanMents: @tinaengel - is this OK?
```suggestion
| `Temperature`    | Decimal                                               | optional                      | This can be used to control the randomness of the output. The value should be a decimal between 0.0 and 2.0. The default value is 1.0. Higher values make the output more random, while lower values make it more focused and deterministic. Note: very high values for temperature (>1.7) may give unexpected results and even internal server errors. |
```

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -656,13 +772,22 @@ Use the microflow `ChatCompletions_Execute_WithHistory` to execute a chat comple
 | `Configuration`          | [Configuration](#configuration-entity)                | mandatory                     | This is an object that contains endpoint and API key.        |
 | `Model`                  | String                                                | only mandatory for **OpenAI** | This is the ID of the model to use; not considered for **Azure OpenAI** configurations. |
 | `ResponseFormat`         | [ENUM_ResponseFormat_Chat](#enum-responseformat-chat) | optional                      | This can be used to specify the format that the model must output. |
+| `Temperature`            | Decimal                                               | optional                      | This can be used to control the randomness of the output. The value should be a decimal between 0.0 and 2.0. The default value is 1.0. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic. Note: high values for temperature (>1.7) may give unexpected results and even internal server errors. |
```
MarkvanMents:
```suggestion
| `Temperature` | Decimal | optional | This can be used to control the randomness of the output. The value should be a decimal between 0.0 and 2.0. The default value is 1.0. Higher values make the output more random, while lower values make it more focused and deterministic. Note: very high values for temperature (>1.7) may give unexpected results and even internal server errors. |
```

content/en/docs/appstore/modules/openai/_index.md
MarkvanMents: Would it be worth making this explicit?
```suggestion
* `ChatCompletionsSession_AddMessage` is used to attach the historical messages to the `ChatCompletionsSession`. The content of such a message corresponds to a system, assistant, or user prompt. In the case of multiple historical messages the latest message has more influence than earlier ones.
```

content/en/docs/appstore/modules/openai/function-calling.md
MarkvanMents: In any case, I think we should be clearer where the validation takes place:
```suggestion
        * If it calls a function, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as described in the request. Note that the model may hallucinate parameters, so they should be validated within the function before being used.
```

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -349,7 +397,41 @@ This is a wrapper for a list of messages comprising the conversation so far.
...
+| `_id`     | The ID of the tool call, generated by the LLM. |
+
+##### 4.1.3.6 `ToolCallFunction` {#toolcallfunction}
+
+This is a specialization of the [AbstractFunction](#abstractfunction) entity and represents the function that the model called.
+The ToolCallFunction entity is only applicable for messages with role `assistant`.
+
+| Attribute  | Description                                    |
+| ---------- | -----------------------------------------------|
+| `Arguments`| The arguments to call the function with, as generated by the model in JSON format. Note that the model does not always generate valid JSON, and may hallucinate parameters not defined by your function schema. Validate the arguments in your code before calling your function. |
```
MarkvanMents: If the function is being called automatically you can't always validate before it is called - so making the recommendation more generic.

```suggestion
| `Arguments`| The arguments to call the function with, as generated by the model in JSON format. Note that the model does not always generate valid JSON, and may hallucinate parameters not defined by your function schema. Validate the arguments before using them. |
```

MarkvanMents (changes requested):
I've made some suggestions. Feel free to commit them if they look good. Or make different changes if you can think of something better.

@tinaengel would it be worth having a quick call about the high-level section at the beginning of `function-calling` as I think I am misunderstanding something.

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,75 @@
...
+### 1.1 High-level flow {#high-level}
+
+The basic steps of function calling are as follows:
+
+1. Invoke the chat completions API with a user prompt and a list of available functions (microflows) with expected input parameters.
+
+    * The model decides whether to call one or more functions based on the user prompt and the available functions.
+
+        * If it calls a function, the content of the assistant's response will be a stringified JSON object containing the input parameters of the function as described in the request. Note that the model may hallucinate parameters, so they should be validated before the function is actually called.
+        * Parse the string into JSON and call the function with its input parameters. The microflow runs in the original user's session and `$currentUser` can for example be used to apply security constraints.
```
MarkvanMents:
```suggestion
        * Parse the string into JSON and call the function (microflow) with its input parameters. The microflow runs in the original user's session and `$currentUser` can for example be used to apply security constraints.
```

content/en/docs/appstore/modules/openai/function-calling.md
MarkvanMents: https://github.com/mendix/docs/pull/7704/files#r1593792140

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,75 @@
...
+* `FunctionCollection_AddFunction` can be used to add a new `Function` to an existing `FunctionCollection`.
+
+Depending on the user prompt and the available functions, the model may suggest multiple tool calls to the same or different functions or there might be multiple API calls followed by new tools calls until the model returns the final assistant's response.
+A way to steer the function calling process is the [ToolChoice](/appstore/modules/openai-connector/#enum-toolchoice) parameter. This optional attribute on the [FunctionCollection](/appstore/modules/openai-connector/#functioncollection) entity controls which (if any) function is called by the model.
+
+{{% alert color="warning" %}}
+Function calling is a very powerful capability, but this also introduces potential risks. Function microflows do not respect any entity access rules for the current end-user. Make sure to retrieve and return only information that the end-user is allowed to view, otherwise confidential information may be visible to the current end-user in the assistant's response.
```
MarkvanMents:
```suggestion
Function calling is a very powerful capability, but this also introduces potential risks. Function microflows do not respect any entity access rules for the current end-user. You can use `$currentUser` to help you ensure you retrieve and return only information that the end-user is allowed to view, otherwise confidential information may be visible to the current end-user in the assistant's response.
```

content/en/docs/appstore/modules/openai/function-calling.md
MarkvanMents: https://github.com/mendix/docs/pull/7704/files#r1593797902

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: It also functions as UUID in other specialization so this is not correct. In my opinion the suggestion to use the description is sufficient

content/en/docs/appstore/modules/openai/_index.md
tinaengel-mx: How about: "If multiple messages are present, they should be ordered chronologically."

MarkvanMents: Great suggestion - I'll put it here and commit it.

```suggestion
* `ChatCompletionsSession_AddMessage` is used to attach the historical messages to the `ChatCompletionsSession`. The content of such a message corresponds to a system, assistant, or user prompt. If multiple messages are present, they should be ordered chronologically.
```

content/en/docs/appstore/modules/openai/_index.md
MarkvanMents: OK - I'll resolve this and not commit the suggestion.

content/en/docs/appstore/modules/openai/_index.md
```diff
@@ -656,13 +772,22 @@ Use the microflow `ChatCompletions_Execute_WithHistory` to execute a chat comple
...
 **Return value**
 
 | Name                    | Type   | Description                                                  |
 | ----------------------- | ------ | ------------------------------------------------------------ |
 | `AssistantResponseText` | String | Assistant message that was generated by the model as a response to a user message. |
 
+The following microflows may be used to construct and handle the required inputs: 
+
+* `ChatCompletionsSession_Create` is used to create the session wrapper that must be passed as input parameter. 
+* `ChatCompletionsSession_AddMessage` is used to attach the historical messages to the `ChatCompletionsSession`. The content of such a message corresponds to a system, assistant, or user prompt. In the case of multiple historical messages the latest message has more influence than earlier ones.
```
MarkvanMents:
```suggestion
* `ChatCompletionsSession_AddMessage` is used to attach the historical messages to the `ChatCompletionsSession`. The content of such a message corresponds to a system, assistant, or user prompt. If multiple messages are present, they should be ordered chronologically.
```

content/en/docs/appstore/modules/openai/_index.md
MarkvanMents: Wrong suggestion committed. Now committed the right suggestion 54022f4b47c43a6cfe084a809db1977edd88fae5

content/en/docs/appstore/modules/openai/function-calling.md
```diff
@@ -0,0 +1,78 @@
...
+
+### 1.1 High-level flow {#high-level}
+
+If you use the `ChatCompletions_Execute_WithoutHistory` or `ChatCompletions_Execute_WithHistory` microflows to do your chat completions, you can use the OpenAI connector to do function (microflow) calling for you in just two steps:
+
+1. Invoke the chat completions API with a user prompt and a list of available functions (microflows) with expected input parameters.
+
+    The model will decide which function (microflow) to call, if any, and will respond based on the information you provide and the response from any function (microflow) it calls.
+
+2. Invoke the chat completions API again and let the model answer your initial prompt with the new information provided by the function.
```
MarkvanMents: Does the developer even have to do this? From the examples in section 1.4 it looks like the connector does this step too and basically the function call is a single-step process.

tinaengel-mx: for with and without history operations, this is handled by the connector. For advanced it isn't

MarkvanMents: OK, I'll remove this and just have it in the detailed description.

MarkvanMents (approved):
Thanks for all your input and help.
This looks great now, so I'll publish it immediately.

PR #7707, opened 2024-05-08
Rght ToC styling

dbreseman:


MarkvanMents (changes requested):
Looks good. Thanks for doing this.
The only thing I'm not sure about is the GitHub logo. Now it is there I suddenly realize that both the editing and the issue reporting go to GitHub.
So I think I'd rather keep the current icons (since I don't think GitHub have a different logo for issues and editor).
Let me know if you want to talk about it. We can also have a meeting with design to see if we can come up with a solution for this.

dbreseman:
Makes sense! Maybe it makes sense to think about changing the tooltips on the icons, then. In any case, that aspect seems low priority and I'll leave it as is for now. I'll close this out and create a separate PR that just fixes the bolding on the active ToC element.

PR #7709, opened 2024-05-08
Update using-mendix-studio-pro-on-a-mac.md

ghost:
added warning to skip step 8 in case it causes issues


ConnorLand:
@mauritselzinga-mx we already have an info bubble at the top of the doc telling people to not use any of this guide if they are using the native Mac beta client.

ConnorLand:
@mauritselzinga-mx for the reason listed above, I'll close this PR. However, when you come back from vacation Slack me and if there is any value we can add to this doc I'll help get it done.

MariaShaposhnikova:
@ConnorLand , I've reopened the PR upon request. It describes a different use case than the one at the beginning of the doc.

PR #7712, opened 2024-05-10
Generating a Software Bill of Materials

grootjans:


content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+title: "SBOM Generation"
+url: /refguide/sbom-generation/
+category: "General Info"
+weight: 60
+description: "Describes how to create an SBOM from Studio Pro, and the underlying format    "
+tags: ["build", "deploy", "deployment package", "command-line", "studio pro", "sbom", "cyclonedx"]
+---
+
+## 1 Introduction
+When building a Mendix app, you make use of a lot of re-usable components such as Widgets, Javascript Action, Appstore modules and more. You want to be aware of what you use in your app, what other components they are dependent on and what licenses they make use of. Also, over time these components may become outdated, or become vulnerable. To ensure you are following all the development practices you would like to adhere to with regards to components, you need to have this information in a ready-to-use and standard format. Studio Pro does this by generating a Software Bill of Materials (SBOM) based on the standardized CycloneDX format.
```
tknippenberg: > Studio Pro does this by generating a Software Bill of Materials (SBOM) based on the standardized CycloneDX format.

At what point does Studio Pro do this?

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+
+## 2 Generating an SBOM
+You can generate an SBOM by using [mxbuild](/refguide/general/mxbuild). This generates an `sbom.json` file as part of your deployment folder or your package `.mda`. Bundling the SBOM together with your deployment ensures that the components that were present during the build are clearly represented.
+
+```bat
+mxbuild --java-home="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot" --java-exe-path="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot\bin\java.exe" --generate-sbom  "C:\Mendix\MyApp\MyApp.mpr"
+```
+
+## 3 SBOM format
+The SBOM is based on the [CycloneDX 1.4 JSON format](https://cyclonedx.org/docs/1.4/json/). Required fields are delivered, and fields such as `version`, `descripttion`. To identify the components from different package managers in a clear way, we use the `purl` field. The content follows the specification defined at: https://github.com/package-url/purl-spec. This includes Mendix modules, NPM packages and Java libraries.
```
tknippenberg: > Required fields are delivered, and fields such as `version`, `descripttion`.
Should `delivered` be between quotes as well? 
Typo in `descripttion`

content/en/docs/refguide/general/sbom-generation.md
grootjans: I'm leaving it generic here, as I believe we are considering adding a menu option to do this. For now, the method is described in the 'Generating an SBOM' section

content/en/docs/refguide/general/sbom-generation.md
grootjans: fixed.

content/en/docs/refguide/general/sbom-generation.md
tknippenberg: I thought currently this was only done when making a build in Mendix Cloud, but I might be wrong. If that's the case, perhaps we should remove the 'by Studio Pro' until it's built.

cwesdorp: Javascript Action by itself is not a re-usable (re-distributable) artefact. For Mendix we have Modules and Widgets, for 3rd party JARs and NPM.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+description: "Describes how to create an SBOM from Studio Pro, and the underlying format    "
+tags: ["build", "deploy", "deployment package", "command-line", "studio pro", "sbom", "cyclonedx"]
+---
+
+## 1 Introduction
+When building a Mendix app, you make use of a lot of re-usable components such as Widgets, Javascript Action, Appstore modules and more. You want to be aware of what you use in your app, what other components they are dependent on and what licenses they make use of. Also, over time these components may become outdated, or become vulnerable. To ensure you are following all the development practices you would like to adhere to with regards to components, you need to have this information in a ready-to-use and standard format. Studio Pro does this by generating a Software Bill of Materials (SBOM) based on the standardized CycloneDX format.
+
+## 2 Generating an SBOM
+You can generate an SBOM by using [mxbuild](/refguide/general/mxbuild). This generates an `sbom.json` file as part of your deployment folder or your package `.mda`. Bundling the SBOM together with your deployment ensures that the components that were present during the build are clearly represented.
```
cwesdorp: "or you package `.mda`"
Suggests the SBOM is also inside the MDA. That is not the case right?

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bat
+mxbuild --java-home="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot" --java-exe-path="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot\bin\java.exe" --generate-sbom  "C:\Mendix\MyApp\MyApp.mpr"
+```
+
+## 3 SBOM format
+The SBOM is based on the [CycloneDX 1.4 JSON format](https://cyclonedx.org/docs/1.4/json/). Required fields are delivered, and fields such as `version`, `description`. To identify the components from different package managers in a clear way, we use the `purl` field. The content follows the specification defined at: https://github.com/package-url/purl-spec. This includes Mendix modules, NPM packages and Java libraries.
```
cwesdorp: description is not a required field. Only name is. At minimal we provide name, version and purl.

quinntracy:
@cwesdorp @tknippenberg Is this PR ready for a language review, or should I wait until the existing comments/questions are resolved?

content/en/docs/refguide/general/sbom-generation.md
grootjans: Good point, that does not make complete sense.

content/en/docs/refguide/general/sbom-generation.md
grootjans: What I tried to convey with the sentence is that 
- required fields are delivered
- also other fields such as description, version, purl. These are not required according to the CycloneDX spec

content/en/docs/refguide/general/sbom-generation.md
grootjans: The SBOM is inside the MDA if you do not specify an alternate location.

content/en/docs/refguide/general/sbom-generation.md
grootjans: I made a slight update, hope this is better.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+
+## 2 Generating an SBOM
+You can generate an SBOM by using [MxBuild](/refguide/mxbuild/). This generates an `sbom.json` file as part of your deployment folder or your package `.mda`. Bundling the SBOM together with your deployment ensures the components that were present during the build are clearly represented.
+
+```bat
+mxbuild --java-home="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot" --java-exe-path="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot\bin\java.exe" --generate-sbom  "C:\Mendix\MyApp\MyApp.mpr"
+```
+
+## 3 SBOM Format
+The SBOM is based on the [CycloneDX 1.4 JSON format](https://cyclonedx.org/docs/1.4/json/). This includes required fields, and information such as `version` and `description`. To identify the components from different package managers, we use the `purl` field. The content follows the specification defined at `https://github.com/package-url/purl-spec`. This includes Mendix modules, NPM packages, and Java libraries.
```
quinntracy: > we use the `purl` field.

Who is "we"? Does Mendix or someone else recommend to use this field? Please clarify, as we want to avoid using personal pronouns.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+weight: 60
+description: "Describes how to create an SBOM from Studio Pro."
+tags: ["build", "deploy", "deployment package", "command-line", "studio pro", "sbom", "cyclonedx"]
+---
+
+## 1 Introduction
+When building a Mendix app, you use many reusable components such as Widgets and App Store modules, which make use of components such as Java libraries and node modules. You should know what you use in your app, what other components they are dependent on, and what licenses they make use of. Over time, these components may become outdated or vulnerable, so it is important to have this information in a ready-to-use, standard format. Studio Pro does this by generating a Software Bill of Materials (SBOM) based on the standardized CycloneDX format.
+
+## 2 Generating an SBOM
+You can generate an SBOM by using [MxBuild](/refguide/mxbuild/). This generates an `sbom.json` file as part of your deployment folder or your package `.mda`. Bundling the SBOM together with your deployment ensures the components that were present during the build are clearly represented.
```
Sakshi-Dhakad, line 15: This isn't true. The SBOM is generated together with the deployment package only when online platform service is used to build a deployment package. Is a SBOM is being created from Studio Pro, it has no link with the .MDA. Would suggest removing the sentence or saying something like the SBOM is based on the current state code/project/branch in Studio Pro
> This generates an `sbom.json` file as part of your deployment folder or your package `.mda`. Bundling the SBOM together with your deployment ensures the components that were present during the build are clearly represented.

content/en/docs/refguide/general/sbom-generation.md
Sakshi-Dhakad: we = Mendix. purl is a common industry wide standard used to identify components

Sakshi-Dhakad: - SBOM also includes Widgets
- Instead of "Mendix modules", I'd suggest saying "Mendix marketplace modules and widgets" since we only detect the components which are identifiable via marketplace ids to my knowledge.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+title: "SBOM Generation"
+url: /refguide/sbom-generation/
+category: "General Info"
+weight: 60
+description: "Describes how to create an SBOM from Studio Pro."
+tags: ["build", "deploy", "deployment package", "command-line", "studio pro", "sbom", "cyclonedx"]
+---
+
+## 1 Introduction
+When building a Mendix app, you use many reusable components such as Widgets and App Store modules, which make use of components such as Java libraries and node modules. You should know what you use in your app, what other components they are dependent on, and what licenses they make use of. Over time, these components may become outdated or vulnerable, so it is important to have this information in a ready-to-use, standard format. Studio Pro does this by generating a Software Bill of Materials (SBOM) based on the standardized CycloneDX format.
```
Sakshi-Dhakad: 1. Replace "App Store modules" with marketplace modules since "App store" is old speak and all our public documentation refers to it as marketplace content
2. One can also make use of java libraries and npm packages if they create custom java or javascript in their mendix app. By saying that we only detect the java and npm modules, we are saying that we won't detect direct dependencies which are dragged with the above use case, which isn't correct. So would suggest keeping it very generic or to lay out both use cases which lead to java and npms
3. What are "node modules"? Did you mean to say "npm" modules?

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+
+```bat
+mxbuild --java-home="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot" --java-exe-path="C:\Program Files\Eclipse Adoptium\jdk-11.0.16.101-hotspot\bin\java.exe" --generate-sbom  "C:\Mendix\MyApp\MyApp.mpr"
+```
+
+## 3 SBOM Format
+The SBOM is based on the [CycloneDX 1.4 JSON format](https://cyclonedx.org/docs/1.4/json/). This includes required fields, and information such as `version` and `description`. To identify the components from different package managers, we use the `purl` field. The content follows the specification defined at `https://github.com/package-url/purl-spec`. This includes Mendix modules, NPM packages, and Java libraries. 
+
+## 4 Example Output
+Since a full SBOM contains all project details, this is only a partial representation of a full example. To simplify, below is one of the Java libraries that is packaged; a full example would contain all App Store modules, Java libraries, and JavaScript libraries under `components`. The `dependencies` field will contain the dependency relationship between the separate components.
```
Sakshi-Dhakad: 1. Misses widgets, 2. Replace 'appstore' with 'marketplace' and 3. We need to use the same terminology, if we are using npm packages, then I'd suggest we stick to that in all cases. So replace "javascript libraries" with "npm packages"
> Example would contain all App Store modules, java libraries and java script libraries

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+}
+
+```
+
+## 5 Supported Features
+The table below describes what components are currently covered in the generated SBOM, and from what version. It is recommended to use version `10.10`, `9.24.22`, or `10.6.9` and above.
+
+| Feature | Description | Version
+| --- | --- | --- |
+| Mendix Modules | `9.24.14`, `10.4.0` | Mendix Modules in your projects and imported modules as, for instance, from the App Store. This does not include Add-on Modules, Solutions, or Extensions. |
```
Sakshi-Dhakad: Are all Mendix modules listed in the SBOM? I thought it was only standard marketplace modules that were identified. As a reader, when I read this, I'd also think that my custom modules will be listed here which isn't the case I think.

Additionally, content from both public and private marketplace will be listed in the SBOM. Many customers ask this question during interviews/feedback sessions, would suggest adding that.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+      "modified": false
+    },
+   
+  "externalReferences": [],
+  "dependencies": []
+}
+
+```
+
+## 5 Supported Features
```
Sakshi-Dhakad, line 73: Below are a few limitations I'd suggest documenting apart from the ones in the table:
1. Marketplace modules, or modules that are imported directly, so not via the marketplace will not be shown in the SBOM. When they contain Jars or other components that are listed, those components will be present in the SBOM. The module will not be listed as a marketplace module, and (should) not have a package url.
2. Widget packages will only have a App Store version guid when they were downloaded from the marketplace in Studio Pro. For widget packages which are part of a module or imported from the local file system there is no App Store version guid available. 

3. Module, widget, npm dependencies are not available.

content/en/docs/refguide/general/sbom-generation.md
```diff
@@ -0,0 +1,83 @@
...
+| Feature | Description | Version
+| --- | --- | --- |
+| Mendix Modules | `9.24.14`, `10.4.0` | Mendix Modules in your projects and imported modules as, for instance, from the App Store. This does not include Add-on Modules, Solutions, or Extensions. |
+| [Java Dependencies](/refguide/managed-dependencies/) | `9.24.14`, `10.4.0` | Java libraries imported into your project using [Managed Dependencies](/refguide/managed-dependencies/), or those manually added in the `userlib` folder. Libraries added through managed dependencies will have access to information of the package manager and will include more details, such as the used license. |
+| [Widgets](/appstore/widgets/) | `10.9.0`, `10.6.6`, `10.10.0`, `9.24.18` | User interface elements downloaded from the Marketplace, such as drop-downs or buttons. |
+| Javascript Actions(/refguide/javascript-actions/) | `10.10` | NPM libraries that are used in your JavaScript actions. |
+
+## 6 Read More
+* [CycloneDX](https://cyclonedx.org/)
+* [Package URL spec](https://github.com/package-url/purl-spec)
```
Sakshi-Dhakad, line 85: You can also add a reference to software composition. The docs url is going to be - https://docs.mendix.com/control-center/software-composition/

Sakshi-Dhakad (changes requested):
Added a few comments

MariaShaposhnikova:
I've updated labels: removed 10.10 as it is already out and we use version labels for future releases (to not merge too early) and changed Internal/Peer Review to Internal WIP, as this PR is still work in progress.

content/en/docs/refguide/general/sbom-generation.md
grootjans: You should be able to change this into 'the purl field is used' and take out the personal pronoun. Thanks for the clarification Sakshi.

content/en/docs/refguide/general/sbom-generation.md
grootjans: Though I think it would be good to use 'marketplace modules' in general, I'll have to experiment to see what happens with custom modules.

content/en/docs/refguide/general/sbom-generation.md
grootjans: the `sbom.json` is generated inside the `mda` by default. For the online services I would expect you specify the `--sbom-output-path=VALUE` parameter, allowing you to generate outside of the `mda`.

content/en/docs/refguide/general/sbom-generation.md
grootjans: I updated the structure, no longer using a personal pronoun.

grootjans: Updated.

content/en/docs/refguide/general/sbom-generation.md
grootjans: I've taken it out, as this is just a repetition of previous paragraphs.

content/en/docs/refguide/general/sbom-generation.md
grootjans: Good point, this was too general and this would not have been reflected. I updated the text.

content/en/docs/refguide/general/sbom-generation.md
grootjans: I can add it with a small PURL as soon as it is published.

content/en/docs/refguide/general/sbom-generation.md
grootjans: 1. Added a clarification in the table, though let's further update this based on the actual behavior
2. Good point. Though I would really like us to move away from that limitation, that is likely a part of the larger topic of cross-mendix-module dependencies.
3. Modules do not have dependencies at this moment, I would not go with documenting that. NPM dependencies of modules are there (I just checked Nanoflow commons, the NPM dependencies are there), Widget dependencies are likely not present, only their used libraries.

 For now I will add:
he field `dependencies` will contain the dependency relationship from Mendix Marketplace Modules to npm modules and java libraries.

content/en/docs/refguide/general/sbom-generation.md
grootjans: 1. Good point, will update
2. Good point, will update
3. Node module is the term for a javascript library, where npm is the 'Node Package Manager', who is able to resolve these dependencies. For clarity sake, let's choose a single term here, given that the SBOM uses the npm package manager as a PURL identifier for now, 'npm modules' is likely the better choice.

PR #7740, opened 2024-05-15
Explained commit order in Developer Portal

tknippenberg:


Luyao-Zhang-1:
@tknippenberg Thanks for sending the PR. Do you think we need to add this information also in this [doc](https://docs.mendix.com/refguide/view-menu/#32-changes)?

tknippenberg:
> @tknippenberg Thanks for sending the PR. Do you think we need to add this information also in this [doc](https://docs.mendix.com/refguide/view-menu/#32-changes)?

Not there as it's only about the "changes" pane, but [here](https://docs.mendix.com/refguide/using-version-control-in-studio-pro/#history) where the "history" panel is mentioned, it could be added!

content/en/docs/developerportal/general/team-server/_index.md
```diff
@@ -54,6 +54,8 @@ For SVN, the command has the form `svn checkout --revision "{REVISION}" --userna
 
 ## 4 Revision History {#revision-history}
 
+{{% alert color="info" %}}For Git-based applications, revisions are sorted according to the commit history, which sometimes do not reflect the chronological order due to Git's decentralized nature and local commits. In contrast, for SVN-based applications, revisions are sorted strictly in chronological order. {{% /alert %}}
```
MariaShaposhnikova: @Luyao-Zhang-1 , @tknippenberg , we were removing SVN mentions everywhere in docs, why do we bring it back here?

MariaShaposhnikova:
```suggestion
{{% alert color="info" %}}For Git-based applications, revisions are sorted according to the commit history, which sometimes does not reflect the chronological order due to Git's decentralized nature and local commits. In contrast, for SVN-based applications, revisions are sorted strictly in chronological order. {{% /alert %}}
```

tknippenberg: I was not aware we are removing SVN mentions everywhere in the docs. For this major version-agnostic page it seems a bit premature to me, as most of the customer base is still on SVN.

The reason I added this piece in general was because of some questions, so I thought it'd be sensible to explain the difference. But if the direction is to remove SVN mentions, also in the Developer Portal documentation, then you can remove that part.

MariaShaposhnikova: @tknippenberg , our docs usually reflect the latest version (if it's not a dedicated refguide9 space, for example), so we deleted mentions of SVN as Mendix 10 users cannot create SVN apps and are encouraged to move to Git.

tknippenberg: I'm not sure how well that policy is suited for tying into specific major-version-dependent features, but it's clear. Then let's remove that part on SVN.

PR #7757, opened 2024-05-17
Add New Users button

dbreseman:


MarkvanMents (approved):
Not the ideal solution, but a good start and works well.
Approving.

PR #7777, opened 2024-05-22
Clarify scheduled event disable/enable KI

emirmx:
The issue also affects normal deployments. Removed mention of `App Settings / Configurations` which may suggest this happens only when development.


MarkvanMents (approved):
Thanks for the change, Emir.
Will approve and merge.

emirmx:
Thanks Mark!

PR #7795, opened 2024-05-24
Proposal for Front Page Redesign

dbreseman:


dbreseman:
Front page redesign handled in https://github.com/mendix/docs/pull/8255

PR #7814, opened 2024-05-29
Fix the style issue

Luyao-Zhang-1:


Luyao-Zhang-1:
Close the PR because it is created to make screenshots.

PR #7838, opened 2024-06-04
Change "Mendix Docs" to "Docs" in MxDock to reflect the new design

Luyao-Zhang-1:


MarkvanMents (approved):
Looks great. Thanks for doing this so quickly.

PR #7843, opened 2024-06-04
Disable on-click link to images and open images as pop-ups

Yiyun333:


layouts/partials/hooks/body-end.html
```diff
@@ -8,3 +8,29 @@
...
+  Copyright (c) 2016-present François Chalifour
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software. -->
+<script src="https://cdnjs.cloudflare.com/ajax/libs/medium-zoom/1.0.7/medium-zoom.min.js" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
MarkvanMents: Should we be embedding the script itself rather than a link - what happens if someone does something malicious to the script in this library?

assets/scss/_styles_project.scss
```diff
@@ -854,3 +865,5 @@ nav.foldable-nav {
     font-size: $font-size-base !important;
     color: $primary !important;
 }
+
+
```
MarkvanMents:
```suggestion
```

MarkvanMents (commented):
Looks really good.
Is there a way to keep the old functionality if Javascript is turned off in the browser. Probably not so important as right-click also lets you open the image in a new tab, but would be nice if it was easy to do.

I've made a comment about picking up a script off the internet, perhaps we can discuss the best way to implement it - but nothing to stop you demoing it tomorrow.

Great work.

Yiyun333:
> Looks really good. Is there a way to keep the old functionality if Javascript is turned off in the browser. Probably not so important as right-click also lets you open the image in a new tab, but would be nice if it was easy to do.
> 
> I've made a comment about picking up a script off the internet, perhaps we can discuss the best way to implement it - but nothing to stop you demoing it tomorrow.
> 
> Great work.

Thanks a lot for the review!

> Is there a way to keep the old functionality if Javascript is turned off in the browser.

I'll have a look and see if there is an easy way to do this.

> I've made a comment about picking up a script off the internet, perhaps we can discuss the best way to implement it 

Yes, I think that's a good idea!

layouts/partials/hooks/body-end.html
Yiyun333: Good point!

PR #7848, opened 2024-06-05
Clarify limitations of private repositories

feddovanede:
Clarify limitations of private repositories.


feddovanede:
Should we add this also to the documentation for Mendix 9 and Mendix 8?

MariaShaposhnikova:
@feddovanede , sorry, I clicked **Merge** and then realized I did not answer about Mendix 9 and 8. We can add this information to Mendix 8 and 9 if necessary. Did customers on 8 and 9 have questions about it? Shall I add it?

feddovanede:
No, we didn’t get any questions from customers about this for Mendix 8 and 9. I just noticed the whole paragraph is missing for Mendix 8 and 9. This might lead customers to those things are supported with Mendix 8 and 9, but they are not. The limitations are the same, regardless of Mendix version.

PR #7850, opened 2024-06-06
Remove blank lines in the front matter

MarkvanMents:
I got fed up of the blank line which is in the front matter. It wasn't breaking anything, but I'd rather the front matter didn't have blank lines in it, just in case Hugo becomes fussier in the future.

Used the regular expression
^(?<!\n)(---\n[\s\S]*?)(\n\n)([\s\S]*?^---\n)
and replaced with 
$1\n$3

Despite what ChatGPT3.5 will say if you ask it, the "(?<!\n)" part is needed to ensure we only make the change at the beginning of the file.


dbreseman (approved):
LGTM!

PR #7863, opened 2024-06-07
[RUN-3787][RUN-3855][RUN-3666] Add documentation for managed dependency improvements (10.12.0)

atennapel:
This MR updates the documentation for multiple different improvements to Java Dependency Management. These features will be released in 10.12.0.


content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -43,7 +43,23 @@ After finding the package of your choice, locate the Snippets part, containing t
...
 {{< figure src="/attachments/refguide/java-programming/managed-dependencies/junit-notation-example.png" class="no-border" >}}
 
-## 3 Dependency Synchronization
+## 3 Adding or Updating Exclusions
+
+{{% alert color="info" %}}
+Exclusions are available in Studio Pro 10.12.0 and above.
+{{% /alert %}}
+
+In certain cases a conflict between transitive dependencies between different configured Java dependencies may appear.
```
lvanengelen: 🎄 replace the first 'between' with 'in' to not repeat the same word.

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -43,7 +43,23 @@ After finding the package of your choice, locate the Snippets part, containing t
...
+To fix this transitive dependencies can be excluded when configuring dependencies.
```
lvanengelen: 🎄 I feel there's a comma missing after 'To fix this'

content/en/docs/refguide/java-programming/managed-dependencies.md
lvanengelen: Be a bit more specific about the reason why you would add an exclusion here (to not be confused with the exclusions in the dependencies tab).

One thing to highlight (either here or in de docs for the dependencies tab) is that they are part of the module when exported, so 'inherited' when imported in another project, while those in the dependencies tab are not (because they are global to a project).

Maybe mention a way that a conflict can occur as an example, e.g. a dependency that is unused but might conflict with other modules? Maybe you know something more realistic?

content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
```diff
@@ -392,9 +392,22 @@ A microflow selected for this setting will start every time a workflow changes i
...
 
-## 8 Deployment Tab {#deployment}
+## 8 Dependencies Tab {#deployment}
 
-This tab allows you to exclude libraries from deployment. For example, you can exclude libraries when you consume a different version of an existing add-on module.
+{{% alert color="info" %}}
+The Deployment tab was renamed to Dependencies in Studio Pro 10.12.0.
+{{% /alert %}}
+
+This tab can be used to view the managed dependencies in your App and to manage the dependencies in the userlib directory.
```
lvanengelen: Maybe make it more explicit by something like '... to view all the managed dependencies in your App in one place and ...' to highlight the benefits of this view.

atennapel: Done.

content/en/docs/refguide/java-programming/managed-dependencies.md
atennapel: Done.

content/en/docs/refguide/java-programming/managed-dependencies.md
atennapel: Done.

content/en/docs/refguide/java-programming/managed-dependencies.md
atennapel: Done.

lvanengelen (approved):

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -43,7 +43,27 @@ After finding the package of your choice, locate the Snippets part, containing t
...
 
-## 3 Dependency Synchronization
+## 3 Adding or Updating Exclusions
+
+{{% alert color="info" %}}
+Exclusions are available in Studio Pro version 10.12.0 and above.
+{{% /alert %}}
+
+Sometimes you may find a conflict in transitive dependencies between different configured Java dependencies.
+For example, if two Java dependencies have the same package name and classes but a different group or artifact then they may both be included in your app, possibly causing conflicts.
```
MarkvanMents: This reads as being quite weak. You **may** find a conflict from dependencies which **may** be included and **possibly** cause conflicts.

Can we make it a bit more explicit - perhaps with an example?

content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
```diff
@@ -392,9 +392,25 @@ A microflow selected for this setting will start every time a workflow changes i
...
+
+This tab can be used to view the managed dependencies in your app in one place and to manage the dependencies in the userlib directory. It contains three tabbed sections.
+
+### 8.1 Overview
+
+This shows all the managed dependencies in your app listed by group and artifact. It shows which versions of the dependencies you have and which modules they are coming from. If your app reports multiple versions of the same group and artifact then the highest version is used, so having multiple versions of a dependency is not necessarily a problem.
+
+### 8.2 Managed Dependency Exclusions
+
+This shows all the managed dependencies in your app listed by package name. This overview includes transitive dependencies. If you have conflicts between different dependencies, you can uncheck here any files which you want to exclude. Ensure you leave at least one dependency which supports any calls made by your app or its dependencies.
```
MarkvanMents: These both say that they **show all the managed dependencies**. I have indicated a difference in what is displayed, but can it be made clearer why they are two different lists?

MarkvanMents (changes requested):
Hi @lvanengelen and @atennapel ,
Thanks for the PR.
I have made some changes in this commit: 3276468e3b10ab39d753371341daa70b6f0aeb4e
I have also made a couple of minor comments which might need addressing - let me know if you think they are valid.

Please let me know if you are happy with the changes, or if there are further things which could be improved.
Thanks
Mark

content/en/docs/refguide/java-programming/managed-dependencies.md
lvanengelen: What about:

'Transitive dependencies of different configured Java dependencies might conflict. For example, a package moved from one dependency to another, but some modules still use the old dependency while others moved to the new dependency. This causes the package to be defined multiple times which is not allowed.'

I think this describes at least one case we ran into.

content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
lvanengelen: The main difference is that 'Managed Dependency Exclusions' shows all dependency artifacts including transitive ones (i.e. dependencies of dependencies) 'Overview' only shows the direct dependencies (so only one level). We mention that ('This overview includes transitive dependencies'), should that be elaborated more?

Another (less important) difference is that the 'Managed Dependency Exclusions' lists actual dependency artifact files on disk while the 'Overview' merely shows the definitions of the dependencies.

MarkvanMents: OK - thanks. How about:
```suggestion
This shows all the direct managed dependencies in your app listed by group and artifact. It shows which versions of the dependencies you have and which modules they are coming from. If your app reports multiple versions of the same group and artifact then the highest version is used, so having multiple versions of a dependency is not necessarily a problem.

### 8.2 Managed Dependency Exclusions

This shows all the managed dependencies in your app listed by package name. This overview includes both direct and transitive dependencies. If you have conflicts between different dependencies, you can uncheck here any files which you want to exclude. Ensure you leave at least one dependency which supports any calls made by your app or its dependencies.
```

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: I've added this in c0ca898e626e71e0ac032c3c61f963bde8e4733a. Can you check it is clear enough. Thanks.

MarkvanMents (changes requested):
Thanks for your response @lvanengelen 
I've committed one change and suggested another. Can you have a look and see if they are good.
Thanks.

content/en/docs/refguide/modeling/app-explorer/app/app-settings/_index.md
lvanengelen: Looks good!

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -49,8 +49,7 @@ After finding the package of your choice, locate the Snippets part, containing t
 Exclusions are available in Studio Pro version 10.12.0 and above.
 {{% /alert %}}
 
-Sometimes you may find a conflict in transitive dependencies between different configured Java dependencies.
-For example, if two Java dependencies have the same package name and classes but a different group or artifact then they may both be included in your app, possibly causing conflicts.
+Transitive (or indirect) dependencies of different configured Java dependencies might conflict. For example, a package used in your app moves to a new version of a dependency; some modules use the new dependency but some still use the old dependency. This causes the package to be defined multiple times which is not allowed.
```
lvanengelen: I'm afraid this is incorrect because different versions of the same dependency are just resolved to the highest resolved to the highest version and things work ok (as long as the versions are api compatible, i.e. have the same major version).

The problem I wanted to convey is a bit more subtle: say library `org.foo:foo:2.0` contains package `frobnicator` but the owner of the library moves the package `frobnicator` to another library `org.foo:bar:1.0`, then some modules might depend on `org.foo:foo:2.0` and some on `org.foo:bar:1.0`. These won't be resolved as being different versions of the same library, so they will both be included in the app. But then the `frobnicator` is duplicated.

content/en/docs/refguide/java-programming/managed-dependencies.md
lvanengelen: I'm afraid that change is incorrect. See [my comment](https://github.com/mendix/docs/pull/7863/commits/c0ca898e626e71e0ac032c3c61f963bde8e4733a#r1640018222)

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: OK - I'll put your original wording back. I'm not sure I can describe the `frobnicator` case accurately enough.
```suggestion
Transitive (or indirect) dependencies of different configured Java dependencies might conflict. For example, if two Java dependencies have the same package name and classes but a different group or artifact then they may both be included in your app, possibly causing conflicts.
```

MarkvanMents (approved):
@lvanengelen and @atennapel 
Thanks for the PR.
I've reverted my incorrect explanation back to what you wrote. I think everything else is agreed so marking as approved ready for publication with 10.12.

atennapel:
Looks good to me. Approved from my side.

content/en/docs/refguide/java-programming/managed-dependencies.md
lvanengelen: I think the wording is ok now, thanks

PR #7880, opened 2024-06-12
Update hermes.md

jdiehl:
@ConnorLand 
@Ahmad-Elsayed 

The link in Studio Pro 10.11.0 points to https://docs.mendix.com/refguide/mobile/hermes - can we update the alias to match that? We will correct the link in 10.12.


ConnorLand:
Sounds good @jdiehl . When I come back next week I'll see if there's something we all can do in order to make sure SP has fewer incorrect links pointing at docs. The issue is that sometimes URLs are coded into SP before docs gets a chance to alter the links to our standards in GitHub.

PR #7881, opened 2024-06-12
[RUN 4121] Add links to remote repositories

akkrishnakumar:
Customers who use custom repositories usually have maven central configured. But sometimes they don’t know how to configure the Gradle Plugins repository because finding the right link to this repository can be a little tedious. Hence, we plan to provide links to these repositories in the documentation for faster development.


content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -90,6 +90,12 @@ For more details, refer to the Gradle documentation on [Declaring repositories](
 
 {{< figure src="/attachments/refguide/java-programming/managed-dependencies/custom-repository.png" class="no-border" >}}
 
+{{% alert color="info" %}}
+Custom Repositories like Nexus, JFrog etc. are not configured to point to central remote repositories by default. You will have to request the IT infrastructure administrators to configure them.
```
emirmx: Can we say `may not be` instead of `are not`? But if we are sure of that, then it is fine :-)

emirmx: Instead of, `You will have to request the IT infrastructure administrators to configure them.`,
would the following be better?
`The following sources need to be configured in your Nexus, JFrog, etc.`

akkrishnakumar: Good point. For JFrog, I know for sure that it is not configured by default for a fresh installation. I presume the same for nexus. 

But being skeptical here is the right way to go because we are not sure. 
I'll make the change.

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -90,6 +90,12 @@ For more details, refer to the Gradle documentation on [Declaring repositories](
 
 {{< figure src="/attachments/refguide/java-programming/managed-dependencies/custom-repository.png" class="no-border" >}}
 
+{{% alert color="info" %}}
+Custom Repositories like Nexus, JFrog etc. maybe not be configured to point to central remote repositories by default. The following sources need to be configured in your custom remote repository.
+1. Maven Repository - https://repo1.maven.org/maven2/
+2. Gradle Plugins Repository - https://plugins.gradle.org/m2/
+{{% /alert %}}
```
MarkvanMents: Hi @akkrishnakumar 
Reading this section of the managed dependencies it is not clear where I might have to configure these.
Do I have to do this as part of the 
`flatDir { dirs 'lib'}` json, or am I supposed to change them within the repository (e.g. `lib`) that I am pointing to.

It also seems to conflict with the default configurations for `gradlePluginPortal` and `mavenCentral`. 

Would an example of how to set up a configuration pointing at the right Maven and Gradle repos?

It is not clear to me how the `lib` directory relates to the repository configuration.

I realize I am not the target audience, but I think we could make this clearer.

We can have a quick call about it if that is easier.

akkrishnakumar: Yeah sure. Can you schedule one for Monday if that is ok ??

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -71,11 +71,16 @@ Platform-supported Marketplace modules created by Mendix have been updated with
...
 
-Custom repositories are configured in the **Repositories** setting of the **Deployment** tab in the [Preferences](/refguide/preferences-dialog/) dialog box. This setting uses the same syntax as Gradle. For internal usage of the platform, some dependencies are required which are also resolved using the configured repositories. For example, to resolve dependencies from a directory `lib`, enter the following:
+Custom repositories are configured in the **Repositories** setting of the **Deployment** tab in the [Preferences](/refguide/preferences-dialog/) dialog box. This setting uses the same syntax as Gradle. For internal usage of the platform, some dependencies are required which are also resolved using the configured repositories. For example, to resolve dependencies using maven from a remote nexus repository on port 8046 of your local network, enter the following:
 
 ```groovy {linenos=false}
-flatDir {
-    dirs 'lib'
+maven {
```
MarkvanMents, line 77: Is there a reason why this is maven and not mavenCentral?

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -86,6 +91,13 @@ gradlePluginPortal()
 mavenCentral()
 ```
```
MarkvanMents, line 91: When I choose custom repositories in 10.10 I only get the `mavenCentral()` repository, not a `gradlePluginPortal()` is this to be expected?

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: Updated in this Commit: 530e758fccf0c682d891dadd3509ecd0a4d755b3

MarkvanMents (commented):
Hi @akkrishnakumar ,
I've updated the text and image in this commit: 530e758fccf0c682d891dadd3509ecd0a4d755b3

Could you review what I've written and see if it makes sense. I've tried to make it clearer (to me, at least!) where things need to be configured.
I have two further questions, below, but they are probably irrelevant.

content/en/docs/refguide/java-programming/managed-dependencies.md
akkrishnakumar: `Maven` is a style of directory structure. The directory structure also contains certain metadata. This allows various tools, like Gradle for example, to resolve the required dependencies by searching in the right (sub-)folders.

Think of it like a dictionary. 

Other styles are `flatDir` which is just a flat file folder structure with almost no metadata and `Ivy` style which is kind of similar to `maven` style but is more primitive compared to `maven`.

This is overly simplified version of what it actually is. 

`MavenCentral` on the other hand is a repository which actually hosts these dependencies in the form of (sub)folders

content/en/docs/refguide/java-programming/managed-dependencies.md
akkrishnakumar: No, You should get both. 

Did you ever change this setting in a previous version of Mendix ? If yes, then this new version simply inherits the setting from earlier versions.

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -71,11 +71,16 @@ Platform-supported Marketplace modules created by Mendix have been updated with
...
 
-Custom repositories are configured in the **Repositories** setting of the **Deployment** tab in the [Preferences](/refguide/preferences-dialog/) dialog box. This setting uses the same syntax as Gradle. For internal usage of the platform, some dependencies are required which are also resolved using the configured repositories. For example, to resolve dependencies from a directory `lib`, enter the following:
+Custom repositories are configured in the **Repositories** setting of the **Deployment** tab in the [Preferences](/refguide/preferences-dialog/) dialog box. This setting uses the same syntax as Gradle. For internal usage of the platform, some dependencies are required which are also resolved using the configured repositories. For example, to resolve dependencies using maven from a remote nexus repository on port 8046 of your local network, enter the following:
```
akkrishnakumar: Is it a good idea to mention a specific repository in the example ? 

Should we make it more generic so that customers will not feel that this feature only works with Nexus ?
If not, then you can resolve this comment.

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -71,11 +71,16 @@ Platform-supported Marketplace modules created by Mendix have been updated with
...
 
 ```groovy {linenos=false}
-flatDir {
-    dirs 'lib'
+maven {
+    url 'http://localhost:8046/artifactory/maven-remote/'
```
akkrishnakumar: Again, this is very specific to Nexus. 
How about
`url 'url-to-your-custom-remote-repository'`

content/en/docs/refguide/java-programming/managed-dependencies.md
```diff
@@ -71,11 +71,16 @@ Platform-supported Marketplace modules created by Mendix have been updated with
...
+    allowInsecureProtocol true
```
akkrishnakumar: This is my mistake, we should remove this. This was added to support `localhost` (or local setup of Jfrog on my system). 

Ideally, we shouldn't give this as an example. Customers are free to configure it this way if they please if they find it in the official documentation of Gradle.

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: Thanks for the explanation.
I'll assume that the customer will know this and not mention it in the documentation.

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: I will pick up your suggestion below and make this more generic.

See https://github.com/mendix/docs/pull/7881/commits/2a86e7b3dad8853854f4099cbce2c610ff4ebcf9

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: I never changed it I don't think. I can't think why I would.
I've just downloaded and opened 10.11 and it still only gives me mavenCentral() as the repo. Does it copy the preferences forward from 10.10? If so, perhaps my preferences have got corrupted/changed at some point in the past?

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: Yes, much better.

Updated in this commit: https://github.com/mendix/docs/pull/7881/commits/2a86e7b3dad8853854f4099cbce2c610ff4ebcf9

content/en/docs/refguide/java-programming/managed-dependencies.md
MarkvanMents: Removed in https://github.com/mendix/docs/pull/7881/commits/2a86e7b3dad8853854f4099cbce2c610ff4ebcf9

MarkvanMents (commented):
Hi Akhil @akkrishnakumar 
I've made the example more generic, please let me know if you think more changes are needed or whether we can now publish this.
Thanks for working with me on this.

content/en/docs/refguide/java-programming/managed-dependencies.md
akkrishnakumar: Yes, that could be the case. 
Nevertheless, I will test it out once again. Will fix it if there is a defect. 
For now you can presume that both will be present.

akkrishnakumar:
LGTM 👍🏽

MarkvanMents (approved):
Agreed with @akkrishnakumar 
Thanks for all your help in making this change as clear as possible.
Approving and merging.

PR #7884, opened 2024-06-13
Ensure navigation is still shown on non-JS browsers (e.g. Google Search).

MarkvanMents:
Google does not have JavaScript enabled when it crawls - therefore it was missing the navigation pane.

It doesn't work very well for non-JS browsers, but is better than not being shown at all.


dbreseman (approved):
LGTM!

MarkvanMents:
Thanks Dana - merging.

PR #7891, opened 2024-06-14
Update system-requirements.md

ghost:
Added change to highlight that we support local installations of windows, or local virtualised versions of windows. (Therefore excluding app virtualisation or other virtualisation methods from being officially supported)

### DIFF
--- a/content/en/docs/refguide/installation/system-requirements.md
+++ b/content/en/docs/refguide/installation/system-requirements.md
@@ -17,6 +17,7 @@ Mendix Studio Pro has the requirements listed in the following sections.
 ### 2.1 Software Specifications {#software}
 
 [Mendix Studio Pro](/refguide/modeling/) 10 is supported on 64-bit versions of Windows 10 release 1809 and above; this also includes Windows 11. 
+We support running on Windows locally, or locally virtualized.
 
 Studio Pro 10.7 and above can run in beta mode on an ARM Mac and Intel Mac running macOS Sonoma 14.0 and above. For information on known limitation, see the [Known Limitations for Mac](#mac-limitations) section below. 
 


content/en/docs/refguide/installation/system-requirements.md
```diff
@@ -17,6 +17,7 @@ Mendix Studio Pro has the requirements listed in the following sections.
 ### 2.1 Software Specifications {#software}
 
 [Mendix Studio Pro](/refguide/modeling/) 10 is supported on 64-bit versions of Windows 10 release 1809 and above; this also includes Windows 11. 
+We support running on Windows locally, or locally virtualized.
```
MarkvanMents, line 20: Hi @mauritselzinga-mx 
I have a couple of questions:

1. Will general readers understand what we mean here?
2. Should we be saying that we _only_ these forms of Windows?
3. Do we not support running on services like [Amazon Workspaces](https://docs.aws.amazon.com/workspaces/latest/adminguide/amazon-workspaces.html)?
4. Will all local virtualization methods work?
5. Should we also give an example of a Windows platform that is not supported.

ghost: 1. That is always the complexity here. it's mostly an addendum to the point above
2. I would not be so restrictive, since it may work, we just can't guarantee how it will work. E.g.: you can spin up an AWS EC2 instance with Studio Pro that works perfectly, it's just very slow. (remotely virtualized)
3. We don't test for this, so cannot support it. 
4. They should
5. No, since this is an extremely long list.

MarkvanMents: OK - I'll just leave it as it is.
Thanks for the PR. I'm approving and merging.

MarkvanMents (approved):
Thanks for the PR Maurits, and for taking the time to respond to my questions.
I'll leave it as it is and approve and merge immediately.

PR #7918, opened 2024-06-20
Indicate we do not support alias with spaces in OQL

passalaqua:


MarkvanMents (approved):
Thanks for the PR @passalaqua - looks good and am approving.
Has this been approved by the development team and can it be merged immediately?

Seems a very strange thing to mention - almost encouraging the use of spaces within aliases. But at least we have caught it and put it right now.

Thanks.

dahfjkg:
Hi @MarkvanMents, now it is approved by the development team, so let's merge

PR #7935, opened 2024-06-25
Add ML Kit QA

Yiyun333:


Yiyun333:
Add upon request from Esteban.

PR #7946, opened 2024-06-26
Update scheduled-events with information about session object

jveldhuizen:
This change should also be applied to the Mendix 9 documentation.

### DIFF
--- a/content/en/docs/refguide/modeling/resources/scheduled-events.md
+++ b/content/en/docs/refguide/modeling/resources/scheduled-events.md
@@ -181,6 +181,6 @@ Hour- and minute-based intervals can only be integer divisors of 24 or 60, respe
 
 Every time a scheduled event is run it produces an entry in the `System.ProcessedQueueTask` table in the database. Over time these accumulate and the table can grow large. Refer to the documentation on [Cleaning Up Old Processed Tasks](/refguide/task-queue/#cleanup) in *Task Queue* to learn how to remove processed entries.
 
-### 5.5 One session for all scheduled events
+### 5.5 One Session for All Scheduled Events
 
-Each runtime node has one specific session in memory which is used for all scheduled events. Changes to this session are visible for all scheduled events on that node. Things like changing the time zone via a Java action in one scheduled event can lead to unexpected behavior in other scheduled events. It is therefore strongly discouraged to change the session for scheduled events.
+Each runtime node has one specific session in memory which is used for all scheduled events. Changes to this session are visible for all scheduled events on that node. Things like changing the time zone via a Java action in one scheduled event can lead to unexpected behavior in other scheduled events. You are therefore strongly discouraged from changing the session for scheduled events.
--- a/content/en/docs/refguide/modeling/resources/scheduled-events.md
+++ b/content/en/docs/refguide/modeling/resources/scheduled-events.md
@@ -183,4 +183,4 @@ Every time a scheduled event is run it produces an entry in the `System.Processe
 
 ### 5.5 One Session for All Scheduled Events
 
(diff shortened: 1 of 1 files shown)


jveldhuizen:
Approval needed from Runtime Core.

hugomendix (approved):
LGTM

PS: If we decide that we want to remedy this in the next major version, we may want to announce that here as well.

content/en/docs/refguide/modeling/resources/scheduled-events.md
```diff
@@ -180,3 +180,7 @@ Hour- and minute-based intervals can only be integer divisors of 24 or 60, respe
 ### 5.4 Cleaning Up Completed Runs of Scheduled Events
 
 Every time a scheduled event is run it produces an entry in the `System.ProcessedQueueTask` table in the database. Over time these accumulate and the table can grow large. Refer to the documentation on [Cleaning Up Old Processed Tasks](/refguide/task-queue/#cleanup) in *Task Queue* to learn how to remove processed entries.
+
+### 5.5 One Session for All Scheduled Events
+
+Each runtime node has one specific session in memory which is used for all scheduled events. Changes to this session are visible for all scheduled events on that node. Things like changing the time zone via a Java action in one scheduled event can lead to unexpected behavior in other scheduled events. You are therefore strongly discouraged from changing the session for scheduled events.
```
MarkvanMents: Hi @jveldhuizen 
I'm not sure about the last sentence. It implies that you can change the session, but I don't believe that it is possible to change that. I think we are talking about changing the session _environment_ (timezone, language, time). I'm not sure how best to say this? Perhaps
```suggestion
Each runtime node has one specific session in memory which is used for all scheduled events. Changes to this session are visible for all scheduled events on that node. Things like changing the time zone via a Java action in one scheduled event can lead to unexpected behavior in other scheduled events. You are therefore strongly discouraged from changing the global session environment for scheduled events.
```

@NielsKunst - I will copy this into the Mendix 9 documentation once this language has been agreed.

MarkvanMents (changes requested):
Hi @jveldhuizen 
Jonathan, I have a small suggestion to clarify what we are talking about. I think the current description could be misinterpreted.

content/en/docs/refguide/modeling/resources/scheduled-events.md
jveldhuizen: You cannot change the session of scheduled events to another session. That's true. But you can change the session object itself; you can change its timezone field, see the method `setTimeZone` on https://apidocs.rnd.mendix.com/10/runtime/com/mendix/systemwideinterfaces/core/ISession.html. If you think that that's more clear with talking about _session environment_, that's fine.

MarkvanMents: > You cannot change the session of scheduled events to another session. That's true. But you can change the session object itself; you can change its timezone field, see the method `setTimeZone` on https://apidocs.rnd.mendix.com/10/runtime/com/mendix/systemwideinterfaces/core/ISession.html. If you think that that's more clear with talking about _session environment_, that's fine.

Ah, I see what you mean. How about I say
```suggestion
Each runtime node has one specific session in memory which is used for all scheduled events. Changes to this session are visible for all scheduled events on that node. Things like changing the time zone via a Java action in one scheduled event can lead to unexpected behavior in other scheduled events. You are therefore strongly discouraged from changing the session object for scheduled events.
```

jveldhuizen: Yes, that's a nice addition.

MarkvanMents (approved):
@jveldhuizen 
Looks good, I assume I can merge this immediately? I'll do that unless I hear from you today.

PR #7956, opened 2024-06-27
Add link to the React doc page

ancajuravle:
Adding also a link to the React docs, since the other one points only to the release notes since it became beta.


ConnorLand (approved):

ConnorLand:
Thanks @ancajuravle , nice improvement.

ConnorLand (approved):

ancajuravle:
Thanks Connor, I will keep in mind the correct link(s) path, I thought that would have been enough, since they were on the same tree level.

PR #7959, opened 2024-06-27
Remove fix note for stable sbom generation which is not actually fixed

emirmx:
The fix for stable sbom generation does not work consistently. Removing the fix note.


MarkvanMents (approved):
Looks good, merging.
Thanks for the PR.

PR #7966, opened 2024-06-28
Update 404 page

dbreseman:


layouts/partials/navbar.html
```diff
@@ -1,66 +1,3 @@
-<!-- NK - commented out to keep navbar from appearing anywhere, 
+<!-- DB - deleted content (previously commented out) to keep navbar from appearing anywhere,
```
MarkvanMents: I wonder if it would be better to leave the content as this is a copy of a layout from Docsy and it makes it easier to see if something is changed in a new version of Docsy?

Let me know when if you want me to review this PR fully.

dbreseman:
Hi @MarkvanMents, ready for your review now! I added in some more structure on the 404.html page so that navbar.html could be left alone. The styling and sizing on different screens isn't exactly the same as PDS, but it's roughly similar. Let me know if anything jumps out that needs adjusting, or if I should coordinate with PDS to get an exact match.

MarkvanMents:
> Hi @MarkvanMents, ready for your review now! I added in some more structure on the 404.html page so that navbar.html could be left alone. The styling and sizing on different screens isn't exactly the same as PDS, but it's roughly similar. Let me know if anything jumps out that needs adjusting, or if I should coordinate with PDS to get an exact match.

Hi @dbreseman, thanks for doing this.

Text is perfect. The image is still a bit smaller than the sprintr versions, and considerably higher up the page. And the **Go Back Home** button is slightly narrower.

I think it is OK, but perhaps get a designer to look at it. If they are happy, then let's merge?

MarkvanMents (approved):
I'll approve this, but perhaps a designer should check it as well?

PR #7967, opened 2024-06-28
Better explanation of why not to use ID as your unique key.

MarkvanMents:
Customers were confused over the description of why you shouldn't use the object ID as the unique key in a published OData entity.
This provides a more business-oriented reason,


MarkvanMents:
Agreed with Clyde - publishing.

PR #7970, opened 2024-07-01
Update React Client

jdiehl:
Clarified that deprecation warnings are shown
@ConnorLand


MarkvanMents (approved):
Hi @jdiehl 
@ConnorLand is away for a couple of days but this change looks simple so I will approve and merge immediately.

PR #7974, opened 2024-07-02
Release manual trigger, new steps, and mini survey

dbreseman:


content/en/docs/releasenotes/deployment/mendix-cloud.md
```diff
@@ -16,6 +16,16 @@ For information on the current status of deployment to Mendix Cloud and any plan
 
 ## 2024
 
+### July 4, 2024
+
+#### Pipelines Improvements
+
+* We added support for two new pipeline steps:
```
Sakshi-Dhakad: I'd just say - "We added two new pipeline steps" -- instead of "support for". We are building it, not supporting steps for example from 3rd party vendors.

content/en/docs/releasenotes/deployment/mendix-cloud.md
```diff
@@ -16,6 +16,16 @@ For information on the current status of deployment to Mendix Cloud and any plan
 
 ## 2024
 
+### July 4, 2024
+
+#### Pipelines Improvements
```
Sakshi-Dhakad: These should be under the label "New Features"

content/en/docs/releasenotes/deployment/mendix-cloud.md
```diff
@@ -16,6 +16,16 @@ For information on the current status of deployment to Mendix Cloud and any plan
...
+
+* We added support for two new pipeline steps:
+    * Use the Start Environment step to start a selected environment.
+    * Use the Promote Package step to move a deployment package from a specified source environment to a specified target environment.
+* It is now possible to trigger pipelines manually. To use this new trigger in your pipeline, select **Manual** as the **Trigger** in the Start Pipeline step of your pipeline design.
+* We will be launching a mini survey for Pipelines soon, asking you some questions about what you like and dislike.
```
Sakshi-Dhakad, line 27: Create a category for Pipeline Improvements and add the following to it:
1. We have removed the ability to edit pipeline name if it's active to avoid confusion. Pipelines names can now be changed after 
2. Deleting a pipeline also deletes the associated pipeline runs. Before deleting a pipeline design, you will be explicitly alerted of this behavior to ensure you are aware.
3. The underlying Mendix Studio Pro version for Pipelines capability has been upgraded to the latest LTS version.
4. Deep links have been added to runs, designs and settings page for easy sharing of record specific details.

Sakshi-Dhakad (commented):
Added some comments

Also @dbreseman - The PipelineId which has been added as a hidden field under Runs tab is the id of the pipeline definition. Is it possible to define it somewhere? I want to make sure users don't get confused with the run id.

content/en/docs/releasenotes/deployment/mendix-cloud.md
dbreseman: @Sakshi-Dhakad my feedback is below:
1. I think there's a typo here—could you clarify? When can pipeline names be changed?
2. The Docs already have a warning alert with this text: "If you delete a pipeline design in the Designs tab, any run history associated with that pipeline design is deleted from the Runs tab." If I understand correctly, this behavior isn't changing, but there's now a new pop-up that makes this harder to miss in the UI. I generally wouldn't include this new pop-up in the release notes because it just reinforces pre-existing functionality in a way that should be self-explanatory, but I can if you think it's important to emphasize to users.
3. Will the switch to the LTS version be apparent to users when they use Pipelines? Otherwise, it's okay to omit this.
4. Thanks for flagging this. I'll make sure it's included.

Sakshi-Dhakad:
1.  Pipeline name can be changed when a pipeline is in inactive state. Hope that clarifies.
  2.  Ok
  3.  Nothing super noticeable, small things like better performance, better UI -- similar to what happens when you upgrade to a new version of apple.
  4.  Ok

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

dbreseman:
@Sakshi-Dhakad Updated now to integrate your feedback: https://github.com/mendix/docs/pull/7974/files. Please let me know if you see anything that needs further adjustments.

PR #7992, opened 2024-07-04
Create Keyboard Shortcuts doc

quinntracy:


MariaShaposhnikova:
It looks good, but do we need to make a child page of the Studio Pro Overview as it used to be a part of it and they are related?

PR #7997, opened 2024-07-05
Kk pmp relnotes

katarzyna-koltun-mx:


katarzyna-koltun-mx:
This PR must be approved by the PMP team. I've asked them to comment in the docs slack channel when the PR is OK to publish.

PR #8003, opened 2024-07-08
Add missing log nodes

grootjans:
Added some comments on the function of a few log nodes that did not have a definition.


MarkvanMents (approved):
Thanks for the PR Robbert Jan.
Approving and publishing after some very minor proofreading.

PR #8021, opened 2024-07-11
[WTF-1862] Handle expression variables with long paths in the linked action property 

weirdwater:
Documentation update to reflect our changes to the behavior of expression arguments for linked Microflow and Nanoflow actions. They now support long paths, allowing Mendix developers to supply argument values over associations. For widget developers this change means that ActionValues may state that they can be executed while they are still waiting for all arguments to be loaded.

- [x] Update Pluggable Widget API ActionValue documentation
- [x] Search for and update Modeling documentation stating that associations cannot be used for supplying Action Expression Argument values.
  - Descriptions of relevant actions refer to the page [On Click Event and Events Section](https://docs.mendix.com/refguide/on-click-event/#3322-microflow-arguments) which does not mention any restriction on long paths.
- [x] Will add a comment whether this is for 10.13 or 10.14.
  - This will be for 10.14

### DIFF
--- a/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/_index.md
+++ b/content/en/docs/apidocs-mxsdk/apidocs/pluggable-widgets/pluggable-widgets-client-apis/_index.md
@@ -56,11 +56,21 @@ export interface ActionValue {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


weirdwater:
Ready for technical review by the team

ConnorLand (approved):

PR #8046, opened 2024-07-16
Add troubleshooting hint for blank page to PDF Document Generation module docs

allard-mx:


content/en/docs/appstore/use-content/modules/document-generation.md
```diff
@@ -413,25 +413,27 @@ If you encounter a `DocGenRuntimeException` exception with a message that is not
...
 
 In case of issues regarding styling, Mendix recommends temporarily adding the page microflow to your app navigation (for details, see step 2 in the [Module Usage and Runtime Issues](#module-usage-runtime-issues) section). This allows you to preview the page in your browser and inspect the applied styles. Mendix recommends using Chrome or Chromium and the [Chrome DevTools](https://developer.chrome.com/docs/devtools/css/) for this, since Chromium is the browser that is used by the document generation service.
 
-In case the resulting PDF document only contains a part of the expected content, verify that the layout used for the page does not include a scroll container. Layouts that do include a scroll container, such as **Atlas_Default**, will not work properly.
+If the resulting PDF document only contains a part of the expected content, verify that the layout used for the page does not include a scroll container. Layouts that do include a scroll container, such as **Atlas_Default**, will not work properly.
+
+If the resulting PDF document contains an additional blank page at the end of the document, verify that the page contains an outer layout grid.
```
MarkvanMents: @allard-mx - Are you saying that the page must always contain an outer layout grid if you are generating a PDF document? Perhaps we should be more clear in section 4 when we say:

> Using the DocumentGeneration_Default layout is not required. You can use any layout of layout type Responsive, or create your own layout, as long as the layout does not include a scroll container. Layouts that do include a scroll container, such as Atlas_Default, will not work properly. 

and add "but there must be a layout to generate the PDF correctly"?

MarkvanMents (changes requested):
Hi @allard-mx 
Can you have a look at my question and let me know if there is more we can say about layouts in section 4 to try and avoid this situation in the first place.

MarkvanMents:
@Luyao-Zhang-1 
I didn't get a response from Allard. Could you remind him, if you have time, otherwise I can pick this up when I'm back.

content/en/docs/appstore/use-content/modules/document-generation.md
allard-mx: The layout grid is not required in all scenarios, but could help when running into the empty page issue. It depends on the exact page setup (in terms of widgets that are used in combination with the applied styling). I've pushed another commit to reflect that the layout grid could offer a potential solution, but is not a hard requirement.

PR #8050, opened 2024-07-16
Update extension docs

ghost:
Update extension docs to point to .NET 8 else the nuget can't install. I think this is the only one that was missed out


MarkvanMents (approved):
Looks good to me. Approving and merging.

PR #8057, opened 2024-07-17
Add a basic description of parameters when using the custom connection type in database replication

passalaqua:
The custom connection option in database replication has no documentation. This PR adds some basic information.

### DIFF
--- a/content/en/docs/appstore/use-content/modules/database-replication.md
+++ b/content/en/docs/appstore/use-content/modules/database-replication.md
@@ -44,21 +44,36 @@ Once a database and mapping have been defined, you can use the **importByMapping
 
 You can create new database connections from the **Databases** page. Each connection has a few basic properties. After creating a database connection, you should synchronize the table information using the button on the overview. This makes sure the connection information is correct, and that the replication module has all the table information from your database.
 
-#### 2.1.1 Type
+On this page you must select the database type before filling in the connection details. Available database parameters vary depending on database type.
 
-This is the type of database you want to import. Available database parameters vary per database type.
+#### 2.1.1 Common Parameters
 
-#### 2.1.2 Host and Port
+The following parameters are available to most database types:
+* Host
+* Port
+* Username
+* Password
 
-This is the host name and port number of the server on which your database is running.
+Depending on the database, some of the following parameters may be available:
+* Database name
+* Instance name
+* Service name (exclusive to Oracle)
+* SID (exclusive to Oracle)
+* Use Integrated Security (e.g. to connect to SQL Server using the your Windows user credentials)
 
-#### 2.1.3 Name and Instance Name
+#### 2.1.2 Custom Connection Parameters
 
-This is the name and instance name of the database to connect to. The **Instance name** is optional.
+If your database type is not listed, it may still be possible to connect using a JDBC compatible driver and the **Custom** connection option. You need to obtain the drivers and place them inside the `userlib` directory and then configure the connection parameters:
 
-#### 2.1.4 Username and Password
-
-These are the credentials necessary to sign in to the database.
+* Username and Password: The credentials given to the driver to connect to the database server.
+* Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a PostgreSQL database).
+* Open/Close escape character: Characters used to quote table and column names (e.g. `"` in PostgreSQL or `[` and `]` in SQL Server).
+* Driver class: The full name of the driver to be loaded (e.g. `org.postgresql.Driver` for PostgreSQL).
+* Escape column names: Indicates if column names must be quoted using the escape characters when querying the database.
+* Escape table names: Indicates if table names must be quoted using the escape characters when querying the database.
+* Escape the schema and table name separately: Indicates if we should quote schema and table names together (e.g. `"public_schema.table_name"`) or separately (e.g. `"public_schema"."table_name"`).
+* Allow AS token for tables: Indicates if `AS` is necessary when aliasing table names (e.g. `SELECT * FROM table AS A` vs `SELECT * FROM table A`).
+* Close connection after query: Indicates if the connection to the database must be closed after every query.
 
 ### 2.2 Mapping Properties
 
--- a/content/en/docs/appstore/use-content/modules/database-replication.md
+++ b/content/en/docs/appstore/use-content/modules/database-replication.md
@@ -49,30 +49,32 @@ On this page you must select the database type before filling in the connection 
 #### 2.1.1 Common Parameters
 
 The following parameters are available to most database types:
+
 * Host
 * Port
 * Username
 * Password
 
 Depending on the database, some of the following parameters may be available:
+
 * Database name
 * Instance name
 * Service name (exclusive to Oracle)
 * SID (exclusive to Oracle)
-* Use Integrated Security (e.g. to connect to SQL Server using the your Windows user credentials)
+* Use Integrated Security (for example, to connect to SQL Server using the your Windows user credentials)
 
 #### 2.1.2 Custom Connection Parameters
 
 If your database type is not listed, it may still be possible to connect using a JDBC compatible driver and the **Custom** connection option. You need to obtain the drivers and place them inside the `userlib` directory and then configure the connection parameters:
 
 * Username and Password: The credentials given to the driver to connect to the database server.
-* Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a PostgreSQL database).
-* Open/Close escape character: Characters used to quote table and column names (e.g. `"` in PostgreSQL or `[` and `]` in SQL Server).
-* Driver class: The full name of the driver to be loaded (e.g. `org.postgresql.Driver` for PostgreSQL).
+* Connection string: A fully formed JDBC URL string (for example, `jdbc:postgresql://localhost:5432/database` for a PostgreSQL database).
+* Open/Close escape character: Characters used to quote table and column names (for example, `"` in PostgreSQL or `[` and `]` in SQL Server).
+* Driver class: The full name of the driver to be loaded (for example, `org.postgresql.Driver` for PostgreSQL).
 * Escape column names: Indicates if column names must be quoted using the escape characters when querying the database.
 * Escape table names: Indicates if table names must be quoted using the escape characters when querying the database.
-* Escape the schema and table name separately: Indicates if we should quote schema and table names together (e.g. `"public_schema.table_name"`) or separately (e.g. `"public_schema"."table_name"`).
-* Allow AS token for tables: Indicates if `AS` is necessary when aliasing table names (e.g. `SELECT * FROM table AS A` vs `SELECT * FROM table A`).
+* Escape the schema and table name separately: Indicates if we should quote schema and table names together (for example, `"public_schema.table_name"`) or separately (for example, `"public_schema"."table_name"`).
+* Allow AS token for tables: Indicates if `AS` is necessary when aliasing table names (for example, `SELECT * FROM table AS A` vs `SELECT * FROM table A`).
 * Close connection after query: Indicates if the connection to the database must be closed after every query.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+- Service name
+- SID
+- Use Integrated Security
+
+#### 2.1.2 Custom Connection
+
+If your database type is not listed, it may still be possible to connect using a JDBC compatible driver and the **Custom** connection option. You need to obtain the drivers and place them inside the `userlib` directory and then configure the connection parameters:
+
+- Username and Password: The credentials given to the driver to connect to the database server;
+- Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a Postgres databas);
```
dahfjkg: *database

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
-#### 2.1.3 Name and Instance Name
-
-This is the name and instance name of the database to connect to. The **Instance name** is optional.
-
-#### 2.1.4 Username and Password
-
-These are the credentials necessary to sign in to the database.
+In this page you must select the database type before filling in the connection details. Available database parameters vary per database type.
+
+#### 2.1.1 Common Parameters
```
dahfjkg, line 49: I suspect that your goal is not to list available databases, but I would still create a short comment for every field.
Especially the following ones:
- Service name and SID — I'd mention that these are only for Oracle and that one excludes the other
- Use Integrated Security — this one is only for SQL Server

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+
+- Username and Password: The credentials given to the driver to connect to the database server;
+- Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a Postgres databas);
+- Open/Close escape character: Characters used to quote table and column names (e.g. `"` in Postgres or `[` and `]` in SQL Server);
+- Driver class: The full name of the driver to be loaded (e.g. `org.postgresql.Driver` for Postgres);
+- Escape column names: Indicates if column names must be quoted using the escape characters when querying the database;
+- Escape table names: Indicates if table names must be quoted using the escape characters when querying the database;
+- Escape the schema and table name separately: Indicates if we should quote schema and table names together (e.g. `"public_schema.table_name"`) or separately (e.g. `"public_schema"."table_name"`);
+- Allow AS token for tables: Indicates if `AS` is necessary when aliasing table names (e.g. `SELECT * FROM table AS A` vs `SELECT * FROM table A`);
+- Close connection after query: Indicates if the connection to the database must be closed after every query.
```
dahfjkg: In the list in 2.1.1 there is nothing in the end of the string. In this list you use `;` and `.` — perhaps those can be dropped as well

dahfjkg (changes requested):

content/en/docs/appstore/use-content/modules/database-replication.md
MarkvanMents: I agree, if these are common parameters, then it shouldn't depend on the database type as to which ones become available (see below).
If some of these are specific to specific cases (even if that case is "not Oracle" then I think it should be in a separate section).

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
-This is the host name and port number of the server on which your database is running.
-
-#### 2.1.3 Name and Instance Name
-
-This is the name and instance name of the database to connect to. The **Instance name** is optional.
-
-#### 2.1.4 Username and Password
-
-These are the credentials necessary to sign in to the database.
+In this page you must select the database type before filling in the connection details. Available database parameters vary per database type.
```
MarkvanMents: Proofread
```suggestion
On this page you must select the database type before filling in the connection details. Available database parameters vary depending on database type.
```

content/en/docs/appstore/use-content/modules/database-replication.md
MarkvanMents:
```suggestion
- Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a PostgreSQL database);
```

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+- Instance name
+- Service name
+- SID
+- Use Integrated Security
+
+#### 2.1.2 Custom Connection
+
+If your database type is not listed, it may still be possible to connect using a JDBC compatible driver and the **Custom** connection option. You need to obtain the drivers and place them inside the `userlib` directory and then configure the connection parameters:
+
+- Username and Password: The credentials given to the driver to connect to the database server;
```
MarkvanMents: Should we mention here not to hard code them in the app? Here is a good link: https://docs.mendix.com/refguide/dev-best-practices/#25-passwords-and-other-secrets (although I'll need to set up a permanent anchor).

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+- Connection string: A fully formed JDBC URL string (e.g. `jdbc:postgresql://localhost:5432/database` for a Postgres databas);
+- Open/Close escape character: Characters used to quote table and column names (e.g. `"` in Postgres or `[` and `]` in SQL Server);
```
MarkvanMents: I think the name of the product in PostgreSQL, so we should say:
```suggestion
- Open/Close escape character: Characters used to quote table and column names (e.g. `"` in PostgreSQL or `[` and `]` in SQL Server);
```

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,34 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+- Driver class: The full name of the driver to be loaded (e.g. `org.postgresql.Driver` for Postgres);
```
MarkvanMents:
```suggestion
- Driver class: The full name of the driver to be loaded (e.g. `org.postgresql.Driver` for PostgreSQL);
```

content/en/docs/appstore/use-content/modules/database-replication.md
MarkvanMents: ➕

MarkvanMents (commented):
Although I haven't used these settings the documentation looks clear to me.
I've made a couple of suggestions.
I will still do another proofread once you have responded (for example, we use `*` for unordered lists) but looking good so far.

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: If anything, I'd add a `;` and `.` to the other list (following what seems to be the AP Stylebook). I couldn't find any mention of which style to use in our documentation, though.

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: I tried adding a short description but all I got was noise. They are either self explanatory while reading the list or only relevant for the database types they appear on (and in which case it should be clear to the user what it is).

The only possible exception is `Use Integrated Security`, which seems to also be available for Informix as well.

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: Actually, it does not look like it's AP's, but someone else's.

MarkvanMents: There is something in Confluence - but it isn't clear. I'll update it.

Unordered lists should either not have any punctuation at the end, or should have a full stop if each item completes a sentence which introduces the list.

In this case, I would probably put a full stop at the end (although this is one thing which I am never 100% sure about😀).

dahfjkg: Usually a label should have a full stop at the end if there are multiple sentences. If it has one sentence, no punctuation is required. For lists, I tend to follow the same rule.

MarkvanMents: Oh, I've found the styling for bullets - I'll Slack you a link.

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: There is no hard coding here. This page is filled after the app is deployed and the credentials are stored in the database (which is just as bad, IMO, but that's how it currently is implemented).

MarkvanMents: OK - then the security is managed for them. I'll resolve this one.

content/en/docs/appstore/use-content/modules/database-replication.md
MarkvanMents: I agree there is no point in repeating the in-app explanation. But perhaps a short note that says "Oracle databases only" or something like that?

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: I replaced them with periods in this list as there is a full description after each entry. The other lists I left period-less.

dahfjkg:
No more comments from my side, so technical verification is complete

content/en/docs/appstore/use-content/modules/database-replication.md
```diff
@@ -44,21 +44,38 @@ Once a database and mapping have been defined, you can use the **importByMapping
...
+* Connection string: A fully formed JDBC URL string (for example, `jdbc:postgresql://localhost:5432/database` for a PostgreSQL database).
+* Open/Close escape character: Characters used to quote table and column names (for example, `"` in PostgreSQL or `[` and `]` in SQL Server).
+* Driver class: The full name of the driver to be loaded (for example, `org.postgresql.Driver` for PostgreSQL).
+* Escape column names: Indicates if column names must be quoted using the escape characters when querying the database.
+* Escape table names: Indicates if table names must be quoted using the escape characters when querying the database.
+* Escape the schema and table name separately: Indicates if we should quote schema and table names together (for example, `"public_schema.table_name"`) or separately (for example, `"public_schema"."table_name"`).
+* Allow AS token for tables: Indicates if `AS` is necessary when aliasing table names (for example, `SELECT * FROM table AS A` vs `SELECT * FROM table A`).
+* Close connection after query: Indicates if the connection to the database must be closed after every query.
```
MarkvanMents: Should we indicate that these are Booleans. We imply that the other parameters are strings by indicating what goes into them.

MarkvanMents (changes requested):
Hi Augusto,
Looks good. I have one further question - if you can let me know what you think, we can resolve it and publish.
Thanks for creating this PR - I think it will help our customers a lot.

content/en/docs/appstore/use-content/modules/database-replication.md
passalaqua: The boolean options are displayed as checkboxes, so I don't think (within their context) that this clarification is necessary.

MarkvanMents (approved):
Thanks for this PR.
Looks good to me.
Approving.

content/en/docs/appstore/use-content/modules/database-replication.md
MarkvanMents: OK - we've discussed. I'll resolve this.

PR #8080, opened 2024-07-23
Add folder naming convention info

quinntracy:


content/en/docs/refguide/modeling/dev-best-practices.md
```diff
@@ -103,15 +103,23 @@ With models built in lower versions of Mendix, however, you will need to manuall
 
 ### 3.3 Folders
 
+#### Folder Names
```
quinntracy, line 106: Are there any symbols that cannot be used to name a folder?

MariaShaposhnikova: Probably depends on the OS, but I would verify it with a dev.

quinntracy: Confirmed by dev that there are no character limitations

PR #8095, opened 2024-07-25
Genaicommons embeddings, knowledgebase & image generation

tinaengel-mx:
Contains changes in
- GenAI Commons
- PgVector Knowledge Base
- OpenAI Connector


katarzyna-koltun-mx:
@tinaengel There are a few Thumbs.db files here, I assume they should be deleted, can you confirm?

PR #8122, opened 2024-08-01
Update _index.md

NielsKunst:
Clarify that a runtime license per environment is required. Customers tend to think that a runtime license per app is needed, while an app has several environments and each environment needs one.


katarzyna-koltun-mx:
Edited and reached out to the PC team for approval.

PR #8132, opened 2024-08-05
Update notif-implement-pwa.md

ThijsSeppen:
adding links to document


ConnorLand:
This is a really nice improvement idea @ThijsSeppen . However, we have to implement anchor links a little differently on the docs site by manually adding an anchor hash in a header and then linking to it. 

I'll make the changes you intended and then link my commit here to show how it works.

ConnorLand:
Implemented here in versions 10 and 9: https://github.com/mendix/docs/commit/a793c8d7131f74d465669327b36ec8f691c9947f

PR #8133, opened 2024-08-05
Update setting-up-google-firebase-cloud-messaging-server.md

ThijsSeppen:
added context that 3.5 is only needed for native apps


ConnorLand:
@ThijsSeppen , I Implemented this change [here](https://github.com/mendix/docs/commit/385aa0a6ac687839fb92f8c041df8a90fd34a27a) across all docs that required it.

PR #8155, opened 2024-08-07
Add 9.24.26 release notes

MariaShaposhnikova:


MariaShaposhnikova:
The date needs to be updated, after that it is good to go.

ConnorLand (approved):

PR #8191, opened 2024-08-16
Support for Azure and pgvector

liamsommer-mx:
We added to our existing page an instruction to setup an Azure PostgreSQL server to use a vector database and connect to your Mendix application.


content/en/docs/appstore/use-content/modules/genai/pg-vector-knowledge-base/vector-database-setup.md
```diff
@@ -70,10 +72,49 @@ You can use the values in the steps below for experimental purposes:
...
+4. Continue with the **Networking** configurations in the next tab.
+   1. Based on your requirements, decide how the database server can be accessed (for testing purposes, it is recommended to use *Public Access*):
+      1. **Public access**: firewall rules need to be added for the IP addresses that are allowed to access the server. Use **Add current client IP address** to add your own IP when running the application locally. For apps running in the Mendix cloud, add the IP of that environment, see [Mendix IP Addresses: Outgoing IP](/developerportal/deploy/mendix-ip-addresses/#outgoing) for a list of addresses to safe-list in this scenario. Alternatively, you can use **Add 0.0.0.0 - 255.255.255.255** so that no IP addresses are blocked. Use this carefully and make sure that this aligns with your security requirements.
+      2. **Private Access**: the server can only be accessed from a **Virtual Network** that needs to be selected (or created). Make sure that your Mendix App is running in the same network.
+5. It is not necessary to configure anything in the **Security** or **Tags** tabs to get the server running.
```
Luyao-Zhang-1: @liamsommer-mx Should the users always skip these two tabs?

liamsommer-mx: @Luyao-Zhang-1 for testing yes. But I rather don't want to recommend to skip those in general (because we don't the requirements for the specific use case). Same for public vs private access.

PR #8196, opened 2024-08-16
Update certificates.md

NielsKunst:
Removed reference to support to prevent incorrect expectations


NielsKunst:
Related discussion: https://mendix.slack.com/archives/CA82XPUQG/p1723817490129769

PR #8199, opened 2024-08-19
Autonumber headings

dbreseman:


MarkvanMents (approved):
As discussed, this looks great. I haven't checked every file, but a quick check looks like we have caught everything, and if a couple of files look odd we can always pick those up when we find them.

Can you present it at the group meeting so that people don't get shocked and can look into their own PRs to ensure that nothing gets broken when they are merged.

PR #8204, opened 2024-08-20
Update database versions in system requirements

basdebakker:
Adds recent versions of supported databases to the system requirements.

### DIFF
--- a/content/en/docs/refguide/installation/system-requirements.md
+++ b/content/en/docs/refguide/installation/system-requirements.md
@@ -223,13 +223,13 @@ Mendix tries to support the most recent and patched database server versions fro
 
 Current support:
 
-* [MariaDB](/refguide/mysql/): 10.4, 10.5, 10.6, 10.11
+* [MariaDB](/refguide/mysql/): 10.4, 10.5, 10.6, 10.11, 11.4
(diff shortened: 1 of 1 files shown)


MarkvanMents:
@basdebakker 
Are all these versions supported by all Mendix 10 versions, or do some versions need a minor version to be supported.
Also, is there a reason that the HANA version has changed completely and that we don't support two different versions?
Thanks

basdebakker:
Ah, what does "supported" mean exactly? We haven't gone back and tested all Mendix 10 releases with those new database versions. But we also haven't made any changes to support them, they just worked, so we don't expect any issues. But if there turns out to be an issue with some older Mendix 10 release, we may just ask you to upgrade Mendix to a recent MTS patch version (like we may for any issue, even if not database related).

As to SAP HANA, they only have docker containers for the latest version, so that is all we can test on. If you report an issue with an earlier version and we cannot reproduce it with the latest one, we may just ask you to upgrade SAP HANA. So while we'll try to help you in any case, we can't really promise support for different versions.

MarkvanMents (approved):
Thanks for the comprehensive answer to my question.
I think, in this case since there are no changes in the product and we expect it to just work, that the best thing is just to implement what you have submitted.
Approving and merging.

PR #8206, opened 2024-08-20
Enable Chroma syntax highlighting

dbreseman:


content/en/docs/community-tools/contribute-to-mendix-docs/markdown-shortcodes.md
```diff
@@ -65,16 +65,22 @@ This alert is indented to match the sub-list item.
 }
 ```
 
-#### HTTP Code Block with Line Numbers Disabled
```
MarkvanMents, line 68: What is the behavior now with line numbers. It sometimes seems to add them (`json {linenos=table,hl_lines=[6],linenostart=22}` adds them) but the default now seems to be _not_ to have them?
Do you now always have to add them using `{linenos=table}`?

content/en/docs/community-tools/contribute-to-mendix-docs/markdown-shortcodes.md
```diff
@@ -65,16 +65,22 @@ This alert is indented to match the sub-list item.
...
+{
+   public static void Main()
+   {
+-      System.Console.WriteLine("Hello, World!");
++      System.Console.WriteLine("Rock all night long!");
+   }
+}
 ```
 
-#### Java Code Block with Lines 4-6 Highlighted, and Line Numbering Starting at 33
```
MarkvanMents, line 75: linenostart still works - I think it is worth having the example.

config/_default/hugo.toml
```diff
@@ -49,8 +49,10 @@ disableKinds = ["taxonomy", "term"]
             unsafe = true
     [markup.highlight]
         # See a complete list of available styles at https://xyproto.github.io/splash/docs/all.html
-        style = "tango"
-        lineNos = true
+        # See config options at https://gohugo.io/getting-started/configuration-markup/#highlight
+        style = "borland" # DB - Changed default style for accessibility (see https://www.ctmartin.me/blog/2019/11/print-accessible-code/)
+        lineNos = false # DB - Turned off line numbers
```
MarkvanMents: This seems like a good change, but we should discuss it and make sure that we identify where line numbers are needed.

Didn't realize that we had the option 😀

MarkvanMents (changes requested):
Looks good, but I think we need to think a bit more about the line numbers and keep some examples for where they are needed. And communicate/discuss it with the team.

But a great find that we can now do the copy with Chroma. Thanks.

dbreseman:
Migrating some notes from Slack for transparency's sake:
- The built-in copy button works fine—when there aren't any line numbers. I imagine there's a workaround for the line number issue, but I wonder how much we actually want line numbers? With Chroma, line highlighting and diffs work very well (pictured below), and that might be a clearer and more robust solution than referring to specific lines by number. Currently, it's set up with lineNos = false.
- There are several accessible themes available, and it's possible to customize them if needed. For an overview, see https://www.ctmartin.me/blog/2019/11/print-accessible-code/.
- I used guessSyntax = true to set it so code blocks without a language specification keep the same styling as the rest of the code blocks, rather than reverting to the Docsy default like Prism does.

content/en/docs/community-tools/contribute-to-mendix-docs/markdown-shortcodes.md
MarkvanMents: Can we keep an example of this syntax with the warning that enabling linenos will disable the copy button. Just in case someone really has this use case?

config/_default/hugo.toml
MarkvanMents: Can we also put something in the comment here to warn the next person that this has an impact on the copy button?
```suggestion
        lineNos = false # DB - Turned off line numbers as the default. Can turn them on individually, if required, but this causes the copy button to disappear.
```

content/en/docs/community-tools/contribute-to-mendix-docs/markdown-shortcodes.md
```diff
@@ -65,16 +65,22 @@ This alert is indented to match the sub-list item.
 }
 ```
 
-#### HTTP Code Block with Line Numbers Disabled
+#### Diff Code Block
 
-```http {linenos=false}
-GET /projects/d92064a5-b1fd-4be4-97db-53fc90201d1c/epics HTTP/1.1
-Authorization: MxToken 7LJE…vk
+```diff
```
MarkvanMents: Can we put a link to the Chroma documentation so it is easy to find the options like `diff` and `linenos=`?

dbreseman: This page is just for demos, but I'll add it to https://mendix.atlassian.net/wiki/spaces/RNDHB/pages/2735898685/Shortcodes+Markdown+and+HTML#Code-Blocks!

content/en/docs/community-tools/contribute-to-mendix-docs/markdown-shortcodes.md
dbreseman: This page is intended to visualize common use cases—maybe it's sufficient to mention that over on Confluence (https://mendix.atlassian.net/wiki/spaces/RNDHB/pages/2735898685/Shortcodes+Markdown+and+HTML#Code-Blocks) instead?

PR #8216, opened 2024-08-22
Added warning for commandline clones

tknippenberg:


tknippenberg:
@blis-mendix , do you think this properly addresses the question we received?

blis-mendix:
> @blis-mendix , do you think this properly addresses the question we received?

Hey new text works!

PR #8222, opened 2024-08-22
Update combobox.md

TimvanSteenbergen:
- Renamed titles drastically increasing readability: Tab 'xyz' and Box 'abc'  
- Added Box 'Advanced'
- Better more complete description of several other elements.


ConnorLand:
Thanks for the PR @TimvanSteenbergen . I will talk to the developers about the best way to present some of this info, especially the cases that only show up in certain configurations. 

After syncing with devs, I will either merge this PR or do the changes elsewhere and show you the commit. This PR has a conflict or two, so that's why I might do the latter option.

PR #8251, opened 2024-08-29
Improve Google Search

MarkvanMents:
Google is choosing the wrong canonical because of duplicate page names.

Updating some page names to avoid duplicates.

### DIFF
--- a/content/en/docs/deployment/general/software-composition.md
+++ b/content/en/docs/deployment/general/software-composition.md
@@ -1,5 +1,5 @@
 ---
-title: "Software Composition (App)"
+title: "Software Composition"
 linktitle: "Software Composition"
 url: /developerportal/deploy/software-composition/
 description: "Describes the Software Composition page in Apps."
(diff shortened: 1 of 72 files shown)


content/en/docs/deployment/general/software-composition.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: "Software Composition"
+title: "Software Composition (App)"
```
Luyao-Zhang-1: I wonder whether it is better to call this "Software Composition (Deployment)".

MarkvanMents: @Luyao-Zhang-1 
Thanks for the review and the comment.
I chose "App" because it is about the composition of the app (and just one app, not the whole portfolio). I think that "Deployment" implies that it is only the composition of an app after it has been deployed, but it is created for all deployment packages - whether they are deployed or not.

I'm torn. Perhaps **Software Composition (Deployment Package)**? Or is that too awkward?

MarkvanMents: Decided to leave this one as was and only rename the Control Center Software Composition page.

MarkvanMents:
Approved by team over Slack.

PR #8254, opened 2024-08-30
Clarify full snapshot format

dbreseman:


dbreseman:
Based on https://mendix.slack.com/archives/C01615RD4NN/p1724926715843269

PR #8256, opened 2024-08-30
Adding Known Issues section to the web-view doc entry

pavel-constantinov:
- add Intelij editor internal files to .gitignore
- add known issues section
- remove ide config files from the repo(please recreate them locally if needed)


.travis.yml
```diff
@@ -7,7 +7,7 @@ os: linux             # optional, this is default
 dist: jammy          # or bionic | xenial | trusty | precise with xenial as default - Recommended here: https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments
 
 git:
-  depth: 20000 # We're cloning a max of xxx commits, so the author date is correct
+  depth: 20000 # We're cloning this many commits as it is roughly covering the last 4 years of commits, thus letting us show the last changed date on the documentation pages.
```
MarkvanMents, line 10: Thanks for clarifying this.

pavel-constantinov: I'm still a little sceptical about us needing this many commits on regular PR builds. I get in when we're publishing a new release, that makes sense. But how does that process work? Do you just push the build artefact to s3 every time a PR is merged? It should be possible to base the depth on the type of build we're running, I hope.

pavel-constantinov: However, if it isn't, it's not that big of a deal. As far as I recall, changing this number to 1 shaved off about a minute of build time at best.

PR #8268, opened 2024-09-02
Updated the Git version 2.43 which we install with 10.6.10+ MTS

mikita-haurykau:
Updated git version 2.43 which we install in 10.6.10+ MTS


MariaShaposhnikova:
@mikita-haurykau , the way it is written now, it reads that we install this version with SP 10.6.10 OR 10.10 and above. How about versions above 10.6.10? 
Did you mean SP 10.6.10 and above, and SP 10.10 and above? 
Does it also mean that 10.7-10.9 are not included?

ngavrikov:
> @mikita-haurykau , the way it is written now, it reads that we install this version with SP 10.6.10 OR 10.10 and above. How about versions above 10.6.10? Did you mean SP 10.6.10 and above, and SP 10.10 and above? Does it also mean that 10.7-10.9 are not included?

Exactly so 
SP 10.6.10 OR 10.10 and above
10.6 got the latest git as it is an MTS. 10.7 - 10.9 won't get this update

MariaShaposhnikova (approved):

PR #8276, opened 2024-09-04
Prepare for release of x-ray mode

MendixBart:
Update docs in preparation of releasing x-ray mode in 10.15.


ConnorLand (approved):

ConnorLand:
@MendixBart nice change! So X-ray mode is simply being removed from the Preferences menu and taken out of Beta? No additional functionality we need to document?

MendixBart:
@ConnorLand That's correct. Other than that it is mostly visual changes, the biggest being that we will be showing data sources by default, but there's no real functional changes.

PR #8306, opened 2024-09-11
Bedrockshowcasedeprecation

strategicalliances:


content/en/docs/appstore/use-content/modules/aws/amazon-bedrock.md
```diff
@@ -128,13 +130,13 @@ To build a simple microflow that uses the ChatCompletions operation to send a si
...
+### Invoking Specific Models by Using the InvokeModel Operation
 
-Most text models can be used with the **ChatCompletions** operation. For an overview of the supported models and capabilities, see [Supported models and model features](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html#conversation-inference-supported-models-features).
+Depending on your needs, you can just reuse the operations inside the **AmazonBedrockConnector (GenAICommons)** section. You can also find guidance on how to implement the required structures in the [GenAICommons](/appstore/modules/genai/) documentation. Most text models can, however, be used with the **ChatCompletions** operation. For an overview of the supported models and capabilities, see [Supported models and model features](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html#conversation-inference-supported-models-features) in the AWS Bedrock documentation.
```
MarkvanMents: @strategicalliances 
Not sure why the "however" is here? It doesn't seem relevant to the beginning of the paragraph which is about reusing existing operations.
Are we saying that most text models can be used with the ChatCompletions operation but there are some which are not?

It seems there are three things here:

1. reuse existing operations
2. use ChatCompletions if you can
3. use a specific model that doesn't support ChatCompletions

These seem to be three independent ways of doing it. Is that correct?

MarkvanMents (commented):
Looks good, but I think the InvokeModel section could be clearer - see my comment.

PR #8323, opened 2024-09-13
Update anti-virus performance tips

MxMarcel:
The description that is there is too open. It is better to advise to add more specific folders rather than the full application.


quinntracy:
@MxMarcel I made a couple small edits but otherwise can merge this PR at any time

PR #8325, opened 2024-09-16
Add schedules

feddovanede:
Add schedules so we can link to it from the Backups page.


dbreseman:
@feddovanede The Archived App and Access Logs section in this page already has an alert about this: "For apps deployed to Mendix Cloud, logs are rotated and archived every day. They are retained for six months." Would it work to just link to that existing heading?

feddovanede:
I messed that 🤦 For Backups we do have a separate paragraph with the retention schedule. If w add it to the Logs as well, the pages would align better, I think?

dbreseman:
Sure, I can shift it to match the format from backups.

PR #8344, opened 2024-09-19
Update log nodes and logging documentation

emirmx:


MarkvanMents (approved):
@emirmx 
Looks good. I have made a few small updates in https://github.com/mendix/docs/pull/8344/commits/cf0062e92523c6603930e6beca364b00830a5fd3 but I hope I haven't changed the meaning.
We don't normally updated Mx 9 documentation unless it has a high impact on the customer - but thanks for doing that as well. I will keep the changes since they are already in the PR.

Are the new log nodes available from 10.0.0 or were they introduced later? It isn't so important for log nodes, but for clarity to the customer it might be nice to mention when they start being used if they haven't been available for all of Mx10. (More important for settings which can actually change things in the runtime!),

emirmx:
> @emirmx Looks good. I have made a few small updates in [cf0062e](https://github.com/mendix/docs/commit/cf0062e92523c6603930e6beca364b00830a5fd3) but I hope I haven't changed the meaning. We don't normally updated Mx 9 documentation unless it has a high impact on the customer - but thanks for doing that as well. I will keep the changes since they are already in the PR.
> 
> Are the new log nodes available from 10.0.0 or were they introduced later? It isn't so important for log nodes, but for clarity to the customer it might be nice to mention when they start being used if they haven't been available for all of Mx10. (More important for settings which can actually change things in the runtime!),

Thanks Mark. They are clearer now :-)
And the lognodes, they should all available by 10.0.0

MarkvanMents:
Thanks Emir. 
Merging.

PR #8354, opened 2024-09-20
Non-Interrupting Boundary Event documentation

davevissermx:
This PR adds the public documentation for the boundary events. The following has been added about the boundary events:

- Workflow Engine description
- Jump To options
- Boundary Events page
- Versioning
- New basic functions options added.


content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
```diff
@@ -77,7 +77,8 @@ You can move elements around in your workflow. To do so, drag and drop an elemen
 
 This moves the whole structure of the dragged element, meaning that the element and its underlying elements are moved too.
 
-You cannot move the first (and only) start event and the last end event.
+- You cannot move the first (and only) start event and the last end event.
+- You cannot move the End of Boundary Event Path Activity
```
Yiyun333: What is the "End of Boundary Event Path Activity"? Do you mean this dot here:
image
I am not sure if we should call it an activity. Perhaps we shoudl just stick to the UI text and call it the "End of boundary path".

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
```diff
@@ -101,6 +102,9 @@ When using cut/copy/paste, note the following:
...
 * Pasting an activity after an **End** or **Jump** activity results in a consistency error as an **End** or **Jump** activity should be placed at the end of a flow.
 * Pasting activity can result in a consistency error when the next activity becomes unreachable by either placing an **End** or **Jump** activity in front of it or when all flows before a merge end with an **End** or **Jump** activity.
+* Pasting a boundary event on an activity adds it to the activity. You can add maximum five boundary events to one activity. For more information on what activities support adding boundary events, see [Boundary Events](/refguide/boundary-events/).
+* Pasting a **Jump** activity in a boundary event is only allowed when there are multiple paths going to the boundary event path end, otherwise a consistency error is thrown.
+* Pasting an **End** event inside a non-interrupting boundary event flow will result in a consistency error. This is because an **End** event is only allowed if no other paths will be executed, which is not the case in a non-interrupting timer boundary event.
```
Yiyun333: What does it mean by “when there are multiple paths going to the boundary event path end“?

Also, regarding this part:
> Pasting an **End** event inside a non-interrupting boundary event flow will result in a consistency error.

I tried this but it's not allowed to add an end event there. So how can it result in a consistency error?

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
```diff
@@ -111,6 +115,7 @@ The following elements cannot be cut, copied or pasted in the workflow editor:
 * Start activity
 * End activity (the final **End activity** which is at the end of the workflow)
 * Outcomes (meaning outcomes of different outgoing flows from an activity that can have several flows, for example, a Decision)
+* End of Boundary Event Path Activity (the final activity inside a non-interrupting boundary event path)
```
Yiyun333: I am not sure if we call this an activity. It's not in the Toolbox. Shall we stick to the UI text and use the following text?

* End of boundary path (the end of a non-interrupting boundary event path)

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
```diff
@@ -120,6 +125,12 @@ To view properties of an element, do one of the following:
 * Right-click an element and select **Properties** from the list of options that opens.
 * Double-click an element.
 
+### Elements that do not have any properties
+
+The following element does not have any properties and context menu.
+
+* End of Boundary Event Path Activity
```
Yiyun333: * End of Boundary Event Path Activity

Same. Can we just call it: End of boundary path?

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
davevissermx: From what I understood this is still possible to add if the user directly alters the model through the SDK/API. And thus adding the end activity to the flow.

davevissermx: And what is meant with this part is that the boundary event should still be finish able aside from the flow which has the jump. The jump is sort-of the end of a flow and thus is there another required.

Yiyun333: > From what I understood this is still possible to add if the user directly alters the model through the SDK/API. And thus adding the end activity to the flow.

Is this common among users? My understanding here is that this is a basic pasting/adding action in the workflow editor, which should not even imply altering the model in another way. Will it make sense to say:

> Pasting an **End** event inside a non-interrupting boundary event path is not allowed.

davevissermx: I don't know if this is very common, It might be done sometimes. 
But to circle back to the section where this text is placed in then I think that your new version should be clear enough. Given with the basic actions we indeed do not get an CE.

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
davevissermx: Good idea, I forgot we did not mention it like that to the user.

content/en/docs/refguide/modeling/application-logic/workflows/perform-workflow-basic-functions.md
davevissermx: Screenshot 2024-09-23 at 11 15 04
We indeed don't use the naming of activity in the ui. End of Boundary Path should be fine

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
```diff
@@ -48,14 +52,18 @@ The **Timer** properties are described in the table below:
...
 * When a timer is set on an in-progress workflow, the workflow continues when the timer expires.
 
 * When a time is set on a paused workflow and when the timer expires, the workflow only continues after the workflow is in progress again.
 
 ### Workflow Incompatibility
 
+{{% alert color="info" %}}
+This section is only applicable when **Timer** is used as a standalone activity on a workflow path.
+{{% /alert %}}
+
```
Yiyun333, line 82: @davevissermx @DmytroKost I added this info alert based on our previous discussion. Can you double-confirm whether this is correct?

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
```diff
@@ -67,6 +75,7 @@ The following cases do not trigger a continuation of the workflow path when time
 * Expiration in a workflow that is aborted.
 * Expiration in a workflow that is incompatible. (After the workflow resumes, the workflow path continues normally.)
 * Expiration in a workflow that is jumped from the timer to a different activity. 
+* Expiration in a workflow that is completed. (It can only occur when **Timer** is used as a [Boundary Event](/refguide/boundary-events/).)
```
Yiyun333: @davevissermx @DmytroKost 

Should I add "non-interrupting" in front of "Boundary Event" here, like in the following?

> It can only occur when **Timer** is used as a **non-interrupting** [Boundary Event](/refguide/boundary-events/).

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
```diff
@@ -13,11 +13,15 @@ aliases:
 **Wait for timer** was renamed to **Timer** in Studio Pro 10.14.0.
 {{% /alert %}}
 
-The **Timer** activity allows you to suspend a workflow path for a configurable duration or until a certain date and time.
+The **Timer** allows you to suspend a workflow path for a configurable duration or until a set date and time. It can be used in the following two ways:
```
Yiyun333: @davevissermx @DmytroKost 
Is the following description still accurate in terms of Timer used as a boundary event? Does a non-interrupting boundary event also suspend a workflow path?

> The **Timer** allows you to suspend a workflow path for a configurable duration or until a set date and time.

davevissermx: Oh we might have missed this, but no this is not correct. The workflow will not be suspended by a non-interrupting Timer boundary event.

If an activity has a Timer boundary event, albeit interrupting Timer or non-interrupting Timer then the workflow will stay in progress. Once the timer executes the workflow will act with behaviour depending on the type of timer event.

The only difference between interrupting/non-interrupting is that the non-interrupting allows the parent activity and the workflow to run simultaneously with the boundary event flow. The interrupting variant aborts all active flows and starts the defined path.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
davevissermx: No in my opinion you do not have to add non-interrupting here. This behaviour is the same for the interrupting variant.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
davevissermx: This looks correct to me.

content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
Yiyun333: [Commit](https://github.com/mendix/docs/pull/8354/commits/abdf3d87fed573bef75f1a8a348d8589df4e6227) reviewed via Slack.

Yiyun333 (approved):

PR #8366, opened 2024-09-24
Update workflow-events.md

recktenwaldfabian:
Add clarification on live state of referenced objects (WorkflowUserTask may be empty for User Task Ended event)

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/workflows/workflow-events.md
+++ b/content/en/docs/refguide/modeling/application-logic/workflows/workflow-events.md
@@ -47,7 +47,7 @@ The **WorkflowRecord** entity represents workflow instance data at a specific ex
 {{< figure src="/attachments/refguide/modeling/application-logic/workflows/workflow-events/workflow-event-entities.png" max-width=100% >}}
 
 {{% alert color="warning" %}}
-All associations to the above-mentioned non-persistent entities (with the exception of a sub **WorkflowRecord**) are associations to "live" objects, whose state may have been updated since the event occurred (e.g. when receiving 'User Task Ended', the WorkflowUserTask may already be deleted).
+All associations to the above-mentioned non-persistent entities (with the exception of a sub **WorkflowRecord**) are associations to "live" objects, whose state may have been updated since the event occurred (for example, when receiving the **User Task Ended** event, the **WorkflowUserTask** may already be deleted).
 {{% /alert %}}
 
 These non-persistent entities are provided as the default input parameters to the microflow specified in the event handler configuration:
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/application-logic/workflows/workflow-events.md
```diff
@@ -47,7 +47,7 @@ The **WorkflowRecord** entity represents workflow instance data at a specific ex
 {{< figure src="/attachments/refguide/modeling/application-logic/workflows/workflow-events/workflow-event-entities.png" max-width=100% >}}
 
 {{% alert color="warning" %}}
-All associations to the above-mentioned non-persistent entities (with the exception of a sub **WorkflowRecord**) are associations to "live" objects, whose state may have been updated since the event occurred.
+All associations to the above-mentioned non-persistent entities (with the exception of a sub **WorkflowRecord**) are associations to "live" objects, whose state may have been updated since the event occurred (for example, when receiving the **User Task Ended** event, the **WorkflowUserTask** may already be deleted).
```
Yiyun333: @recktenwaldfabian 
Does "the **WorkflowUserTask**" here mean "the **WorkflowUserTask** entity" or "an instance of the **WorkflowUserTask** entity"? 

Also the **WorkflowUserTask** entity is a persistable entity. What's the relation of the added example to the existing context? - Asking mainly from a linguistic perspective :)

recktenwaldfabian: This refers to the instance of the WorkflowUserTask which is referenced by the WorkflowActivityRecord_UserTask association from the WorkflowActivityRecord instance which is passed to the event handler.
It is an example for a "live" objects which has its state changed. One case of a changed state is that the "live" object has been deleted already and no longer exists.

Yiyun333: @recktenwaldfabian Thanks for the clarification! 
Will it make sense if I add "instance" after "WorkflowUserTask" like in the following?

> (for example, when receiving the **User Task Ended** event, the **WorkflowUserTask** instance may already be deleted).

recktenwaldfabian: I assumed that's already clear from the context, however It's a bit clearer like that.

Yiyun333 (approved):

content/en/docs/refguide/modeling/application-logic/workflows/workflow-events.md
Yiyun333: @recktenwaldfabian 
Thanks! I've approved the PR. I'll merge the PR after the Travis build has passed.

Yiyun333:
Verified by the team as well.

PR #8398, opened 2024-10-01
PMP updated MTS notes

katarzyna-koltun-mx:


content/en/docs/releasenotes/private-platform/1.6.md
mengdix, line 1: Line 2 (title) -- remove mention of MTS, standardize titling format

mengdix: Titling format has changed between 1.11 and 1.12, before the titling was more verbose, later it was simply version number.

Additionally, previous versions had mentioned "Mendix Private Platform" instead of "Private Mendix Platform"

Issue #8394, opened 2024-09-30
managing-app-signing-keys


RCatersels:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [managing-app-signing-keys](https://docs.mendix.com/refguide8/managing-app-signing-keys/)

Document link: [managing-app-signing-keys.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide8/mobile/managing-app-signing-keys.md)

## My Issue/Suggestion

The link https://slproweb.com/products/Win32OpenSSL.html is no longer valid because V1.1.1 is already some time depricated and advised not to use it anymore because it will no longer get security updates. So this version you can no longer find on this website.
Mendix should support Windows OoenSSL V3 in the Mendix cloud,
Regards,
Ronald

ConnorLand:
Hi @RCatersels ,

Thank you for reaching out to Mendix Documentation. I am tracking this issue, and the docs will be improved as soon as possible. I'll get back to you when I resolve the matter so you can see the change.

Thank you,

Connor

ConnorLand:
Fixed here: https://github.com/mendix/docs/pull/8406

Thank you for bringing this to our attention. If you spot any more erroneous OpenSSL documentation, please reach out.

Connor

PR #8406, opened 2024-10-01
Fix OpenSSL documentation instances 

ConnorLand:
This applies to distributing native apps documentation as well as cloud deployment. This PR solves issue #8394 .


MarkvanMents (approved):
Looks great.
Thanks for following up on this.
Approving.

PR #8412, opened 2024-10-02
Explanation of 'changes on disk' dialog added, focusing on 'file comparison'

tknippenberg:


tknippenberg:
@blis-mendix can you confirm whether I correctly described the functionality, specifically wrt the 'Modified' status?

blis-mendix:
Yeah seems good

content/en/docs/refguide/installation/system-requirements.md
```diff
@@ -109,6 +109,18 @@ To run a Mendix app, Mendix Studio Pro uses the following ports by default. If y
 
 For more information on ports and modifying Studio Pro's default ports, see [Configurations](/refguide/configuration/) and the [Troubleshooting Common Mobile Issues](/refguide/mobile/getting-started-with-mobile/prerequisites/#troubleshooting) section of *Native App Prerequisites and Troubleshooting*.
 
+### File Comparison
+
+{{% alert color="info" %}}
+Comparing files on disk with the original is currently not supported on macOS.
+{{% /alert %}}
+
+By default, Studio Pro points to an executable shipped together with TortoiseGit to compare files on disk. To customize this, navigate to **Preferences** > **Version control** > **General** > **File comparison** > **Executable** and choose your preferred tool.
```
quinntracy: @tknippenberg Could this first sentence be rephrased to "By default, Studio Pro uses an executable bundled with TortoiseGit for file comparison" or "By default, Studio Pro uses an executable bundled with TortoiseGit to compare files on disk"? Checking to ensure this wording wouldn't change the meaning.

tknippenberg: Thanks for checking. It's quite complicated haha.

There are two things:
* SP on a new device sets a default path to that executable. It's not checked whether TortoiseGit is installed, we just assume so.
* when installing TortoiseGit it by default installs to that path.

I think there is a subtle difference between "SP uses.." and "SP points to..", but if you think it's clear I'm fine either way!

quinntracy:
@MariaShaposhnikova This can be merged after review/approval

content/en/docs/refguide/installation/system-requirements.md
quinntracy: Thanks for providing additional details and helping clarify! I'll adjust appropriately using this feedback.

PR #8414, opened 2024-10-02
Update authentication token docs

EnasAbdelrazek:


content/en/docs/refguide/runtime/session-management.md
```diff
@@ -68,6 +68,8 @@ Below Mendix version 10.9.0, there is no support for authentication tokens.
 
 By default, authentication tokens will not be used in online Mendix apps version 10.9.0 and above. However, you can use them to remember the end-user by writing your own flow using the new API, `login2`, and setting the `useAuthToken` parameter to `true`.
 
+Online apps still utilize the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to maintain the session data changes. If that setting is disabled, the authentication token is used to keep users logged in, but it does not maintain the changes.
```
MarkvanMents: Hi @EnasAbdelrazek
I find this confusing.

1. The stuff about using the authentication token only applies from Mx 10.9.0, so I think we should add this.
2. I don't understand why data will be lost if the authentication token is used. Surely, for an online app, the data will be committed to the database and so won't be lost. I think talking about "not maintaining data changes" is unclear and will worry the customer.

Is there somewhere on Slack I can see a discussion of this, or can we have a discussion there or in a call. Thanks.

EnasAbdelrazek: I understand the confusion. I was referring to uncommitted changes. I should have explained it more clearly. We can also have a call if that helps.

MarkvanMents: @EnasAbdelrazek - thanks for getting back to me. Would this be OK as a rewrite?
```suggestion
Online apps still utilize the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to maintain uncommitted data which changes during the session. If that setting is disabled, in Mendix apps version 10.9.0 and above where `useAuthToken` is set to `true` the authentication token is used to keep users signed in but uncommitted changes are not maintained.
```

EnasAbdelrazek: Yeah it is much better now but we can add more details at the end. 

Online apps still utilize the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to keep maintain uncommitted data which changes during the session. If that setting is disabled, in Mendix apps version 10.9.0 and above where `useAuthToken` is set to `true` the authentication token is used to keep users signed in but uncommitted changes are not maintained after SessionTimeout.

It should also be reviewed from one of our team.

content/en/docs/refguide/runtime/session-management.md
```diff
@@ -68,6 +68,8 @@ Below Mendix version 10.9.0, there is no support for authentication tokens.
 
 By default, authentication tokens will not be used in online Mendix apps version 10.9.0 and above. However, you can use them to remember the end-user by writing your own flow using the new API, `login2`, and setting the `useAuthToken` parameter to `true`.
 
+Online apps still utilize the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to maintain uncommitted data which changes during the session. If that setting is disabled, in Mendix apps version 10.9.0 and above where `useAuthToken` is set to `true` the authentication token is used to keep users signed in but uncommitted changes are not maintained when a session is considered expired after [SessionTimeout](/refguide/custom-settings/#SessionTimeout)
```
WimJongeneel: I think its still a bit cryptic what "when a session is considered expired" means for those without in-dept knowledge of our platform. I think we could improve this by mention why a session can expire, that this normally means that a user gets logged out, and that we reload the application when that happens.

EnasAbdelrazek: What do you think about this change?

```suggestion
Online apps still utilize the [EnableKeepAlive](/refguide/tricky-custom-runtime-settings/#session-duration) setting to maintain uncommitted data which changes during the session. If that setting is disabled, in Mendix apps version 10.9.0 and above where `useAuthToken` is set to `true`, uncommitted changes will be lost if an action is performed after the session expires, which occurs after the [SessionTimeout](https://github.com/refguide/custom-settings/#SessionTimeout). In this case, the authentication token is used to reinitialize the session to keep user signed in, after which the application is reloaded.
```

WimJongeneel: Yes, looks fine for me

WimJongeneel (approved):

MarkvanMents (approved):
@EnasAbdelrazek and @WimJongeneel - looks good.
Approving.
I'll merge this at the end of today as it seems to be approved and refer to the current release(s). Let me know if there is a reason not to publish.

PR #8418, opened 2024-10-03
Shortcode to exclude text from being translated by Shutto

Luyao-Zhang-1:


Luyao-Zhang-1:
Will update the glossary instead to exclude certain words being translated.

PR #8431, opened 2024-10-04
Documentation for OQL

dahfjkg:
This is the documentation for OQL following the new structure.

It contains the original Clauses PR and changes from [OQL expressions](https://github.com/mendix/docs/pull/8434/commits) and [OQL expression syntax](https://github.com/mendix/docs/pull/8520/commits)


content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+	ON Req.CustomerName = Cust.LastName
+```
+| LastName | Number |
+| -------- | ------ |
+| Doe      | 1      |
+| Moose    | 2      |
+| NULL     | -1     |
+
+
+### `WHERE` clause
```
pijuskri: Good to mention that `WHERE NULL` is equivalent to `WHERE FALSE`

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+OQL clauses are reserved words that structure the OQL query into sections. Each section and, therefore, clause is responsible for a different aspect of the query. Every query has 2 mandatory clauses: `SELECT` and `FROM`.
+
+```sql
+SELECT LastName, Address
+FROM Sales.Customer
+```
+
+These basic clauses define which data needs to be retrieved, and which entity or entities should be used as a source. Other clauses are not mandatory. They are used to specify limitations and other rules on the data being retrieved.
+
+Other clauses are not mandatory, but if used, they should be present in a fixed order:
```
pijuskri: "Other clauses are not mandatory" is repeated in line 34

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+```sql
+SELECT FirstName FName, LastName LName
+FROM Sales.Customer
+```
+
+Alias is an alternative name to replace the column name in the result. When the name attribute is retrieved, the result column is "Name". With an alias, you can specify another result column name, like "Customer_Name". An alias cannot contain spaces.
+
+
+{{% alert color="info" %}}
+In some scenarios, aliases in `FROM` are mandatory. Aliases are mandatory when defining a view entity.
```
pijuskri: "In some scenarios, aliases in `FROM` are mandatory." should probably be moved to the FROM clause section

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+```
+
+| RequestNumber | CustomerCount |
+| ------------- | ------------- |
+| 1             | 1             |
+| 2             | 1             |
+| -1            | 0             |
+
+{{% alert color="info" %}}
+Please note that if you use a subquery as an expression in `FROM`, then such subquery should always return  a single column, and the number of rows should be at most one. If the subquery returns more than one row or a number of columns different than one, that will lead to an exception during runtime.
```
pijuskri: "in FROM" should be "in SELECT"

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+SELECT *
+FROM Sales.Customer Cust, Sales.Request Req
+WHERE Cust.LastName = Req.CustomerName
+```
+| sales$customer.ID | FirstName | LastName | sales$request.ID | CustomerName | Number |
+| ---------------   | -----     | -----    | ---------------- | ------------ | ------ |
+| 562949953421521   | John      | Doe      | 1688849860264073 | Doe          | 1      |
+| 562949953421923   | Jane      | Moose    | 1688849860264231 | Moose        | 2      |
+
+It is possible to specify columns in `SELECT`. To avoid ambiguity in case opf duplicate attribute names, it is possible to specify entity name or alias. Respectively, `<module>.<entity>/<attribute>` or `<alias>/<attribute>`. It is possible to retrieve alll attributes of a particular entity using `<alias>/*`
```
pijuskri: A lot of overlap with what is already mentioned in `SELECT` clause.

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+
+In addition to standard SQL syntax, OQL supports joins over associations.
+
+The syntax is as follows:
+
+	{ INNER | { { LEFT | RIGHT | FULL } [ OUTER ] } } JOIN
+	entity_path | entity_name | ( sub_oql_query ) [ [ AS ] from_alias ]
+	[ ON <constraint> ]
+
+##### entity_path
```
pijuskri: Should mention `ON` which is mandatory when long path is not used

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+
+#### Syntax
+
+The syntax is as follows:
+
+```sql
+WHERE <constraint>
+```
+
+`<constraint>` is an expression for which the value always equals true. Expressions consist of simple comparisons using operators, functions, keywords or system variables.
```
pijuskri: What do you mean by "for which the value is always true"? Perhaps it should have been "which is of type BOOLEAN"?

pijuskri: Are system variables parameters? Do we call them that way?

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+```
+| FirstName | LastName |
+| --------- | -------- |
+| John      | Doe      |
+
+### `GROUP BY` clause
+
+The `GROUP BY` clause is used to group OQL query results into summary rows based on the values of one or multiple attribuites. It is possible to filter aggregated results further using a `HAVING` clause.
+
+#### Syntax
```
pijuskri: Good to explain what is allowed in the expression (any columns of entities in scope) and HAVING constraint (aggregates of columns in group by). But this might not be neccesary as this behaviour is explained later.

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+	COUNT(*) > 1
+	AND
+	EXISTS (
+		SELECT *
+		FROM Sales.Storage stor
+		WHERE stor/AvailableStorage >= SUM(loc/Stock)
+	)
+```
+
+### `ORDER BY` clause
```
pijuskri: Perhaps explain the ordering rules per datatype

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
```diff
@@ -0,0 +1,998 @@
...
+```
+| LastName     |
+| ------------ |
+| Doe          |
+| Moose        |
+| Elk          |
+
+### `LIMIT` and `OFFSET` clauses
+
+With the `LIMIT` and `OFFSET` clauses, a portion of the result of a query can be returned. It is recommended to combine `LIMIT` and `OFFSET` clauses with `ORDER BY` because otherwise it can be impossible to predict which portion exactly weill be returned.
```
pijuskri: Perhaps a better phase would be "because the return the order of rows without any order by is inconsistent and can lead to unpredictable output"

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+* `SELECT SUM(TotalAmount) FROM Sales."Order" WHERE IsPaid = true` – retrieves the sum of the total prices of all paid orders (`Order` needs to be wrapped in quotes, see the [Reserved Words](#reserved-oql-words) section below)
+
+{{% alert color="info" %}}
+OQL queries do not take security into account out-of-the-box. This means that you can use OQL to manually define custom security expressions. In some cases, handling security yourself using OQL—instead of using the out-of-the-box security of XPath—may result in faster queries.
+{{% /alert %}}
+
+Try your OQL example online with the [OQL Playground](https://service.mendixcloud.com/p/OQL) demo app. 
+
+## OQL Clauses
+
```
MarkvanMents: I think this is more appropriate for the OQL landing page. So we have the structure:

    OQL Landing Page (what is OQL where do you use it, how to use it - or links to pages on how to use it
        OQL Clauses
        OQL Expressions
        (other OQL docs?)

I've put a dummy OQL Landing page there for now - we can use some of what is in the existing OQL page, but I think we should ideally have a completely new page?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+3. `WHERE`
+4. `GROUP BY`
+5. `HAVING` (allowed only in combination with `GROUP BY`)
+6. `ORDER BY`
+7. `LIMIT`
+8. `OFFSET`
+
+### `SELECT` clause
+
+The `SELECT` clause specifies which entity attributes or other specified data must be retrieved. The `SELECT` clause consists of the term `SELECT` and one or more expressions. Each expression must be separated by a comma. Each expression defines a column in the result. Each expression can have an alias, which will be the name of the column in the result.
```
MarkvanMents: You say that you have `SELECT` plus expressions. However, you explain these expressions in this document, not in the **OQL Expressions** document. I think this is a bit confusing - is there another word we can use (sub-clauses, or definitions?) instead of **expressions** here?

Also, you say that _Each expression defines a column in the result._ However, the `*` expression defines multiple columns in the result, doesn't it?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+OQL clauses are reserved words that structure an OQL query into sections. Each section, and therefore clause, is responsible for a different aspect of the query. Every query has 2 mandatory clauses: `SELECT` and `FROM`. For example:
+
+```sql
+SELECT LastName, Address
+FROM Sales.Customer
+```
+
+These basic clauses define which data needs to be retrieved and which entity or entities should be used as a source. Other clauses are not mandatory. They are used to specify limitations and other rules on the data being retrieved.
+
+Other clauses are not mandatory, but if used, they should be present in a fixed order:
```
MarkvanMents:
```suggestion
These basic clauses define the data that needs to be retrieved and which entity or entities should be used as a source. Other clauses are not mandatory. They are used to specify limitations and other rules on the data being retrieved. If they are used, they should be present in the following order:
```

MarkvanMents (commented):
Part way through the review
Just noting here that I've moved the document to match the structure we discussed in Confluence. Will continue working on this tomorrow.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+JOIN Sales.Customer/Sales.Customer_Request/Sales.Request
+```
+
+This query returns all attributes of objects of `Sales.Request` that are associated to `Sales.Customer` objects (see TODO)
+
+```sql
+SELECT Sales.Request/*
+FROM Sales.Customer
+JOIN Sales.Customer/Sales.Customer_Request/Sales.Request
+```
```
MarkvanMents: Using SQL for the language here has a weird effect on the markdown formatting as `/*` is not part of the SQL standard. We may have to not specify a language on these code blocks.
image

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+```sql
+SELECT FirstName FName, LastName LName
+FROM Sales.Customer
+```
+
+Alias is an alternative name to replace the column name in the result. When the name attribute is retrieved, the result column is "Name". With an alias, you can specify another result column name, like "Customer_Name". An alias cannot contain spaces.
+
+
+{{% alert color="info" %}}
+In some scenarios, aliases in `FROM` are mandatory. Aliases are mandatory when defining a view entity.
```
MarkvanMents: Is this the only case of a mandatory use of an alias, or just an example?
If there are more, then can we define them all?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+{{% /alert %}}
+
+#### Selecting all attributes using `*`
+
+`*` (asterisk) specifies that all attributes from all entities in the `FROM` clause should be returned. `entity_name/*` and `from_alias/*` specify that all attributes of the specified entity or expression of the `FROM` clause should be returned. `entity_name` can be optionally put in double quotes. If the entity name is a [reserved OQL word](/refguide/oql/#reserved-oql-words) (like `Order` or `Group`), double quotes are mandatory.
+
+##### Examples
```
MarkvanMents: Do we need four examples. Can we just have two examples: the simplest `SELECT *` and the most complex?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+SELECT *
+FROM Sales.Customer
+```
+
+This query returns all attributes of objects of `Sales.Customer` entity all attributes of associated objects of `Sales.Request` (see TODO)
+
+```sql
+SELECT *
+FROM Sales.Customer
+JOIN Sales.Customer/Sales.Customer_Request/Sales.Request
```
MarkvanMents: Should we have an example of JOIN before we have documented it? I think this would be better in the **JOIN** section.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+It is possible to use an aggregate function:
+
+```sql
+SELECT COUNT(*) AS CustomerCount
+FROM Sales.Customer
+```
+
+| CustomerCount |
+| ----------    |
+| 5             |
```
MarkvanMents: I think these should be in the expressions doc. We don't need them here.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1000 @@
...
+
+Another possibility is subqueries. A subquery can refer to other tables and expressions in `FROM`.
```
MarkvanMents: We currently have a whole document on subqueries (https://docs.mendix.com/refguide/oql-subqueries/). I think we should just move that into the OQL section of the documentation and refer to that rather than having examples here?

MarkvanMents (commented):
Hi @dahfjkg and @pijuskri 
I have done some proofreading in these commits: https://github.com/mendix/docs/pull/8431/files/8786adfd64abf1412b116d6433dd8457a11e7712..239aebdb960b3c3a018fedc47c43b47f5021afb3.

I have also made various suggestions in some comments.

Can you have a look at what I've done so far and see if you think I'm on the right track. We can discuss it further in a call if we need to discuss my changes.

If any of your earlier discussions are resolved can you mark them as resolved so it is clear what is still to do?

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
dahfjkg: By this I mean the same as the author of the original docs: nothing that makes sense 😅

dahfjkg: System variables are a different thing https://docs.mendix.com/refguide/xpath-keywords-and-system-variables/#system-variables

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
dahfjkg: It is explained right after, so no reason to repeat

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
pijuskri: But how can they be used in OQL?

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
pijuskri: This can be resolved then

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1022 @@
...
+| 562949953421521 | John  | Doe   |
+| 562949953421683 | Jane  | Doe   |
+| 562949953421777 | Jane  | Doe   |
+| 562949953421923 | Jane  | Moose |
+
+#### Expressions
+
+It is possible to use more complex expressions in `SELECT`. That is explained in detail in [OQL Expressions](/refguide/oql-expressions). Here are some examples:
+
+It is possible to use a function of an attribute. For instance:
```
MarkvanMents: Not sure about this. I initially didn't like the word "function" but I see that SQL has a lot of functions and that we are using it here in a similar way.
But the function isn't a function of the attribute (unlike a method in OOP) so perhaps 
```suggestion
It is possible to apply a function to an attribute. For instance:
```

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1022 @@
...
+```sql
+SELECT Sales.Customer/LastName
+FROM Sales.Customer
+```
+
+An alias can be used to replace entity name:
+
+```sql
+SELECT Cust/LastName
+FROM Sales.Customer Cust
```
MarkvanMents: Would this be clearer if we included `AS`?
```suggestion
FROM Sales.Customer AS Cust
```

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1022 @@
...
+
+This query is equivalent to the previous one, but it uses table aliases
+
+```sql
+SELECT Req/*
+FROM Sales.Customer Cust
+JOIN Cust/Sales.Customer_Request/Sales.Request Req
+```
+
+#### Selecting Distinct Values with `DISTINCT`
```
pijuskri: This section is being referenced as #Distinct by oql-expressions

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1022 @@
...
+FROM Sales.Customer Cust, Sales.Request Req
+WHERE Cust.LastName = Req.CustomerName
+```
+| sales$customer.ID | FirstName | LastName | sales$request.ID | CustomerName | Number |
+| ---------------   | -----     | -----    | ---------------- | ------------ | ------ |
+| 562949953421521   | John      | Doe      | 1688849860264073 | Doe          | 1      |
+| 562949953421923   | Jane      | Moose    | 1688849860264231 | Moose        | 2      |
+
+It is possible to specify columns in `SELECT`. To avoid ambiguity in case opf duplicate attribute names, it is possible to specify entity name or alias. Respectively, `<module>.<entity>/<attribute>` or `<alias>/<attribute>`. It is possible to retrieve alll attributes of a particular entity using `<alias>/*`
+```sql
```
MarkvanMents: I think we will have to leave the syntax highlighting off this as the `*` messes up the formatting and it looks weird (https://meta.stackoverflow.com/questions/401610/code-block-containing-asterisks-displaying-in-italics).
So
```suggestion
```
```

**Review by MarkvanMents — COMMENTED** [review]
Got a bit further, but still have only got as far as JOIN.
Sorry this is taking so long - need to do other things at the same time!

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 7 above)
Comment (dahfjkg) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Like this: `select Name from Module.Person where Module.Person_USER = '[%CurrentUser%]'`

There is a test for this

Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: This can be resolved

File: content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md (hunk 17 above)
Comment (dahfjkg) [reply content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md]: I removed this example, but other ones are used to illustrate `TableName/*`

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 10 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: This can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 7 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: This can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 4 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 3 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 10 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 1 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Can be resolved

File: content/en/docs/refguide/modeling/data/oql/oql-clauses.md (hunk 2 above)
Comment (pijuskri) [reply content/en/docs/refguide/modeling/data/oql/oql-clauses.md]: Can be resolved

File: content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md [hunk 24]
```diff
@@ -0,0 +1,1130 @@
...
+{ INNER | { { LEFT | RIGHT | FULL } [ OUTER ] } } JOIN
+entity_path | entity_name | ( sub_oql_query ) [ [ AS ] from_alias ]
+[ ON <constraint> ]
+```
+
+##### entity_path
+
+`entity_path` specifies the entity to join and the path from an earlier defined entity in the `FROM` clause to this entity.
+
+The example path `Crm.Customer/Crm.Customer_Address/Crm.Address` defines an association path from the entity **Crm.Customer** to a new entity **Crm.Address**.
```
MarkvanMents:
```suggestion
The example path `Crm.Customer/Crm.Customer_Address/Crm.Address` defines an association path from the entity **Crm.Customer** to an associated entity **Crm.Address**.
```

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1130 @@
...
+GROUP BY Brand, City
+```
+
+| Brand  | SumStock |
+| ------ | -----    |
+| Cinco  | 5        |
+| Rekall | 9        |
+| Rekall | 3        |
+| Veidt  | 24       |
+| Veidt  | 2        |
```
MarkvanMents: Why does this table not include the `City` column?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1130 @@
...
+OFFSET 2
+```
+
+| Brand  | City      | LocationNumber |
+| ------ | -----     | -----          |
+| Rekall | Zwolle    | 3              |
+| Rekall | Utrecht   | 4              |
+| Veidt  | Utrecht   | 5              |
+
+### Subqueries
```
MarkvanMents: Will this replace the current [OQL Subqueries page](https://docs.mendix.com/refguide/oql-subqueries/)?

That page has subquery in HAVING. Are we deliberately not documenting this because subqueries in GROUP BY are database dependent?

MarkvanMents (commented):
Hi there,
I've nearly finished my review. Sorry it is taking so long, but I keep being called away to do other things.
I have a couple more questions which are related to parts of the text. See below:

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
dahfjkg: GROUP by is database dependent, but HAVING should work in every database. I'm pushing some examples for that.

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
pijuskri: This can be resolved

content/en/docs/refguide/modeling/domain-model/oql/_index.md
```diff
@@ -0,0 +1,23 @@
...
+
+The Mendix Object Query Language (OQL) is a relational query language inspired by [SQL](https://en.wikipedia.org/wiki/Sql). The major advantage of OQL is that it uses Mendix entity and association names instead of actual database table names. That way, it is possible to create queries that use names from the data model of your Mendix app without thinking of how that data model is represented in the database.
+
+In addition, OQL can use predefined relations (associations) to easily join objects without having to calculate which columns should be coupled. Despite these differences, many SQL keywords also work in OQL.
+
+These are some examples of OQL queries:
+
+* `SELECT LastName FROM Sales.Customer` – retrieves the names of all customers
+* `SELECT FirstName FROM Sales.Customer WHERE LastName = 'Jansen'` – retrieves the first name of all customers with name "Jansen"
+* `SELECT SUM(TotalAmount) FROM Sales."Order" WHERE IsPaid = true` – retrieves the sum of the total prices of all paid orders (`Order` needs to be wrapped in quotes, see the [Reserved Words](#reserved-oql-words) section below)
```
pijuskri: There is no `reserved-oql-words` section

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
pijuskri: This can be resolved

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
pijuskri: This can be resolved

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
dahfjkg: TBH, I think it is obvious without explanation. Data grids support sorting, so users should be familiar with how it works

content/en/docs/refguide/modeling/domain-model/oql/_index.md
```diff
@@ -21,3 +21,79 @@ OQL queries do not take security into account out-of-the-box. This means that yo
 {{% /alert %}}
 
 Try your OQL example online with the [OQL Playground](https://service.mendixcloud.com/p/OQL) demo app. 
+
+## Reserved Words {#reserved-oql-words}
+
+Words with a specific purpose in OQL are reserved. If you use reserved words for entity, variable or attribute names in an OQL query, they must be wrapped in double quotes `" "`. For example, in the OQL query `SELECT AVG(TotalPrice) FROM Sales."Order" WHERE IsPaid = 1`, `Order` needs to be wrapped in quotes because it is a reserved word, as it can be used for `ORDER BY`.
+
+Here is a list of all OQL reserved words:
```
pijuskri, line 31: Consider my notes on reserved words.
https://mendix.atlassian.net/wiki/spaces/DS/pages/3107586121/Undocumented+features+reserved+words

Some of the keywords in the list are used as function names, but are not strictly reserved. They are only used with function call syntax and so will not cause issues when used as regular identifiers. Might be worth to mention in a separate "keywords" list.

content/en/docs/refguide/modeling/domain-model/oql/_index.md
pijuskri: Can be resolved

content/en/docs/refguide/modeling/data/oql/oql-clauses.md
pijuskri: Can be resolved

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,136 @@
...
+FROM (SELECT * FROM Module.Person) P
+JOIN (SELECT * FROM Module.City) C
+ON P/Residence = C/Name
+```
+
+If attribute names are specified explicitly, retrieving duplicate columns with `*` is not allowed, the same as in OQL v1. In this case, use aliases.
+
+The following is **not allowed**:
+
+```sql
```
pijuskri, line 166: This is actually allowed, see test in `SubqueryInJoin:339`

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,136 @@
...
+    SELECT Name
+    FROM Module.Person
+    ORDER BY Name
+    LIMIT 20
+)
+```
+
+### `ORDER BY` without `LIMIT` and `OFFSET` in View Entity is not allowed
+
+For view entities, having `ORDER BY` without `LIMIT` and/or `OFFSET` is not allowed even for the top level query. Similarly to the case of subqueries, view entitiy results are not accessed directly, but are often subject of further filtering and ordering. TRherefore, it would not make sense to have `ORDER BY` directly in the view entity definition unless it is used to select a slice of the result by the means of `LIMIT` and/or `OFFSET`.
```
pijuskri: Maybe can be more specific and say "view entitiy results are not accessed directly, but are used as subqueries in OQL or xpath queries"

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,136 @@
+---
+title: "OQL v2 features"
+url: /refguide/oql-v2/
+weight: 100
+---
+
+## OQL v2 features
```
pijuskri: GIven its a breaking change, i think this is important to mention also
https://mendix.atlassian.net/browse/DAT-3753

pijuskri:
Passed technical review

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1234 @@
...
+5. `HAVING` (allowed only in combination with `GROUP BY`)
+6. `ORDER BY`
+7. `LIMIT`
+8. `OFFSET`
+
+### `SELECT` clause
+
+The `SELECT` clause specifies which entity attributes or other specified data must be retrieved. The clause returns all the values of objects which match the `SELECT` clause.
+
+The `SELECT` clause consists of the term `SELECT` and one or more column definitions. Each column definition must be separated by a comma. Each column definition defines a column or a set of columns in the result. Each single value column definition can have an alias, which will be the name of the column in the result.
```
MarkvanMents, line 45: I think we need to say somewhere that a "column" is the same as an "attribute"?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1234 @@
...
+#### Selecting All Attributes Using `*`
+
+Using `*` (asterisk) in the SELECT clause specifies that the values of all attributes from all entities in the `FROM` clause should be returned.
+
+Specifying `entity_name/*` and `from_alias/*` specify that the values of all attributes of the specified entity or expression of the `FROM` clause should be returned.
+
+`entity_name` can optionally be put in double quotes. If the entity name is a [reserved OQL word](/refguide/oql-clauses/#reserved-oql-words) (like `Order` or `Group`), double quotes are mandatory.
+
+{{% alert color="warning" %}}
+Specifying all attributes will also return attributes which are normally hidden in the Domain Model, such as the `ID` of each object.
```
MarkvanMents, line 97: Is this a warning or just info?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1234 @@
...
+		WHERE stor/AvailableStorage >= SUM(loc/Stock)
+	)
+```
+
+### `ORDER BY` clause
+
+The `ORDER BY` clause specifies the sort order used on columns returned in a `SELECT` statement. Multiple columns can be specified. Columns are ordered in the sequence of the items in the `ORDER BY` clause.
+
+{{% alert color="info" %}}
+This clause can include items that do not appear in the `SELECT` clause, except when `SELECT DISTINCT` is specified or when a `GROUP BY` clause exists. When `UNION` is used, the column names or aliases must be those specified in the `SELECT` clause of the first part of the query.
```
MarkvanMents: I think we should add a link to UNION documentation (not sure where that is).

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1234 @@
...
+
+The syntax is as follows:
+
+```sql
+[ LIMIT number ] [ OFFSET number ]
+```
+
+#### `LIMIT` clause
+
+`LIMIT` specifies how many rows must be returned.
```
MarkvanMents: Isn't it the maximum number of rows that are returned. It could be less than LIMIT

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1234 @@
...
+```
+
+| RequestNumber | CustomerCount |
+| ------------- | ------------- |
+| 1             | 1             |
+| 2             | 1             |
+| -1            | 0             |
+
+{{% alert color="info" %}}
+If you use a subquery as an expression in `SELECT`, then such subquery should always return a single column, and the number of rows should be at most one. If the subquery returns more than one row or a number of columns different than one, that will lead to an exception during runtime.
```
MarkvanMents: Could we just say a single value? Or saying the same thing in a shorter way:
```suggestion
If you use a subquery as an expression in `SELECT`, then such subquery should always return a single row and column. If the subquery returns more than one row or column, that will lead to an exception during runtime.
```

MarkvanMents (commented):
I have reviewed the clauses and _index documents again. My changes are here: https://github.com/mendix/docs/pull/8431/files/7915620e4eee2a89c1fde7f26fb10609adfcd284..04aec8a20f5c22496066d15e04c18b5c43641930 and I have left a couple more questions/suggestions.

I will try to do oql-v2 tomorrow and also the expression syntax.

At some stage it would be good to combine everything into one PR - but we can discuss this later.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,224 @@
+---
+title: "OQL Version 2 Features"
+linktitle: "OQL V2 Features"
```
MarkvanMents: Is this really "features". I would say it is "changes" or "differences".
You could even call it something like "Switching to OQL Version 2"?

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,224 @@
...
+```sql
+SELECT *
+FROM (SELECT LastName AS N FROM Module.Person) P
+JOIN (SELECT Name AS N FROM Module.City) C
+ON P/Residence = C/Name
+```
+
+### `ORDER BY` in Subquery
+
+You must now have a `LIMIT` and `OFFSET` in subquery containing `ORDER BY`. Using `ORDER BY` in subquery makes sense only when it is combined with `LIMIT` and `OFFSET`. Without the limitations, database engines do not guarantee that the row order in the subquery will be preserved in the outer query.
```
MarkvanMents: Is this **LIMIT and OFFSET** or **LIMIT and/or OFFSET**. You say both in the document.

MarkvanMents (commented):
I've now proofread oql-v2 in this commit: 2954bd39a0f67f15694198b75779f57eb0ae2e58
I have a couple of comments which need clarification.
I will try to read the third PR, #8520 before the end of the week.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
dahfjkg: I reduced the number of examples

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
dahfjkg: More than that: we should add UNION documentation itself 😅

dahfjkg:
I see that I still need to:
1.  Document `UNION`
2. Explain more bugfixes in OQL v2

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+Instead, you can use the long path in the `JOIN` pattern as follows:
+
+```sql
+SELECT COUNT(*)
+FROM Module.Person AS P
+JOIN Module.Person/Module.Person_City/Module.City AS C
+GROUP BY C/Name
+```
```
MarkvanMents: Will this work, or does COUNT(*) need an **AS**?

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,253 @@
...
+
+In OQL v2, such queries are no longer allowed. You must always provide an alias for subqueries that do not result in a named column. The query above can be rewritten as follows:
+
+```sql
+SELECT
+  Name,
+  (
+    SELECT COUNT(*)
+    FROM Module.Person
+    WHERE Module.Person/City = Module.City/Name
```
MarkvanMents, line 100: Will this work, or does COUNT(*) need an AS?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1258 @@
...
+
+ A subquery can be used as a column in the `SELECT` clause. It can refer to other tables and expressions in `FROM`.
+
+```sql
+SELECT
+	Req/Number AS RequestNumber,
+	(
+		SELECT COUNT(*)
+		FROM Sales.Customer AS Cust
+		WHERE Cust/LastName = Req/CustomerName
```
MarkvanMents, line 1276: Will this work, or does COUNT(*) need an AS?

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1258 @@
...
+Example:
+
+```sql
+SELECT COUNT(*) AS LocationCount, SUM(Stock) as CityStock, City AS City
+FROM Sales.Location AS Location
+GROUP BY City
+HAVING
+	SUM(Stock) <= (
+		SELECT COUNT(*)
+		FROM Sales.Location
```
MarkvanMents, line 1444: Will this work, or does COUNT(*) need an AS?

MarkvanMents (commented):
Reviewed for **COUNT (*)** as discussed on Slack.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
hugomendix: It will need an `AS`.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
hugomendix: Same here; requires `AS`, because otherwise `COUNT(*)` would be a nameless column.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
hugomendix: `AS` needed.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
hugomendix: Same asnwer; `AS` needed.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
pijuskri: `AS` is not required here, the column name will be the alias of the subquery `PersonCount`, the internal subquery column name has no effect on the main queries column names.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
pijuskri: `AS` is not required, `HAVING` and `WHERE` are not affected by changes to mandatory column names. It only applies to the main `SELECT` query and the attributes/columns of it.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
pijuskri: `AS` is not required, same reason as other case with subquery in attribute

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
hugomendix: Thanks. I missed the fact that it was nested from the diff.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
dahfjkg: In OQL v2, AS should be required

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -1027,6 +1027,128 @@ OFFSET 2
 | Rekall | Utrecht   | 4              |
 | Veidt  | Utrecht   | 5              |
 
+## `UNION` Clause {#oql-union}
+The clause takes multiple select queries and combines their results into a single result set.
+The resulting set by default only includes distinct rows. The `ALL` keyword can be used to include all rows. Rows are considered distinct if the combination of the column values is distinct from all other rows. Comparison logic of values is the same as the `DISTINCT` keyword of a `SELECT` clause.
+
+All select queries must define the same number of columns in the same order and their datatypes should match. The resulting column names will those used in the very first `SELECT` clause.
```
dahfjkg: There is probably a word missing in `The resulting column names will those used`

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -1027,6 +1027,128 @@ OFFSET 2
 | Rekall | Utrecht   | 4              |
 | Veidt  | Utrecht   | 5              |
 
+## `UNION` Clause {#oql-union}
```
dahfjkg, line 1056: We should add an example of how `UNION` works with association columns

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1157 @@
...
+
+```sql
+CAST ( expression AS data_type )
+```
+
+##### expression
+
+`expression` specifies the expression to convert.
+
+##### data_type
```
passalaqua, line 705: We still support CAST to FLOAT (and FLOAT is still a valid keyword in OQL). We have only deprecated it as a valid result type for queries, but intermediate calculations should still work.

MarkvanMents: Deprecated or removed entirely?
Is FLOAT deprecated (will it go in Mx 11, for example). I can't really see any point in keeping it if we have DECIMAL.

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
MarkvanMents:
```suggestion
All select queries must define the same number of columns in the same order and their datatypes should match. The resulting column names will be those used in the very first `SELECT` clause.
```

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
pijuskri: Is there a use case for a new user of OQL to use `FLOAT` instead of `DECIMAL` ?

passalaqua: > Deprecated or removed entirely?

As I said:
> We have only deprecated it as a valid result type for queries, but intermediate calculations should still work.

And indeed, we still have tests for it and it still works, AFAICS.

Either we document here or in the list of keywords, but completely omitting mentioning something that would otherwise break your query if you use as an identifier is not a good idea, IMO. I just wrote the comment here because `CAST` is the only place I remember still accepts `FLOAT`.

dahfjkg: I pushed a mention of `FLOAT`

passalaqua: Looks good to me!

pijuskri: Wouldn't a better place for the mention of `FLOAT` be in the `CAST` section, given that is the only valid use case?

dahfjkg: If we are to move away from Float, I'd not mention any use cases of it at all

pijuskri: Sorry, I reread Augusto's comment. Given the reason to mention it is to explain why its a keyword, then the current placement is good. Nothing else needs to be added.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,278 @@
+---
+title: "OQL Version 2 Features"
+linktitle: "Switching to OQL Version 2"
+url: /refguide/oql-v2/
+weight: 100
+---
+
+## Introduction
+
+In Studio Pro version 10.17, we introduced OQL v2. You will have to switch to OQL v2 in order to enable View entities. See the [View entities](/refguide/app-settings/#enable-view-entities) app setting for more information.
```
sbont: Note that we've renamed this setting to "OQL version 2", and the anchor will be #oql-version-2.
See also our branch for View Entities at https://github.com/mendix/docs/compare/development...sbont:docs:dux-1-view-entities
(Also, I assume you might want to update the Studio Pro release version...)

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -58,14 +58,16 @@ There is no direct support for `DATETIME` literals. For functions that take `DAT
...
 ## System variables
 
-XPath [system variables](/refguide/xpath-keywords-and-system-variables/) can be used in OQL with the format:
+Most XPath [system variables](/refguide/xpath-keywords-and-system-variables/#system-variables) can be used in OQL with the format:
 
 ```sql
 '[%SystemVariable%]'
 ```
 
 These variables can be used the same way as other expressions.
```
pijuskri, line 67: To be more explicit, we should mention that system variables like `%CurrentUser%` are resolved to id's of objects with corresponding type (`LONG` ?). Same with date system variables, are they just strings? How do intervals work (like `%DayLength%`)?

dahfjkg: Pushed an update

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1218 @@
...
+
+Literals represent values that are constant and are part of the query itself. The supported literals are detailed below:
+
+| Format | Example         | Data Type            | Description                                   |
+|--------|-----------------|----------------------|-----------------------------------------------|
+|        | `TRUE`, `FALSE` | `BOOLEAN`            | Conditional constants                         |
+| 's*'   | 'my_string'     | `STRING`             | String literal                                |
+| d+     | 5               | `INTEGER` and `LONG` | Natural number literal                        |
+| d+.d+  | 5.3             | `DECIMAL`            | Real number literal                           |
+|        | `NULL`          | N/A                  | NULL literal to represent non-existent values. OQL does not support the keyword `empty`. In XPath, it has identical purpose as `NULL`. |
```
MarkvanMents: @pijuskri 
Is it a bit confusing to talk about XPath here? Or are you saying that in XPath you can't use `NULL` and have to use `empty` and in OQL you can't use `empty` and have to use `NULL`?

In that case I think it would be better to say:
```suggestion
|        | `NULL`          | N/A                  | NULL literal to represent non-existent values. OQL uses `NULL` where the equivalent XPath expression would use `empty`. |
```

dahfjkg: The behavior is that XPath supports both NULL and empty, and OQL supports only NULL, so your proposed change is not entirely correct.

The idea here was to highlight similarities and differences to `XPath Keywords and System Variables` page. We already received some feedback from customers participating in beta testing who are familiar with XPath and who expected some things to work the same way in OQL.

Now that I think of it, system variables have lots of similarities between XPath and OQL and are definitely worth being described from that perspective, but Keywords in XPath are quite different from Literals in OQL, so It's also fine to omit mentioning XPath here.

MarkvanMents: @dahfjkg 
Thanks for getting back to me, Pavel.
I wonder if it is useful to have the comparison somewhere. Would it be worth having an additional page which describes the differences (and similarities) between XPath and OQL? 

I'm happy to put something about XPath here (or get rid of it totally), but if we had it in another place that might be better - and resolve the issues that customers have raised?

dahfjkg: The other issues are being addressed by implementing new features, and aside from those and the ones in XPath Keywords and System Variables, not that many points come to mind at the moment. Maybe we can consider that for the future, but I don't think that there is enough reason to work on that now. I'd say let's keep it mentioned here.

MarkvanMents: @dahfjkg - so if we keep it, and my description above is wrong, how about:
```suggestion
|        | `NULL`          | N/A                  | NULL literal to represent non-existent values. OQL only uses `NULL` where the equivalent XPath expression can also use `empty`. |
```
?

MarkvanMents (approved):
Thanks for all the hard work on this.
I will approve now, as I have no outstanding questions. But please let me know if you want to add anything else, or if this fails to get into 10.19.😀

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,278 @@
...
+
+In OQL v1, the result of the arithmetic operation will always be of type pf the first attribute in the expression because it is handled by the database. Therefore, the result would depend on the underlying database engine.
+
+When handling numeric types in OQL v2 (Integer, Long, and Decimal), the result of the operation is always the most precise attribute type, using the following precedence:
+
+* Decimal (highest)
+* Long
+* Integer
+
+If any side of the operation is of a non-numeric type, no casting is performed, and the result is handled by the , as in OQL v1. See [Expression syntax](/refguide/oql-expression-syntax/#type-coercion) for more information.
```
pijuskri: There seems to be a missing word after `the result is handled by the `. @dahfjkg was it supposed to be "database"?

pijuskri:
Since this PR is ready to be merged, for new content I created a separate PR targeting the oql-clauses branch on Pavels repository.
https://github.com/dahfjkg/docs/pull/3/files

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1491 @@
...
+It is possible to build paths over multiple associations. Associated entities can be reached from both directions. System associations `System.owner` and `System.changedBy` can also be used in such association paths, assuming they are enabled for the entity. For example:
+
+```sql
+SELECT
+	LastName AS CustomerName,
+	Sales.Customer/Sales.Request_Customer/Sales.Request/System.owner/System.User/Name AS UserName
+FROM Sales.Customer
+```
+
+In the case when the association [multiplicity type](/refguide/association-properties/#multiplicity) is one-to-many or many-to-many, every attribute over association will result in multiple results per object. If there are multiple attributes over association in the `FROM` clause, the result will be a cartesian product of associated objects meaning that there will be a row for every combination of associated objects. If you want to avoid that effect, consider rewriting the query using [`JOIN`](/refguide/oql-clauses/#join).
```
MarkvanMents:
```suggestion
In the case when the association [multiplicity type](/refguide/association-properties/#multiplicity) is one-to-many or many-to-many, every attribute over association may result in multiple results per object. If there are multiple attributes over association in the `FROM` clause, the result will be a cartesian product of associated objects meaning that there will be a row for every combination of associated objects. If you want to avoid that effect, consider rewriting the query using [`JOIN`](/refguide/oql-clauses/#join).
```

content/en/docs/refguide/modeling/domain-model/oql/oql-clauses.md
```diff
@@ -0,0 +1,1491 @@
...
+
+For example, when a `Sales.Customer` with `LastName` "Doe" has two associated objects in `Sales.Request` with `Number` 1 and 2, the following query will return 4 rows:
```
MarkvanMents:
```suggestion
For example, when a `Sales.Customer` with `LastName` "Doe" has two associated  `Sales.Request` objects with a `Number` attribute with values 1 and 2, the following query will return 4 rows:
```

MarkvanMents (commented):
Looks like a good addition. Have made a couple of proposals for changing the text.

content/en/docs/refguide/modeling/domain-model/oql/oql-v2.md
```diff
@@ -0,0 +1,294 @@
+---
+title: "OQL Version 2 Features"
+linktitle: "Switching to OQL Version 2"
+url: /refguide/oql-v2/
+weight: 100
+---
+
+## Introduction
+
+In Studio Pro version 10.19, we introduced OQL v2. It can be enabled by setting the [OQL version 2](/refguide/app-settings/#oql-version-2) app setting to *Yes*. You will have to switch to OQL v2 in order to enable [View entities](/refguide/view-entities/).
```
MarkvanMents, line 10: @dahfjkg and @pijuskri 
The link to app-settings doesn't work and I can't find any other PR where this is added.
Is there supposed to be an update to app-settings in this PR?

MarkvanMents: OK - found the link in another PR!

PR #8434, opened 2024-10-04
Documentation for OQL expressions

pijuskri:
- Documentation for OQL expressions following the new structure
- Includes aggregates and parameters
- Needs team review
- To be merged with other OQL documentation MR's


content/en/docs/refguide/modeling/data/oql/oql-expressions.md
```diff
@@ -0,0 +1,221 @@
...
+
+Aggregations are functions that reduce a list of values from a retrieved column (or columns) into a singular value. They can be used as an attribute in a `SELECT` clause or in a condition in `HAVING`. 
+
+When combined with a `GROUP BY` clause, aggregations can be used with any column normally available in `SELECT`, not just those specified in the `GROUP BY` clause. Aggregation is performed over groups separately in this case.
+
+## Syntax
+
+```sql
+  COUNT ( * )
+  | { COUNT | AVG | MAX | MIN | SUM } ( [ DISTINCT ] attribute_path )
```
dahfjkg: Let's add `STRING_AGG` here

content/en/docs/refguide/modeling/data/oql/oql-expressions.md
```diff
@@ -0,0 +1,221 @@
...
+```sql
+SELECT Sum(Stock) AS ProductCount FROM Sales.Product
+```
+| ProductCount |
+|:------------:|
+|     103      |
+
+The sum of all products in stock:
+```sql
+SELECT SUM(Stock) AS StockSum FROM Sales.Product
```
dahfjkg: This example is repeated twice

content/en/docs/refguide/modeling/data/oql/oql-expressions.md
```diff
@@ -0,0 +1,221 @@
...
+
+The highest stock number:
+```sql
+SELECT MAX(Stock) as StockMax FROM Sales.Product
+```
+| StockMax |
+|:--------:|
+|    54    |
+
+Using `MAX` on a string column gives the last alphabetically sorted name:
```
dahfjkg: TBH, this is probably too technical

content/en/docs/refguide/modeling/data/oql/oql-expressions.md
```diff
@@ -0,0 +1,221 @@
...
+```sql
+SELECT Name
+FROM Module.Person
+WHERE
+    Age > $param 
+    AND
+    Active = true
+```
+
+If the value of `$param` is not provided, the query will be equivalent to:
```
dahfjkg: I think this is not entirely correct. If a parameter is empty, that part of the expression should be replaced with true, like this:
```
WHERE
    true 
    AND
    Active = true
```
This is even more relevant in case of `OR`, so maybe we should use `OR` in our examples

dahfjkg (changes requested):

content/en/docs/refguide/modeling/data/oql/oql-expressions.md
pijuskri: Agreed, i dont think the example is neccesary for more advanced users and this behaviour is mentioned in the documentation

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
+---
+title: "OQL Expressions"
+url: /refguide/oql-expressions-new/
+---
+
+## Introduction
+
+An OQL expression is a query building block that returns a value or a list of values. Expressions can be one of the following:
+
+* a constant
```
MarkvanMents, line 10: Can this also be a list of constants? Or can only some expressions return lists?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+* a function
+* a combination of attribute names, constants, and functions connected by operators
+* a subquery
+
+OQL expressions can be used in `WHERE`, `SELECT`, `GROUP BY`, `UNION`, and `JOIN` clauses.
```
MarkvanMents: Should probably link to the OQL Clauses doc here.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+
+## Aggregations
+
+Aggregations are functions that reduce a list of values from a retrieved column (or columns) into a single value. They can be used in the following ways:
+
+* as an attribute in a `SELECT` clause
+* as a condition in a `HAVING` clause
+* When combined with a `GROUP BY` clause, aggregations can be used with any column normally available in `SELECT`, not just those specified in the `GROUP BY` clause. Aggregation is performed over groups separately in this case.
```
MarkvanMents: I don't understand this at all - and there is no example. How can you put an aggregation into a GROUP BY clause if the column isn't mentioned in the GROUP BY clause?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+
+```sql
+SELECT COUNT(*) AS ProductCount FROM Sales.Product
+```
+
+| ProductCount |
+|:------------:|
+|      4       |
+
+The same result can be retrieved by using `COUNT` on a single table:
```
MarkvanMents: On a single **attribute**?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+| Name     | Stock |
+|----------|-------|
+| Cheese   | 5     |
+| Milk     | 54    |
+| Tomatoes | 44    |
+| Tomatoes | NULL  |
+
+#### COUNT
```
MarkvanMents, line 96: Should these examples be in the same order as we explain the expressions (i.e. with AVG first)?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+
+The average stock per product:
+
+```sql
+SELECT AVG(Stock) AS ProductCount FROM Sales.Product
+```
+
+| ProductCount |
+|:------------:|
+|    34.333    |
```
MarkvanMents: This is only right because one of the **Tomatoes** rows has NULL. If it had a number in then the average would be wrong 😀 because we aren't filtering on distinct product names.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+
+```sql
+SELECT STRING_AGG(Name, ',') as ProductNames FROM Sales.Product
+```
+
+|         ProductNames          |
+|:-----------------------------:|
+| Cheese,Milk,Tomatoes,Tomatoes |
+
+#### Distinct
```
MarkvanMents: I think we should link to the clauses doc here.
Also, DISTINCT isn't an expression - the example is about COUNT so I think we should move this to the COUNT examples.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+
+| ProductCount |
+|:------------:|
+|      3       |
+
+## Parameters
+
+Parameters are external variables that are referenced to by name in an OQL query. To use a defined parameter in a query, prepend the `$` sign to the name of the parameter. 
+
+If you use undefined in `IN` and `LIKE` comparison expressions, the condition always returns `true`. In other cases, undefined parameters cause an exception.
```
MarkvanMents, line 219: We don't have any documentation for `IN` and `LIKE`. Are these expressions going to be added here, or will they be in a different document?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+```sql
+SELECT Name
+FROM Sales.Person
+WHERE
+    Age > $param 
+    OR
+    Job = 'Sales'
+```
+
+If the value of `$param` is not provided, the query will be equivalent to:
```
MarkvanMents: Not sure exactly what "not provided" means in this context.
Does it mean that the parameter has not been created and therefore is not in the current scope, or can you create a variable but still not initialize it.
If it is the first, then perhaps say "If the parameter `$param` has not been created"?

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+SELECT COUNT(DISTINCT Name) AS ProductCount FROM Sales.Product
+```
+
+| ProductCount |
+|:------------:|
+|      3       |
+
+## Parameters
+
+Parameters are external variables that are referenced to by name in an OQL query. To use a defined parameter in a query, prepend the `$` sign to the name of the parameter.
```
MarkvanMents, line 217: Do we have some documentation on how these are created - I guess they are just Java variables, but how does that work in data sets?
Perhaps this would become obvious if I was actually creating some OQL?

MarkvanMents (commented):
Hi @pijuskri and @dahfjkg 
I've proofread the whole of this document.
I think the style matches the clauses document quite well, I'm not worried that they will seem too different.
I've asked some questions for clarification. 
I'll go back to the clauses document while you digest my changes and come back to me.
Let me know if you want to set up a meeting.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: The query can reference a parameter and then the parameter mapping needs to be provided when calling the query, say in a java action. I wouldn't say "created" is the best word to use here, parameters are quite temporary.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: Yes they will be documented in `oql-expression-syntax`

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: Yes, did not notice the order, will change it.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: I misformated this. It should not be part of the "ways to use aggregations" list. 
The comment about group by is for aggregates the `SELECT` clause. Ill clarify this

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: There can be a list of of constants, but only when used with the `IN` operator. Don't think that single use case is important enough to mention here. Some expressions (like subquery) can return multiple values, so its useful to mention that here.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: The best we have now is the datasets documentation. There arent any visuals or examples there and honestly I dont know how that works either.
For java actions its partially explained in the [API docs](https://apidocs.rnd.mendix.com/8/runtime/com/mendix/systemwideinterfaces/connectionbus/requests/types/ITextGetRequest.html#createParameterMap()).

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
MarkvanMents: OK, Thanks.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
MarkvanMents: OK, thanks.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
MarkvanMents: So could we say
```suggestion
If the value of `$param` is not provided as a parameter to the query, the query will be equivalent to:
```

MarkvanMents (commented):
I've responded to some of your replies.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: Thats true. Making this query distinct would require a subquery, so a change of the description is better.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
pijuskri: Yes that seems more clear

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,264 @@
...
+An OQL expression is a query building block that returns a value or a list of values. Expressions can be one of the following:
+
+* a constant
+* a function
+* a combination of attribute names, constants, and functions connected by operators
+* a subquery
+
+OQL expressions can be used in `WHERE`, `SELECT`, `GROUP BY`, `UNION`, and `JOIN` clauses. For more information, see [OQL clauses](/refguide/oql-clauses/).
+
+## Aggregations
```
dahfjkg: I'm referring to this in Clauses using tag #aggregates. Please add it here

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,265 @@
...
+---
+
+## Introduction
+
+An OQL expression is a query building block that returns a value or a list of values. Expressions can be one of the following:
+
+* a constant
+* a function
+* a system variable
+* a combination of attribute names, constants, system variables and functions connected by operators
```
dahfjkg: A subquery can also be a part of "a combination of attribute names, constants, ...", so I'd put this line last and mention subquery in it

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,265 @@
...
+* a subquery
+
+OQL expressions can be used in `WHERE`, `SELECT`, `GROUP BY`, `UNION`, and `JOIN` clauses. For more information, see [OQL clauses](/refguide/oql-clauses/).
```
dahfjkg: 1. I think `HAVING` too;
2. For `JOIN` clauses, I'd rephrase it as "and `ON` conditions of `JOIN` clauses"

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,265 @@
...
+
+```sql
+SELECT COUNT(Stock) AS ProductCount FROM Sales.Product
+```
+
+| ProductCount |
+|:------------:|
+|      3       |
+
+There are duplicate values in the `Name` column, which might not want to be counted separately. [Distinct](/refguide/oql-clauses/#Distinct) can be used to get the number of unique rows:
```
dahfjkg: Let's explain how `DISTINCT` works with `AVG` and `SUM` too

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -0,0 +1,265 @@
...
+* as an attribute in a `SELECT` clause
+* as a condition in a `HAVING` clause 
+
+When combined with a `GROUP BY` clause, aggregations in `SELECT` can be used with any column normally available, not just those specified in the `GROUP BY` clause. Aggregation is performed over groups separately in this case.
+
+### Syntax
+
+```sql
+  COUNT ( * )
+  | { COUNT | AVG | MAX | MIN | SUM | STRING_AGG } ( [ DISTINCT ] attribute_path )
```
dahfjkg: This syntax is not entirely correct: `DISTINCT` is allowed with all functions except `STRING_AGG`

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -100,7 +103,17 @@ SELECT AVG(Stock) AS ProductCount FROM Sales.Product
 
 | ProductCount |
 |:------------:|
-|    34.333    |
+|    36.75     |
+
+There are duplicate values in the `Stock` column, which can be ignored by using [DISTINCT](/refguide/oql-clauses/#Distinct). The query below returns the average unique stock:
+
+```sql
+SELECT AVG(Distinct Stock) AS ProductCount FROM Sales.Product
```
dahfjkg: Let's use more representative aliases for different aggregate functions. Both in queries and in result tables.

dahfjkg:
Passed technical review

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
MarkvanMents: I've reordered them as the initial description has it ordered COUNT (*) and then COUNT | AVG | MAX - so it seems better to present the descriptions in this order.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
MarkvanMents: Thanks for the explanation. I've added a link to the API Docs.

dahfjkg:
This can be closed as it is merged into https://github.com/mendix/docs/pull/8431

MarkvanMents:
No unresolved comments. Merged into #8431 so closing this PR.

PR #8442, opened 2024-10-08
[APPEXT-857] Add release note for IWebView.PostMessage API in 10.15.1

bvanreeven:


MarkvanMents (approved):
This was released but had the wrong label and so never got merged.
Merging now.

PR #8453, opened 2024-10-10
Add info box about Web vs Classic version in Consumed REST Service

quinntracy:


content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services-beta.md
```diff
@@ -52,6 +52,12 @@ Download [Studio Pro](https://marketplace.mendix.com/link/studiopro/) and add th
 
 Use the Consumed REST Service to configure a `GET`, `POST`, `PUT`, `PATCH`, or `DELETE` request for your app. 
 
+{{% alert color="info" %}}
+
+As of Mendix 10.16.0, you can configure a Consumed REST service using the Web Version or the Classic Version. It is recommended to use the Web Version, as the Classic version will be removed in a future release.
```
aimiliatheologimendix: Please change the `Web version` to `New Version`. It was decided by the team that the customers are not interested in the underlined technology that we are using, so it's better to just call it New

PR #8460, opened 2024-10-11
Make it explicit project is not removed

feddovanede:
Make it explicit the project is not removed when a Free App is archived.


MarkvanMents (approved):
Thanks Feddo. Looks good. Approving and merging.

PR #8476, opened 2024-10-15
Added information how to analyze and shrink MPR file

tknippenberg:


tknippenberg:
@grootjans @robertvmx can you proofread this and comment whether you think more/other information is needed?

tknippenberg:
Converted back to draft as we want to do some further work on `analyze-mpr` tooling before publishing.

PR #8501, opened 2024-10-18
Describe clearing environment better DEP-6444

MarkvanMents:


content/en/docs/deployment/mendix-cloud-deploy/environments-details.md
```diff
@@ -87,8 +87,13 @@ To clear your environment, follow these steps:
 
 1. Click **Clear Environment**.
 2. Select one of the following options:
-    * **Only clear the database** – This empties all data from your database. After you confirm the deletion, the application is stopped, the existing database is deleted, a new database is created, and the application is restarted. Ensure you have a backup of any data that you want to keep.
+
+    * **Only clear the database** – This removes all data from your database tables. Ensure you have a backup of any data that you want to keep.
+
+        After you confirm the deletion, the application is stopped, the existing data is removed from the database, and the application is restarted. Note that the database retains some information, such as logs and system data.
+
     * **Clear the full environment (model and database)** – This clears all data from your database and file storage. It also removes your app from this environment. Clear the full environment if you want to deploy a different app to the environment.
```
MarkvanMents, line 97: Do we need to say anything here, or does the database instance also get removed here so there are no logs left?

feddovanede: If you clear the full environment, we do the same clearing of the database as when selecting the database only option. Additionally, we also clear the app.

MarkvanMents: @feddovanede - 
The **Only clear the database** option does not mention the file storage, but this one does.
Do both options clear the file storage - in which case we should perhaps mention that on the first option as well.
Would this be clearer:
```suggestion
    * **Clear the full environment (model and database)** – This clears all data from your database and file storage in the same way as described above. It also removes your app from this environment.
    
        Use **Clear the full environment…** if you want to deploy a different app to the environment.
```

feddovanede: No, only the full clean clears the file storage. The Database Only option only clears the database, and doesn't touch the file storage.

MarkvanMents: OK, I'll leave this as is and merge the PR.

PR #8506, opened 2024-10-21
Move Selenium doc from Integration to Front End

quinntracy:


content/en/docs/howto/front-end/selenium-support.md
```diff
@@ -1,8 +1,10 @@
 ---
 title: "Configure Selenium Support"
-url: /howto/integration/selenium-support/
+url: /howto/front-end/selenium-support/
 weight: 16
 description: "Describes how to use naming conventions and avoid pitfalls using Selenium."
```
ConnorLand: Replace description with something better that features Selenium earlier in the sentence. These descriptions are auto-pulled onto the parent doc which gives an overview of what each doc does. Maybe something like

"Describes Selenium best practices, including advice on how to name widgets and avoid common pitfalls."

ConnorLand: pitfalls -> best practices (see below)

content/en/docs/howto/front-end/selenium-support.md
```diff
@@ -1,8 +1,10 @@
...
+aliases:
+    - /howto/integration/selenium-support/
 ---
```
ConnorLand, line 8: And the last thing I think you should do is rename "pitfalls" to "best practices". When discussing painful parts of the UX experience, we want to be as transparent about problems/solutions as possible, while also conveying the most confidence in our product possible.

PR #8508, opened 2024-10-21
Move Integration Extensions How-tos to Refguide

quinntracy:


MariaShaposhnikova:
Hi @quinntracy ,
Overall it looks good! Some suggestions for improvements:

1. Some images appear very big, could they be resized?
2. Could you add the **Read More** section to the newly moved docs? 
3. Perhaps the External database connector documentation could use a bit more explanation how three documents relate to each other? We have /refguide/external-database-connection/, /refguide/use-the-external-database-connector/, /appstore/modules/external-database-connector/.

quinntracy:
@MariaShaposhnikova Which **Read More** section are you referring to?

PR #8520, opened 2024-10-22
Documentation for OQL expression syntax

pijuskri:
Documentation for OQL expression syntax following the new structure.
Includes data types, literals, operators, functions.
Needs team review
To be merged with other OQL documentation MR's


content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+| `OR`     | Logical disjunction      | Logical    |
+| `AND`    | Logical conjunction      | Logical    |
+
+Binary operators are used with this syntax:
+```sql
+	expression operator expression
+```
+Where `operator` is any available binary operator. Both `expression` operands should be of compatible types for the operator and compatible with the other operand.
+
+#### Type coercion precedence
```
dahfjkg: I'm referring to this using `#type-coercion` tag. Please ad it here.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+Returns true is the left `expression` is greater or equal to the right. Both operands should be numeric.
+#### OR
+Returns true if at least one input `expression` returns true. Both operands need to be of type `BOOLEAN`.
+#### AND
+Returns true if both input `expression` return true. Both operands need to be of type `BOOLEAN`.
+
+Its main use is to make complex `WHERE` conditions with a combination of input values. In the query below large orders or smaller orders with a high value are selected:
+
+```sql
+SELECT LastName, Number, Price FROM Sales.Order WHERE Number > 5 OR (Price > 5 AND Number > 2)
```
dahfjkg: In an example with parentheses, I'd put `OR` in parentheses and `AND` outside to illustrate how they affect the precedence. Currently, the parentheses do not affect the result.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+```sql
+	SELECT Revenue, Cost FROM Sales.Finance WHERE Revenue IS NOT NULL 
+```
+| Revenue | Cost |                                                       
+|---------|------|
+| 10      | 7    |
+
+### CASE
+
+The `CASE` expression is a conditional expression, similar to if/else statements in other programming languages. Each condition is an expression that returns a Boolean result. If the condition's result is true, the value of the `CASE` expression is the result that follows the condition, and the remainder of the `CASE` expression is not processed. If the condition's result is not true, any subsequent `WHEN` clauses are examined in the same manner. If no `WHEN` condition yields true, the value of the `CASE` expression is the result of the `ELSE` clause. If the `ELSE` clause is omitted and no condition is true, the result is null.
```
dahfjkg: I'd say it's more similar to `CASE` expression in other languages (the ones that support it)

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+`expression` specifies the string to convert.
+
+#### Example
+The function is useful to enforce consistent case for all strings, especially for comparisons. The query below would return no results in case-sensitive databases, as there is only a "Doe":
+
+```sql
+SELECT * FROM Sales.Customer WHERE LastName = 'doe'
+```
+
+Using `LOWER` this inconsistency can be fully avoided:
```
dahfjkg, line 867: True, but we would also not use the index anymore. I think it's worth mentioning

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+| `IS`     | Tests if a value is `NULL`                                      |
+
+#### LIKE
+Matches an `expression` to the pattern after the operator. 
+
+##### Syntax
+```sql
+expression LIKE pattern
+```
+Where `expression` is of type `STRING` and `pattern` is a string literal or parameter. A `NULL` pattern is treated as an empty string.
```
dahfjkg: I'd mention explicitly that having a function on the right side of like does not work, but on the left side it's fine

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+SELECT End, Revenue FROM Sales.Period
+WHERE End < RANGEEND($range_past)
+```
+| End                 | Revenue |
+|---------------------|---------|
+| 2024-06-02 15:12:45 | 10      |
+
+### REPLACE
+
+The `REPLACE` function replaces all occurrences of a specified string value with another string value. The function supports limited and unlimited `STRING` types. Arguments of other types are not supported.
```
dahfjkg: I'd mention the first parameter, something along the lines of: The `REPLACE` function takes an input string and replaces all occurrences within it of a specified string value with another string value

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,937 @@
...
+ROUND ( expression , length )
+```
+
+##### expression
+
+`expression` is any numeric expression to be rounded. If `expression` is `NULL`, the function will return `NULL`.
+
+##### length
+
+`length` specifies the amount of decimals to which the `expression` must be rounded. Must be of numeric type. If the value is of type `DECIMAL`, there is database specific rounding. In PostgreSQL and MySQL, `length` is rounded. With other databases, `length` is floored. If the `length` is `NULL`, the function result will be `NULL`.
```
dahfjkg: > In PostgreSQL and MySQL, `length` is rounded

I would mention explicitly that it is rounded to the closest value of the required scale. Similarly, I would explain a bit more what flooring means.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
pijuskri: The simple syntax is more like switch statements, while the extended is more like if else. I can mention that

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -752,7 +761,7 @@ The function is useful to enforce consistent case for all strings, especially fo
 
 ```sql
 SELECT * FROM Sales.Customer WHERE LastName = 'doe'
-```
+```f
```
dahfjkg: Probably a typo

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: TBH, now it sounds even more complicated, so let's keep the previous version after all. It is somewhat similar to if/else after all 😅

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -763,6 +772,8 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
 |----|----------|-----------|
 | -  | Doe      | John      |
 
+Note that this query caj no longer take advantage of an index for `LastName` for comparison, resulting in a performance decrease.
```
dahfjkg: `caj` is a typo

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: "is rounded to the nearest natural value" is not entirely correct, I  think. `ROUND(1.128, 2) == 1.13`, so it gets rounded not to a natural value, but to a value of requested scale

pijuskri: Misunderstanding here, im writing about the value of `LENGTH` itself getting rounded if its a decimal and not an integer. I will rewrite this to be more clear.

dahfjkg: Oh, indeed a misunderstanding. TBH, I would consider leaving this whole part out because it is a very specific detail that only makes things confusing, as we just saw.

pijuskri: Yeah perhaps i write too many details sometimes... This can be left out

dahfjkg:
Passed technical review

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+
+| Format | Example         | Data Type            | Description                                   |
+|--------|-----------------|----------------------|-----------------------------------------------|
+| -      | `TRUE`, `FALSE` | `BOOLEAN`            | Conditional constants                         |
+| 's*'   | 'my_string'     | `STRING`             | String literal                                |
+| d+     | 5               | `INTEGER` and `LONG` | Natural number literal                        |
+| d+.d+  | 5.3             | `DECIMAL`            | Real number literal                           |
+| -      | `NULL`          | -                    | NULL literal to represent non-existent values |
+
+Where `d` is a number, `s` is any character, * indicates zero or more characters, + indicated one or more characters.
```
MarkvanMents, line 37: Is "+" always a numeric character?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+'[%SystemVariable%]'
+```
+
+These variables can be used the same way as other expressions. This query gets the names of all `Person` objects that are associated with the current user:
+```sql
+select Name from Person where Person_USER = '[%CurrentUser%]'
+```
+
+## Operators
+Operators perform common operations and do not use parenthesis in their syntax. They take `expression` as input, which can be other operator results, functions, columns and literals.
```
MarkvanMents, line 61: You mean you cannot force precedence using parentheses, or something else?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+| ID | LastName | Number | Price |                                                         
+|----|:---------|--------|-------|
+| -  | Doe      | 7      | 1.5   |
+| -  | Doe      | 2      | 5     |
+| -  | Moose    | 3      | 8.2   |
+
+
+The operator can be used to modify an attribute in SELECT.
+```sql
+SELECT (Number + 5) FROM Sales.Order
```
MarkvanMents, line 134: Why does this return `LastName`?
And you say above that "Operators … do not use parentheses in their syntax" and then show an example which has parentheses around the operator expression.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+|--------------------|---------------------------------------|
+| `%`                | Matches zero or more of any character |
+| `_`                | Matches one of any character          |
+
+In order to search for special characters, they should be escaped with the `\` escape character (including `\` itself).
+
+##### Examples
+Presume we have 3 strings for column `House`: `Apartment`, `Tenement` and `Flat`. We can select all string ending with "ment" with this condition:
+```sql
+Select House FROM House LIKE '%ment'
```
MarkvanMents, line 350: This looks a bit ambiguous. Do the entity and attribute have the same name?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+| Apartment |
+| Tenement  |
+
+A certain length of string can be enforced with the use of the `_` operator This query matches any string that has 4 of any character ending with "mend"":
+```sql
+Select House FROM House LIKE '____ment' 
+```
+| House     |                                                         
+|-----------|
+| Apartment |
```
MarkvanMents, line 366: Wouldn't this be rejected as it has five characters before "ment"?
```suggestion
```

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,948 @@
...
+	expression operator expression
+```
+Where `operator` is any available binary operator. Both `expression` operands should be of compatible types for the operator and compatible with the other operand.
+
+#### Type coercion precedence {#type-coercion}
+
+Binary operations perform type casting when operand types are differing. For operations involving only numeric types, data types are always upcasted to ensure data types are matching. The result type will also be the operand type with the highest precedence. This is done according to this ordering:
+- `DECIMAL`
+- `LONG`
+- `INTEGER`
```
MarkvanMents: Should we refer to CAST here?

MarkvanMents (commented):
Nearly finished reviewing this. Still need to do from COALESCE.
Looks good. Just a couple of questions/comments.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: I guess you mean that they aren't functions and so parameters (`expressions`) are not in parentheses. Like it isn't `+(a, b)` but is `a+b`.
I think we could state this more clearly?
```suggestion
Operators perform common operations and, unlike functions, do not put their parameters in parentheses. They take `expression` as input, which can be other operator results, functions, columns and literals.
```

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1047 @@
...
+This interval can be used to calculate the average revenue per month:
+
+```sql
+SELECT Revenue : DATEDIFF(MONTH, End, Start ) as avg_revenue FROM Sales.Period
+```
+
+| avg_revenue |                                                        
+|-------------|
+| 2           |
+| 10          |
```
MarkvanMents: Nice example.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1047 @@
...
+| `QUARTER`    | 1, 2, 3 or 4                                     | 3                                           |
+| `MONTH`      | 1 to 12                                          | 7                                           |
+| `DAYOFYEAR`  | 1 to 366                                         |                                             |
+| `DAY`        | 1 to 31                                          | 5                                           |
+| `WEEK`       | 1 to 53 (depends on the database implementation) |                                             |
+| `WEEKDAY`    | 1 to 7 (1 = Sunday, 7 = Saturday)                | 6                                           |
+| `HOUR`       | 0 to 23                                          | 16                                          |
+| `MINUTE`     | 0 to 59                                          | 34                                          |
+| `SECOND`     | 0 to 59                                          | 20                                          |
+| `MILISECOND` | 0 to 999                                         | 356                                         |
```
MarkvanMents, line 789: I corrected this to `MILLISECOND` above. Should it be
```suggestion
| `MILLISECOND` | 0 to 999                                         | 356                                         |
```
?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1047 @@
...
+
+##### timezone
+
+`timezone` specifies the time zone to use for the retrieval. This parameter is optional and defaults to the local time zone. It should be a string literal containing an IANA time zone. GMT offset time zones are not supported.
+
+#### Examples{#oql-datepart-example}
+
+| datepart     | Definition                                       | Example (Friday July 1, 2005, 16:34:20.356) |
+|--------------|--------------------------------------------------|---------------------------------------------|
+| `YEAR`       |                                                  | 2005                                        |
```
MarkvanMents, line 779: Are there years where this won't work? e.g. 835 CE or 3500 BCE?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1047 @@
...
+The function is useful to enforce consistent case for all strings, especially for comparisons. The query below would return no results in case-sensitive databases, as there is only a "Doe":
+
+```sql
+SELECT * FROM Sales.Customer WHERE LastName = 'doe'
+```
+
+Using `LOWER` this inconsistency can be fully avoided:
+
+```sql
+SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
```
MarkvanMents, line 870: Would it be a better practice to LOWER both sides of the = expression? (even through it is unnecessary in this example).
```suggestion
SELECT * FROM Sales.Customer WHERE LOWER(LastName) = LOWER('doe')
```

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -0,0 +1,1047 @@
...
+
+The syntax is as follows:
+
+```sql
+REPLACE ( expression, pattern, replacement )
+```
+
+`expression` specifies the string to be searched.
+
+`pattern` specifies the pattern to search for. In the function output, all occurrences of the pattern will be replaced with the value of `replacement`.
```
MarkvanMents, line 975: Does this have to be a specific string, or are regular expressions allowed?

MarkvanMents (commented):
I have finished my review of this one as well.
Can you have a look at [my changes](https://github.com/mendix/docs/pull/8520/files/58a2be04898e96199d5977014ce5300263820f74..e1a302fb2d14048f167a7702650750f4d9541af3), and comments?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: + is not a character at all. Here Pijus uses regex syntax, but apparently does not explain it clear enough.

In regex, "d+" means "one or more occurrences of d"

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: The change looks good

dahfjkg:
This can be closed as it is merged into https://github.com/mendix/docs/pull/8431

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: How exactly do you mean?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: I think it really depends on the database, but I would rather not include such details

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: I'd say for understanding how LOWER works this example is good because if you have LOWER on both sides, who knows how it transforms the string

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: 🤦

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: OK, let's leave it at this - we can't cover everything.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Good point. Let's leave it.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Good change.

MarkvanMents:
All review comments applied in #8431. Closing.

PR #8550, opened 2024-10-28
PMP release notes

katarzyna-koltun-mx:


mengdix (commented):
Release notes 1.15
Line 24:
Correction: not disabling for all deployed apps at the same time. Rather, disabling it moving forward. So no retroactive effect. Suggest:
"now supports disabling the default MxAdmin account for apps at deployment.";
Additionally, needs a bit more clarification, which I'm happy to add:
"This setting effectively removes the ability to access the app using the default system admin account, and is only recommended when combined with default use of an SSO module to avoid lockout."
Line 26:
Grafana metrics are not a new feature, please move to Improvements
Line 36
Add mention that the Platform settings import and export functionality have been restored. However, the change is what you wrote already.
Line 52
replace "supported" with "compatible"

mengdix (approved):
Good!

PR #8557, opened 2024-10-28
Explain generate full debug info

emirmx:
Can be merged when MR 26526 is merged.


MariaShaposhnikova:
@emirmx , thanks for the PR! Do you refer to an internal MR? Is it for 10.16.0?

emirmx:
> @emirmx , thanks for the PR! Do you refer to an internal MR? Is it for 10.16.0?

Yes internal MR the implements the setting. It will probably included in 10.17.

PR #8560, opened 2024-10-29
Move Catalog how-tos to Refguide

quinntracy:


MariaShaposhnikova:
@quinntracy , the new place looks good, but I suggest urls of the documents to be changed to refguide to avoid confusion (It now says how-tos). In this case alliases are added to the metadata.

Issue #8568, opened 2024-10-30
translation-generator


RCatersels:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [translation-generator](https://docs.mendix.com/refguide/translation-generator/)

Document link: [translation-generator.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide/modeling/mendix-ai-assistance/translation-generator.md)

## My Issue/Suggestion
It is not clearly stated that only empty values will be translated. I would suggest to add this in 2.1.

Yiyun333:
Hi @RCatersels,

Thanks a lot for your feedback. I've created a PR to add this information: https://github.com/mendix/docs/pull/8569/files.
The PR will be merged by the end of the day and I'll close this issue then. Please let us know if you have any further feedback!

Best,
Yiyun

PR #8569, opened 2024-10-30
Add a note to clarify only untranslated texts are translated

Yiyun333:
To address:
https://github.com/mendix/docs/issues/8568


Yiyun333:
Reviewed by PM.

PR #8590, opened 2024-11-01
Note difference in size between backup and database

MarkvanMents:
Added disclaimer that backup size does not equate to the database size


helina-belete (commented):
Looks good! The only minor suggestion I have (now looking at this again) is to replace the word "backup" with "snapshot" so it shows:

`The Snapshot Size does not match the size of the actual database within the application. This does not contain a copy of the database but rather contain instructions on how to recreate the database.`

PR #8601, opened 2024-11-04
[DUX-143] Add link to View Entity survey in editor

adeel2023:
Customers can fill out a survey about OQL View entities. We added a separate hidden document for this so that Studio Pro 10.17.0 can link to it directly. In the future, once the survey closes after some time, we will replaces the text with "Use the idea forum to submit feedback"

This should be merged before Studio Pro 10.17.0 is released.


MarkvanMents (approved):
Looks great. Thanks for the PR.
Approving.
It will be merged with all the rest of the 10.17 documentation at the same time as 10.17 is released.

PR #8603, opened 2024-11-04
References to "column" should be to "attribute"

MarkvanMents:


content/en/docs/refguide/runtime/data-storage/attributes-type-migration.md
```diff
@@ -24,16 +24,16 @@ Key | Means
 --- | ---
 **&#x2713;** | Conversion always possible.
 **\*<sup><small>note</small></sup>** | Conversion is not always possible, or data will be changed during conversion. See related note for more information. If conversion is not possible, the behavior is the same as for "**X**", below.
-**X** | Conversion not possible. The original column will be removed and a new column will be created with default values for the existing rows.
+**X** | Conversion not possible. The original attribute will be removed and a new attribute will be created with default values for the existing rows.
```
basdebakker: In this case I prefer the original wording. The attribute is still there, though with a different type. It is the database column that is removed and recreated (by the runtime).

MarkvanMents: Fixed in 9607637e527d25e790ede5e0720b268532e1cd43

MarkvanMents:
Approved over Slack.

PR #8621, opened 2024-11-06
Update teamcenter-extension.md - Nov 6

wanchoorohan:


content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
```diff
@@ -43,7 +43,7 @@ The compatibility among Teamcenter Extension, Teamcenter Connector, and Studio P
 | ------------- | ------------- | ------------- |
 | 1.0.0 | 3.6.1, V 3.6.0, V 3.5.0 | 10.6.5 thru 10.7 |
 | 2.0.0 | 2406.0.0 | 10.12.1 and above |
```
MarkvanMents: Is this only patch versions of 10.12? It says below that 10.13 thru 10.15 are not supported?
If so, perhaps
```suggestion
| 2.0.0 | 2406.0.0 | Only 10.12, patch versions 1 and above |
```

content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
```diff
@@ -241,29 +241,19 @@ Once you finished import mapping, click **Generate** to create microflows for th
...
-| Error message                                           | How to Solve the Error                                       |
-| ------------------------------------------------------- | ------------------------------------------------------------ |
-| `TeamcenterToolkit.BOMWindow` no longer exists          | <ol><li>Update all references to <br/>`TcConnector.BOMWindow`.</li><li>Search for `TeamcenterToolkit.BOMWindow` to find `BOMapping` parameters where the entity is used, and change it to `TcConnector.BOMWindow`.</li></ol> |
-| `TeamcenterToolkit.top_line` no longer exists           | Update all associations to `TcConnector.top_line`.           |
-| The selected Java action parameter […] no longer exists | Set all the Java action parameters again.                    |
-
-### Resolving the Errors – Solution 2 {#solution-2}
-
 {{% alert color="info" %}}
 Using Solution 2 has an advantage: after completing the procedure, the integrations will appear on the **History** tab.
```
MarkvanMents: This is now Solution 1 - so
```suggestion
Using Solution 1 has an advantage: after completing the procedure, the integrations will appear on the **History** tab.
```

content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
```diff
@@ -241,35 +240,37 @@ Once you finished import mapping, click **Generate** to create microflows for th
...
 {{% /alert %}}
 
-As an alternative to solution 1, you can perform the following steps:
+Perform the following steps:
```
MarkvanMents, line 262: I think the solution numbering has gone a bit wrong, so I've removed it from here.

MarkvanMents (commented):
Hi @rohanwanchoo123 ,
I've proofread the changes in these two commits: https://github.com/mendix/docs/pull/8621/files/332f1e1fe5e9aad6173a45671afae8c89fb68756..47c2c332faed95361b6d8e705c09ac8d5a26580c
I've also made a couple of suggestions.
Please can you let me know if the suggestions are good and you are happy with my changes and I can merge and publish this PR.

content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
wanchoorohan: Makes sense

content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
wanchoorohan: Makes sense

content/en/docs/appstore/platform-supported-content/modules/teamcenter-extension.md
wanchoorohan: This is for all versions starting from 10.12.1 and above. It'll not work with 10.12.0. Due to certain exceptions 10.13 - 10.15 aren't supported either which is why we put a special note at the bottom. 

Should we put an "!" or notification so the developer doesn't missing seeing the note?

MarkvanMents: So it does work in 10.16 and above. I think it is best to put this in the table and not have a note at the bottom. Then customers will definitely not have an excuse to miss it.
If I've understood it correctly it should be
```suggestion
| 2.0.0 | 2406.0.0 | 10.12, patch versions 1 and above and 10.16.0 and above |
```
And then we can delete the note about 10.13 to 10.15?

wanchoorohan: I guess that's fine. However, we do want to talk about the possible workaround if they use the unsupported studio pro versions. Should we still keep that statement in the note?

MarkvanMents (approved):
Approved and merging.

PR #8649, opened 2024-11-13
Limit to Premium

feddovanede:
The application data replication setting will only be available for customers with a Premium subscription.

The setting will be applied to all environments that are provisioned, for existing environments the setting will not be updated. The setting is applied to environments with a Standard, Premium or Premium Plus resource pack, it will not be applied to environments with a Basic plan.


katarzyna-koltun-mx:
@feddovanede Reviewed - shall I publish directly or wait for the go ahead?

katarzyna-koltun-mx (approved):

PR #8674, opened 2024-11-19
Update import-a-large-excel-file.md

mahendra-mahajan-mendix:
Introduced a new way to create a relatively big Entity without any manual effort using Data Importer extension.

create-entity-using-excel-input


quinntracy:
HI @mahendra-mahajan-mendix, is this tied to a release or can it be merged once reviewed?

PR #8682, opened 2024-11-20
10.17.0 Studio Pro Release Notes

MariaShaposhnikova:


Robinfr:
Would it be possible to add the release note from #8696 as well?

PR #8683, opened 2024-11-21
Add documentation for optional parameters

Robinfr:
In 10.17, we are introducing optional parameters for microflows/nanoflows. Updated the documentation accordingly.


Yiyun333 (approved):

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
```diff
@@ -25,6 +25,18 @@ The data type of a parameter defines the type of the value that it expects. See
 
 Default: *Object*
 
+### Argument {#argument}
+
+{{% alert color="info" %}}
+This feature was introduced in Studio Pro 10.17.0.
+{{% /alert %}}
+
+Determines whether the argument is optional or required. When the argument is **optional**, it can be ommitted when calling the microflow/nanoflow. In case it is **required**, an argument must be passed.
```
EnasAbdelrazek: It would be helpful to add a screenshot explaining how to make an argument optional and set a default value.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
```diff
@@ -25,6 +25,18 @@ The data type of a parameter defines the type of the value that it expects. See
...
+
+#### Default Value
+
+When an argument is set to **optional**, a default value can be set. The default value is used when the argument is omitted, not when the value is empty.
```
EnasAbdelrazek: Is it clear to users what empty value means, or should we add something like (null) for clarification?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
Yiyun333: @EnasAbdelrazek I can add the screenshot but can you please provide me with one? I tried to look for a 10.17 build but couldn't find one yet.

Robinfr: I was pondering this, but I didn't see a screenshot for anything else in this documentation so decided against it. What do the tech writers think?

I have as screenshot right here 😄 

Screenshot 2024-11-20 at 10 11 36

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
Robinfr: I used the words "empty value" because it was the same words uses in other places. Afaik we typically only use `NULL` for database operation, and we call `null` values in the client "empty". But again, this is something the tech writers know better than me.. So maybe they can comment on this? 😃

Yiyun333: @EnasAbdelrazek Sounds good to me! Is it correct that arguments can be of different data types? Then, it would indeed be better if we could clarify what "empty" means here. 
@Robinfr What do you think?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
Yiyun333: Hi both, sorry, I didn't see Robin's replies when I was adding mine.
I will add the screenshot that Robin provides here. I like the example in the screenshot :)

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
Yiyun333: Hi both,

I checked the UI, `empty `is the one that's used in the UI. So it's actually better to just stick to `empty` here.

Yiyun333:
Hi @EnasAbdelrazek @Robinfr:

I've added the screenshot and made some more changes in this commit based on the most recent feedback: https://github.com/mendix/docs/pull/8683/commits/c27d9a72ffd30a24e118d4b4d6cdce007eb246b7.

Here is a screenshot from my local build:
image

Please let me know what you think. Thanks!

Yiyun333:
> Hi @EnasAbdelrazek @Robinfr:
> 
> I've added the screenshot and made some more changes in this commit based on the most recent feedback: [c27d9a7](https://github.com/mendix/docs/commit/c27d9a72ffd30a24e118d4b4d6cdce007eb246b7).
> 
> Here is a screenshot from my local build: image
> 
> Please let me know what you think. Thanks!

Approved by Robin over Slack.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/parameter.md
Robinfr: @Yiyun333 I actually realized that the screenshot has an incorrect "Default value". It cannot have a string default value since the "Data type" is boolean 😅 I've attached a correct one here:

Screenshot 2024-11-27 at 10 53 25

Yiyun333: @Robinfr I had some doubts before but I convinced myself that it's possible 😅. What an interesting possibility, I thought. Haha, okay, I just changed the screenshot to the correct one.

PR #8693, opened 2024-11-22
Domain Model Generator supports adding attributes to an entity

Yiyun333:


content/en/docs/refguide/modeling/mendix-ai-assistance/domain-model-generator.md
```diff
@@ -33,6 +33,8 @@ Describe your application or its main goals in the chatbox, and Maia will use th
 In this dedicated chat, only requests related to domain model generation will be properly handled. If you have other questions, close this chat and go back to the general [Maia Chat](/refguide/maia-chat/) interface.
 {{% /alert %}}
 
+After a domain model is generated and if the dedicated chat is still active, you can still ask Maia to add more entities and associations to the generated domain model. In Studio Pro 10.17 and above, it also supports deleting or editing what are already in the generated domain model. This includes adding more attributes to an existing entity. Have a try!
```
Yiyun333: Hi @richardburleigh-mx,
In Studio Pro 10.17 and above, does it also support deleting what's in the generated domain model?

richardburleigh-mx: > does it also support deleting what's in the generated domain model?

Thank you @Yiyun333  for asking. No, we have decided not to allow deleting items at this time, to protect the integrity of customer projects.

xding78: @richardburleigh-mx @Yiyun333 , this is XInrong. about the second sentence "In Studio Pro 10.17 and above ...", I thought users can already add stuff on top of already generated stuff. I fail to see what's new in 10.17 and above.

xding78: > > does it also support deleting what's in the generated domain model?
> 
> Thank you @Yiyun333 for asking. No, we have decided not to allow deleting items at this time, to protect the integrity of customer projects.

In fact, I think Studio Pro "not deleting" entities automatically is a design principle. And when users manually delete an entity, if it's in use, users will first see a confirmation. This confirmation includes the usage of the entity. So that users can make an informed decision.

Yiyun333: @xding78 I just tested "add stuff on top of already generated stuff" in 10.16 and it's not supported there. I think according to the MxAssist joint sprint review last week, it's mentioned there that it would be supported in 10.17 and above? 
@richardburleigh-mx Can you please confirm again? Thanks!

xding78: @Yiyun333 @richardburleigh-mx. This is Xinrong. I think since we will remove the "Create only" chapter, we should mention creating (adding) in this block as well. For the second sentence, I suggest: 

"In Studio Pro 10.17 and above, you can not only add more to what are already generated in the generated domain model, but also edit it. This includes adding or editing attributes to existing entities. Have a try!"

richardburleigh-mx: Thank you for the contributions @xding78 !

Also for clarification, the editing is currently limited to addition. So it's "editing" in the sense that if I ask `"Create an entity named Registrations"`, I can then say `"add the attribute RegistrationNumber to the Registrations entity"`.

Also, to give some emphasis to what was mentioned earlier, Studio Pro 10.17 doesn't YET support modifying existing domain models (we have an early version with this but decided to delay it to ensure maximum integrity of customer projects). Users will not be able to click the "Generate Domain Model" button if their domain modal is already populated.

However, Maia is aware of everything it has done in the chat, including what entities were created, so can extend these with new objects / capabilities.

Yiyun333: @richardburleigh-mx @xding78 

Thank you both so much for all the feedback and clarification! Based on Richard's comment, I decided not to use the word "edit" but to mention exactly what will be supported.

Here is the latest version:
image

I would say this PR is okay to merge tomorrow for the 10.17 release.

PR #8705, opened 2024-11-26
Deletion of Unused Free Apps (to be released on December 1)

ghost:
Scheduled release Sunday 1 December


MarkvanMents:
@loket4 
In the [Control Center documentation](https://docs.mendix.com/control-center/apps-overview/#introduction) we say that you should search on App ID. Is this being changed to Project ID?

I think we need somehow to link Project ID and App ID in the documentation if this isn't the case.

content/en/docs/releasenotes/developer-portal/_index.md
```diff
@@ -19,6 +19,17 @@ To see the current status of the Mendix Portal, see [Mendix Status](https://stat
 
 ## 2024
 
+### December 1, 2024
+
+#### Improvements
+
+* We expanded the response of our [Projects API](/apidocs-mxsdk/apidocs/projects-api/) to include details on the company owning a particular app.
+
```
MarkvanMents, line 27: @loket4 - do we not want to mention that we renamed App ID to Project ID? Seems like quite an important thing to say?

MarkvanMents (commented):
@loket4 
Should we add that we renamed App ID to Project ID as part of the release note?
I've made various changes to support this change, can you quickly review.
Thanks.

MarkvanMents (approved):
Agreed to discuss App->Project release note with PM outside this PR.
Approving and merging.

PR #8723, opened 2024-11-28
Update link to widget data source in data-view.md

ragutislt:
Updates the link which should point to the page 'Listen to widget source' instead of 'Microflow source'


MarkvanMents (approved):
Hi Aurimas
Thanks for spotting this error and submitting a PR to change it.
Approving and merging.
Thanks for making the Mendix documentation better.
Mark

PR #8728, opened 2024-11-29
Add section on empty values in XPath expressions

basdebakker:


content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
```diff
@@ -177,6 +177,27 @@ Conversions work in the following way:
...
+
+* `[Attribute1 = Attribute2]` is `true` if both attributes are not empty and the values are equal. It is `false` if both attributes are empty.
+* `[Attribute1 != Attribute2]` is `true` if both attributes are not empty and the values are not equal. It is `false` if one attribute is empty and the other is not.
+
+There are some exceptions that are treated as special cases:
+
+* `[Attribute = empty]` is `true` if the attribute is empty.
+* `[Attribute != empty]` is `true` if the attribute is not empty.
+* `[Attribute != 'value']` is `true` if the attribute does not match the constant `'value'` or if the attribute is empty.
+* `[not(Attribute = 'value')]` is `true` if the attribute does not match the constant `'value'` or if the attribute is empty.
```
MarkvanMents: Can we not make this simpler and say
```suggestion
* `[Attribute = 'value']` is `false` if the attribute does not match the constant `'value'` or if the attribute is empty.
```
Otherwise the double-negative makes it more difficult to follow?

MarkvanMents (commented):
@basdebakker - I've made some changes but hopefully not messed with the logic.
And I've made one suggestion which (I think) makes it clearer.
Let me know what you think.

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
basdebakker: No. While the changed version is correct, it says something quite different. The changed version already follows from the general rule, in that it is false if the attribute is empty. But the original version is an exception to the general rule.

Though now that I'm writing this, I realize this only makes sense when you know the complete general rule, which is more than I've written. The thing is, in SQL almost any operation on `NULL` just returns `NULL`, so if `Attribute` is NULL, then `Attribute = 'value'` is NULL and then `not(Attribute = 'value')` is also NULL, which (when used as a condition) is treated as `false`.

Therefore, if you know SQL, you would expect `[not(Attribute = 'value')]` to be `false` when the attribute is empty. But because Mendix treats this case specially, it is actually `true`.

On the other hand, `[Attribute = 'value']` is not treated specially and behaves exactly as you would expect if you know SQL.

MarkvanMents: Ah - OK. So my change is wrong. I won't commit it. Sorry to cause confusion.
I also changed the text to say `true` instead of `match` - is that also misleading? I found it clearer, but perhaps I should go back to "match" as being more indicative of what the database is actually doing?

basdebakker: I think the "is `true`" is fine. I'm less sure about "attribute does not match the constant". I would say "attribute is not equal to the constant". To me, the verb "match" suggests some kind of pattern matching.

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
```diff
@@ -177,6 +177,27 @@ Conversions work in the following way:
...
+
+If the value of an attribute in the database is empty, this is interpreted as "unknown" and any comparison of that attribute to another attribute will not match, even if both are empty.
+This follows the semantics of `NULL` values in SQL, to which XPath expressions are translated.
+
+Here are two examples:
+
+* `[Attribute1 = Attribute2]` is `true` if both attributes are not empty and the values are equal. It is `false` if both attributes are empty.
+* `[Attribute1 != Attribute2]` is `true` if both attributes are not empty and the values are not equal. It is `false` if one attribute is empty and the other is not.
+
+There are some exceptions that are treated as special cases:
```
dahfjkg: I would mention explicitly that all these exceptions are related to the situation when one of the sides is constant. Or even split this in two subsections: comparison of attributes when one of them is empty and comparison of attribute to constant.

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
```diff
@@ -177,6 +177,27 @@ Conversions work in the following way:
 * A String is converted to a Boolean by comparing it to the string `'true'` in a case insensitive way.
 * A String is converted to a Decimal or Integer/Long by parsing it as a number.
 
+### Empty values
+
+If the value of an attribute in the database is empty, this is interpreted as "unknown" and any comparison of that attribute to another attribute will not match, even if both are empty.
+This follows the semantics of `NULL` values in SQL, to which XPath expressions are translated.
+
+Here are two examples:
```
dahfjkg: I would also mention the behavior of `<`, `>`, `<=`, `>=`. As I understand, those are always false if one side is empty.

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
dahfjkg: Also, we have a section on `empty` and `NULL` in XPath Keywords and System Variables document. Maybe the part about constants belongs there?

dahfjkg (changes requested):

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
basdebakker: The idea was that this falls under "any comparison".

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
basdebakker: 1. How about we change this line to "If one of the sides of the comparison is a constant, there are some exceptions that are treated as special cases:"?
2. Does "the part about constants" refer to the last two examples below? Those are not about the keywords. Maybe a reference from there to here?

dahfjkg: 1. Sounds good
2. True, then indeed maybe adding a reference would work

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-expressions.md
dahfjkg: I would also add examples of what happens when a boolean is null: in that case, both `[flag]` and `[not(flag)]` are false

basdebakker: Mendix does not allow boolean values to be empty.

dahfjkg:
Passed team review

MarkvanMents (approved):
Thanks for the PR. Sorry the final review got delayed.
Approving.

MarkvanMents:
Approved in Slack.

PR #8732, opened 2024-11-29
Add information about deleting list variables in loops and the new temporary Microflows.RemoveVariableOnDeleteObjectsActivity setting

jveldhuizen:


content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -62,6 +62,7 @@ The following custom settings can be configured:
...
 | <a id="JavaKeyStorePassword" href="#JavaKeyStorePassword">JavaKeyStorePassword</a> | Password for the default Java keystore. | changeit |
 | <a id="LongLivedSessionTimeout" href="#LongLivedSessionTimeout">LongLivedSessionTimeout</a> | This setting is the same as `SessionTimeout`, but specific to offline-first progressive web apps. | 604800000 (7 days) |
+| <a id="MicroflowsRemoveVariableOnDeleteObjectsActivity" href="#MicroflowsRemoveVariableOnDeleteObjectsActivity">Microflows.<wbr>RemoveVariableOnDeleteObjectsActivity</a> | Defines if the pre-Mendix 10.17 behavior should be enabled where variables were removed after execution of a **Delete object(s)** microflow activity. This is applicable to loops where list variables were silently unusable after a **Delete object(s)** microflow activity. Newly added items were not propagated to activities using these variables. That's fixed in Mendix 10.17. This setting is meant to temporarily bring back the old behavior. | `false` |
```
emirmx: This is only applicable to list variables, right? Normal variables still become inaccessible after delete?

Also, I think, we can move the part starting with `This is applicable to loops...` to the warning block in the main section.

jveldhuizen: We think it's only noticeable in case of list variables, but in fact, internally, this change keeps object (non-list) variables too, but makes them empty internally. They are still accessible in loops, like before, but they are useless as they are empty afterwards (instead of deleted before this fix). The end result is the same, but there is a risk that the Runtime behaves differently on unexpected places. Therefore, this setting has to bring the old behavior back for single object variables too. It's still difficult what to document. We think people will only really see a change in behavior for lists. But we cannot rule out that it has a broader impact. So the setting is broad (list and single object variables), but the documentation is more about the list variables.

I come back on your second remark next week. That part can indeed be a littlebit more clear. It's at least missing that the part where these variables are still useful, is in loops.

jveldhuizen: I changed the text in the warning block a little bit. Can you review it again?

emirmx: Looks goo!

emirmx:
For me this looks good!

emirmx:
> For me this looks good!

I approve the changes from our side.

MarkvanMents:
Thanks for this PR - sorry about the delay in getting back to you.
I've done a proofread in a6a18d3f148e2a6506f0ad9e0435cf6b68f9482a and have removed all the references to "variables" and changed them to list **references**. Variables are something different
image

@emirmx / @jveldhuizen does it still give the right information, and can I merge it?

jveldhuizen:
@MarkvanMents, thanks for your response. But I don't fully understand the difference between variables and references. Internally, all names of things where you can refer to in microflow activities or expressions (with $name) are called variables. Internally, there is no difference between input parameters, create variable activities and names of activity resuls (like Retrieve). Everywhere you can use them, you can use them all, based on the type. We are handling them all equal. Is it clear for end-users to keep different names for them inside a microflow?
But now I'm thinking about it, is it true that you want to reserve the 'variable' keyword for variables which you can change? I'm still afraid that having multiple words for things you can refer to in expressions at the same way, is a little bit confusing. In my eyes, you have variables you can change (the ones you create via 'Create variable') and variables you cannot change (return types of other activities, $currentSession, $currentUser, $latestError, etc.). How do you call these last things b.t.w., like $currentSession?
What will happen if we make it possible to allow `Change variable` on lists in the future? Then the hard split in names is not possible anymore. But OK, we will not do that I think, so there won't be a problem in reality.
If you already have this distinction in the documentation, it's good to have it here as well.

MarkvanMents:
Hi Jonathan - would it help to have a quick call?
My point is that there are things in Mendix microflows called variables which are basically primitive data types (integer, string, etc) and are only used as temporary storage in memory.
A list is not one of these things, and anything which is stored in a list is not one of these things, it is an object (perhaps non-persistable, but still an object). So to a user of a microflow I want to avoid using the word "variable" when I'm referring to an object or a list as you cannot create it using "create variable" or manipulate it using "change variable". So I use things like "list name" or "object name". In this case I think "reference to a list" is perhaps OK - but perhaps I should look at saying "name" instead?

I guess that in the backend (Java and C#) everything is a variable, whether it is a primitive or an object (class) reference - but this is not how the user sees it which adding actions to a microflow and we should use their terminology rather than ours.

Does that make sense - I'm happy to chat about it, although I'm pretty busy today.

MarkvanMents:
Since 10.17 is live, is my rewrite clear enough to publish, or do you have concerns about it that need to be addressed immediately?

jveldhuizen:
Thanks for the explanation. It's fine to publish this now.

MarkvanMents (approved):
Thanks for the PR.
Merging.

PR #8750, opened 2024-12-04
Apply zoom effect only to images without a "link" parameter

Yiyun333:


MarkvanMents (approved):
Looks good. Tested and seems to work.
Thanks for investigating and picking this up, @Yiyun333

PR #8751, opened 2024-12-04
Re-structure the offline installation section of the documentation

rodrigocmoreira:
Today users would have a hard time succeeding when trying to do an offline installation

### DIFF
--- a/content/en/docs/refguide/installation/install.md
+++ b/content/en/docs/refguide/installation/install.md
@@ -77,7 +77,7 @@ The prerequisites are the following:
 
     | Studio Pro 10.0.0 - 10.9.0 | Studio Pro 10.10.0 and above |
     | --- | --- |
-    | Gradle 7.6.X (7.6.3 or above) | Gradle 8.5  |
+    | Gradle 7.6.3 or above | Gradle 8.5  |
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/installation/install.md
```diff
@@ -77,30 +77,46 @@ The prerequisites are the following:
 
     | Studio Pro 10.0.0 - 10.9.0 | Studio Pro 10.10.0 and above |
     | --- | --- |
-    | Gradle 7.6 or above | Gradle 8.5 or above |
+    | Gradle 7.6.X (7.6.3 or above) | Gradle 8.5  |
```
quinntracy: @rodrigocmoreira Are there scenarios where anything below Gradle 7.6.3 could be used, or does it always have to be 7.6.3 or above? If the latter, I would adjust to just say "Grade 7.6.3 or above" rather than using an X placeholder.

rodrigocmoreira: Technically, it's better if the user always goes to 7.6.3 or above. With older 10.X versions we still install 7.6 but users are recommended to upgrade to 7.6.3 whenever possible.

quinntracy: In that case, I would change it to say Gradle 7.6.3 or above to ensure users make the upgrade

PR #8780, opened 2024-12-10
Snowflake Cortex AI

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Notes from review with Chris:

1. Add information about available functions to https://docs.mendix.com/appstore/modules/genai/ (to be verified with Lina and team)
2. ANOMALY DETECTION is an ML function so it shouldn't be listed as a GenAI function. I may move it to the Snowflake RestSQL doc instead. Also, the intro shouldn't mention ML functions if I do.
3. Add info from the latest sprint demo to the showcase app section.

linaaaf (commented):
PO OK :)

linaaaf (approved):

katarzyna-koltun-mx:
@linaaaf I've made some changes after feedback from Chris, let me know if you have any additional comments.

Issue #8779, opened 2024-12-10
Broken images in page document templates


Andries-Smit:
https://docs.mendix.com/refguide/document-templates/

![image](https://github.com/user-attachments/assets/15d2278a-9dac-4fc9-a35f-394338a0a9e7)

Yiyun333:
Hi @Andries-Smit,

Thanks a lot for reporting this! The issue is fixed in this PR: https://github.com/mendix/docs/pull/8782. You will be able to see the fixed images online by the end of the today.

Best,
Yiyun

PR #8782, opened 2024-12-10
Fix rendering issue with images in an alert box

Yiyun333:
To address:
https://github.com/mendix/docs/issues/8779


MarkvanMents (approved):
Look s good Yiyun. Solves the issue when I build the site using this branch.
Thanks.

PR #8785, opened 2024-12-11
React moving into GA

ConnorLand:
I need to comb all docs and ensure that when React goes GA in 10.18, the docs have no more beta flags.


ConnorLand:
This is ready to go

PR #8788, opened 2024-12-11
Domain Model Generator updates in 10.18

Yiyun333:


Yiyun333:
Reviewed by the team.

PR #8820, opened 2024-12-23
Add 4XL-5XLDB CRP

feddovanede:
Add 4XL-5XLDB CRP, with 4XL app resources and 5XL database resources.


MarkvanMents (approved):
@feddovanede 
Looks good. Merging.
**Is there any reason that all our current packs have the number 21 at the end. We are just going into 2025 and if these are going to remain our packs going forward, then perhaps a name without the (20)21 in it might make us look more up-to-date?**

PR #8835, opened 2024-12-30
Remove duplicate header

dahfjkg:
image


MarkvanMents (approved):
Thanks for spotting this and correcting it for us.
Approving and merging.
@MariaShaposhnikova FYI

Issue #8498, opened 2024-10-17
rollback-object


Auke-Enexis:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [rollback-object](https://docs.mendix.com/refguide/rollback-object/)

Document link: [rollback-object.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/rollback-object.md)

## My Issue/Suggestion

In the introduction, the following is phrased.

"The Rollback object activity can be used to undo changes (that have not been committed) made to the object in the part of the flow preceding the activity."

This true but incomplete. It is easier to say it rolls back all changes, not persisted to the database.

The current phrase implicates that it only rolls back uncommited changes in the current (sub/parent)flow

MarkvanMents:
Thank you for your suggestion. I have updated the documentation and it will be published later today.
Sorry about the delay in responding to you, but I hope it won't put you off reporting any other improvements that can be made to the Mendix documentation.
Thanks for wanting to improve the Mendix documentation.
Mark

PR #8837, opened 2024-12-31
Improve rollback description - see issue rollback-object #8498

MarkvanMents:
#8498


akkrishnakumar:
I approve this change.

Reason being - 

The one mentioned in the docs (“made to the object in the part of the flow preceding the activity”), it feels like it refers to the flow in the current executing MF. This could be misleading because any changes made in the parent MF will also be reverted/rollbacked if not committed.

But the one suggested by the customer (rolls back all changes, not persisted to the database), it clearly says that All Changes , be it in the current flow or parent flow will be reverted/rollbacked if not committed. This seems to be more correct, IMO

MarkvanMents:
Thanks for the review @akkrishnakumar.
Merging this change now.

PR #8840, opened 2025-01-02
[RUN-4280] Add release note for commons-io upgrade

atennapel:
We upgraded commons-io in 9.24.30 and 8.18.33, but we did not add a release note at the time. Because of a customer question related to this we would now like to add the release note retroactively.


MarkvanMents (approved):
@atennapel 
Thanks for the PR, Albert.

Issue #8662, opened 2024-11-15
export-to-excel


jduc-smartwave:
Page link: [export-to-excel](https://docs.mendix.com/refguide/using-the-excel-exporter/)

Document link: [export-to-excel.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide/modeling/integration/use-excel-documents/export-to-excel.md)

## My Issue/Suggestion

In the microflow that calls the Java Action "GenerateExcelDoc", if you set apply entity access = true you may encounter an unwanted exception due to OQL unable to retrieve data.

It might be great to add a note that warns about this potential issue, and to avoid it, ensure that Apply entity access = false

MarkvanMents:
Hi @jduc-smartwave 
Sorry for the delay in following up on this issue.
The developers are not aware of this exception and would like to know more about it. Before adding this to the documentation, they would like to know why it is happening as it is probably a bug.
Could you send more information, ideally a stack trace, about what is going on so that the developers can investigate it.
This issue is probably not the best place to do this, so I suggest you [raise a support ticket](https;//support.mendix.com) where you can securely attach the information.
If you can mention this issue there, and let me know the support ticket number here, then we can link the two things together.

Thanks for wanting to improve Mendix.

PR #8848, opened 2025-01-02
Export to Excel microflow should not apply entity access..

MarkvanMents:
Addresses #8662


MarkvanMents:
Closing in favour of #8865

PR #8861, opened 2025-01-06
Update connector-guide-best-practices.md

mahendra-mahajan-mendix:
add a caveat for unlimited string for sorting or grouping by operations.


ConnorLand:
Saw it was appstore and assigned myself, but then realized @quinntracy is the connector go-to.

PR #8877, opened 2025-01-08
File uploader new context object committed before upload

Andries-Smit:
When a new context object is not committed before the multiple upload starts, the upload of the file will trigger the commit via the associated to the object.  When context is not in the DB many simultaneous request might cause the the commit to happen multiple times. Cause an error:

```
com.mendix.systemwideinterfaces.connectionbus.data.UniqueConstraintViolationRuntimeException: Duplicate column value violates its unique constraint: 
integrity constraint violation: unique constraint or index violation ; 
SYS_PK_10349 table: "fileuploader$fileuploadcontext" (SQL State: 23505, Error Code: -104) 
Detail Message: org.hsqldb.HsqlException: integrity constraint violation: unique constraint or index violation ; 
SYS_PK_10349 table: "fileuploader$fileuploadcontext"
```


takumagott:
Note: Please wait with reviewing. We will see if we can provide a quick fix for this issue next sprint. If so, we will not need this disclaimer.

Andries-Smit:
Should be fixed in file uploader  v1.0.2

PR #8884, opened 2025-01-08
Mentioned data snapshot upload when uploading app with test data

tknippenberg:


tknippenberg:
Closed PR. Apparently reproduction was incorrect, and previously documented behavior is correct.

PR #8922, opened 2025-01-15
Update: react-readiness for idx modules

Karuna-Mendix:


Karuna-Mendix:
Approved by Jaap Franke on slack. Merging now.

PR #8931, opened 2025-01-16
Update security-findings-faq.md

mruiserrmendix:
According to SDM (Arjen) from Runtime, Mendix 9 and 10 already use Dojo 1.7.3 which has a fix for this issue.


MarkvanMents (approved):
Thanks for the PR,
I've updated it to be clearer to customers using older versions of Mendix.
Luckily it encourages them to move to the LTS version of Mx 9.
Approving and merging.

PR #8941, opened 2025-01-21
SCIM: Add info about upgrading the SCIM

Karuna-Mendix:


harshita-raj25:
Merge into production

Karuna-Mendix:
> Merge into production

@harshita-raj25, let me know when the module will be released. I will release the PR after that.

JaapF (commented):
Looks good to me, you can merge

PR #8947, opened 2025-01-22
Default access in beta access rule editor

WimJongeneel:


MarkvanMents (approved):
Thanks for this @WimJongeneel 
I have added back the pre-10.19 behavior as many customers will be on 10.18 for the next 6 months as it is MTS.
Approving.

EnasAbdelrazek (approved):

PR #8958, opened 2025-01-23
Update the security overview docs 

EnasAbdelrazek:
For 10.20:
- Document the Page, Nanoflow, microflow access tabs
- Document the Export to Excel functionality


WimJongeneel:
Screenshot does still have the old export icon

EnasAbdelrazek:
@WimJongeneel Could you please update the screenshot?

content/en/docs/refguide/modeling/security/security-overview.md
```diff
@@ -46,7 +46,51 @@ Multiple columns are shown for entities with XPath constraints. Access rules wit
...
+
+1. Entity Access: This sheet contains the entity access rules data, including:
+    1. User Role
+    2. Module: The module containing the entity.
+    3. Entity: The name of the entity.
+    4. Member: The attribute or association name.
+    5. Kind: Specifies whether the member is an association or an attribute.
+    6. Type: The type of the member.
+    7. XPath: The XPath constraint.
+    8. XPath Caption:  The XPath constraint caption if exists.
```
WimJongeneel: " if exists" feels a bit strange, maybe something like ", when set"?

content/en/docs/refguide/modeling/security/security-overview.md
```diff
@@ -46,7 +46,51 @@ Multiple columns are shown for entities with XPath constraints. Access rules wit
...
+    9. Access: The member access right (None, Read, ReadWrite).
```
WimJongeneel: "access level" might sound nicer then "member access right"

content/en/docs/refguide/modeling/security/security-overview.md
```diff
@@ -46,7 +46,51 @@ Multiple columns are shown for entities with XPath constraints. Access rules wit
...
+
+2. Document Access: This sheet includes the accessible pages, microflows and nanoflows, including:
+    1. Document type: The type of the accessible document (Page, Microflow, Nanoflow).
```
WimJongeneel: I don't know if we should document that this (and some other) fields are translated and will be different based on your SP language.

content/en/docs/refguide/modeling/security/security-overview.md
```diff
@@ -46,7 +46,51 @@ Multiple columns are shown for entities with XPath constraints. Access rules wit
...
+    2. User Role
+    3. Module: The module containing the the page, nanoflow, or microflow.
+    4. Document: The name of the page, nanoflow, or microflow.
+
+3. Module Roles: This sheet provides the mappings between user roles and module roles in each module, including:
```
WimJongeneel: I don't think "in each module" is relevant here, it is just all mappings

content/en/docs/refguide/modeling/security/security-overview.md
EnasAbdelrazek: True, But that applies for everything in docs :)

WimJongeneel (approved):

MariaShaposhnikova:
@EnasAbdelrazek and @WimJongeneel , there is a comment that screemshots contain an old icon. Do they need an update?

MariaShaposhnikova:
@EnasAbdelrazek and @WimJongeneel , on the second thought, do we actually need a screenshot per each tab? Perhaps, we can use one screenshot in the introduction as an example of what the whole page looks like.

content/en/docs/refguide/modeling/security/security-overview.md
```diff
@@ -27,15 +27,9 @@ The **Security Overview** summarizes the app's security for a selected user role
 The **Security Overview** page has the following tabs:
 
 * [Entity access](#entity-access)
-* **Page access**
-* **Microflow access**
-* **Nanoflow access**
-
-{{% alert color="info" %}}
-In the beta release of the **Security Overview** page, only the **Entity access** tab is available. The other tabs will be available in future versions of Studio Pro.
```
MariaShaposhnikova, line 35: @EnasAbdelrazek , is Security Overview still in Beta?

WimJongeneel: Hi, yes it is still in Beta

WimJongeneel:
> @EnasAbdelrazek and @WimJongeneel , on the second thought, do we actually need a screenshot per each tab? Perhaps, we can use one screenshot in the introduction as an example of what the whole page looks like.

If you think its better you could remove all but the `static/attachments/refguide/modeling/security/app-security/user-roles/entity-security-overview.png` screenshot (that is the view you see when opening the overview).

EnasAbdelrazek:
> @EnasAbdelrazek and @WimJongeneel , there is a comment that screemshots contain an old icon. Do they need an update?

It is up to date.

MariaShaposhnikova (approved):

PR #8966, opened 2025-01-27
Removed unmaintained device capabilities list

jdiehl:
@ConnorLand


ConnorLand:
I think that it's useful to inform devs about which what supported per browser, but maybe this table is not achieving that goal. @jdiehl 

What's the most normal way a dev checks which features are compatible with which platforms/browsers? A site like https://caniuse.com/, or the official Mozilla docs? 

Because I think we can delete our table, but if we do than we do need to provide devs with a bit more guidance on the following; 
(1) which common PWA features depend per browser
(2) how they devs check these features using 3rd party tools.

jdiehl:
@ConnorLand I have added a link to mdn documentation

ConnorLand (approved):

PR #8975, opened 2025-01-28
[RUN-4426] Add com.mendix.core.LenientDateTimeParsing to Runtime Customization page

atennapel:
The `com.mendix.core.LenientDateTimeParsing` runtime setting was added in 10.0.0 but was never added to the Runtime Customization page, this PR rectifies that oversight.


emirmx:
Checking from runtime side. Looks good to me.

content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -46,6 +46,7 @@ The following custom settings can be configured:
...
 | <a id="commendixcoreisClusterSlave" href="#commendixcoreisClusterSlave">com.<wbr>mendix.<wbr>core.<wbr>isClusterSlave</a> | Set to `true` in a high-availability scenario when this is *not* the [Cluster Leader](/refguide/clustered-mendix-runtime/#cluster-leader-follower). The buildpack will usually enforce this setting, but it may need to be set for some on-premises deployments. | `false` |
+| <a id="commendixcoreLenientDateTimeParsing" href="#commendixcoreLenientDateTimeParsing">com.<wbr>mendix.<wbr>core.<wbr>LenientDateTimeParsing</a> | The `parseDateTime` function will use more lenient parsing, as it did in Mendix 9, when set to `true`. For example `parseDateTime("yyyyMMdd", "2021-11-10")` will return `2020-11-01` as a date. If set to `false`, which is the default, this example will throw an error. It is recommended to not enable this setting to avoid unexpected results. | `false` |
```
MarkvanMents: @atennapel 
I think there is a typo - 2021-11-10 should return 2021-11-10 not 2021-11-01? Or am I missing something.
I wonder if it would be even clearer to say that 

> this example will throw an error as it is expecting `20211110`

just for completeness?

MarkvanMents (changes requested):
Hi @atennapel 
I think there is a typo in the example - and I've made a suggestion for making the information more complete.

content/en/docs/refguide/runtime/custom-settings/_index.md
atennapel: No, it's not a typo, that's a consequence of the lenient datatime parsing. It's unexpected but the "-1"s in the datatime string will be interpreted as "-1" and substract from the year, shifting the other values. This is why we now use strict datetime parsing by default in Mx10, to avoid these unexpected results. We added this runtime setting just in case there's a customer that really wants to use lenient parsing.

> this example will throw an error as it is expecting 20211110
Yes that would be fine to add.

MarkvanMents:
```suggestion
| <a id="commendixcoreLenientDateTimeParsing" href="#commendixcoreLenientDateTimeParsing">com.<wbr>mendix.<wbr>core.<wbr>LenientDateTimeParsing</a> | When set to `true`, the `parseDateTime` function will use more lenient parsing, as it did in Mendix 9. For example `parseDateTime("yyyyMMdd", "2021-11-10")` will return `2020-11-01` as a date if this is set to true, which is probably not the intended result.. If set to `false`, which is the default, this example will throw an error as it is expecting `20211110`. It is recommended to not enable this setting to avoid unexpected results. | `false` |
```

MarkvanMents (approved):
Looks good,
Thanks for the PR
Approving and merging.

PR #8999, opened 2025-01-31
Mendix SSO: Using Page URL

Karuna-Mendix:


Karuna-Mendix:
Got approval via Slack. Ready to be merged when informed.

PR #9013, opened 2025-02-03
User type info update for the IAM modules.

Karuna-Mendix:
https://mendix.atlassian.net/browse/TW-2522

### DIFF
--- a/content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
+++ b/content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
@@ -32,16 +32,28 @@ This approach is for end-users who are already set up in your app. For new end-u
 
 Outlined below is an example of a module that can be used to update UserType attribute. You will need to adapt the module logic for classifying your own internal and external end-users. 
 
+### Assigning UserType for Existing Users of IAM Modules
+
+The simplest method to set the user type is by using the IAM modules, which require only configuration without the need to develop a microflow. Alternatively, you can build a custom microflow as described in the [Populating UserType for Existing Users of an App](#using-microflow) section below.
+
+When connecting your app with an IdP, set up the user type through the capabilities of the OIDC SSO, SCIM, or SAML module. The user type is now configured in the User Provisioning, which is integrated into the OIDC SSO, SCIM, and SAML modules. This means you can directly configure end-users of your application as `internal` or `external` in the **User Provisioning** tab of your app. Based on this configuration, users are updated each time they log in. These modules allow you to set the user type per IdP as the source of your end-users, assuming that separate IdPs are used for `internal` and `external` users.
+
+For more information, refer to the User Provisioning section of the following modules:
+
+* [OIDC SSO](/appstore/modules/oidc/#custom-provisioning-rt)
+* [SAML](/appstore/modules/saml/#custom-provisioning-rt)
+* [SCIM](/appstore/modules/scim/#user-provisioning)
+
 ### Domain model
 
-In the example below, our aim is to update UserType attribute of UserReportInfo entity. However, the entity `UserReportInfo` is protected in the System module and has no access rules. As a result, it cannot be exposed directly in the UI pages. 
+In the example below, our aim is to update UserType attribute of `UserReportInfo` entity. However, the entity `UserReportInfo` is protected in the System module and has no access rules. As a result, it cannot be exposed directly in the UI pages. 
 Therefore, the approach we take is to create a new non-persistable entity, `UserTypeReport`, which we will populate based on the values of `UserReportInfo` to show in the UI.
 
 {{< figure src="/attachments/howto/monitoring-troubleshooting/populate-user-type/usertypereport.png" class="no-border" >}}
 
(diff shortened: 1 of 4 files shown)


content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
```diff
@@ -75,3 +75,9 @@ Therefore, the approach we take is to create a new non-persistable entity, `User
     {{< figure src="/attachments/howto/monitoring-troubleshooting/populate-user-type/user-type-report.png" class="no-border" >}}
 
 7. The report can be exported into an Excel file.
+
+### Assigning UserType for Existing Users of an IAM Modules
+
+When connecting your app with an IdP, the easiest way to set up the user type is through the capabilities of the OIDC SSO, SCIM, or SAML module. The user type is now configured in the User Commons, which is integrated into the OIDC SSO, SCIM, and SAML modules. This means you can directly configure end-users of your application as `internal` or `external` in the **User Provisioning** tab of your app. Based on this configuration, users are updated each time they log in.
```
JaapF: I think we agreed some time ago to not mention user Commons in user documention. Reason was that it's mostly an "internal" building block to create common functionalities.
So my suggestion would be to remove the 2nd sentence and tweak the 3rd sentence.

JaapF: Plesae add something like:

These modules allow you to set the Usertype _per IdP_ as the source of your end-users
We assume that you are using separate IdP's for your internal versus your external users.

content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
```diff
@@ -75,3 +75,9 @@ Therefore, the approach we take is to create a new non-persistable entity, `User
     {{< figure src="/attachments/howto/monitoring-troubleshooting/populate-user-type/user-type-report.png" class="no-border" >}}
 
 7. The report can be exported into an Excel file.
+
+### Assigning UserType for Existing Users of an IAM Modules
```
JaapF: Can we put this method first in the document?

And maybe some introduction like:
- easiest way to set user type is using the IAM modules. Just config, no need to develop a microflow.
- alternatively build your own microflow as described in section x.y.z

content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
```diff
@@ -32,16 +32,28 @@ This approach is for end-users who are already set up in your app. For new end-u
 
 Outlined below is an example of a module that can be used to update UserType attribute. You will need to adapt the module logic for classifying your own internal and external end-users. 
 
+### Assigning UserType for Existing Users of IAM Modules
+
+The simplest method to set the user type is by using the IAM modules, which require only configuration without the need to develop a microflow. Alternatively, you can build a custom microflow as described in the [Populating UserType for Existing Users of an App](#using-microflow) section below.
```
MarkvanMents: Perhaps put the list of IAM modules at this point rather than (or as well as) at the end of the section.
Customers might not immediately link SAML to IAM, or might not know what IAM options we offer.
For example
> The simplest method to set the user type is by using the IAM modules, which require only configuration without the need to develop a microflow. Mendix provides the following IAM modules:
> {list of modules}
> 
> Alternatively …

content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
```diff
@@ -32,16 +32,28 @@ This approach is for end-users who are already set up in your app. For new end-u
 
 Outlined below is an example of a module that can be used to update UserType attribute. You will need to adapt the module logic for classifying your own internal and external end-users. 
 
+### Assigning UserType for Existing Users of IAM Modules
```
MarkvanMents: I think this should be an H2 section above the **## Approach** section above. It isn't really part of that section and interrupts the flow.

Then have the whole of the **## Approach** section after the IAM section and call it something like **## Assigning UserType Using a Microflow**

I'm also not sure that this is really for "Existing Users", it is for all users who use SSO through an IAM module. It is also set up for new users (according to the documents above).

MarkvanMents (changes requested):
Hi Karuna,
Content looks good, but I've made some suggestions about where the section should go.
I think the new content is it's own H2 section rather than being an H3 as part of another section.
Other than that, I think it is good to go.
Mark

content/en/docs/howto/monitoring-troubleshooting/populate-user-type.md
Karuna-Mendix: done

Karuna-Mendix:
> Hi Karuna, Content looks good, but I've made some suggestions about where the section should go. I think the new content is it's own H2 section rather than being an H3 as part of another section. Other than that, I think it is good to go. Mark

Thank you Mark! I am implementing your feedback.

Karuna-Mendix (commented):
Implementing the feedback.

Karuna-Mendix (commented):
done

PR #9014, opened 2025-02-03
Replace LoginAction with authentication microflows for published services

joostverhoog:
Mendix 10, does not use LoginAction for published services, so for apps that use that, users must reimplement that functionality in authentication microflows.


adeel2023:
Looks good.

PR #9024, opened 2025-02-04
Clarify data validation behavior during commits

emirmx:


MariaShaposhnikova:
Hi @emirmx , are these changes for a specific release?

emirmx:
> Hi @emirmx , are these changes for a specific release?

No. They are not. But we still need to review them within our team.

atennapel:
Approved from the Runtime side.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -51,6 +51,10 @@ Indicates whether or not to execute the commit event handlers of the objects.
 
 Default: **Yes**
 
+{{% alert color="warning" %}}
+Validations, except unique validations, are not performed when committing without events. See [data validation](/refguide/setting-up-data-validation/) for more information.
```
MarkvanMents: I wonder if it is better to turn this around since we are talking about "With Events". Something like
```suggestion
Validations, except unique validations, are implemented as event handlers. Therefore they are not performed when **With events** is set to **No**. See [Setting Up Data Validation](/refguide/setting-up-data-validation/) for more information.
```

content/en/docs/refguide/modeling/domain-model/setting-up-data-validation.md
```diff
@@ -13,7 +13,7 @@ This document explains how you can set up data validation with Mendix. Make sure
 
 ## Data Validation on Entity Level
 
-This section explains how you can add validation rules to the domain model of your module. Validation rules are always triggered when changes to an object are committed. 
+This section explains how you can add validation rules to the domain model of your module. Validation rules are triggered when changes are submitted, changed in microflows, and when new objects are committed with events.
```
MarkvanMents: Why have you split out new objects here? You can also choose *Commit without events* for changes in microflows which means that they won't be run there either.

MarkvanMents (changes requested):
Hi @emirmx 
Thanks for the PR.
I've made a suggestion for a text update in the committing objects document.

In setting up data validation, I think that we could be more clear about when the validation rules are triggered - the way it is written now implies that the with events only applies to new objects, but I think it applies in all cases?

Also, in setting up data validation, we don't mention anywhere that **unique** is always applied, even without events being run. I think that would be a useful addition (we could even explain that it is applied at the database level and so doesn't use a separate event). Let me know if you agree and you want me to add that.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
emirmx: The thing is, `implemented as event handlers` only applies for commit. runtime also performs validation when changing objects, e.g. with `change object` action (even without committing). These does not depend on events. If you think above change does not confuse these cases I think it is fine.

content/en/docs/refguide/modeling/domain-model/setting-up-data-validation.md
emirmx: Ok, let me create a table so I do not miss any cases:


| Action                                   | Validations except Unique | Unique validation |
|------------------------------------------|---------------------------|-------------------|
| Change Object                            | Yes                       | No                |
| Commit without Events (New object)       | No                        | Yes               |
| Commit without Events (Committed object) | No                        | Yes               |
| Commit with Events (New object)          | Yes                       | Yes               |
| Commit with Events (Committed object)    | No                        | Yes               |

For `Change Object (with commit)` the cases for `Change Object (without committing)` is combined with the `Commit with Events`.

I tried to capture above table with the explanation :-)

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Rewritten: 1ef121cd53af339e47fe0ebe0a72309396ac2a5b

content/en/docs/refguide/modeling/domain-model/setting-up-data-validation.md
MarkvanMents: Rewritten: 1ef121cd53af339e47fe0ebe0a72309396ac2a5b

MarkvanMents (approved):
Updated and approving.
Thanks for the PR - difficult to describe edge cases, but I think this will really help the customer.

PR #9067, opened 2025-02-18
Replace <sup> HTML with Unicode superscript for readability

MarkvanMents:


MarkvanMents:
Reviewed by Connor and approved over Slack.

PR #9104, opened 2025-02-25
Clarify location of M2EE password

MarkvanMents:


content/en/docs/refguide/runtime/monitoring-mendix-runtime.md
```diff
@@ -27,7 +27,12 @@ The request needs to be of the **POST** type with **No Authorization** and the f
 * Content-Type: **application/json**
 * X-M2EE-Authentication: **yourM2EEPassword_Base64Encoded**
 
-The M2EE password is NOT the super administrator password, but a separate password. If you have the application deployed *on premises*, you can set this password in the **settings.yaml** file, which is located in the **Apps/YourProject** folder. If you are *running the application from Studio Pro*, the M2EE password is set automatically by Mendix, and you can retrieve it from the environment variables of your application process.
+    The M2EE password is NOT the super administrator password, but a separate password. You can obtain it as follows:
+
+    * If you have the application deployed *on premises*, you can find this password in the **settings.yaml** file, which is located in the **Apps/YourProject** folder. It is the `ServerPassword` value, located under `Runtime`.
```
jveldhuizen:
```suggestion
    * If you have the application deployed *on premises*, you can find this password in the **Settings.yaml** file, which is located in the **Apps/YourProject** folder. It is the `ServerPassword` value, located under `Runtime`. 
```

content/en/docs/refguide/runtime/monitoring-mendix-runtime.md
```diff
@@ -27,7 +27,12 @@ The request needs to be of the **POST** type with **No Authorization** and the f
 * Content-Type: **application/json**
 * X-M2EE-Authentication: **yourM2EEPassword_Base64Encoded**
 
-The M2EE password is NOT the super administrator password, but a separate password. If you have the application deployed *on premises*, you can set this password in the **settings.yaml** file, which is located in the **Apps/YourProject** folder. If you are *running the application from Studio Pro*, the M2EE password is set automatically by Mendix, and you can retrieve it from the environment variables of your application process.
+    The M2EE password is NOT the super administrator password, but a separate password. You can obtain it as follows:
+
+    * If you have the application deployed *on premises*, you can find this password in the **settings.yaml** file, which is located in the **Apps/YourProject** folder. It is the `ServerPassword` value, located under `Runtime`. 
+    * If you are *running the application from Studio Pro*, the M2EE password is set automatically by Mendix, and you can retrieve it from the environment variables of your application process.
```
jveldhuizen:
```suggestion
    * If you are *running the application from Studio Pro*, the M2EE password is set automatically by Mendix, and you can retrieve it from the `M2EE_ADMIN_PASS` environment variable of your application process (`javaw.exe`).
```

However, if I'm allowed to say more, I don't know why this split was made between Studio Pro and Service Console. The Service Console has that Settings.yaml file were the `ServerPassword` is stored, where Studio Pro is creating that password on-the-fly I think. But in both situations, you have a `javaw.exe` process with the `M2EE_ADMIN_PASS` environment variable. So I'm fine with it when you merge both options to one with only the `M2EE_ADMIN_PASS` environment variable way and when you remove the `Settings.yaml` option.

Furthermore, for the Windows Service Console, the admin port isn't the same as configured in Studio Pro. It's configured inside the Service Console in another form and named 'Admin server port'.

Furthermore, is it clear that the explanation behind *on premises* in this case means the Windows deployment with the Windows Service Console and not https://docs.mendix.com/developerportal/deploy/unix-like/?  
But nice is, as the documentation for https://github.com/mendix/m2ee-tools/blob/master/examples/full-documented-m2ee.yaml#L157 shows, the needed password is set in a environment variable too. And it's the same environment variable as under Windows, as the Runtime server has to pick it up, so it's almost since Mendix 2 or 3 the same variable. So we could say: in all situations you have to find the `javaw.exe` or `java` ([see here](https://github.com/mendix/m2ee-tools/blob/master/src/m2ee/config.py#L364)) process and find the `M2EE_ADMIN_PASS` environment variable for that process.  
So if you want, you can describe three situations (Studio Pro, Windows Service Console and m2ee-tools) in one sentence instead of splitting them.

content/en/docs/refguide/runtime/monitoring-mendix-runtime.md
MarkvanMents: Rejecting this change - rewriting the whole section

content/en/docs/refguide/runtime/monitoring-mendix-runtime.md
MarkvanMents: Thanks for this comprehensive information, @jveldhuizen. Great that you are thinking more broadly than just my changes here.
I have tried to summarize the detail above in this commit: e6c313b3ef724e324e851cea0617c1aca32b236b - can you have a look and see if I've understood it. There may be other things we can add to help customers, but if they are doing this I think we have to assume they are quite technically knowledgeable.

MarkvanMents (commented):
Hi @jveldhuizen 
Thanks for the thorough review and thinking of more use cases and how to simplify the information.
Can you look at my change and see if it reflects the way that customers can get the password for all of their on-premises and local deployments.
Thanks
Mark

content/en/docs/refguide/runtime/monitoring-mendix-runtime.md
jveldhuizen: Thanks! I think this is clear enough and enough general.

MarkvanMents:
Thanks for your review. Merging.

PR #9112, opened 2025-02-27
Replace image with code snippet: Build Microflow Actions with Java

Karuna-Mendix:
Closing this PR and changes are done via [this](https://github.com/mendix/docs/pull/10050) PR


content/en/docs/howto/extensibility/howto-connector-kit.md
```diff
@@ -59,9 +59,40 @@ To create a generic action using type parameters, follow these steps:
 
 The Java implementation still uses strings to specify the name of an entity, which means that you can upgrade your existing Java actions to use these new parameter types without having to refactor your existing code.
 
-Finally, here's the actual Java implementation of the action defined:
+Finally, here is the actual Java implementation of the action defined:
 
-{{< figure src="/attachments/howto/extensibility/howto-connector-kit/join_objects_javacode.png" alt="Java implementation join object" class="no-border" >}}
```
MarkvanMents, line 64: You can delete all the images now that you aren't using them any more. It makes the PR more complete and clearer about what the intention is.
(Although there is a monthly job which will do that automatically if you miss any)

content/en/docs/howto/extensibility/howto-connector-kit.md
```diff
@@ -59,9 +59,40 @@ To create a generic action using type parameters, follow these steps:
...
+{
+	private final java.lang.String Entity;
+	private final IMendixObject Object1;
+	private final IMendixObject Object2;
+
+	public JoinObjectsInList(
+		IContext context,
+		java.lang.String _entity,
+		IMendixObject _object1,
+		IMendixObject _object2
```
MarkvanMents, line 75: Do you know why these are now things like `_entity` rather than `Entity` as they were in the original image. I don't think it matters, and perhaps these are generated anyhow as they aren't in the section between BEGIN USER CODE and END USER CODE?

content/en/docs/howto/extensibility/howto-connector-kit.md
```diff
@@ -99,7 +130,27 @@ In the Java implementation for this action, you'll see the following details for
...
+	{
+		super(context);
+		this.ResultEntity = _resultEntity;
+		this.DefaultObject = _defaultObject;
+		this.InitializationMicroflow = _initializationMicroflow;
+		this.ListSize = _listSize;
+	}
+```
 
 The `executeAction` method is where all the magic happens:
```
MarkvanMents: > where all the magic happens

Looks like this document could do with a thorough review as well as I'm not sure whether we should be saying this.

But we can pick that up later - and perhaps give it to Quinn to do?

content/en/docs/howto/extensibility/howto-connector-kit.md
```diff
@@ -184,7 +235,38 @@ Implement the action in Java as follows:
...
+		try (InputStream is = new ByteArrayInputStream(this.InputString.getBytes(StandardCharsets.UTF_8))) {
+			
+			// Import the string by executing the mapping
+			List<IMendixObject> objects = Core.integration().importStream(getContext(), is, this.ImportMapping, null, false);
+			
+			// Return first object created in mapping
+			return objects.get(0);
+
+		} catch (Exception e) {
+			throw new CoreException("Failed to import JSON string: " + e.getMessage(), e);
```
MarkvanMents: This seems to be different from the original image - but if it runs correctly for the dev, then great.

MarkvanMents (approved):
Hi Karuna,
I have noted a couple of questions, but if the developer has created this and the Java does what it is supposed to do then they aren't important.

Generally a great improvement. Thanks.

content/en/docs/howto/extensibility/howto-connector-kit.md
Karuna-Mendix: I will continue to do the proofreading for the rest of the doc in this same PR.

content/en/docs/howto/extensibility/howto-connector-kit.md
Karuna-Mendix: Hi @MarkvanMents, I confirmed with the developer. Those are just names and do not affect the code.

PR #9120, opened 2025-02-28
Removed info box with duplicate content

jdiehl:
The content was wrong (MTS is supported until 3 months after the LTS release)
Also fixed the order of current MTS versions

@ConnorLand

### DIFF
--- a/content/en/docs/releasenotes/studio-pro/lts-mts.md
+++ b/content/en/docs/releasenotes/studio-pro/lts-mts.md
@@ -51,8 +51,8 @@ An LTS version is the last minor release of a major version. Its release is alig
 
 These are the current LTS versions of Studio Pro:
 
-* [9.24](/releasenotes/studio-pro/9.24/) (started in March, 2023)
 * [8.18](/releasenotes/studio-pro/8.18/) (started in March, 2021)
+* [9.24](/releasenotes/studio-pro/9.24/) (started in March, 2023)
 
 Release notes for the current LTS versions are marked with an LTS badge (<text class="badge badge-pill badge-lts" style="margin-left:0px">LTS</text>) in the left sidebar.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/releasenotes/studio-pro/lts-mts.md
```diff
@@ -95,15 +95,11 @@ An MTS is a minor version released approximately every 6 months after a new majo
 This is the current MTS version of Studio Pro:
 
 * [10.18](/releasenotes/studio-pro/10.18/) (started in December, 2024)
-* [10.6](/releasenotes/studio-pro/10.6/) (started in December, 2023)
 * [10.12](/releasenotes/studio-pro/10.12/) (started in June, 2024)
+* [10.6](/releasenotes/studio-pro/10.6/) (started in December, 2023)
 
 Release notes for the current MTS version are marked with an MTS badge (<text class="badge badge-pill badge-mts" style="margin-left:0px">MTS</text>) in the left sidebar.
 
-{{% alert color="info" %}}
```
ConnorLand, line 103: @MariaShaposhnikova Jonathan is deleting this because, quote "MTS is supported until 3 months after the LTS release". 

Just wanted to confirm -- does this sync with our existing Support agreements?

MarkvanMents: I think we need to say something about when they expire? Not just delete the whole section?

content/en/docs/releasenotes/studio-pro/lts-mts.md
```diff
@@ -95,15 +95,11 @@ An MTS is a minor version released approximately every 6 months after a new majo
 This is the current MTS version of Studio Pro:
 
 * [10.18](/releasenotes/studio-pro/10.18/) (started in December, 2024)
-* [10.6](/releasenotes/studio-pro/10.6/) (started in December, 2023)
 * [10.12](/releasenotes/studio-pro/10.12/) (started in June, 2024)
+* [10.6](/releasenotes/studio-pro/10.6/) (started in December, 2023)
```
MarkvanMents: Shouldn't these be in numerical order to make it easier to see what is going on?

content/en/docs/releasenotes/studio-pro/lts-mts.md
MariaShaposhnikova: @ConnorLand , We will have an LTS in Mendix 10 when Mendix 11 is released.  I agree with Mark, we can add a note about when they expire rather than delete the section.

MariaShaposhnikova: We can also change an example there from 9 to Mendix 10 and 11.

MariaShaposhnikova: I have confirmed that this information is incorrect. However, a bit further down in the documentation we have the correct one: "Support for an MTS version ends three months after the LTS version of that major version is released." I will keep this note deleted to avoid duplication.

PR #9127, opened 2025-03-03
Added version remark to client api

jdiehl:
@ConnorLand


MarkvanMents:
@Yiyun333 
Looks like a sensible change - you added this, I think, so I'm assigning it to you. But feel free to pass it to Connor if it is more his.

ConnorLand (approved):

PR #9148, opened 2025-03-05
Updated Custom Role as deprecated for old custom permission

Nidhi251289:
Old Custom permissions needs to be deprecated. Hence, the API is also adapted to accommodate this change


katarzyna-koltun-mx (approved):

PR #9160, opened 2025-03-07
Update react readiness

jdiehl:
@ConnorLand


MarkvanMents (approved):
Hi Jonathan,
Thanks for the PR. Runtime is actually my area. You don't usually need to tag anyone in specific on these PRs as we go through the backlog regularly and assign them to the right person. 
If something is urgent, it is usually better to use the Slack #docs channel as that will trigger us to look at the unassigned PRs.

PR #9163, opened 2025-03-07
Explain that remote entities can cause synchronization errors

jdiehl:
in particular, if the remote service is down

@ConnorLand


MarkvanMents:
@jdiehl please do not tag Connor on everything. This one is for someone else to deal with.

quinntracy (approved):

PR #9173, opened 2025-03-10
Data grid docs update (WC-2503, WC-2794, WC-2796, WC-2846)

iobuhov:


gjulivan:
NOTE:
add limitation on pagination: can’t use pagination for datasource MF/NF
workaround: user can create their own paging button and pass the current page data into their MF/NF datasource parameters

PR #9182, opened 2025-03-10
Direct Associations

MarkvanMents:


content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,59 @@
...
+* You can choose to implement direct associations for some of your associations, while leaving others as association tables
+
+## Limitations of Direct Associations
+
+You cannot replace association tables with direct associations for all types of association in your app. Direct associations have the following limitations: 
+
+* You can only use direct associations for one-to-many (1-*, default owner) and one-to-one (1-1, owner both) associations – you cannot use it for many-many associations.
+* You can only use direct associations between persistable and remote entities. You cannot use them on view entities or non-persistable entities.
+
+{{% todo %}}Originally said "remote persistent" - I don't think we document two different types of remote entity?{{% /todo %}}
```
pijuskri: Im not that knowledgeable on remote entities, but the documentation mentions the ability to make non-persistant remote entities 
https://docs.mendix.com/refguide/external-entities/#general

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,59 @@
...
+* In many cases, using direct associations can speed up queries and statements
+* They use less space in the database
+* You can choose to implement direct associations for some of your associations, while leaving others as association tables
+
+## Limitations of Direct Associations
+
+You cannot replace association tables with direct associations for all types of association in your app. Direct associations have the following limitations: 
+
+* You can only use direct associations for one-to-many (1-*, default owner) and one-to-one (1-1, owner both) associations – you cannot use it for many-many associations.
+* You can only use direct associations between persistable and remote entities. You cannot use them on view entities or non-persistable entities.
```
pijuskri: I think it would be good to be a more explicit here:

> Direct associations are only allowed between persistable entities, between local entities and from a local to a remote entity. You cannot use them on view entities, non-persistable entities or between remote entities.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,59 @@
...
+
+Enabling direct associations is simple, and has the following features:
+
+* The choice is reversible – you can decide to revert to using association tables (but see things to think about [before switching to direct associations](#before))
+* Direct associations are available in Mendix versions 10.21 and above
+* You can decide whether to enable it for a specific association, or [make it the default](/brokenlink) for all new associations
+
+For more information, see the [Association Storage](/refguide/association-properties/#storage) section of *Association Properties*.
+
+{{% todo %}}Do we want to say this without a firm timeline?{{% /todo %}}
```
pijuskri: I realize now that it would be better not to include planned features as a promise in the documentation. This is something i would like to hear pavels opinion on as well.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,59 @@
...
+
+The development pipeline includes plans to bulk migrate the database to avoid long downtime for large databases.
+
+## Before Switching to Direct Associations{#before}
+
+Before deciding to switch from an association table to a direct association, bear the following in mind:
+
+* Associations have to be rewritten to the database so migration can take a long time, especially where you have a large amount of data already stored in your database 
+* Queries are not always faster, and might not be faster in your use case
+* If you have written any custom SQL, this might break – existing XPaths and OQL queries will not be affected
```
pijuskri: Maybe to clarify: 
>If you have written any custom SQL that accesses mendix database tables directly....

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Updated.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: @dahfjkg - do you agree?

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: I've put it back in.

MarkvanMents (commented):
Addressed review comments.

content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
```diff
@@ -115,12 +116,25 @@ An object can only be deleted if it is not associated with any other object (or
...
+
+{{% alert color="info" %}}
+Association storage was introduced in Mendix 10.21.
+{{% /alert %}}
+
+In Mendix 10.20 and below, associations are always stored in association tables. This enables you to change the multiplicity easily if your design changed. In Mendix 10.21 and above, you can choose to store associations directly in the owning entity, thus removing the need for an extra association table. For more information, see [Association Storage Options](/refguide/association-storage/).
+
+| Association storage | Description | Notes |
+| --- | --- | --- |
+| Direct associations | The associations to other entities are stored as hidden attributes of the owner entity | Mendix 10.21 and above<br />Not available for many-to-many associations |
```
basdebakker: I'm not a fan of phrasing it as "hidden attributes". That doesn't mean anything. Entities, attributes and associations are an abstraction. They are stored as database tables and columns. I would say "stored as columns in the table of the owner entity".

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+* In many cases, using direct associations can speed up queries and statements
+* They use less space in the database
+* You can choose to implement direct associations for some of your associations, while leaving others as association tables
+
+## Limitations of Direct Associations
+
+You cannot replace association tables with direct associations for all types of association in your app. Direct associations have the following limitations: 
+
+* You can only use direct associations for one-to-many (1-*, default owner) and one-to-one (1-1, owner both) associations – you cannot use it for many-many associations.
+* Direct associations are only allowed between persistable entities or between a persistable entity and a persistable external entity. You cannot use them on view entities, non-persistable entities, or between external entities.
```
basdebakker: Not sure I'd call that a limitation. Those types of associations are not stored in the local database, so one cannot say that for these "you cannot replace association tables", as they don't have any in the first place.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+
+* The choice is reversible – you can decide to revert to using association tables (but see things to think about [before switching to direct associations](#before))
+* Direct associations are available in Mendix versions 10.21 and above
+* You can decide whether to enable it for a specific association, or [change the default](/refguide/app-settings/#miscellaneous) for all new associations
+
+For more information, see the [Association Storage](/refguide/association-properties/#storage) section of *Association Properties*.
+
+{{% todo %}}Do we want to say this without a firm timeline?{{% /todo %}}
+
+The development pipeline includes plans to bulk migrate the database to avoid long downtime for large databases.
```
basdebakker: The only plan we have is to make it easier in Studio Pro to change the setting on many associations. This would just be a convenience feature to avoid design time clicks and would in no way reduce the deployment downtime.

content/en/docs/refguide/modeling/domain-model/associations/_index.md
```diff
@@ -79,18 +79,22 @@ Your mouse pointer should not activate a white dot. This indicates the creation
 
 In Mendix 10.3.0 and above, you can reverse an association direction by right-clicking the association and selecting **Reverse direction**.
 
+{{% alert color="warning" %}}
+Reversing the association will result in the loss of existing associations, but the objects themselves will not be deleted.
```
dahfjkg: > Reversing the association will result in the loss of existing associations

Can we rephrase this so it is more clear that the first mention of "association" refers to the model, and the second to data?

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Introduction
+
+Association storage options were introduced in Mendix 10.21 to give you more control over how associations are implemented in the underlying database.
+
+## Association Tables vs. Direct Associations
+
+<!--- Everything except simple associations is in tables – what do you mean "except simple associations?" --->
+
+Prior to Mendix 10.21, all associations were stored in association tables. This had the advantage that you didn't have to worry about the [multiplicity](/refguide/association-properties/#multiplicity) or [navigability](/refguide/association-properties/#navigability) of the associations. You could change things as your domain model evolved.
```
dahfjkg: > Prior to Mendix 10.21, all associations were stored in association tables

That is not true for Owner and ChangedBy. Those can be used in XPaths, so I guess users perceive them as associations

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+
+In Mendix 10.20 and below associations are always implemented as association tables.
+
+In Mendix 10.21 and above, the following defaults apply:
+
+* **New projects** – one-to-many and one-to-one associations are implemented as direct associations
+* **Upgraded projects** – for projects which are upgraded from an older version of Mendix, all new associations continue to be implemented as association tables
+
+{{% alert color="info" %}}
+In your app settings you can [change the default](/refguide/app-settings/#miscellaneous) for all new associations.
```
dahfjkg: That is implied, but I would mention explicitly that the setting does not affect existing associations.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+{{% /alert %}}
+
+## Advantages of Direct Associations
+
+Because they don't have association tables, using direct associations can bring the following advantages:
+
+* In many cases, using direct associations can speed up queries and statements
```
dahfjkg: > queries and statements

This language is a bit technical, and there is no way to run a statement from Mendix (we are working on data modifications in OQL though). I would mention something closer to everyday Mendix development like data grids and data modifications.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+* You can only use direct associations for one-to-many (1-*, default owner) and one-to-one (1-1, owner both) associations – you cannot use it for many-many associations.
+* Direct associations are only allowed between persistable entities or between a persistable entity and a persistable external entity. You cannot use them on view entities, non-persistable entities, or between external entities.
+
+## Enabling Direct Associations
+
+Enabling direct associations is simple, and has the following features:
+
+* The choice is reversible – you can decide to revert to using association tables (but see things to think about [before switching to direct associations](#before))
+* Direct associations are available in Mendix versions 10.21 and above
+* You can decide whether to enable it for a specific association, or [change the default](/refguide/app-settings/#miscellaneous) for all new associations
```
dahfjkg: These options are not exclusive, so I would make them separate points

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
dahfjkg: IMO, users open docs primarily to understand existing functionality, so I agree with Pijus

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
```diff
@@ -0,0 +1,70 @@
...
+
+The development pipeline includes plans to bulk migrate the database to avoid long downtime for large databases.
+
+## Before Switching to Direct Associations{#before}
+
+Before deciding to switch from an association table to a direct association, bear the following in mind:
+
+* Associations have to be rewritten to the database so migration can take a long time, especially where you have a large amount of data already stored in your database 
+* Queries are not always faster, and might not be faster in your use case
+* If you have written any custom SQL that accesses Mendix tables directly, this might break – existing XPaths and OQL queries will not be affected
```
dahfjkg: Using " – " here is probably correct, but for me as a native Russian speaker it means that what follows adds details to what stands before it. As there are more of us working in software, let's use "but" instead 😅

content/en/docs/refguide/modeling/domain-model/_index.md
```diff
@@ -69,6 +69,8 @@ For more information on how the underlying database behaves and why data cannot
 
 ### Implementation Example
 
+{{% todo %}}Add an OrderLine-Order association which is a direct association{{% /todo %}}
```
MarkvanMents: @pijuskri and @dahfjkg I'd like to add an implementation example here. I can write it but my question is - _what is the column name for a direct association? Would it be module$orderid, for example. What happens if you have two direct associations between the same two tables._

content/en/docs/refguide/modeling/domain-model/associations/_index.md
MarkvanMents: Change attempted, please let me know if it looks OK.

```suggestion
Reversing the association in the domain model will result in the loss of existing association links between objects, but the objects themselves will not be deleted.
```

content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
MarkvanMents: Change made.
My intention when talking about them being "hidden" is that it isn't possible in OQL to actually retrieve the ID directly and manipulate it (use it to retrieve an object directly from the database). Is this something we need to say (and is it true - if not, how can customers retrieve the IDs from this column?)

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: How about
```suggestion
Prior to Mendix 10.21, all associations (except `Owner` and `ChangedBy`, where selected) were stored in association tables. This had the advantage that you didn't have to worry about the [multiplicity](/refguide/association-properties/#multiplicity) or [navigability](/refguide/association-properties/#navigability) of the associations. You could change things as your domain model evolved.
```

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Done

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Good point. How about
```suggestion
* In many cases, using direct associations can speed up populating data grids and changing entities
```

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: So miss out the second bit? I think we need to mention it somewhere and it is a limitation in that the option doesn't appear for other entities (albeit there is a very good reason for that).
```suggestion
* Direct associations are only allowed between persistable entities or between a persistable entity and a persistable external entity.
```

content/en/docs/refguide/modeling/domain-model/_index.md
basdebakker: The column name is generated from the association name, so would be like `module$orderline_order`. If there are two associations, they must have different names, so they would map to different column names.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: How about:
```suggestion
* You can [change the default](/refguide/app-settings/#miscellaneous) for all new associations
* You can enable it for specific associations
```

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Removed.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: I've removed this as we don't normally document future improvements - although we can tell individual customers if they ask about it.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Done.

MarkvanMents (commented):
Thanks for the comprehensive review. I've implemented some, but have made suggestions for the others, so if you could have a look and give the thumbs up (or add a note if you have a better suggestion) that would be great.

content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
basdebakker: Whatever they can or cannot do with this ID in OQL, there is no change compared to having it in an association table. Association storage is just that: storage. So I don't think we need to say something about it here.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
basdebakker: How about something like:

Association storage only applies to associations stored in the local database, those are associations between persistable entities or between a persistable entity and a persistable external entity. Associations from view entities, non-persistable entities or external entities are not stored in the local database, so they do not have an association storage option.

MarkvanMents:
```suggestion
* Association storage only applies to associations stored in the local database, that is associations between persistable entities or between a persistable entity and a persistable external entity. Associations from view entities, non-persistable entities, or external entities are not stored in the local database, so they do not have an association storage option.
```

content/en/docs/refguide/modeling/domain-model/_index.md
MarkvanMents: Thanks for the info. I'll let you know when I've updated this as it will need a review.

MarkvanMents (commented):
Thanks for the responses. If I don't hear anything before Monday I'll assume my other suggestions are good and commit them.

content/en/docs/refguide/modeling/domain-model/_index.md
```diff
@@ -96,10 +96,25 @@ The association `Order_Customer` is stored in the table `module$order_customer`
...
+
+The entity `OrderLine` is stored in the table `module$orderline` which is shown below. It is similar to the table of the entity `Order`. `Orderline` also has an association, `OrderLine_Order` associating each order line with the order it belongs to. This is implemented as a direct association, so the information is stored in the `module$orderline` table, as shown below.
+
+| id | module$orderline_order | productid | quantity |
+| --: | --: | --- | --: |
+| 22 | 5 | X23592 | 1 |
+| 23 | 5 | X23613 | 7 |
+| 55 | 8 | Z97D22 | 2 |
+| 57 | 8 | A49TS3 | 2 |
+
```
MarkvanMents, line 117: @basdebakker @dahfjkg and @pijuskri 
I've added this example of a direct association. Can you look at it and see if I've understood it correctly.
Thanks.

content/en/docs/refguide/modeling/domain-model/_index.md
MarkvanMents: Direct association implementation example added in d07cac1a2d6689866ae21458368f9bdd43603230

content/en/docs/refguide/modeling/domain-model/_index.md
basdebakker: The example looks good to me. I just wonder whether the top level domain model page is the right place to dig into these implementation details. (I realize that this is already the case, I have the same doubts about the existing content.)

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
basdebakker: The documentation sometimes calls them "system associations".  Also, they aren't capitalized. See https://docs.mendix.com/refguide/entities/#properties. So maybe "except the system associations `owner` ..."?

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
dahfjkg: Maybe like this?
```
In many cases, using direct associations can speed up populating data grids, changing entities and other actions that involve database interactions
```
Or even
```
In many cases, using direct associations can speed up data retrieval and modification
```

content/en/docs/refguide/modeling/domain-model/_index.md
MarkvanMents: @basdebakker thanks for your comment. It is always great to get another view on things.
I discussed this with Arjan H. and he thought it was quite important as he feels that customers don't immediately link the domain model with actual implementation of storage.
I will put something on my backlog to review this, and perhaps move implementation details to a sub-document. But I'll publish it now as it is, and look at the improvements for after the 10.21 release.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: Thanks for the review. Missed the proper references, sorry. I'll change it to
```suggestion
Prior to Mendix 10.21, all associations (except the system associations `owner` and `changedby`, where selected) were stored in association tables. This had the advantage that you didn't have to worry about the [multiplicity](/refguide/association-properties/#multiplicity) or [navigability](/refguide/association-properties/#navigability) of the associations. You could change things as your domain model evolved.
```

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
MarkvanMents: I like the shorter version, so how about:
```suggestion
* In many cases, using direct associations can speed up data retrieval and modification
```

MarkvanMents (commented):
Thanks for all the suggestions. I've committed the approved changes, and made further suggestions. Let me know if they look OK.

content/en/docs/refguide/modeling/domain-model/associations/association-storage.md
dahfjkg: Looks good

MarkvanMents:
Ready to be merged. Thanks for all the review comments.

PR #9213, opened 2025-03-18
BYOIDP

Karuna-Mendix:
https://mendix.atlassian.net/browse/TW-2345

### DIFF
--- a/content/en/docs/control-center/security/set-up-sso-byoidp.md
+++ b/content/en/docs/control-center/security/set-up-sso-byoidp.md
@@ -157,7 +157,7 @@ When you have completed your IdP setup, you can perform a test sign-in before ac
 You can test in one of two ways:
 
 * Use the **Test Link** for your configuration. It will redirect you to the login page of your IdP. Enter the credentials of a user known to your IdP. If the test succeeds, the Mendix Portal landing page opens.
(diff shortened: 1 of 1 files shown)


content/en/docs/control-center/security/set-up-sso-byoidp.md
```diff
@@ -136,9 +136,9 @@ Click **Previous** if the test indicates that you have not entered the correct c
 
 ### Mapping Between IdP and Mendix{#attribute-mapping}
 
-The data from your IdP may have different attribute names from the Mendix identity.
+Attribute mapping is a critical step in configuring SSO for the Mendix Platform. Incorrect mappings can lead to failed authentications, misconfigured user profiles, and access issues.
 
-Configure which data from your IdP is mapped to the attributes of the Mendix identity. Configure at least a **Foreign ID**, **Username**, and **First Name** or **Last Name**.
+The data from your IdP may have different attribute names from the Mendix identity. Configure which data from your IdP is mapped to the attributes of the Mendix identity. Configure at least a **Foreign ID**, **Username**, and **First Name** or **Last Name**.
```
JaapF, line 141: 'from the Mendix identity'. I don't understand what this means, so I'm assuming customers will also not understand.

My best guess would be
"Your IdP may not exactly follow the standard claims suggested by OIDC specs. So you need to configure which claims from your IdP's ID-tokens should be mapped to the attributes of the Mendix identity. Configure at least ..."

JaapF:
The PR doesn't address the issue as was flagged in the Jira ticket.
The comment was that attribute mapping and testing is scattered across a few sections.
That needs some rework.

It's a nice-to-have improvement, since most customers don't have an issue with understanding.
I was hoping you could improve the document structure. as suggested in the jira

PR #9218, opened 2025-03-18
Fix on prerequisites

MXClyde:
Documentation incorrectly states that Contributor role on subscription will also suffice for deployment, but reality is that Owner role is required.


MXClyde:
Closing this in favor of a new one

PR #9222, opened 2025-03-19
Change from port fix to cherry pick

blis-mendix:
We are changing "Port Fix" to "Cherry Pick" in Version Control in Studio Pro so we also aligned document. 

Also removed mentions of multiple commits and taking author/email as those are not valid.

Added new images updated existing one.


blis-mendix:
Jira ticket with code for ref.: [SOLP-2255](https://mendix.atlassian.net/browse/SOLP-2255)

MariaShaposhnikova:
@blis-mendix , thank you for the PR! For which SP version is this change? As I understand, this is for a future release.

blis-mendix:
Hi @MariaShaposhnikova, you know what it's a mix as I think right now.
"Port Fix -> Cherry Pick" will be merged to version 11.0
But removed content like mentions of multiple commits and taking author/email  should be applied to 10 also.

Should I than create second PR with only those being removed for 10?

PR #9220, opened 2025-03-18
Release - https://mendix.atlassian.net/projects/DES/versions/24171/ta…

Nidhi251289:
https://mendix.atlassian.net/projects/DES/versions/24171/tab/release-report-all-issues.

Also, contains some changes on Global operator Documentation based on feedback from customers.

PR #9228, opened 2025-03-19
Fix PR 9220

katarzyna-koltun-mx:


katarzyna-koltun-mx:
A fix for https://github.com/mendix/docs/pull/9220/ - the original contained a number of files that shouldn't have been included.

PR #9232, opened 2025-03-20
Remove custom runtime settings

Robinfr:
These runtime settings have been removed from Mendix as determining the device type is now done client-side.


content/en/docs/refguide9/runtime/custom-settings/_index.md
```diff
@@ -225,8 +225,6 @@ The settings below influence the behavior of the Mendix web client.
 | Name | Description | Default Value |
 | --- | --- | --- |
 | <a id="EnableKeepAlive" href="#EnableKeepAlive">EnableKeepAlive</a> | Defines whether the web client sends a keep alive request every SessionTimeout/2 milliseconds, to prevent a session timeout. Each click in the browser also acts as KeepAlive. Disabling this property will result in a user being logged out automatically after 10 minutes of inactivity, even if the browser remains open. | true |
-| <a id="PhoneUserAgentRegEx" href="#PhoneUserAgentRegEx">PhoneUserAgentRegEx</a> | Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a phone. The regular expression is matched against the user-agent header sent by the client's web browser. | Android, Mobile (iPhone, iPod, BlackBerry) |
-| <a id="TabletUserAgentRegEx" href="#TabletUserAgentRegEx">TabletUserAgentRegEx</a> | Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a tablet. The regular expression is matched against the User-Agent header sent by the client's web browser. | Android, iPad |
```
MarkvanMents, line 229: @Robinfr - which version of Mx 9 does this apply to?

content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -269,8 +269,8 @@ The settings below influence the behavior of the Mendix web client.
...
-| <a id="TabletUserAgentRegEx" href="#TabletUserAgentRegEx">TabletUserAgentRegEx</a> | Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a tablet. The regular expression is matched against the User-Agent header sent by the client's web browser. | Android, iPad |
+| <a id="PhoneUserAgentRegEx" href="#PhoneUserAgentRegEx">PhoneUserAgentRegEx</a> | *Only used in Mendix versions 10.20.x and below*<br />Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a phone. The regular expression is matched against the user-agent header sent by the client's web browser. | Android, Mobile (iPhone, iPod, BlackBerry) |
+| <a id="TabletUserAgentRegEx" href="#TabletUserAgentRegEx">TabletUserAgentRegEx</a> | *Only used in Mendix versions 10.20.x and below*<br />Defines the regular expression that is used to determine whether a user is visiting a Mendix application from a tablet. The regular expression is matched against the User-Agent header sent by the client's web browser. | Android, iPad |
```
MarkvanMents: I think we need to tell people about this if they are using earlier versions of Mendix 10.

MarkvanMents (commented):
Thanks for the PR @Robinfr 
We still need to document what happens in older versions as not everyone is using 10.21.
Can you also let me know which 9.24 version has this change?
I'm assuming this is in 10.21 as there is a related release note. Is this correct? Do we need more explanation, or is it clear why these are no longer needed?

Robinfr:
@MarkvanMents I don't believe that's relevant, we've backported this change to all 10 MTS versions, and LTS 9. It will be released over the next weeks for those versions.

Robinfr:
I also ensured that there are no customers using this for anything but the default value, so no one is really affected by this :)

MarkvanMents:
Hi Robin, 
This is in the release notes as a breaking change so I really feel that customers need to know the full impact. Obviously we can take them out in Mx 11, but there may be customers who are aware of these settings and will find it strange if they are removed - even if, as you say, they aren't using them.

I'd rather not make any assumptions about what customers are using, but document what is there. Any customer can decide to use any version of Mendix which is available, even if we advise them to use the latest MTS and LTS versions.

I can make all the changes now, so we don't have to keep having new PRs for each patch version where this change is made, if you can let me know the target versions for the change.

Robinfr:
@MarkvanMents We are not assuming anything. We have done our due diligence and analysed usage both in development and cloud environments, and found effectively no usage. We are still working on merging all the backports, but these are the likely versions:

- 10.21.0
- 10.18.5
- 10.12.16
- 10.6.22
- 9.24.34

MarkvanMents:
Thanks for the update. I'm not saying that you have missed anything, but I can imagine a customer seeing the release note and then wondering what those values did and whether they had used them, but then not being able to find any reference to them.

It is probably over cautious, but for the sake of a few words added to the docs, I'd rather do it that way. I'll add that information.

MarkvanMents (approved):
Thanks for the PR and additional information.
Approving and is ready for 10.21 release.

Robinfr:
@MarkvanMents I see you changed the text to deprecated now, but this seems to imply that it still exists as functionality, but they are removed entirely. If you want to mention something, maybe we should just mention that these options only exist before those versions?

MarkvanMents:
> @MarkvanMents I see you changed the text to deprecated now, but this seems to imply that it still exists as functionality, but they are removed entirely. If you want to mention something, maybe we should just mention that these options only exist before those versions?

@Robinfr - you are right. I meant that they were deprecated in the previous versions, but it wasn't very clear. I've changed the wording in this commit: https://github.com/mendix/docs/pull/9232/commits/4817eb5145417bfd546dd1aa93b37679b0020267

PR #9233, opened 2025-03-20
Update .NET version in 9.24 ref guide

GulamY:
Starting from 9.24.34 the dotnet prerequisite changed to NET8. These changes reflect that in the documentation

The publication of these changes should match with the release of 9.24.34

### DIFF
--- a/content/en/docs/refguide9/general/install.md
+++ b/content/en/docs/refguide9/general/install.md
@@ -102,11 +102,11 @@ It is possible to prepare the prerequisite installers beforehand, so that the Me
 5. Rename the following dependencies:
     * Microsoft .NET Desktop Runtime
         * For Studio Pro versions 9.0.0 through 9.24.33, rename the Microsoft .NET Desktop Runtime 6.0.x
-            * On x64, rename *windowsdesktop-runtime-6.0.35-win-x64.exe* to *windowsdesktop-runtime-6.  0-x64.exe*
-            * On ARM64, rename *windowsdesktop-runtime-6.0.35-win-arm64.exe* to *windowsdesktop-runtime-6.  0-arm64.exe*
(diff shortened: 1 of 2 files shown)


content/en/docs/refguide9/general/install.md
```diff
@@ -96,7 +100,9 @@ It is possible to prepare the prerequisite installers beforehand, so that the Me
 3. Create a folder with the name **Dependencies** in the same location where the Mendix Studio Pro installer was placed.
 4. Download the prerequisites listed in the [Troubleshooting](#troubleshooting) section above and move them into the **Dependencies** folder.
 5. Rename the following dependencies:
-    * The Microsoft .NET Desktop Runtime 6.0.x executable (*dotnet.exe*) to *windowsdesktop-runtime-6.0-x64.exe*
+    * | Studio Pro 9.0.0 - 9.24.33 | Studio Pro 9.24.34 and above |
+    | --- | --- |
+    | The Microsoft .NET Desktop Runtime 6.0.x executable (*dotnet.exe*) to *windowsdesktop-runtime-6.0-x64.exe* | The Microsoft .NET Desktop Runtime 8.0.x executable (*dotnet.exe*) to *windowsdesktop-runtime-8.0-x64.exe* |
```
rodrigocmoreira: Hi Gulam, can you do it like we did for version 10? We need to be very specific with the version to prevent problems in the future.

You can check here: https://docs.mendix.com/refguide/install/#offline

If we updated master to also use 8.0.14 we need to update the document of version 10 and 11 as well.

GulamY: I will align the text. 10 and 11 are on an older version of the runtime, so it is still accurate

MariaShaposhnikova (approved):

PR #9258, opened 2025-03-26
Do not use direct associations in imported modules

MarkvanMents:


content/en/docs/releasenotes/studio-pro/10/10.21.md
```diff
@@ -51,114 +51,116 @@ This improvement simplifies widget code. Widget developers no longer need to set
...
+* Opportunity for big performance improvements for apps with a lot of data. We added support for an alternative way of storing associations for one-to-many and one-to-one associations. This stores associations in the same table as the parent entity instead of a separate table. This is released as an opt-in for associations, as the subsequent data migration can take a long time (depending on your app's data). To use this feature, change the **Association Storage** setting on an association from **Association table** to **Direct association**. For more information, see [Association Storage Options](/refguide/association-storage/).
+    * You should not use direct associations in modules which are designed to be imported into apps (for example, Marketplace modules) as this might conflict with the app the module is being imported into.
```
dahfjkg, line 61: This restriction should be a temporary one, AFAIK. Shall we add "For now, ..."? Because people will read it and then remember that this is how it is supposed to be

MarkvanMents: @dahfjkg - good point. But will this be resolved for 10.21, or will the ability only be added in later versions.

dahfjkg: Only in later versions

MarkvanMents: This is the Mx 10.21 Release Note and this will apply to this version always. So I think we will leave it as is.

MarkvanMents:
Approved over Slack.

PR #9281, opened 2025-03-31
Clarify the Betas are Public

grootjans:


content/en/docs/releasenotes/studio-pro/lts-mts.md
```diff
@@ -16,21 +16,21 @@ In the 9.24.1 example, the first number (**9**) indicates the major version.
...
 For Mendix 10 and 11, we are planning to follow this release timeline:
 
 | Release | Date |
 | --- | --- |
 | 9.24 LTS | March 2023 |
-| 10 Beta 1 | April 2023 |
-| 10 Beta 2 | May 2023 |
+| 10 Public Beta 1 | April 2023 |
+| 10 Public Beta 2 | May 2023 |
 | 10.0 GA | June 2023 |
```
MarkvanMents: @grootjans Do we still need these three?
I think we only need the MTS and LTS for Mx 10?

grootjans: Fair point, you can take those out.

PR #9286, opened 2025-04-01
Add release notes for Native Template 14

vadymv-mendix:


ConnorLand:
@vadymv-mendix is this the last NT planned for Mendix 10? Because y'all just released one, and we can't keep making so many new RN pages.

vadymv-mendix:
@ConnorLand yes, the next major version release will be for StudioPro 11

ConnorLand (approved):

PR #9327, opened 2025-04-09
Change the location of the Navigation settings after [CTRL-3192]

envialon:
The Navigation setting has been moved to the Work Environment panel.


quinntracy:
Hi @envialon , is this for 10.22?

envialon:
@quinntracy yes it will be in 10.22

quinntracy (approved):

PR #9331, opened 2025-04-09
Correct System.SynchronizationFailure to System.SynchronizationError

ConnorLand:
Thijs brought this to the attention of MOO and I


ConnorLand:
This is currently wrong in the docs, but it's not super critical to merge now. Therefore I think we should park it as a PR until the merge for all versions (even 9 and 10).

PR #9333, opened 2025-04-09
Remove duplicate {#widgets} key

Maismaus:
There were two {#widgets} keys, so the last one redirected to the first one higher up on the page.


MarkvanMents (approved):
Hi Marius,
Thanks for noticing this and raising a PR. It is great that you want to improve the Mendix documentation.
Change looks good - approving and merging.
Yours
Mark

PR #9343, opened 2025-04-10
Opcua connector v1.1.0

jowussler:
updated the documentation of the platform supported OPC UA connector to reflect the changes of the latest 1.1.0 release that happened today.


content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
```diff
@@ -218,7 +221,7 @@ For more advanced cases, use the [read action](#read-action) as described below.
 
 The read action allows you to read specific attributes of a node. The request object for the action is `ReadNodeRequest`, which contains a list of `ReadNodeReadValueIDs`.
 
-Each `ReadNodeReadValueID` object specifies the attributes of the nodes you want to read. Upon executing the read action, a `ReadNodeResponse` object is returned. This object contains a list of `ReadNodeResponseResults`, each corresponding to the attributes requested, in the same order they were specified.
+Each `ReadNodeReadValueID` object specifies the attributes of the nodes you want to read. Upon executing the read action, a `ReadNodeResponse` object is returned which contains a list of `ReadNodeResponseResults`, each having a `ReadNodeResponseReadValueId` attached to it so you can match the requested attributes with the results.
```
MarkvanMents: Not sure I understand the response.
Is the response a list of objects that you can manipulate as a list in Studio Pro, or a single object with associated other objects.

And is ReadNodeResponseReadValueId an attribute of the ReadNodeResponseResults, or an object with a 1-1 association?

content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
```diff
@@ -249,7 +252,7 @@ The Write action allows you to write to specific attributes on a node. The reque
 
 The `WriteNode` objects describe how and what to write to a node. 
 
-The `Write node` action returns a `WriteNodeResponse` object. The object contains a list of `WriteNodeStatusCode` that contains the statuses of the written attributes in the same order as the request. Check the statuses to know whether the action succeeded.
+The `Write node` action returns a `WriteNodeResponse` object. The object contains a list of `WriteNodeStatusCode` that contains the statuses of the written attributes in the same order as the request. Check the statuses to see whether the action succeeded. Each `WriteNodeStatusCode` has a `WriteResponseWriteValue` object attached that allows you to match the resulting status code with the requested write action.
```
MarkvanMents: Not sure I understand the response.
If it is ordered, I guess this must be a list of objects that you can manipulate as a list in Studio Pro. If it is an object associated to a list, then I don't see how it can be ordered as I believe there is no guarantee on the order of associations.

MarkvanMents (changes requested):
Hi @jowussler 
Thanks for the PR.
I have proofread it, but have a couple of questions about your use of the term object to describe a list.
Feel free to Slack me if you want to discuss this face-to-face rather than having a back-and-forth in the GitHub comments.
Thanks
Mark

content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
jowussler: so our java action will return one ReadNodeResponse object and this object will have one or ReadNodeResponseResults objects attached to it for each ReadNodeReadValueId object that the ReadNodeRequest had. 
And since the order of the ReadNodeResponseResults coming back from our java action does not match the order of the ReadNodeReadValueId going into the java action, we added these new obejcts called ReadNodeResponseReadValueId. They are attached with a 1-1 with the ReadNodeResponseResults and contain information that is also present in the ReadNodeReadValueId to make it possible to match the response with the request. Otherwise you would not know which read value belongs to which written node.

content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
jowussler: so this is basically the same thing just for the write action. Each WriteNodeStatusCode will have one WriteResponseWriteValue to help identify which WriteNodeStatusCode belongs to which WriteNodeWriteValue from the original request. 
Those are the hoops we have to jump through because on the way from mendix to java, the order cannot be preserved.

content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
MarkvanMents: OK. So I think it would be more accurate (and perhaps clearer) to say:
```suggestion
Each `ReadNodeReadValueID` object specifies the attributes of the nodes you want to read. Upon executing the read action, a `ReadNodeResponse` object is returned which has associations with a list of `ReadNodeResponseResults` objects. Each `ReadNodeResponseResults` object has a 1-1 association with a `ReadNodeResponseReadValueId` object so you can match the requested attributes with the results.
```

content/en/docs/appstore/use-content/platform-supported-content/modules/opc-ua-connector.md
MarkvanMents: OK. So I think it would be more accurate (and perhaps clearer) to say:
```suggestion
The `Write node` action returns a `WriteNodeResponse` object. This object has associations with a list of `WriteNodeStatusCode` objects that contains the statuses of the written attributes. Each `WriteNodeStatusCode` has a 1-1 association with a `WriteResponseWriteValue` object that allows you to match the resulting status code with the requested write action. Check the statuses to see whether the action succeeded. 
```

MarkvanMents (commented):
@jowussler 
I've made some suggestions to make the use of the domain model clearer. Do these look good?

jowussler:
@MarkvanMents 
Both suggestions look good to me and I committed them.

MarkvanMents (approved):
Approving and merging.

PR #9344, opened 2025-04-10
[RUN-4366] Add missing release note for 10.18.5

atennapel:


MarkvanMents (approved):
@atennapel 
Thanks for the PR.
Leaving text to match all the other release notes.

PR #9355, opened 2025-04-11
[DUX-478] Add View Entity Document Changes to Existing Documentation

atmosfer:
This merge request updates the current View Entity documentation by adding a new section that covers View Entity Document Changes. The new content extends the existing documentation to include relevant details about the recent changes in the view entity structure and behavior.

PR #9356, opened 2025-04-11
Cherry pick view entities

MariaShaposhnikova:


MariaShaposhnikova:
Tried cherry picking a commit from https://github.com/mendix/docs/pull/9355, but did not work as expected.

PR #9378, opened 2025-04-15
Update preferences-dialog.md

ghost:
Removed German. We don't support it in the released version


quinntracy:
@mauritselzinga-mx Is this for Mx11?

ghost:
@quinntracy both Mx 10 and 11

quinntracy (approved):

MariaShaposhnikova:
@mauritselzinga-mx , for what version of 10 will it be, for 10.22?

PR #9379, opened 2025-04-15
Update table with date format tokens (Mx 11)

WimJongeneel:
- Removed the warning for `MM` tokens in nanoflows because we are fixing this in Mx 11
- Added a note that this was fixed in Mx 11
- Expanded the documentation on tokens that format differently based on their length
- Removed the "Y week in year" token as it is wrong, should be `w` (already present in the table)
- Added a (hopefully) useful link from the language settings to the table with formatting tokens

Need a review from R&D.

Some of the changes are going to apply to Mx 10, we will make a separate PR for that.


Yiyun333:
@WimJongeneel Thanks for sending this PR. Please let me know when it's ready for tech writer's review.

EnasAbdelrazek:
Would it be helpful to include examples that show the differences while using MMMM/MMM vs LLLL/LLL, since these differences are only matters in certain languages?

exec-astraea:
Month examples in languages that support genitive case:

`Ukrainian`

1. `LLLL` квітень
2. `MMMM` квітня 
3. `LLL` квіт. 
4. `MMM` квіт.

`Polish`
1. `LLLL` kwiecień
2. `MMMM` kwietnia
3. `LLL` kwi 
4. `MMM` kwi

EnasAbdelrazek (approved):

Yiyun333 (approved):

PR #9383, opened 2025-04-15
SOLP-2255 rename Port fix to Cherry pick

mikita-haurykau:


MariaShaposhnikova:
@mikita-haurykau , for what version is this change? 10.22? In this case I would keep information on port fix, as all version below 10.22 will have it named as port fix.

PR #9436, opened 2025-04-23
10.22 Studio Pro release notes

MariaShaposhnikova:


MariaShaposhnikova:
This PR is good to be merged after the date is updated (and latest releases if it is released before Beta 1).

PR #9450, opened 2025-04-25
Update how-to-set-up-your-partner-profile.md

melvingrootenboers:
The parter finder & profile has been completely revamped. This describes how to set up a partner profile in the new way. 

Btw: this is part of the Community Section, but I would like to have a dedicated Partner section as a main section in the documentation. How can this be done?

### DIFF
--- a/content/en/docs/community-tools/how-to-set-up-your-partner-profile.md
+++ b/content/en/docs/community-tools/how-to-set-up-your-partner-profile.md
@@ -1,59 +1,80 @@
 ---
-title: "Set Up Your Partner Profile"
+title: "Setting Up Your Partner Profile"
 url: /community-tools/how-to-set-up-your-partner-profile/
 weight: 5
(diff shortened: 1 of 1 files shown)


MarkvanMents:
HI @melvingrootenboers 
Thanks for the PR. I have a couple of questions:

1. Is this active now, or is there a future release date for this change.
2. We try to keep the documentation in the same sort of structure as the navigation within the product. I see that Partners are part of Community in the MxDock navigation. While we are open to discussing creating new sections, we don't want to have too many top-level choices for customers, nor to get out of alignment with the product navigation. However, we are looking at whether our documentation should become more aligned with personas and this might lead to a restructuring which would be in line with what you are asking. I will make sure we involve you when we have time to address this (we are currently overwhelmed with changes for the imminent Mx11 release.

MarkvanMents (approved):
Thanks for the PR @melvingrootenboers 
Thanks for the review @NicoletaComan 
I'll assume that this has already gone live as I didn't get a response to my question. So approving and merging

PR #9467, opened 2025-04-29
Add information about design property category

hendrik-jacobs-mendix:
We introduced a category to the design properties, that allows users to optionally group their design properties within the styling pane.

This will be available from Mendix Studio Pro 11.0.0-beta.2


ConnorLand (approved):

ConnorLand:
I will break the Mx11 info into its own versioned child doc after this PR and beta2 merges.

PR #9471, opened 2025-04-29
Resolve Mx11 todo with warning

ConnorLand:


jdiehl:
Can you mention that it will be available starting with 11.1?

ConnorLand:
> Can you mention that it will be available starting with 11.1?

Yes, if that's info we are happy being public about, I will put that in.

ConnorLand:
> Can you mention that it will be available starting with 11.1?

Done.

PR #9480, opened 2025-04-30
Kk pmp relnotes

katarzyna-koltun-mx:


mengdix (approved):
Line 57 pls update "Studio Pro 10.14" to "10.18";

Rest of it is good to go!

PR #9482, opened 2025-05-01
[RUN-4667] Add docs for extra Gradle arguments option (10.18.7, 10.23.0)

atennapel:


quinntracy:
Hi @atennapel , will this also be part of 11 Beta 2?

atennapel:
Yes, the exact same change was merged for 11-beta2

atennapel:
Here is the PR for Mx 11: https://github.com/mendix/docs/pull/9486

emirmx:
Looks good, approved from runtime side.

quinntracy:
@atennapel to confirm, is this option available for just 10.23 and above? Or 10.18.7 AND 10.23 and above? If it is for both, we will need to separate this into two PRs, as we don't know the timeline of the 10.18.7 release.

atennapel:
Both 10.18.7 and 10.23.0 (because it was backported to 10.18.x).

quinntracy:
> Both 10.18.7 and 10.23.0 (because it was backported to 10.18.x).

Thank you for clarifying. This PR will be for 10.23, then I will update the docs for 10.18.7 separately afterward.

quinntracy (approved):

PR #9491, opened 2025-05-02
Remove Max Retrieve Size runtime setting REDO

ConnorLand:
Formerly https://github.com/mendix/docs/pull/9342/ but now clean with no conflicts.


ConnorLand:
I'll need to confirm the runtime table change made it into Mx11 docs appropriately, but that's for later.

PR #9504, opened 2025-05-08
Revert previous Travis fix

MarkvanMents:
Travis builds started failing again with the same error as before on npm.
Reverting the fix to see if that works.


MarkvanMents:
Doesn't stall on npm step any more.
Merging to development.

PR #9505, opened 2025-05-08
Clarify behavior of math operators in XPath

dahfjkg:
Requires tech review

### DIFF
--- a/content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-operators.md
+++ b/content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-operators.md
@@ -29,10 +29,10 @@ The following mathematical operators are supported:
 | `*` | Multiplication | `6 * 4` | 24 |
 | `div` | Division | `8 div 4` | 2 |
 
-Mathematical operators can be used in Studio Pro. In Java code, they are disabled by default forn the `XPathQuery` API, but can be enabled using the `allowMathOperators` method. For example:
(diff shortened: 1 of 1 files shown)


MarkvanMents (changes requested):
@dahfjkg 
Hi Pavel,
Does this still need a technical review, or can I publish it?

dahfjkg:
Hi @MarkvanMents, thanks for looking into this! The task is still in review on our side, so let's wait for that

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-operators.md
```diff
@@ -29,6 +29,14 @@ Additionally, the following operators are supported in Java code only:
 | `*` | Multiplication | `6 * 4` | 24 |
 | `div` | Division | `8 div 4` | 2 |
 
+Mathematical operators can be used in Studio Pro. In Java code, they are disabled by default forn the `XPathQuery` API, but can be enabled using the `allowMathOperators` method. For example:
```
basdebakker: Is the first part correct? The point is that we only allow mathematical operators in XPath constraints that are defined in the runtime part of the model, like in microflow retrieve actions. Not in XPath queries that are sent by the client. But I don't know if Studio Pro has ways to define XPath queries that will actually be sent by the client when running the app. (If not, why do we even have a client API that allows executing XPath queries?)

(And there is a typo in `forn`.)

content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-operators.md
```diff
@@ -29,6 +29,14 @@ Additionally, the following operators are supported in Java code only:
...
+
+```java
+core.createXPathQuery("//Module.Entity[attr1 + attr2 = 42]")
```
basdebakker: The public API is in `Core` with a capital `C`.

basdebakker:
Approved by datastorage team.

MarkvanMents (approved):
Thanks for the PR.
Looks good.
Can it be published immediately?

PR #9509, opened 2025-05-08
Explain sorting in view entities

dahfjkg:
Requires team review

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
+++ b/content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
@@ -68,6 +68,25 @@ Age is determined for each customer by calculating the difference in years betwe
 
 The original customer view included address information, but most database optimizers will see that this information is not relevant when counting customers by age, so this information is excluded when retrieving the data. However, the information is still present and can be generated, if requested. 
 
+## Sorting of view entity results {#sorting}
+
+Similarly to database views, vien entities are representations of data that do not imply any particular order of the results. Due to that, it is not possible to use `ORDER BY` clause in the view entity OQL to sort the results of the query. Data retrieved from a view entity can be sorted on retrieval, like regular entities. For example, using [Sorting](/refguide/retrieve/#sorting) functionality of a microflow Retrieve activity.
+
+### `ORDER BY` in combination with `LIMIT` and `OFFSET`
+
+`ORDER BY` clause can be used in a view entity in combination with `LIMIT` or `OFFSET` clauses. That can be done to define a slice of data to retrieve. In that case, you still should not rely on the order of the output. If it is required to have the results in particular order, they can be sorted on retrieval.
+
+For example, the following OQL query defines a view entity `Books.Bestseller`, which contains data of 10 most sold books. When using that view entity in your app, you should still explicitly specify sorting of the results.
+
+```sql
+FROM Books.Book
+SELECT Name AS Name,
+       ISBN AS ISBN,
+       Sold AS Sold
+ORDER BY Sold DESC
+LIMIT 10
+```
+
 ## View Entities and Parameters
 
 You may have previously used the OQL module or datasets to execute OQL queries. There, you use parameters to configure the query to retrieve the correct data. View entities do not have parameters, but you can dynamically specify what data you want to retrieve.
--- a/content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
+++ b/content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
@@ -68,15 +68,15 @@ Age is determined for each customer by calculating the difference in years betwe
 
 The original customer view included address information, but most database optimizers will see that this information is not relevant when counting customers by age, so this information is excluded when retrieving the data. However, the information is still present and can be generated, if requested. 
 
-## Sorting of view entity results {#sorting}
+## Sorting of View Entity Results {#sorting}
 
-Similarly to database views, vien entities are representations of data that do not imply any particular order of the results. Due to that, it is not possible to use `ORDER BY` clause in the view entity OQL to sort the results of the query. Data retrieved from a view entity can be sorted on retrieval, like regular entities. For example, using [Sorting](/refguide/retrieve/#sorting) functionality of a microflow Retrieve activity.
+Similar to database views, view entities represent data without enforcing any specific order. As a result, it is not possible to use the `ORDER BY` clause in the view entity OQL to sort query results. However, data retrieved from a view entity can be sorted on retrieval (like regular entities) using options like the [Sorting](/refguide/retrieve/#sorting) feature in a Retrieve microflow activity.
 
-### `ORDER BY` in combination with `LIMIT` and `OFFSET`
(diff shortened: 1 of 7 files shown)


quinntracy:
Hi @dahfjkg are these updates tied to an upcoming release, or can they be merged after review? In addition, please indicate in the comments when the PR has completed team review so we can do a language review.

MarkvanMents:
@dahfjkg 
Hi - this still seems to be waiting a technical review. Do you know when this will be done?

dahfjkg:
As soon as it is picked up by someone in the team, so should be soon. To answer the previous question by @quinntracy, this is not tied to any release.

content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
```diff
@@ -68,6 +68,25 @@ Age is determined for each customer by calculating the difference in years betwe
 
 The original customer view included address information, but most database optimizers will see that this information is not relevant when counting customers by age, so this information is excluded when retrieving the data. However, the information is still present and can be generated, if requested. 
 
+## Sorting of view entity results {#sorting}
+
+Similarly to database views, vien entities are representations of data that do not imply any particular order of the results. Due to that, it is not possible to use `ORDER BY` clause in the view entity OQL to sort the results of the query. The outcome of a view entity can be sorted on retrieval. For example, using [Sorting](/refguide/retrieve/#sorting) functionality of a microflow Retrieve activity.
```
pijuskri, line 73: Just to be less confusing on what is not allowed, i would reword this to:
`...it is not possible to only use `ORDER BY` clause...`

pijuskri: Instead of `The outcome of a view entity` it could be better to say `Data retrieved from a view entity`

pijuskri: To make it extra obvious, we can end `can be sorted on retrieval.` with `..., like regular entities`.

pijuskri:
Approved for technical review

dahfjkg:
@MarkvanMents this is good to go on our side

content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
```diff
@@ -68,6 +68,25 @@ Age is determined for each customer by calculating the difference in years betwe
 
 The original customer view included address information, but most database optimizers will see that this information is not relevant when counting customers by age, so this information is excluded when retrieving the data. However, the information is still present and can be generated, if requested. 
 
+## Sorting of View Entity Results {#sorting}
+
+Similar to database views, view entities represent data without enforcing any specific order. As a result, it is not possible to use the `ORDER BY` clause in the view entity OQL to sort query results. However, data retrieved from a view entity can be sorted on retrieval (like regular entities) using options like the [Sorting](/refguide/retrieve/#sorting) feature in a Retrieve microflow activity.
```
MarkvanMents: Not sure what you mean by "database views".
Do you mean "As with persistable entities"?
Or even, "As with other types of entity".

content/en/docs/refguide10/modeling/domain-model/entities/view-entities/use-view-entities.md
```diff
@@ -46,6 +46,25 @@ Age is determined for each customer by calculating the difference in years betwe
 
 The original customer view included address information, but most database optimizers will see that this information is not relevant when counting customers by age, so this information is excluded when retrieving the data. However, the information is still present and can be generated, if requested. 
 
+## Sorting of View Entity Results {#sorting}
+
+Similar to database views, view entities represent data without enforcing any specific order. As a result, it is not possible to use the `ORDER BY` clause in the view entity OQL to sort query results. However, data retrieved from a view entity can be sorted on retrieval (like regular entities) using options like the [Sorting](/refguide/retrieve/#sorting) feature in a Retrieve microflow activity.
```
MarkvanMents: See comment on other file.

MarkvanMents: And
```suggestion
Similar to database views, view entities represent data without enforcing any specific order. As a result, it is not possible to use the `ORDER BY` clause in the view entity OQL to sort query results. However, data retrieved from a view entity can be sorted on retrieval (like regular entities) using options like the [Sorting](/refguide10/retrieve/#sorting) feature in a Retrieve microflow activity.
```

MarkvanMents (changes requested):
@dahfjkg 
Hi Pavel. I've reviewed this now.
Have one question on some terminology, and one place where there is a typo on the link.
My other changes are in cffd8dbcbd309cdfbd8cf0a6e24404baf64223e7

Can you have a look at my comments?
Thanks.
Mark.

content/en/docs/refguide/modeling/domain-model/entities/view-entities/use-view-entities.md
dahfjkg: I mean SQL views, the feature of relational databases. As this is not a Mendix concept, maybe it requires more details. Many people are confused with why it is not possible to have `ORDER BY` in a view entity, so I am trying to provide as much justification as possible

dahfjkg: Reference to that is based on how we define view entiities [here](https://docs.mendix.com/refguide/view-entities/#introduction): `This concept is analogous to views commonly found in relational database systems.`

MarkvanMents: I still think it is better to describe it here in terms of retrieving lists of objects in Mendix than what happens with SQL? I think just saying in the intro that they are like database views is enough reference to that - people who know about SQL views will know about the ordering issues, people who don't will know it more from the side of what happens with a retrieve action.

So would you be happy with:
```suggestion
Similar to other types of Mendix entity, view entities represent data without enforcing any specific order. As a result, it is not possible to use the `ORDER BY` clause in the view entity OQL to sort query results. However, data retrieved from a view entity can be sorted on retrieval (like regular entities) using options like the [Sorting](/refguide/retrieve/#sorting) feature in a Retrieve microflow activity.
```

Or can you not apply a sort on a retrieve action?

dahfjkg: The thing is that people who expect order by to work are exactly the people who come from SQL background, and they expect it to work just as in SQL. That is why the fact that it does not actually work for DB views is a very strong argument IMO. But from the platform perspective, your suggestion makes total sense, so let's go with it

MarkvanMents (approved):
Looks good. Approving. Thanks for the PR and the explanations.
Can this be published immediately?

PS I think I'm happier that we have gone with the change.
The thing is that the OQL **does** return things in the order you specify. In my mind, it is only because the platform is treating it like a normal entity that you lose the order again and it isn't at all related to anything you are doing with OQL/SQL. I wouldn't want customers to worry that using raw OQL loses the ordering.

dahfjkg:
@MarkvanMents yes, would be great if we could publish this soon

PR #9512, opened 2025-05-09
Update Linux documentation to fit refined support scope

MXClyde:
Update Linux documentation to fit refined support scope


katarzyna-koltun-mx:
@MXClyde I've reviewed the text, but could you provide more info on why you renamed and moved the file?

I'm away on PTO next Mon-Wed - @MarkvanMents I've tagged you in case this needs to be addressed while I'm away, otherwise I'm going to follow up after I'm back.

MXClyde:
@katarzyna-koltun-mx We only support Debian 10 now. Which effectively means we only support Linux (and no other Unix-like systems). The new name therefore is more accurate.

MarkvanMents:
Closing this in favour of #9526 to retain history.

PR #9517, opened 2025-05-13
Renamed menu item

tknippenberg:


quinntracy:
Hey @tknippenberg , is this change already in the most recent version of SP (11 beta 1), or is this a future change?

tknippenberg:
> Hey @tknippenberg , is this change already in the most recent version of SP (11 beta 1), or is this a future change?

It's already the case in 10.22 and I expect it has always been incorrect on this Docs page 😅

quinntracy:
> > Hey @tknippenberg , is this change already in the most recent version of SP (11 beta 1), or is this a future change?
> 
> It's already the case in 10.22 and I expect it has always been incorrect on this Docs page 😅

In that case, I'll get it merged now :grin:

PR #9545, opened 2025-05-16
Update documentation to add another Maia domain name

Yiyun333:


Yiyun333:
Verified via Slack: This PR is not needed because `logic.maia.mendix.com` is a subdomain of `maia.mendix.com`, so it does not need to be mentioned separately.

PR #9566, opened 2025-05-22
Update Maia for Pages docs for Mendix 11 beta 2.

MendixBart:
In the latest changes, we're making sure Maia knows about the page (even after closing a session) and can modify widgets by setting certain properties. This means some of the information, specifically around the limitations, is no longer relevant. These changes will be released with Mendix 11 beta 2, but will also be available in Mendix 10.23.


Yiyun333:
@MendixBart Thanks a lot for sending us the PR! I can create another PR for 10.23 after I've reviewed the content in this PR. I think it's worth mentioning this improvement as well in both the 11 beta 2 and 10.23 release notes. Could you please check with the team and let me know if we should add release notes? I can add them to our release note PRs if you could provide me the text.

MendixBart:
@Yiyun333 Hey, it would indeed be good to add some release notes. Here's a suggestion, feel free to change the wording. I think we can list this as an improvement.

Maia for Pages now knows about the contents of your page. It can add widgets to existing pages, as well as modify widgets.

Yiyun333 (approved):

PR #9579, opened 2025-05-23
Replace instances of whitelist/blacklist

NicoletaComan:


MarkvanMents (approved):
Looks great. I can see anything you've missed.
Thanks for picking this up.
Approving.

PR #9584, opened 2025-05-26
Added maintenance criterium to pwa vs native

jdiehl:
Please replicate in mx10 and mx9 docs


ConnorLand:
Thanks for the PR @jdiehl , will duplicate and merge. 

Also, TIL that [criterium](https://en.wikipedia.org/wiki/Criterium) is a bike race and criterion is the singular for criteria. Latin is wild.

ConnorLand (approved):

PR #9588, opened 2025-05-27
Update release notes for Native Template version 7 and 11

YogendraShelke:


ConnorLand:
@UrazAkgultan is helping me check if the release dates are correct for 11.1.0 vs 11.06 etc

PR #9589, opened 2025-05-27
[DAT-4171] Make warning about not using direct associations in marketplace modules more explicit

pijuskri:


MarkvanMents (approved):
Thanks for the PR.
Looks good.
Approving and merging.

PR #9591, opened 2025-05-27
[SOLP-3381] Update Import due to security issues

blis-mendix:


content/en/docs/refguide/modeling/import-and-export/_index.md
```diff
@@ -25,6 +25,17 @@ You can import and export the following:
 
 ## Importing Various Elements
 
+{{% alert color="warning" %}}
+Studio Pro prevents packages from unpacking content to undesired locations. Contact the content publisher to get the issue fixed.
```
MarkvanMents: Not sure I understand the bit about contacting the content publisher. In the examples below the user can choose the location for the import - so they don't need to contact anyone to unpack it somewhere else.

blis-mendix: Hey! This comment is to the whole chapter 2, part you talk about is in chapter 2.1 as it's specific for the solution based apps while in for example 2.2 you can't choose location.

We added a generic check during import so if we detect that content has anything we consider illegal we inform user and they can contact publisher of such module/template/widget etc.

MarkvanMents: Ah, that makes sense.
Perhaps we could say
```suggestion
Studio Pro prevents packages from unpacking content to undesired locations. If you cannot change the location, contact the content publisher to get the issue fixed.
```
Or specify the types of import where you would need to do this?

Is this urgent, or can we wait until Maria comes back to pick this up?

PR #9593, opened 2025-05-27
Rephrased sentence

tknippenberg:


content/en/docs/refguide/version-control/on-premises-git.md
```diff
@@ -118,7 +118,7 @@ Now that you have a Git app on your server, you can download it to another direc
 
     {{< figure src="/attachments/refguide/version-control/on-premises-git/sign-in-dialog.png" alt="Sign In dialog" class="no-border" >}}
 
-1. Enter username (it can be anything except empty) and use the PAT you saved earlier as the password. 
+1. Enter username (it can be anything except empty) and use the PAT you used earlier as the password.
```
MarkvanMents: PAT you **created** earlier (see line 83)?

tknippenberg: Yeah that could also work!

Right now I had an issue with a user being confused whether they should use the Mendix PAT here, as they were migrating to private Git and therefore created a Mendix PAT. But in this step they needed the PAT for their Git provider ofc.
If you have other ideas how to make it clearer 'which PAT' the user should use, I'm all ears.

kasjama:
Indeed, please specify that this is the Git PAT, not the Mendix PAT

content/en/docs/refguide/version-control/on-premises-git.md
MariaShaposhnikova: @tknippenberg , could we say "use PAT from your Git provider you created earlier"?

tknippenberg: Yes, that sounds good.

PR #9614, opened 2025-06-02
Update epics-api.md

ghost:
Disable 'Try it out' button on the api documentation page.


NicoletaComan (approved):
Thank you, @loket4, for providing the info. 
I've rephrased the note so as to remove "below", which is generally inaccessible.
When should this go live?

ghost:
The documentation is behind the implementation on this particular topic. On the other hand, this is not a high-stake feature. This change to the docs can be published at your earliest convenience.

NicoletaComan:
> The documentation is behind the implementation on this particular topic. On the other hand, this is not a high-stake feature. This change to the docs can be published at your earliest convenience.

Thank you for confirming!
The changes are now live.

PR #9633, opened 2025-06-05
Update environments-details.md

NielsKunst:
Correct to match Cloud Portal banner when cancelled


OlufunkeMoronfolu (changes requested):
Hello @NielsKunst ,

Thanks for sharing this. However, the new UX grammatical change does not align with the Mendix language guide, so we will need to decline this.

I will reach out to you privately to discuss this further. Thanks.

OlufunkeMoronfolu:
Note: Change finalized with Tjalling. UX update will be done soon, thus rendering this PR not applicable.

PR #9652, opened 2025-06-10
Add info on MTS, LTS, Beta releases

MariaShaposhnikova:


MariaShaposhnikova:
I am closing this one as there were updates on LTS and MTS since it was created and not all information in the PR is correct.

PR #9655, opened 2025-06-10
Add more keywords for OQL in Mendix 11

passalaqua:
These keywords are not functional, but will no longer be accepted as unquoted identifiers starting with Mendix 11.

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/oql/_index.md
+++ b/content/en/docs/refguide/modeling/domain-model/oql/_index.md
@@ -38,9 +38,9 @@ Here is a list of all OQL reserved words:
 
 `BOOLEAN`, `BY`
 
-`CASE`, `CAST`, `COUNT`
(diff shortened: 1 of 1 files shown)


MarkvanMents (changes requested):
Hi Augusto (@passalaqua)
Thanks for the PR.
The description of this PR talks about Mendix 11, but the changes are targeting the Mendix 10 documentation.
Is it easy for you to make these changes for Mendix 11 instead (refguide/ not refguide10/) or would you like me to move the changes to Mx 11 documentation?

passalaqua:
@MarkvanMents I think I changed to the correct file now. Can you confirm it's good?

MarkvanMents (approved):
Thanks for the PR.
Can I merge immediately, or should I wait for GA?

passalaqua:
I think it can be merged immediately. It's only going to throw errors after GA is out, but the words are already supposed to be considered reserved and should not be used.

PR #9657, opened 2025-06-11
Add guidance to not rename mpr files when using version control

MxMarcel:
We do not support renaming the mpr file when you are using Git version control. This should be clear from the documentation.

I added some comments to make this more clear.

### DIFF
--- a/content/en/docs/developerportal/general/settings/_index.md
+++ b/content/en/docs/developerportal/general/settings/_index.md
@@ -27,13 +27,13 @@ These tabs are only available for users with the **App Settings** permission:
 * **History**
 * **Story Archive**
 
-## Page Header
+## Page Header {#settings-page-header}
 
 The page header displays the following information:
 
 * The app image – You can change the image by clicking it.
 * The app name – You can change the name of the app by clicking the pencil icon next to it.    
-    Renaming the app in the Mendix Portal does not rename it in Studio Pro. The app name in Studio Pro is tied to its *.mpr* file. However, changing the name of the *.mpr* file is not supported, and we strongly advise against it. For more information, refer to the [MPR Format](/refguide/version-control/#mpr-format) section in *Version Control*. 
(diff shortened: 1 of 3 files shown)


content/en/docs/developerportal/general/settings/_index.md
```diff
@@ -35,6 +27,22 @@ These tabs are only available for users with the **App Settings** permission:
 * **History**
 * **Story Archive**
 
+## Page Header
+
+The page header displays the following information:
+
+* The app image – You can change the image by clicking it.
+* The app name – You can change the name of the app by clicking the pencil icon next to it.    
+    Renaming the app in the Mendix Portal does not automatically rename it in Studio Pro. Refer to [Version Control](/refguide/version-control/#mpr-format) in the Studio Pro guide for information on renaming the *.mpr* file.
```
MariaShaposhnikova: @NicoletaComan , is this correct, can we test it? The version control documentation says that you can't rename the *.mpr* file and can only rename the app in the Mendix Portal (Sprintr is an old (internal) term for it).

NicoletaComan: @MariaShaposhnikova, yes, you can change it on the Mendix Portal.
Per the Slack thread, changing it on the Mendix Portal does not trigger a change in Studio Pro, unless the *.mpr* file is also renamed (which is not recommended). 
Should I simply remove the second part altogether to avoid any confusion?

MariaShaposhnikova: @NicoletaComan , indeed, renaming it will not change the .mpr file name, and users should not rename the .mpr file. I just suggest we make it clear here, to me "Refer to [Version Control](/refguide/version-control/#mpr-format) in the Studio Pro guide for information on renaming the *.mpr* file." sounds like I will be offered a step by step guidance on renaming the file, but instead I will only get a warning that this cannot be done.

MariaShaposhnikova: Perhaps, we can say something like: "Renaming the app in the Mendix Portal does not rename it in Studio Pro, which is connected to renaming an _.mpr_ file. Renaming an .mpr file is a known limitation and we strongly advice against it. For more information, see the [MPR Format](/refguide/version-control/#mpr-format) section in *Version Control*."

NicoletaComan: @MariaShaposhnikova, that's true, it can definitely be misleading.
I've updated the content.

PR #9662, opened 2025-06-11
[WTF-2293]: Restructure Pluggable Widgets API Docs by SP version

weirdwater:
Reorganizes the Pluggable Widgets API according to Studio Pro version.

URLs have been left the same as much as possible. The exception being documentation for Mendix 10, these URLs can be considered new.


alihcsumer (approved):

weirdwater:
@ConnorLand, this is now ready for TW review.

ConnorLand:
I'm happy with the nav consistency. It also preserves the order of existing API docs in the Studio Pro API categories: 
image

ConnorLand:
Thanks for resolving the link errors @weirdwater . As long as the aliases are set correctly, I think everything should be fine.

ConnorLand:
Last things that might need to be done with this:

* Alias implementation? The dev told me it was ok, because the URLs are mostly unchanged and the product points to the right places. Also we don't have anything in the redirects doc last time I checked.
* Title fixes to take out versions (in progress).

weirdwater:
Hi @ConnorLand, I cleaned up the titles and index pages. Also fixed a typo.

ConnorLand (approved):

PR #9683, opened 2025-06-16
Oncommit webhook info update

### LINKED ISSUE #9558: Automatic creation WebHooks
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [pipelines](https://docs.mendix.com/developerportal/deploy/pipelines/)

Document link: [pipelines.md](https://github.com/mendix/docs/blob/development/content/en/docs/deployment/mendix-cloud-deploy/pipelines.md)

## My Issue/Suggestion

It seems as if the following paragraph in the documentation needs clarification:

"To trigger pipelines based on Teamserver push (Git), Mendix automatically creates a webhook on your behalf. You can see this webhook if you click Webhooks in the [navigation pane](https://docs.mendix.com/developerportal/#navigation-pane). Do not delete this webhook; deleting it would cause pipeline run failures for pipelines that rely on the Teamserver push (Git) trigger type." 

I cannot judge whether WebHooks are still used for the implementation or not. However, they are not visible in my case, so at least the last two sentences in the paragraph are superfluous.

OlufunkeMoronfolu:
Based on a recent change, the webhook created by pipelines for the "On commit" trigger will not be visible on the webhooks page for new apps - the webhook will still be visible for legacy apps. Thus making the current info stale.

This PR fixes that.

### DIFF
--- a/content/en/docs/deployment/general/webhooks.md
+++ b/content/en/docs/deployment/general/webhooks.md
@@ -162,9 +162,7 @@ Make sure to use the correct key names when using this payload information to ca
 ## Teamserver Push (Git)
 
 {{% alert color="info" %}}
-This webhook is not visible for newly created apps. It is only visible for existing apps.
-
-For apps where the webhook is visible, do not delete it; deleting it would cause pipeline run failures for pipelines that rely on the Teamserver push (Git) trigger type.
+This webhook is not visible to first-time pipeline users. For existing pipeline users who can view this webhook, do not delete it. Deleting it causes failures for pipelines that rely on the Teamserver push (Git) trigger.
 {{% /alert %}}
 
 When you push a model change to the [Git Team Server](/developerportal/general/team-server/), and the webhook responds to the event **Teamserver push (Git)**, request content is sent to the configured endpoint. The request content contains a payload with the following format:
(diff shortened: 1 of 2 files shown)


OlufunkeMoronfolu:
@IshJos

IshJos (approved):

content/en/docs/deployment/general/webhooks.md
```diff
@@ -161,6 +161,10 @@ Make sure to use the correct key names when using this payload information to ca
 
 ## Teamserver Push (Git)
 
+{{% alert color="info" %}}
+This webhook is not visible to first-time pipeline users. For existing pipeline users who can view this webhook, do not delete it. Deleting it causes failures for pipelines that rely on the Teamserver push (Git) trigger.
```
MarkvanMents: Looks good. Two things to think about, but you don't have to change if you think they are OK:

* Is it new users or does this also apply to new pipelines for existing users?

* Perhaps making this two paragraphs (1. you may not see it 2. if you see it don't delete it) might be better.

MarkvanMents (approved):
Looks good.
I've added a couple of things to think about, but generally is OK and you don't have to change anything if you are happy with it.

content/en/docs/deployment/general/webhooks.md
OlufunkeMoronfolu: Thanks for the review:
- The webhook is created once for pipeline users. So, it only affects new pipeline users
- Yes, thought about this. Will push a quick fix for it

PR #9685, opened 2025-06-16
[WOR-2536]: Update docs for interrupting boundary events GA

davevissermx:
This MR updates the docs for the GA release of interrupting boundary events

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
+++ b/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
@@ -82,10 +82,14 @@ For an existing boundary event, when you change its type from non-interrupting t
 
 {{< figure src="/attachments/refguide/modeling/application-logic/workflows/boundary-events/security-dialog.png" alt="Security Dialog when changing type" width="450">}}
 
+The dialog is shown due to execution limitations on ongoing boundary events. For example the parent activity will be in progress when a non-interrupting BE starts. Once the type is changed it's parent activity will stay in progress, which would allow the parent to have more than the maximal allowed interrupting boundary events.
+
 After you confirm the change:
 
-* The boundary event is re-created as the specified type. As a result, new IDs are created. These IDs are used by the Mendix Runtime for conflict detection analysis.
-* The current activities on the boundary event path including the boundary event itself become incompatible and need to be restarted.
+* The boundary event is re-created as the specified type. And the boundary event will be scheduled after redeploy of the workflow, as soon as it becomes in progress. 
+* The workflow will become incompatible once a boundary event is changed that has already been executed. The workflow will become incompatible by the following reasons:
+    * If the old boundary event was non-interrupting, we'll get [Non-interrupting boundary event path removed](/refguide/workflow-versioning/#non-interrupting-boundary-event-path-removed) conflict.
+    * If the old boundary event was interrupting, we'll get the [Current Activity Removed](/refguide/workflow-versioning/#current-activity-removed) conflict.
 
 ### Rearranging Boundary Events
 
@@ -121,24 +125,24 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
 When there are multiple boundary events attached to an activity and an interrupting boundary event is executed, all the scheduled boundary events will be aborted and all the boundary events that have already started will continue to run until the entire workflow ends.
 {{% /alert %}}
 
-## Nested jump rules
-Boundary events come with a specific set of rules for jumps when they are nested, the variations are working meant for both a modelled jump and a runtime ad-hoc jump. In the following table are the differences specified.
+## Jump rules
+Boundary events come with a specific set of rules for jumps, these rules are integrated in both types of jumps such as [Jumping to other activities in design time](/refguide/jump-activity/) and and [jumping in running workflow instances](/refguide/jump-to/). The rules are as follows:
 
-|Parent | Child | Description|
-| -------- |-------- | ------- |
-|Interrupting | Interrupting | When an interrupting boundary event is nested inside an interrupting boundary event we are allowed to either jump out of the boundary event flow or end the workflow.|
-|Interrupting | Non-Interrupting | When an interrupting boundary event is nested inside an non-interrupting boundary event we are only allowed to jump out of the boundary event flow to it's parent flow.|
-|Non-Interrupting | Interrupting | When a non-interrupting boundary event is nested inside an interrupting boundary event we are only allowed to jump in the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
-|Non-Interrupting | Non-Interrupting | When a non-interrupting boundary event is nested inside an non-interrupting boundary event we are only allowed to jump inside the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
+- Jump to the inside of a boundary event: not allowed (and never will be regardless of how they are organized)
+- Jump to the outside of a boundary event: only allowed from within an interrupting boundary event.
+- Jump within a boundary event: allowed
 
 ## Boundary Event Variables
 
-Boundary events have dedicated variables that can be used to get direct access to the values of the parent activity if it is either a user task or Call workflow activity. You can get information such as the parent activity's `DueDate`, which can be used in the boundary event flow and its expressions. For instance, you can use the expression `addDays($ParentTask/DueDate, -2)` to configure a timer boundary event so that it is triggered two days before the due date of its parent user task.
+Boundary events have dedicated variables that can be used by activities in the path of the boundary event to get direct access to the values of the parent activity if it is either a user task or Call workflow activity. You can get information such as the parent activity's `DueDate`, which can be used in the boundary event flow and its expressions. For instance, you can use the expression `addDays($ParentTask/DueDate, -2)` to configure a timer boundary event so that it is triggered two days before the due date of its parent user task.
 
 The list of variables is described below: 
 
 * `$ParentTask` – the parent user task of the attached boundary event
 * `$CalledWorkflowInstance` – the parent Call workflow activity of the attached boundary event
+
+## Current Limitation {#limitation}
+- When a workflow instance has a running timer boundary event which is set to recur. And for example in the running instance we do a manual jump to the parent activity of the running workflow. The already scheduled boundary event will not be cancelled whilst the new one is scheduled. When the old one executes in the mean time the new one will be disregarded. This is a limitation that will be fixed in the future.
 
 ## Read more
 
--- a/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
+++ b/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
@@ -4,19 +4,15 @@ url: /refguide/workflow-boundary-events/
 weight: 20
 ---
 
-{{% alert color="info" %}}
-Interrupting timer boundary events are currently in beta. Certain features or attributes are subject to change and may contain bugs. For more information, see [Beta and Experimental Releases](/releasenotes/beta-features/).
-{{% /alert %}}
-
 ## Introduction
 
 Based on Business Process Model and Notation (BPMN) 2.0, boundary events are a type of event that is attached to the boundary of an activity (such as a task or a sub-process) to handle exceptional situations or are triggered by certain behaviors. 
 
 There are two main types of boundary events:
 
-* Non-interrupting boundary events: These events do not interrupt the ongoing activity. When triggered, they allow the activity to continue while simultaneously starting a new path from the boundary event. Use non-interrupting boundary events when the parent activity should remain active, but you would like to do something in parallel. For example, after 2 days, a reminder should be sent to the assigned user. As per BPMN 2.0 specification, non-interrupting boundary events are visualized as two dashed circles with an icon in the center.
+* **Non-interrupting boundary events**: These events do not interrupt the ongoing activity. When triggered, they allow the activity to continue while simultaneously starting a new path from the boundary event. Use non-interrupting boundary events when the parent activity should remain active, but you would like to do something in parallel. For example, after 2 days, a reminder should be sent to the assigned user. As per BPMN 2.0 specification, non-interrupting boundary events are visualized as two dashed circles with an icon in the center.
 
-* Interrupting boundary events (currently in beta): When these events are triggered, they interrupt the activity they are attached to, meaning that this activity will be aborted. The process flow is redirected to the boundary event's outgoing sequence path. Use interrupting boundary event in situations where further execution of the activity (and other following activities) is not required and an alternative path should be taken. For example, use an interrupting boundary event to start an escalation or a fast-track path when an activity is not completed 2 days after the due date. Or when the assigned user does not make a decision within 5 days, you want to abort the user task and continue the process with a pre-set decision. As per BPMN 2.0 specification, interrupting boundary events are visualized as two solid circles.
+* **Interrupting boundary events**: When these events are triggered, they interrupt the activity they are attached to, meaning that this activity will be aborted. The process flow is redirected to the boundary event's outgoing sequence path. Use interrupting boundary event in situations where further execution of the activity (and other following activities) is not required and an alternative path should be taken. For example, use an interrupting boundary event to start an escalation or a fast-track path when an activity is not completed 2 days after the due date. Or when the assigned user does not make a decision within 5 days, you want to abort the user task and continue the process with a pre-set decision. As per BPMN 2.0 specification, interrupting boundary events are visualized as two solid circles.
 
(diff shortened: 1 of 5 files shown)


content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/timer.md
```diff
@@ -38,14 +38,22 @@ The **Caption** describes what happens in this element. It is displayed under th
...
 By default, it is set to **No**, which means that the timer boundary event is non-interrupting. When it is set to **Yes**, the timer boundary event is interrupting. For more information, see [Boundary Events](/refguide/workflow-boundary-events/).
 
+{{% alert color="info" %}}
+The following section is only applicable when **Timer** is used as a non-interrupting timer boundary event.
+{{% /alert %}}
+
+The **Recurrence** property allows a non-interrupting timer boundary event to execute multiple times when the specified interval has elapsed. The following parameters can be set.
+| Type | Description |
+| --- | --- |
+| Interval | You can set a certain duration for the timer. With the **Repeat every** setting, you can indicate the number of seconds, minutes, hours, days, weeks or months the timer's duration is. Possible values for the setting are:<br /><ul><li>Minutes</li><li>Hours</li><li>Days</li><li>Weeks</li><li>Months</li> </ul> |
```
DmytroKost: 1. Repeat every doesn't have seconds option
2. "With the **Repeat every** setting ... the timer's duration is". You are referring to the interval between occurrences, which is somewhat different from the first execution duration. I think it will be good to indicate that.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -125,6 +121,16 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
 When there are multiple boundary events attached to an activity and an interrupting boundary event is executed, all the scheduled boundary events will be aborted and all the boundary events that have already started will continue to run until the entire workflow ends.
 {{% /alert %}}
 
+## Nested jump rules
+Boundary events come with a specific set of rules for jumps when they are nested, the variations are working meant for both a modelled jump and a runtime ad-hoc jump. In the following table are the differences specified.
```
DmytroKost: "a modelled jump and a runtime ad-hoc jump", I would put it "a Jump activity and and ad-hoc jump" with references to the corresponding articles.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -125,6 +121,16 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
 When there are multiple boundary events attached to an activity and an interrupting boundary event is executed, all the scheduled boundary events will be aborted and all the boundary events that have already started will continue to run until the entire workflow ends.
 {{% /alert %}}
 
+## Nested jump rules
```
DmytroKost: Let's rename it to just "Jump rules", since we might want to expend the scope of cases in this section to also include:
1. Jump rules related to jumping in or out of boundary events to the main path
2. Jumps between sibling or unrelated (not nested) boundary events

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -125,6 +121,16 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
...
+Boundary events come with a specific set of rules for jumps when they are nested, the variations are working meant for both a modelled jump and a runtime ad-hoc jump. In the following table are the differences specified.
+
+|Parent | Child | Description|
```
DmytroKost: We don't need to specify what kind of path is the child/parent path (parallel path, boundary event or main path). It doesn't matter when it comes to the jump options, because we currently don't allow to perform an ad hoc jump into the parent path or out of the child path (we will add an option to jump out of interrupting boundary events later, and if it's already implemented for the Jump activity we should specify it).
I suggest to organize the rules in the following way:
  - Jump inside a boundary event: not allowed (_and never will be regardless of how they are organized_)
  - Jump outside a boundary event: not allowed (will be allowed in the future for interrupting boundary events but only to the parent/grandparent path)
  - Jump within a boundary event: allowed

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -134,12 +140,6 @@ The list of variables is described below:
 * `$ParentTask` – the parent user task of the attached boundary event
 * `$CalledWorkflowInstance` – the parent Call workflow activity of the attached boundary event
 
-## Current Limitation {#limitation}
```
DmytroKost, line 137: Let's mention one limitation that becomes relevant mainly with the introduction of recurrence. The limitation is that a scheduled timer boundary event is not cancelable by another instance of the same boundary event. A more concrete example will be an ad hoc jump from MUT to itself. If we have a scheduled timer and then we perform a jump to self, effectively restarting the activity, a new timer will be scheduled. However, the old timer will still be pending, and if it expires first the new timer will be ignored. There might be other cases where the same behavior can be observed but I don't think we need to list them all. We are also planning or removing this limitation in the future.

DmytroKost:
The points regarding changing the boundary event type have to be completely changed, since they don't describe what happens accurately: https://github.com/mendix/docs/pull/9685/files#diff-5f3afa7b2f6cd6055dc4e5f9c58bc87e5668ac5991eaff9ab6b9ab37cad53d44R87

Point 1 says "The boundary event is re-created as the specified type. As a result, new IDs are created. These IDs are used by the Mendix Runtime for conflict detection analysis.", where the last sentence is incorrect, new IDs are not used for conflict detection. Instead, we should indicate that the boundary event will be scheduled with the specified type right after redeploy as soon as the workflow is in progress.

Point 2 says "The current activities on the boundary event path including the boundary event itself become incompatible and need to be restarted.". The workflow becomes incompatible rather then the activities. And that will happen only if the boundary event has already started, if it was scheduled it will be effectively ignored. I think it's useful to point to the 2 conflict types that will occur:
- If old boundary event was non-interrupting, we'll get "Non-interrupting boundary event path removed" conflict. I see that we haven't documented this new type of conflict, so we should add it to https://docs.mendix.com/refguide/workflow-versioning/ with the following mitigation options: restart, abort, continue, jump to the end of boundary event path.
- If old boundary event was interrupting, we'll get https://docs.mendix.com/refguide/workflow-versioning/#current-activity-removed conflict

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
PaulJongenMx: Comment to section: #### Implications of Changing the Boundary Event Type
Pls explain why we are doing it this way? I think it's mainly because of the possible migrations of inflight processes that would not make much sense.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -125,6 +121,16 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
 When there are multiple boundary events attached to an activity and an interrupting boundary event is executed, all the scheduled boundary events will be aborted and all the boundary events that have already started will continue to run until the entire workflow ends.
 {{% /alert %}}
 
+## Nested jump rules
+Boundary events come with a specific set of rules for jumps when they are nested, the variations are working meant for both a modelled jump and a runtime ad-hoc jump. In the following table are the differences specified.
+
+|Parent | Child | Description|
+| -------- |-------- | ------- |
+|Interrupting | Interrupting | When an interrupting boundary event is nested inside an interrupting boundary event we are allowed to either jump out of the boundary event flow or end the workflow.|
```
PaulJongenMx: @Yiyun333 pls check in this table the use of "we are allowed" or similar. Do we use this or we use "you" or make it neutral. Note: this is for the entire table.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -125,6 +121,16 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
...
+|Non-Interrupting | Interrupting | When a non-interrupting boundary event is nested inside an interrupting boundary event we are only allowed to jump in the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
+|Non-Interrupting | Non-Interrupting | When a non-interrupting boundary event is nested inside an non-interrupting boundary event we are only allowed to jump inside the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
+
 ## Boundary Event Variables
 
 Boundary events have dedicated variables that can be used to get direct access to the values of the parent activity if it is either a user task or Call workflow activity. You can get information such as the parent activity's `DueDate`, which can be used in the boundary event flow and its expressions. For instance, you can use the expression `addDays($ParentTask/DueDate, -2)` to configure a timer boundary event so that it is triggered two days before the due date of its parent user task.
```
PaulJongenMx: "... used to get direct access..." => "... used by activities in the path of the boundary event to get direct access..."

PaulJongenMx (commented):
only few comments. Well written explanation!

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
DmytroKost: Right, the main reason we're doing this is indeed because of execution limitations that we impose on ongoing boundary events. For example, the parent activity will be in-progress when a non-interrupting BE starts, and after it's type is changed to interrupting the parent activity will still stay in progress thus allowing to have more than one active interrupting boundary event. And vice versa, when an interrupting BE is changed to a non-interrupting the parent activity is already aborted, and since a non-interrupting BE cannot end the workflow it will hang forever unless aborted.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
Yiyun333: @PaulJongenMx @davevissermx We are usually a bit careful with using the word "allow" in general. When we use it, we use "Mendix" or the feature itself as the subject. In this case, I'll use "it is possible to xxx". I can make this change during my review but do let me know if you have a different opinion.

DmytroKost: @Yiyun333 the format of this section will change as a result of [this ](https://github.com/mendix/docs/pull/9685#discussion_r2151731622) discussion. However, we will still have to refer to a jump option as either possible or not. @davevissermx please, take the suggested phrasing into account.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -82,10 +82,14 @@ For an existing boundary event, when you change its type from non-interrupting t
 
 {{< figure src="/attachments/refguide/modeling/application-logic/workflows/boundary-events/security-dialog.png" alt="Security Dialog when changing type" width="450">}}
 
+The dialog is shown due to execution limitations on ongoing boundary events. For example the parent activity will be in progress when a non-interrupting BE starts. Once the type is changed it's parent activity will stay in progress, which would allow the parent to have more than the maximal allowed interrupting boundary events.
```
DmytroKost, line 85: I don't think we refer to boundary events as BE in the documentation

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -121,25 +125,25 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
...
 
-|Parent | Child | Description|
-| -------- |-------- | ------- |
-|Interrupting | Interrupting | When an interrupting boundary event is nested inside an interrupting boundary event we are allowed to either jump out of the boundary event flow or end the workflow.|
-|Interrupting | Non-Interrupting | When an interrupting boundary event is nested inside an non-interrupting boundary event we are only allowed to jump out of the boundary event flow to it's parent flow.|
-|Non-Interrupting | Interrupting | When a non-interrupting boundary event is nested inside an interrupting boundary event we are only allowed to jump in the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
-|Non-Interrupting | Non-Interrupting | When a non-interrupting boundary event is nested inside an non-interrupting boundary event we are only allowed to jump inside the nested non-interrupting boundary event flow, this follows the same rules as the parallel split outcomes.|
+- Jump to the inside of a boundary event: not allowed (and never will be regardless of how they are organized)
```
DmytroKost, line 131: I don't think we should say that it will never be allowed (I highlighted it with italic font, meaning for internal but should've used square brackets :) )

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -121,25 +125,25 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
...
+- Jump to the outside of a boundary event: only allowed from within an interrupting boundary event.
```
DmytroKost, line 132: It's not allowed to jump outside interrupting BEs yet with an ad hoc jump. I think we can specify that we plan to introduce this option in the future

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -121,25 +125,25 @@ An interrupting boundary event path must end with an **End** event or a **Jump**
...
 
 The list of variables is described below: 
 
 * `$ParentTask` – the parent user task of the attached boundary event
 * `$CalledWorkflowInstance` – the parent Call workflow activity of the attached boundary event
 
+## Current Limitation {#limitation}
+- When a workflow instance has a running timer boundary event which is set to recur. And for example in the running instance we do a manual jump to the parent activity of the running workflow. The already scheduled boundary event will not be cancelled whilst the new one is scheduled. When the old one executes in the mean time the new one will be disregarded. This is a limitation that will be fixed in the future.
```
DmytroKost, line 145: 1. This will happen regardless if it's a recurrent BE (no need to specify "which is set to recur")
2. "we do a manual jump to the parent activity of the running workflow" - we do a manual jump from a Multi-Input User Task to itself when its a parent activity of the boundary event, or a jump from the parent activity back in its executed path that leads to the activity to be executed for the second time (while a boundary event timer is scheduled and not yet expired).

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -233,3 +233,12 @@ You can do one of the following:
 * The workflow can be restarted, for example, by using the **DefaultWorkflowAdmin** page in the Workflow Commons.
 * The Administrator can use **Mark-as-Resolved** to fix this issue (the currently running activity within the removed path will be aborted).
 * The app developer can revert the change (which adds the path back) and deploy it.
+
+#### Non-interrupting boundary event path removed
```
DmytroKost, line 237: Please add this conflict to the [Conflict Mitigation Matrix](https://docs.mendix.com/refguide/workflow-versioning/#conflict-mitigation-matrix) (as the last row after Parallel Split Path Removed)

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -233,3 +233,12 @@ You can do one of the following:
...
+
+When an app developer removes a non-interrupting boundary event which was already executed, and then deploys this change. The currently running workflow instances that are executing activities within that path cannot continue execution.
```
DmytroKost, line 239: "a non-interrupting boundary event which was already executed" should be "an ongoing non-interrupting boundary event". @Yiyun333 What do you think is the preferred way to refer to a boundary event that has started its execution (meaning the workflow path that is guarded by the boundary event timer has started):
1. Started boundary event
2. Ongoing boundary event
3. Active boundary event
4. Current boundary event

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
```diff
@@ -233,3 +233,12 @@ You can do one of the following:
...
+
+* The workflow can be aborted, for example, by using the **DefaultWorkflowAdmin** page in the Workflow Commons.
+* The workflow can be restarted, for example, by using the **DefaultWorkflowAdmin** page in the Workflow Commons.
+* The Administrator can use **Mark-as-Resolved** to fix this issue (the currently running activity within the removed boundary event path will be aborted).
+* The app developer can revert the change (which adds the boundary event path back) and deploy it.
```
DmytroKost, line 244: Note: in this situation the conflict will be resolved and the boundary event will be re-scheduled again

content/en/docs/refguide/modeling/application-logic/workflows/workflow-versioning.md
Yiyun333: @DmytroKost I would use "ongoing boundary event".

DmytroKost:
All the changes are moved to https://github.com/mendix/docs/pull/9713

PR #9694, opened 2025-06-17
Add Migration note - Removal of EmailModule

akkrishnakumar:


akkrishnakumar:
Waiting for approval from Runtime Team

jveldhuizen (approved):

MarkvanMents (commented):
Hi @akkrishnakumar adn @jveldhuizen 
Is this for 11.0 GA?
Also, the API is still here https://apidocs.rnd.mendix.com/11/runtime/com/mendix/modules/email/EmailModule.html - should we remove it if it isn't supported any more?

jveldhuizen:
> Hi @akkrishnakumar adn @jveldhuizen Is this for 11.0 GA? Also, the API is still here https://apidocs.rnd.mendix.com/11/runtime/com/mendix/modules/email/EmailModule.html - should we remove it if it isn't supported any more?

Yes, but we missed the 11.0.0 deadline. So we already communicate that we don't support it anymore, but we will actually delete the API in 11.1. The API-docs will be automatically updated when 11.1 is released. We want to mention this already now, as documenting breaking changes is more visible in the main 11.0 documentation and the 10-11 migration document than only in some 11.1 release note.

akkrishnakumar:
> > Hi @akkrishnakumar adn @jveldhuizen Is this for 11.0 GA? Also, the API is still here https://apidocs.rnd.mendix.com/11/runtime/com/mendix/modules/email/EmailModule.html - should we remove it if it isn't supported any more?
> 
> Yes, but we missed the 11.0.0 deadline. So we already communicate that we don't support it anymore, but we will actually delete the API in 11.1. The API-docs will be automatically updated when 11.1 is released. We want to mention this already now, as documenting breaking changes is more visible in the main 11.0 documentation and the 10-11 migration document than only in some 11.1 release note.

We might be able to make it to Mx 11 release. I'm currently creating an MR. Let's see if we can sneak it into the 11 release.

MarkvanMents (approved):
I'll tag this for 11.0 GA and approve.
Thanks for the PR and responding to my question.

akkrishnakumar:
@MarkvanMents . We changed our mind. We decided to not merge the `EmailModule` removal MR. Instead, we will go forward with this PR. It mentions that we removed support, which means that this functionality will not work like before. In the next release, we will totally remove this API. 
I hope this will not confuse customers

MarkvanMents:
@akkrishnakumar - I've looked at the text and I think that will be clear to the customer.

Is it worth saying that it will be removed in the future. We could say either:
"The class will be removed completely in Mendix version 12.0"
or
"The class will be removed in a future Mendix release"?

Or would you rather not make a commitment at present?

akkrishnakumar:
I would not make a commitment right now

PR #9721, opened 2025-06-24
Add render-codeblock hook to tell Shutto to ignore translating codeblocks.

MarkvanMents:
Uses documentation here: https://gohugo.io/render-hooks/code-blocks/

With a bit of help from GPT to remind me there was a relevant render hook.


MarkvanMents:
Merging so I can test translations in production.

PR #9735, opened 2025-06-26
Update query-over.md

emirmx:


emirmx:
I tried to make it understandable but the terminology and the wording could be a bit off.

emirmx:
One more thing. The only real issue here is that  originally instead of `Specializations cannot be retrieved..` it should be `Generalizations cannot be retrieved...`. But I wanted to make it more clearer.

But if you think the original wording is fine then you can also only replace `Specialization` with `Generalizations`.

MarkvanMents:
Thanks - I'm having a look now, but trying to read the whole document so I have some context.
This is definitely difficult to get your head around if you are not an experienced DB developer/admin so any improvement is good!

content/en/docs/refguide/modeling/domain-model/associations/query-over.md
```diff
@@ -104,9 +104,8 @@ Here is an example inheritance:
 
 {{< figure src="/attachments/refguide/modeling/domain-model/associations/query-over/limitation.png" class="no-border" >}}
 
-In this example, a list of **Specializations** cannot be retrieved when using a standard by-association retrieve in a microflow if the input is the specialization.
-
-However, there is a workaround for this limitation: The list of Specializations can be retrieved with a Java action using the Java API. This Java action needs two parameters: the **Specialization** and a Boolean **Reverse** via this code snippet:
+In this example, if a standard by-association retrieve in a microflow is used starting from a `Specialization` this will return the `Specialization` that the starting point `Specialization` points to.
```
MarkvanMents: I think we need to be clearer that you do _not_ get the generalization. How about:
```suggestion
In this example, if a standard by-association retrieve in a microflow is used starting from a `Specialization` this will return the `Specialization` that the starting point `Specialization` points to and not the list of `Generalization` which are associated via the `Generalization_Specialization` association.
```

content/en/docs/refguide/modeling/domain-model/associations/query-over.md
```diff
@@ -104,9 +104,8 @@ Here is an example inheritance:
 
 {{< figure src="/attachments/refguide/modeling/domain-model/associations/query-over/limitation.png" class="no-border" >}}
 
-In this example, a list of **Specializations** cannot be retrieved when using a standard by-association retrieve in a microflow if the input is the specialization.
-
-However, there is a workaround for this limitation: The list of Specializations can be retrieved with a Java action using the Java API. This Java action needs two parameters: the **Specialization** and a Boolean **Reverse** via this code snippet:
```
MarkvanMents, line 109: I think it is worth putting a text description of the Java as well.

But on a more general note, however - although my programming days are about 30 years in the past, I'm not so happy with the class name or that the reverse is a parameter to it.
Would it not be better to have a class called RetrieveGeneralizationViaAssociation which is just dedicated to doing that (and perhaps name the object something more meaningful than `B`?

Perhaps this is clearer to a Java developer, but this might be needed by a low-code developer whose Java skills are as weak as mine.

If the customer wants to implement a more general-purpose version, then they can adapt the simple version?

MarkvanMents (commented):
@emirmx 

Thanks for the PR and the explanation of why it is needed.

I think I understand the issue and have made a couple of comments below. Let me know if you think they make sense.

content/en/docs/refguide/modeling/domain-model/associations/query-over.md
emirmx: Yes, this looks more understandable.

content/en/docs/refguide/modeling/domain-model/associations/query-over.md
emirmx: Yes, there seems to be a disconnect in names between the Java action and the model.

MarkvanMents (approved):
Thanks for the PR.
Approving and merging.

PR #9764, opened 2025-07-01
Boundary events scheduling and jumping updates

DmytroKost:
For 11.1.0

### DIFF
--- a/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
+++ b/content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
@@ -62,7 +62,7 @@ To configure the properties of a boundary event, double-click the event to open 
 
 #### Adding a Boundary Event to an Ongoing Activity
 
-When a boundary event is added to an ongoing activity, any workflow instances currently executing that activity will schedule the new boundary event accordingly. The only exception occurs when an ongoing boundary event is removed, the workflow is redeployed, and then the removal is reverted. In this case, the boundary event will not be rescheduled.
+When a boundary event is added to an ongoing activity, any workflow instances currently executing that activity will schedule the new boundary event accordingly. The only exception occurs when an ongoing boundary event is removed from the ongoing activity, the workflow is redeployed, and then the removal is reverted. In this case, the re-added boundary event will not be scheduled.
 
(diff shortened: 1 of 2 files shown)


content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -60,6 +60,10 @@ To add a boundary event to the [above-listed activities](#supported-activities),
 
 To configure the properties of a boundary event, double-click the event to open its properties dialog box. For more information on how to configure the properties of a timer boundary event, see [Boundary Properties](/refguide/timer/#boundary-properties).
 
+#### Adding a Boundary Event to an Ongoing Activity
+
+When a boundary event is added to an ongoing activity, any workflow instances currently executing that activity will schedule the new boundary event accordingly. The only exception occurs when an ongoing boundary event is removed, the workflow is redeployed, and then the removal is reverted. In this case, the boundary event will not be rescheduled.
```
Yiyun333: @DmytroKost 

I would use the following version but I am not sure if my understanding is correct. So I am writting my suggested version here and I've marked my changes in bold. Please let me know what you think of it:

> When a boundary event is added to an ongoing activity, any workflow instances currently executing that activity will schedule the new boundary event accordingly. The only exception occurs when an ongoing boundary event is removed **from the ongoing activity**, the workflow is redeployed, and then the removal is reverted. In this case, the **new** boundary event will not be **scheduled**.

DmytroKost: @Yiyun333 Your version is technically correct and I find it more clear. The point is that you can add a new boundary event by just adding it normally or though an undo of previously removed boundary event. In the latter case you are also adding a "new" boundary event, because firstly if was removed (it's not there) and then Undo adds it back (it is there). In that case, though, the "new" re-added boundary event will not be scheduled.

Yiyun333: @DmytroKost Thanks a lot for the explanation! In this case, "re-added" is better than "new" actually and I'll use "re-added" instead. See commit https://github.com/mendix/docs/pull/9764/commits/2ed08fbbfb35f2487722a88e59c23c14b90bde6a.

PR #9780, opened 2025-07-03
mendix-cloud-request-headers: case-insensitive header names

pommi:
Customers are not always aware that HTTP header keys are case-insensitive and only check for 1 case, which results in failures for other cases.


OlufunkeMoronfolu (approved):
Made some style changes. Approving now!

PR #9801, opened 2025-07-08
Add note about Translations GA

GulamY:
Add note about translations GA


MarkvanMents (approved):
Approving and merging

PR #9807, opened 2025-07-09
Update supported databases in system requirements

dahfjkg:
To be published with 11.2 release

MariaDB 10.5 and MySQL 8.0 will be announced as to be dropped, but it should be safe to already remove them from the requirements as they are not supported by the vendors.

### DIFF
--- a/content/en/docs/refguide/installation/system-requirements.md
+++ b/content/en/docs/refguide/installation/system-requirements.md
@@ -216,12 +216,12 @@ Mendix tries to support the most recent and patched database server versions fro
 
 Current support:
 
-* [MariaDB](/refguide/mysql/): 10.4, 10.5, 10.6, 10.11, 11.4
+* [MariaDB](/refguide/mysql/): 10.6, 10.11, 11.4, 11.8
 * [Microsoft SQL Server](/developerportal/deploy/mendix-on-windows-microsoft-sql-server/): 2019, 2022
 * [Azure SQL](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-compatibility-level?view=sql-server-2017): v12 compatibility mode 140 or higher
-* [MySQL](/refguide/mysql/): 8.0, 8.4
+* [MySQL](/refguide/mysql/): 8.4
 * [Oracle Database](/refguide/oracle/): 19, 21c
-* PostgreSQL: 12, 13, 14, 15, 16, 17
+* PostgreSQL: 13, 14, 15, 16, 17
 * [SAP HANA](/refguide/saphana/): 2.00.076.00.1705400033
 
 {{% alert color="warning" %}}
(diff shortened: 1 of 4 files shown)


MarkvanMents:
Hi Pavel,
Thanks for the PR.
Will there also be a release note for each of these changes?
I notice that there are multiple Mendix versions here but this is tied to the 11.2 release. Is that OK?

content/en/docs/refguide/installation/system-requirements.md
```diff
@@ -216,12 +216,12 @@ Mendix tries to support the most recent and patched database server versions fro
 
 Current support:
 
-* [MariaDB](/refguide/mysql/): 10.4, 10.5, 10.6, 10.11, 11.4
+* [MariaDB](/refguide/mysql/): 10.6, 10.11, 11.4, 11.8
 * [Microsoft SQL Server](/developerportal/deploy/mendix-on-windows-microsoft-sql-server/): 2019, 2022
 * [Azure SQL](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-database-transact-sql-compatibility-level?view=sql-server-2017): v12 compatibility mode 140 or higher
-* [MySQL](/refguide/mysql/): 8.0, 8.4
-* [Oracle Database](/refguide/oracle/): 19, 21c
```
pijuskri, line 223: Do you think we should do this change in a separate MR like we discussed in the standup?

dahfjkg:
> Hi Pavel, Thanks for the PR. Will there also be a release note for each of these changes? I notice that there are multiple Mendix versions here but this is tied to the 11.2 release. Is that OK?

Hi @MarkvanMents, we will have release notes for these, yes. I removed Oracle 23 from this PR as Pijus requested in case we do not manage to add support for it in  11.2

pijuskri:
Approved for technical review

MarkvanMents (commented):
Hi @dahfjkg and @pijuskri ,
I'm concerned that we are removing DB2 from Mx8 and Mx9. Is this because the versions we tested are no longer supported by IBM and we don't want to test against a later version.

Perhaps no one is using DB2 any more, but I think we need a bit more explanation in a release note as to why DB2 has gone completely in versions where we used to support it.

I think the deprecation in 8 and 9 and removal in 10 is a better way to do this. But if the only version we support is no longer supported by IBM that is a good reason - as long as we explain it in a release note.

dahfjkg:
> Hi @dahfjkg and @pijuskri , I'm concerned that we are removing DB2 from Mx8 and Mx9. Is this because the versions we tested are no longer supported by IBM and we don't want to test against a later version.
> 
> Perhaps no one is using DB2 any more, but I think we need a bit more explanation in a release note as to why DB2 has gone completely in versions where we used to support it.
> 
> I think the deprecation in 8 and 9 and removal in 10 is a better way to do this. But if the only version we support is no longer supported by IBM that is a good reason - as long as we explain it in a release note.

Hi Mark, this is actually an oversight from our side. DB2 should remain in the list for 8 and 9

MarkvanMents (approved):
Thanks for the PR.
Approved for publication with Mendix 11.2.

MarkvanMents (approved):
Thanks for the PR. Approving.

PR #9818, opened 2025-07-09
Update Antivirus Warning documentation on performance impact

rodrigocmoreira:


rodrigocmoreira:
Hi docs, I believe this documentation should not be on content/en/docs/refguide/modeling/app-performance/antivirus-exclusion.md since this is not related to app performance, but is a general advice

MarkvanMents:
@rodrigocmoreira 
Hi Rodrigo,
I agree that this isn't about App performance but about the performance of Studio Pro. There seems to be already some documentation about this here: https://docs.mendix.com/refguide/performance-tips/#antivirus-software
Should I move the entire text into this **Performance Tips** document? Seems like the best place for it? I don't think it is worth keeping it as a separate document as it isn't that long.

rodrigocmoreira:
> @rodrigocmoreira Hi Rodrigo, I agree that this isn't about App performance but about the performance of Studio Pro. There seems to be already some documentation about this here: https://docs.mendix.com/refguide/performance-tips/#antivirus-software Should I move the entire text into this **Performance Tips** document? Seems like the best place for it? I don't think it is worth keeping it as a separate document as it isn't that long.

Hi Mark, I think the documents should not be merged, today this document is already referenced by studio pro, so the user can click in a link that will bring him to this documentation page, another thing we need to make sure that this move would not affect older versions that already point to this link.

MarkvanMents:
> > @rodrigocmoreira Hi Rodrigo, I agree that this isn't about App performance but about the performance of Studio Pro. There seems to be already some documentation about this here: https://docs.mendix.com/refguide/performance-tips/#antivirus-software Should I move the entire text into this **Performance Tips** document? Seems like the best place for it? I don't think it is worth keeping it as a separate document as it isn't that long.
> 
> Hi Mark, I think the documents should not be merged, today this document is already referenced by studio pro, so the user can click in a link that will bring him to this documentation page, another thing we need to make sure that this move would not affect older versions that already point to this link.

I can make a redirect so that the links continue to work. Would that be OK for you? They can only point to the whole doc and not a particular anchor.

The other option is to move this document to be under the performance tips but keep the URL.

Any preference?

rodrigocmoreira:
> > > @rodrigocmoreira Hi Rodrigo, I agree that this isn't about App performance but about the performance of Studio Pro. There seems to be already some documentation about this here: https://docs.mendix.com/refguide/performance-tips/#antivirus-software Should I move the entire text into this **Performance Tips** document? Seems like the best place for it? I don't think it is worth keeping it as a separate document as it isn't that long.
> > 
> > 
> > Hi Mark, I think the documents should not be merged, today this document is already referenced by studio pro, so the user can click in a link that will bring him to this documentation page, another thing we need to make sure that this move would not affect older versions that already point to this link.
> 
> I can make a redirect so that the links continue to work. Would that be OK for you? They can only point to the whole doc and not a particular anchor.
> 
> The other option is to move this document to be under the performance tips but keep the URL.
> 
> Any preference?

Hi @MarkvanMents , if it's possible to set up a redirect in the documentation, that would be great. Even merging could work, as long as the redirect guides the user to the correct section.

MarkvanMents (approved):
@rodrigocmoreira I've moved the documentation to a more appropriate place and updated the text appropriately. Can you check it looks OK, then I can merge it.
Thanks for the PR.

MarkvanMents:
No negative response from @rodrigocmoreira - merging.

PR #9848, opened 2025-07-15
[UFC-1708] Add docs for Autofocus page property

brambrink:
Hi team,

This MR adds documentation for the newly introduced **Autofocus** page property, which will be introduced in 11.1. 

If you need any additional information, let me know!


WimJongeneel (approved):

WimJongeneel:
Maybe we can split the advice with regards to accessibility from the documentation of the factual behavior by moving it into an info box? More of a suggestion, I'll approve anyway.

ConnorLand (approved):

PR #9856, opened 2025-07-15
Add info on log retrieval

NicoletaComan:


NicoletaComan:
Confirmed via Slack (DM with Mario) that the docs are ready for publishing.

PR #9902, opened 2025-07-24
Add London to exception list

feddovanede:
As the UK is no longer part of the EU, replicating the data to Frankfurt means replicating to a different political region as well.


OlufunkeMoronfolu (approved):
LGTM 🚀

PR #9962, opened 2025-08-11
UIA Update private service documentation

MxMurshed:
- Add limitation for apps using Self-Signed certificate and PWA enabled
- Add support for health check endpoint

### DIFF
--- a/content/en/docs/marketplace/platform-supported-content/services/private-service.md
+++ b/content/en/docs/marketplace/platform-supported-content/services/private-service.md
@@ -57,6 +57,7 @@ The required resources depend on demand. For instance, to be able to generate 5 
 The following artifact is available for installing the service:
 
 * The Docker image for the PDF Document Generation service 
+* If you need to setup a health check you can use the health check endpoit included in the service at the path `/health`. This endpoint returns a status code of `200` and message `OK` if everything is working correctly.
 
 #### Installing through Docker
 
--- a/content/en/docs/marketplace/platform-supported-content/services/private-service.md
+++ b/content/en/docs/marketplace/platform-supported-content/services/private-service.md
@@ -29,7 +29,7 @@ These are the limitations of using the private PDF Document Generation service:
 * Setup, management, and monitoring of the service is your responsibility.
 * The service is open to all applications that can access it. If additional access restrictions are required, you need to set these up at the network level and configure them.
 * You cannot import custom certificate authorities. Apps that use a self-signed or internal certificate are only supported when you disable certificate validation in the service.
-* The service does not support apps with PWA enabled when using a self-signed and disabling certificate validation in the service.
+* The service does not support apps that are published as a Progressive Web App (PWA) when using a self-signed certificate and disabling certificate validation in the service.
 * You are responsible for setting up a retry mechanism in the application to handle failures or timeouts of the service.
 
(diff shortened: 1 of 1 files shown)


content/en/docs/marketplace/platform-supported-content/services/private-service.md
```diff
@@ -29,6 +29,7 @@ These are the limitations of using the private PDF Document Generation service:
 * Setup, management, and monitoring of the service is your responsibility.
 * The service is open to all applications that can access it. If additional access restrictions are required, you need to set these up at the network level and configure them.
 * You cannot import custom certificate authorities. Apps that use a self-signed or internal certificate are only supported when you disable certificate validation in the service.
+* The service does not support apps with PWA enabled when using a self-signed and disabling certificate validation in the service.
```
allard-mx: I think `certificate` is missing after `when using a self-signed`

content/en/docs/marketplace/platform-supported-content/services/private-service.md
```diff
@@ -56,6 +57,7 @@ The required resources depend on demand. For instance, to be able to generate 5
 The following artifact is available for installing the service:
 
 * The Docker image for the PDF Document Generation service 
+* If you need to setup a health check you can use the health check endpoit included in the service at the path `/health`. This endpoint returns a status code of `200` and message `OK` if everything is working correctly.
```
allard-mx: `endpoit` should be `endpoint` here

allard-mx: Maybe we can move this line into a separate "Health check" section right after the current "Installing through Docker" section, so section 2.2.2 when looking at the current docs? 

I think it could deserve its own section, as it does not seem a separate artifact to be, but is now is mentioned after the "The following artifact is available for installing the service" line

content/en/docs/marketplace/platform-supported-content/services/private-service.md
allard-mx: Maybe `apps that are published as a Progressive Web App (PWA)` instead of just `PWA`, as that matches exactly how the option is shown in Studio Pro. What do you think?

content/en/docs/marketplace/platform-supported-content/services/private-service.md
MxMurshed: @allard-mx Not sure. under installing with Docker `2.2.2` or another section all together `2.3` since it's not only for Docker but mostly for Kubernetes deployments?

allard-mx: Yeah good point, agree that it should not be linked to Docker specifically. However, I think chapter 2.2 is currently about installing the service in general, not specifically Docker. As section 2.2.1 is currently "Installing through Docker", I think section 2.2.2 should still work. What do you think?

MxMurshed: true indeed

MxMurshed:
@NicoletaComan This one is ready from our side :)

NicoletaComan (approved):

NicoletaComan:
Thank you, @MxMurshed!
I've added some small changes, and am merging this PR.

PR #9984, opened 2025-08-15
Review Resources docs

quinntracy:


content/en/docs/refguide/modeling/resources/java-actions.md
```diff
@@ -8,7 +8,7 @@ description: "Describes using Java Actions to extend the functionality of your M
 
 ## Introduction
```
quinntracy, line 9: There are a variety of parameter types that are not documented (Boolean, date and time, Integer, etc.). Do these need to be added to the Parameters section (similar to the table in the [Types ](https://docs.mendix.com/refguide/javascript-actions/#type) section in JavaScript Actions)

PR #9986, opened 2025-08-15
Add CAST to DECIMAL with parameters to OQL docs

dahfjkg:
Not tied to a release. Requires technical review.

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
+++ b/content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
@@ -738,6 +738,16 @@ The table below describes which `CAST` conversions are supported:
 
 Converting `DATETIME` or `BOOLEAN` to `STRING` returns different format per database.
 
+##### `DECIMAL` precision
+
+`DECIMAL` data type can have precision and scale as parameters:
+
+- `DECIMAL(<precision>, <scale>)` — in the case when both parameters are specified, those values are used as precision and scale of the resulting data type
+- `DECIMAL(<precision>)` — when only precision is specified, scaale is set to 0. The resulting data type in that case is `DECIMAL(<precision>, 0)`
+- `DECIMAL` — when no parameters are specified, default values are used. Precision is set to 28, and scale is set to 8: `DECIMAL(28, 8)`
+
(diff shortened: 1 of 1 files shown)


pijuskri:
Approved by team for technical review

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -736,7 +736,19 @@ The table below describes which `CAST` conversions are supported:
...
+
+⁴See [`DECIMAL` precision](#dec-prec), below, for further information.
+
+##### `DECIMAL` precision{#dec-prec}
+
+`DECIMAL` data type can have precision and scale as parameters. This will have the following impact on the way data is converted:
+
+* `DECIMAL(<precision>, <scale>)` – when both parameters are specified, those values are used as precision and scale of the resulting data type
+* `DECIMAL(<precision>)` – when only precision is specified, scale is set to 0. The resulting data type in that case is `DECIMAL(<precision>, 0)`
+* `DECIMAL` – when no parameters are specified, default values are used. Precision is set to 28, and scale is set to 8: `DECIMAL(28, 8)`
```
MarkvanMents, line 749: I'm very surprised that a different default for **scale** is applied when only precision is specified. And especially that the default is 0 as that will drop all the fractional part and only give an integer value. I would expect the default for scale to be the same, whether I specified a precision or not.

Is this correct?

MarkvanMents (approved):
Thanks for the PR.
Looks good, but I have one question about implementation.
Let me know if my change look OK and I can merge.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
dahfjkg: Yes, this is correct. When only precision is specified, it is a common thing to default scale to 0, for instance in [SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver17). The discrepancy here is that without any parameters DECIMAL is (28, 8), but that should match the default in Runtime. I suggest we keep this the way it is and think how to improve if there are customer questions.

PR #9994, opened 2025-08-20
Update distributing-native-apps.md

Praveen-MX:
Certificates cannot be uploaded in NBUI anymore. So removed the "certificate will be used by the Mendix Native Mobile Builder to sign your apps." part.


ConnorLand (approved):

ConnorLand:
@Praveen-MX and this only affects SP 11 docs right, not SP10?

PR #10004, opened 2025-08-20
Updated  `Limitations` for CRS

adeel2023:
Updates `Limitations` for Consumed Rest Service in refguids (10 and 11).
Executes action item from: https://mendix.atlassian.net/browse/REST-93

### DIFF
--- a/content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
+++ b/content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
@@ -34,7 +34,7 @@ The ability to import an OpenAPI/Swagger contract is currently in beta.
 * To use the request response to create a data structure automatically in your domain model, the response data should be in JSON format. It is possible to process other formats, such as XML or raw text, but you will need to extract the data you are looking for in a microflow.
 * It is currently not possible to configure parameters in the Query Parameters tab.
 * Streaming responses from APIs are currently not supported.
-* When the Send REST request action is executed in a microflow and the statusCode is in `2xx` range, but different than `200`, you can use the latestHttpResponse to get the StatusCode and Content of the request that was made.
+* A response can be mapped in a microflow only for responses with status code `200`. If a Send REST request action is executed in a microflow and the status code is in 2xx range, but different than `200`, use the latestHttpResponse to get the status code and response content.
 
 ### Prerequisites 
 
--- a/content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
+++ b/content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
@@ -32,7 +32,9 @@ The ability to import an OpenAPI/Swagger contract is currently in beta.
 ### Limitations
 
 * To use the request response to create a data structure automatically in your domain model, the response data should be in JSON format. It is possible to process other formats, such as XML or raw text, but you will need to extract the data you are looking for in a microflow.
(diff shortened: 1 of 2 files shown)


aimiliatheologimendix:
@quinntracy could you maybe take a look?

content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
```diff
@@ -32,7 +32,9 @@ The ability to import an OpenAPI/Swagger contract is currently in beta.
 ### Limitations
 
 * To use the request response to create a data structure automatically in your domain model, the response data should be in JSON format. It is possible to process other formats, such as XML or raw text, but you will need to extract the data you are looking for in a microflow.
-* It is currently not possible to configure parameters in the Query Parameters tab
+* It is currently not possible to configure parameters in the Query Parameters tab.
+* Streaming responses from APIs are currently not supported.
+* Implicit response-mapping is performed only for responses with status code `200`.
```
aimiliatheologimendix: I checked the docs MR and I would also suggest something different I think, because we never mention the term implicit mapping in the docs. Another example can be:

`When the Send REST request action is executed in a microflow and the statusCode is in 2xx range, but different than 200, you can use the latestHttpResponse to get the StatusCode and Content of the request that was made`

content/en/docs/refguide10/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
```diff
@@ -36,6 +36,9 @@ The ability to import an OpenAPI/Swagger contract was introduced as a beta featu
 ### Limitations
 
 * To use the request response to create a data structure automatically in your domain model, the response data should be in JSON format. It is possible to process other formats, such as XML or raw text, but you will need to extract the data you are looking for in a microflow.
+* It is currently not possible to configure parameters in the Query Parameters tab.
+* Streaming responses from APIs are currently not supported.
+* Implicit response-mapping is performed only for responses with status code `200`.
```
aimiliatheologimendix: Same as my comment above

quinntracy:
@aimiliatheologimendix No problem. Is this for the 11.2 release, or can it be merged after review?

aimiliatheologimendix:
@quinntracy this will be nice to get merged asap. It's not really related to a specific release but these are limitations that we found out adding them to the docs will help the users

content/en/docs/refguide/modeling/integration/rest-services/consumed-rest-services/consumed-rest-service.md
quinntracy: Could these two points be combined? For example, "Implicit response mapping is performed only for responses with status code `200`. If the Send REST request action is executed in a microflow and the status code is in 2xx range, but different than 200, use the latestHttpResponse to get the status code and response content."

aimiliatheologimendix: They could be, but implicit mapping is not a term that the customer knows 🐒 

We could combine them but maybe change the first sentence to: `Response can be mapped in a microflow only for responses with status code 200. etc....`

quinntracy: Oh sorry, I misread! Yes, this should work. If you agree, I can update both docs with this and then get it merged today.

aimiliatheologimendix: Yes that would be great!

PR #10027, opened 2025-08-27
Add limitation about editable behavior without store value in Combo Box

samuelreichert:
### Description

This PR updates the Combo Box widget documentation to include a new limitation regarding the use of the widget without a store value attribute.

Specifically, when a Combo Box is placed inside a data view and no store value attribute is configured, the Editability setting of the data view (set to Never or Conditionally) will not be applied to the Combo Box. This occurs because the widget has no target attribute to evaluate whether it should be editable.

A workaround is documented, recommending that developers manually configure the editability setting for each Combo Box individually in such cases.

---

### Changes

Added a new section under Limitations in the Combo Box widget docs:

Documented the issue with missing store value attributes and editability.

Provided a clear workaround.

Added a note that this is a known limitation.


ConnorLand:
Thanks for the new Limitations section and workaround @samuelreichert !

ConnorLand (approved):

PR #10035, opened 2025-08-28
Document authtoken timeout setting

WimJongeneel:


MarkvanMents (commented):
@WimJongeneel 
In which version was this change made?
Do we still have to mention the old version for people who are using 11.0 and 11.1?

EnasAbdelrazek (approved):

WimJongeneel:
Hi @MarkvanMents, the new alias will be released with 11.3. However, it's just an alias, the old setting will keep existing and we currently have no plans to deprecate it.

MarkvanMents (approved):
Looks good. Thanks for the PR.
I've added information about which versions this applies to in this commit: 72f85a352366fe944343c964af82a818977b8f11
Approving ready for publication with Mendix 11.3

PR #10041, opened 2025-08-29
Add documentation for the OQL module

passalaqua:


content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
```diff
@@ -0,0 +1,90 @@
...
+description: "Describes the configuration and usage of the OQL module, which is available in the Mendix Marketplace."
+---
+
+## Introduction
+
+The [OQL Module](https://marketplace.mendix.com/link/component/66876) executes [OQL](https://docs.mendix.com/refguide/oql/) queries and returns the results as a list of Mendix objects, a count of results in the database or a CSV file.
+
+### Typical Use Cases
+
+This module allows users to run arbitrary OQL queries without having to use Java actions or [View Entities](/refguide/view-entities/).
```
dahfjkg: I would mention here that view entities are best practice

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
```diff
@@ -0,0 +1,90 @@
...
+## Available OQL Actions
+
+All actions support defining [Named Parameters](#named-parameters) for queries.
+Only `SELECT` queries are supported and all queries are executed without access rules.
+
+### ExecuteOQLStatement {#executeoqlstatement}
+
+This action executes an OQL query, either directly or by loading one from a dataset. If you wish to use a dataset, you need to provide its fully qualified name (example "MyFirstModule.Dataset", instead of just "Dataset").
+
+The result of executing this query is converted into objects of the `returnEntity` type, which is then returned.
```
dahfjkg: I'd mention that the result is converted into a *list* of objects

passalaqua: Replaced with a more explicit `Returns` part.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
passalaqua: Should this be a warning or info block?

dahfjkg: Warning would be too much IMO, I'd make it info or simply part of the text

dahfjkg (approved):

dahfjkg:
@MarkvanMents this PR passed technical review and is ready for proofreading

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
```diff
@@ -0,0 +1,118 @@
...
+* `returnEntity`: Entity type to be used to store results
+* `amount`: If not `empty` or 0, will limit the number of results to at most this amount. It is recommended to use an explicit sorting order in the OQL query when using amount
+* `offset`: If not `empty` or 0, will offset the results of the query by this quantity. It is recommended to use an explicit sorting order in the OQL query when using offset
+* `preserveParameters`: If false, all parameters defined for this query will be cleared
+
+It returns the following:
+
+* A list of Mendix objects of type `returnEntity` using the following rules:
+    * Each query result column is stored in an attribute or association of the same name
+    * Attributes and associations of the `returnEntity` that are not present in the query are given the default value
```
MarkvanMents, line 63: Is this the correct interpretation of "will be ignored"

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
```diff
@@ -0,0 +1,118 @@
...
+* `statement`: OQL Query or fully qualified name of a Dataset (module and Dataset name) to be executed
+* `returnEntity`: Entity type to be used to store results
+* `amount`: If not `empty` or 0, will limit the number of results to at most this amount. It is recommended to use an explicit sorting order in the OQL query when using amount
+* `offset`: If not `empty` or 0, will offset the results of the query by this quantity. It is recommended to use an explicit sorting order in the OQL query when using offset
+* `preserveParameters`: If false, all parameters defined for this query will be cleared
+
+It returns the following:
+
+* A list of Mendix objects of type `returnEntity` using the following rules:
+    * Each query result column is stored in an attribute or association of the same name
```
MarkvanMents, line 62: What happens to columns which don't have a matching attribute or association, or columns which have the wrong data type for the attribute or association (or even where you try to update an association with an attribute value)?

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
```diff
@@ -0,0 +1,118 @@
...
+This action takes the following parameters:
+
+* `statement`: OQL Query or fully qualified name of a Dataset (module and Dataset name) to be executed
+* `returnEntity`: A FileDocument or one of its specializations where the resulting file will be stored
+* `removeNewLinesFromValues`: Indicates if new lines inside string values should be replaced with spaces
+* `zipResult`: Indicates if the resulting file should be compressed inside a ZIP file
+* `exportHeaders`: Indicates if the first line of the result should contain a header with the names of each column
+* `separatorChar`: Indicates what character should be used to separate columns in the result
+* `quoteChar`: Indicates what character should be used to quote string values. May be left empty if `escapeChar` is defined
+* `escapeChar`: Indicates what character should be used to escape spaces and other special characters if string values are unquoted. Only applicable if `escapeChar` is not defined
```
MarkvanMents: Typo, I think?
```suggestion
* `escapeChar`: Indicates what character should be used to escape spaces and other special characters if string values are unquoted. Only applicable if `quoteChar` is not defined
```

But how is `quoteChar` escaped if it appears within a quoted string?

MarkvanMents (commented):
I've proofread this to try to make it clearer (to me, at least).
See 5adda0f4c926402ee4caf33f03be6646759fa8ef

I've also left a couple of questions/comments.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
passalaqua: You get an error. Every column in the query must be present in the return entity and must be compatible. Not all attributes and associations of the return entity must be present in the query as columns, however.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
passalaqua: What do you mean by "default value" for an association in Mendix?

As for attributes, yes, they'll have whatever value a fresh object gives it. If there's a set default value, that one will be used.

passalaqua: One can argue that no set value is the default value for associations, so your statement is correct.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
passalaqua: We follow https://www.ietf.org/rfc/rfc4180.txt and duplicate the quote character (So an OQL column with the text `"abc"` is printed into a CSV with quoteChar `"` as `"""abc"""`.)

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
MarkvanMents: I'll mention this, just for completeness. See 6c6d7ebe9fb2919465e0c4c33783c992c1452086

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
MarkvanMents: OK, I'll leave this in then. Thanks for confirming.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
MarkvanMents: Can you confirm that you did mean `quoteChar` here and I should commit the suggestion?

Do you think we need to be explicit about escaping the quote character, or is this so standard as to not be worth it?

MarkvanMents (changes requested):
Looking good.
There is just one question I'd like you to confirm. Thanks.

content/en/docs/marketplace/platform-supported-content/modules/oql-module.md
passalaqua: > Can you confirm that you did mean quoteChar here and I should commit the suggestion?

Yes, thank you for noticing.

> Do you think we need to be explicit about escaping the quote character, or is this so standard as to not be worth it?

It's pretty standard, but I admit it was not obvious to me at first either, so I wouldn't mind if we made it explicit in the `quoteChar` point.

MarkvanMents (approved):
Ready for immediate publication.
Thanks for working with me. 
Approving.

PR #10059, opened 2025-09-04
Add release notes for the private PDF Document Generation service v1.0.2

MxMurshed:
We have released the new version `1.0.2` and it's now latest.


MarkvanMents (approved):
Thanks for the PR.
I've made a very slight change in formatting. 
Approving and merging.

PR #10071, opened 2025-09-08
Add documentation for OQL UPDATE statement for attributes

dahfjkg:
Documentation for functionality to be released in 11.3

Requires technical review by the team


content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
```diff
@@ -0,0 +1,118 @@
...
+    Location = Module.Customer_Address/Module.Address/City,
+    
+
+```
+
+### OQL `UPDATE` Limitations
+
+* At the moment, it is only possible to update attributes, not associations.
+* If a subquery or a long path over a many-to-one or many-to-many association is used as `expression`, it can result in muptile values. In that case, a database-level exception will occur when running the statement.
+* In case of inheritance, all attributes should belong to the same inheritance level. It is not possible to update attributes of a generalization and a specialization in the same statement.
```
basdebakker: This is not correct. You can update all attributes in one statement. What you cannot do is use the value of an attribute declared in one of the involved entities in the expression for an attribute declared in a different entity.

content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
```diff
@@ -0,0 +1,118 @@
...
+```sql
+UPDATE Module.Order SET CustomerName = 'Mary' WHERE ID IN
+  ( SELECT ID FROM Module.Order
+    INNER JOIN Module.Customer ON Module.Customer/CustomerID = Module.Order/CustomerID
+    WHERE Module.Customer/Name = 'Mary' )
+```
+
+## General limitations for OQL statements {#oql-limitations}
+
+* OQL statements can be used only with persitable entities.
```
basdebakker: persi[s]table

content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
```diff
@@ -0,0 +1,118 @@
...
+    ),
+    Location = Module.Customer_Address/Module.Address/City,
+    
+
+```
+
+### OQL `UPDATE` Limitations
+
+* At the moment, it is only possible to update attributes, not associations.
+* If a subquery or a long path over a many-to-one or many-to-many association is used as `expression`, it can result in muptile values. In that case, a database-level exception will occur when running the statement.
```
basdebakker: s/muptile/multiple/

content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
basdebakker: I was incomplete. To run into this, all the attributes mentioned must be attributes that are themselves being updated. So you cannot do `SET SpecializationAttribute = GeneralizationAttribute, GeneralizationAttribute = SpecializationAttribute`. But just one of those at a time is fine.

basdebakker:
Approved from the technical side.

content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
```diff
@@ -0,0 +1,150 @@
+---
+title: "OQL Statements"
+url: /refguide/oql-statements/
+beta: true
+weight: 210
```
MarkvanMents: I wonder if this would be better _before_ **Switching to OQL Version 2** rather than after?

Not a major issue.

MarkvanMents (approved):
Thanks for the PR @dahfjkg 
I've done a quick proofread in this commit: d45c83ea32c9edd2bbb47a7a2cbacbcb39eb19bc and left one very minor comment.
Am approving and is ready to be merged with 11.3 - does not depend on the answer to my comment.
Mark

content/en/docs/refguide/modeling/domain-model/oql/oql-statements.md
dahfjkg: Hi Mark, yes, that would make total sense

MarkvanMents: Done

PR #10087, opened 2025-09-10
Create escalation-flowchart

dperruzzi:
final escalation flow


NicoletaComan:
Hi, @dperruzzi. Thank you for your contribution. I have two asks, please:
1. Would you be able to provide some context around this PR?
What is the business case that prompted it? Why and where is it needed?
2. Could you please sign the CLA?

Thank you!

MarkvanMents:
@NicoletaComan 
Thanks for picking this up.
Dan is in Support and I believe this should replace the image in the support escalation document: https://docs.mendix.com/support/escalation-management-process/#escalation-process-overview

Might be better and more readable to be vertical?

Dan will also be creating another PR - I have had some PMs about it while you were on leave. I'll tell him that you will be picking this up.

NicoletaComan (approved):

PR #10093, opened 2025-09-11
Update dev-best-practices.md

ghost:
Updated the description for why an app should be single purpose


MarkvanMents:
@Yiyun333 - can you pick this up - I think it would normally be for Quinn?

Yiyun333 (approved):

PR #10105, opened 2025-09-16
Correct reference to Microflow section

hugomendix:
In the table in 2.4.2 Decision Method there's a hyper-link to the Microflow section.

However, there are 2 headings named microflow and even though the second one is tagged as `#microflow` the hyper-link refers to the first, which is incorrect.

I've now tagged the first one differently, but I have no means to verify that/if this actually fixes the problem.


MarkvanMents (approved):
Well-spotted! Thanks for correcting that.
I've done a local build and the link now goes to the right place
(The problem is that if there is no specific anchor, Hugo (the site builder) creates a link for all the headings to be used in the ToC. And, in this case, the created link had the same name as the link we defined and the browser will go to the first anchor with that name.)

Approving and merging.

PR #10159, opened 2025-09-30
Updated marketplace links with public links

mendix-nleimer:
- Replaced links to private marketplace with public listings in installation guide


mendix-nleimer:
@katarzyna-koltun-mx Pls. review & approve as discussed

PR #10173, opened 2025-10-02
Dux-487 string agg support in view entities

eng-myousif:
Support STRING_AGG aggregate function in View Entities via Studio Pro OQL editor.

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
+++ b/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
@@ -79,7 +79,7 @@ STRING_AGG ( attribute_path, separator )
 `separator` is any expression of type `STRING`. 
 
 {{% alert color="info" %}}
-This aggregate function is only supported in Java actions.
+This aggregate function is only supported in Java actions and View entities.
 {{% /alert %}}
 
 ### Examples
--- a/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
+++ b/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
@@ -78,10 +78,6 @@ STRING_AGG ( attribute_path, separator )
 
 `separator` is any expression of type `STRING`. 
 
-{{% alert color="info" %}}
-This aggregate function is only supported in Java actions and View entities.
-{{% /alert %}}
-
 ### Examples
 
 In the following examples, the `Sales.Product` entity has five objects with `Name` and `Stock` attributes:
--- a/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
+++ b/content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
@@ -78,6 +78,8 @@ STRING_AGG ( attribute_path, separator )
 
 `separator` is any expression of type `STRING`. 
 
+This aggregate function is supported in View Entities and Datasets starting from Mendix 11.2.0, whereas previously it was only available in Java actions.
+
 ### Examples
 
 In the following examples, the `Sales.Product` entity has five objects with `Name` and `Stock` attributes:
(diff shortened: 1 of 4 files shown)


content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -79,7 +79,7 @@ STRING_AGG ( attribute_path, separator )
 `separator` is any expression of type `STRING`. 
 
 {{% alert color="info" %}}
-This aggregate function is only supported in Java actions.
+This aggregate function is only supported in Java actions and View entities.
```
passalaqua: At this point we might as well invert and say "This aggregate function is not supported in Datasets".

passalaqua: Actually, scratch that, it's also accepted in Datasets, by the looks of it. Maybe we can just remove this alert box altogether.

eng-myousif: Thanks for the comment, I have now removed the alert altogether. Ready for review.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
```diff
@@ -78,10 +78,6 @@ STRING_AGG ( attribute_path, separator )
 
 `separator` is any expression of type `STRING`. 
 
-{{% alert color="info" %}}
-This aggregate function is only supported in Java actions.
-{{% /alert %}}
-
```
MarkvanMents, line 84: Which Mendix version was it supported in View Entities?
11.3?
I think we need to be clear which versions don't support it.

MarkvanMents (changes requested):
@eng-myousif 
Hi Mo,
Thanks for the PR.
Which version is this change made in?
I think we should add it to the release notes as well. And indicate the versions which _don't_ support it.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
eng-myousif: Thanks @MarkvanMents for your comment, It is started to be supported in View Entities starting from 11.2.0,
What about something like this:
"This aggregate function is supported in View Entities and Datasets starting from Mendix 11.2.0, whereas previously it was only available in Java actions."

eng-myousif: I have also now added a commit to add it to the release notes of 11.2.0, let me know if it is ok.

MarkvanMents: Great suggestion - feel free to add that text - I don't think it needs to be in an alert box since for anyone on the latest version it isn't going to be an issue.

content/en/docs/releasenotes/studio-pro/11/11.2.md
```diff
@@ -17,6 +17,7 @@ weight: 98
 * You can now upload a PDF document in Maia for Workflows. This allows you to generate workflows based on PDF documents in addition to text and image inputs. 
 * The **Changes** pane is now automatically updated if files on disk are changed outside Studio Pro. 
 * We added support for saving files using the new Client API by calling `saveFile` from `mx-api/data`. For more information, see the [Client API documentation](/apidocs-mxsdk/apidocs/client-api/).
+* We added support for string_agg aggregate function in View Entities and Datasets.
```
MarkvanMents: Great.
I notice that everywhere else we say "STRING_AGG" in capitals. Do you think we should say that here as well.
I'll take your lead on this as you know what customers will be looking for.

MarkvanMents (changes requested):
Great suggestions and change.
Please add the note back in, using your suggested text.

content/en/docs/refguide/modeling/domain-model/oql/oql-expressions.md
eng-myousif: Thanks Mark, i have added the suggested text

content/en/docs/releasenotes/studio-pro/11/11.2.md
eng-myousif: Good catch, i have corrected the casing to be `STRING_AGG`

MarkvanMents (approved):
Looks great.
Thanks for the PR and working with my suggestions.
I'll publish this today as it is already released.

PR #10179, opened 2025-10-03
renaming mendix tech to SISW

OlufunkeMoronfolu:


layouts/partials/print/print-footer.html
```diff
@@ -2,7 +2,7 @@
     <div class="container">
         <div class="row">
             <div class="col-4 text-left">
-                <span>© Mendix Technology BV 
+                <span>© Siemens Industry Software Netherlands B.V.
                     <script type="text/javascript">
                     var d=new Date();
```
MarkvanMents, line 7: While we are doing this, should we use the same code as the normal footer partial, so that we have consistency in how we are doing the date?

MarkvanMents (approved):
Looks good.
I have one additional comment, but approving anyhow.

layouts/partials/print/print-footer.html
OlufunkeMoronfolu: Yes, we definitely should. Updating right away!

PR #10182, opened 2025-10-03
SCIM updates

Karuna-Mendix:


content/en/docs/marketplace/platform-supported-content/modules/scim.md
```diff
@@ -14,6 +14,10 @@ SCIM is an abbreviation for System for Cross-domain Identity Management; a proto
 
 The SCIM module allows you to integrate your app with the Joiner, Mover, Leaver (JML) process in your organization. It enables the assignment of tasks to users before their first login to the app. Additionally, based on the information in your IdP (for example, user groups), the SCIM module automatically creates and deactivates users in your app. This functionality helps you to control Mendix user licensing costs.
 
+{{% alert color="warning" %}}
+SCIM module versions 1.0.0 to 2.0.0 were released as protected modules. Starting from version 3.0.0, the module is unprotected and treated as a new module. Direct migration to unprotected version of the module is not supported. As a result, transitioning requires manual reconfiguration.
```
JaapF: "is treated as a new module"
I would not be sure hwta this means. Also because it is passive tense.
Can we make it active sentence.
Maybe: "Studio Pro treats the unprotected version as a new module in your app model".
If my understanding is correct than maybe also say:
"As a result, transituioning requires you to remove previous version of SCIM and add and configure version 3.0.0 as if iut was a new module."

Karuna-Mendix: done in this commit: https://github.com/mendix/docs/pull/10182/commits/ff08ef2f38303a3dc947d9568e3378bd10997b86

PR #10190, opened 2025-10-06
[RUN-4425] Add clarification about object commits

atennapel:


lvanengelen-mx:
Approved from the Runtime team side.

PR #10195, opened 2025-10-07
OIDC SSO:  `private_key_jwt` info update

Karuna-Mendix:


content/en/docs/marketplace/platform-supported-content/modules/oidc.md
```diff
@@ -334,7 +334,7 @@ In this case, the OIDC client is the app you are making.
     The options are:
     * `client_secret_basic`: Your app will use the HTTP Basic Authentication scheme to authenticate itself at your IdP. This is the default. The `client_secret_basic` makes use of the `client-id` and `client-secret`.
     * `client_secret_post`: Your app will authenticate itself by including its `client_id` and `client_secret` in the payload of token requests. (Older versions of the OIDC SSO module used this method.)
-    * `private_key_jwt`: This method, introduced in version 4.1.0, uses asymmetric key cryptography (algorithm) for authentication. This is the best option for security. When you select the `private key` option, you can configure the following fields:
+    * `private_key_jwt`: This method, introduced in version 4.1.0, uses asymmetric key cryptography (algorithm) for authentication. This is the best option for security. However, Entra ID does not support this method. When you select the `private key` option, you can configure the following fields:
```
JaapF: Given that 70% of our customers are using Entra, I think it makes sense to make this a explicit alert box.

JaapF: Also, we need to be more concise. 
Let's say:

"Private_key_jwt is not yet supported with Entra ID. This is due to some specifics in Microsoft's implementation of private_key_jwt which require an enhancement of the OIDC SSO module." 

or something like that. EntraID does support it, but not in same way as Okta.

PR #10218, opened 2025-10-14
[RUN-4836] Add AbsoluteSessionTimeout runtime setting (11.4.0)

atennapel:
This setting is added in 11.4.0.


content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -72,6 +72,7 @@ The following custom settings can be configured:
...
+| <a id="AbsoluteSessionTimeout" href="#AbsoluteSessionTimeout">AbsoluteSessionTimeout</a> | Defines after how much time a session becomes invalid (in milliseconds). After that timeout, a session becomes eligible for removal. As opposed to the `SessionTimeout`, the `AbsoluteSessionTimeout` timeout is not affected by user interactions. The session will be destroyed at requests or until the next time a scheduled task runs to clean up the active sessions. If the setting is not set then absolute session timeout is disabled, this is the default. | |
```
jveldhuizen:
```suggestion
| <a id="AbsoluteSessionTimeout" href="#AbsoluteSessionTimeout">AbsoluteSessionTimeout</a> | Defines after how much time a session becomes invalid (in milliseconds). After that timeout, a session becomes eligible for removal. As opposed to the `SessionTimeout`, the `AbsoluteSessionTimeout` timeout is not affected by user interactions. The session will be destroyed at requests and periodically when sessions are cleaned up. If the setting is not set then this behavior is disabled, which is the default. | |
```

jveldhuizen:
Approved from Runtime Core side.

content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -72,6 +72,7 @@ The following custom settings can be configured:
...
+| <a id="AbsoluteSessionTimeout" href="#AbsoluteSessionTimeout">AbsoluteSessionTimeout</a> | Defines after how much time a session becomes invalid (in milliseconds). After that timeout, a session becomes eligible for removal. As opposed to the `SessionTimeout`, the `AbsoluteSessionTimeout` timeout is not affected by user interactions. The session will be destroyed at requests and periodically when sessions are cleaned up. If the setting is not set then this behavior is disabled, which is the default. | Disabled. *Introduced in Mendix 11.4.0.* |
```
MarkvanMents: @atennapel 
What happens if both `SessionTimeout` and `AbsoluteSessionTimeout` are set?
I guess this is only an issue if `SessionTimeout` is (a lot) less than `AbsoluteSessionTimeout`.

atennapel: - `SessionTimeout` is always active
- If `AbsoluteSessionTimeout` is enabled then the regular `SessionTimeout` will still work as before
- This means that inactive sessions will still be deleted from the database, even before the `AbsoluteSessionTimeout`

MarkvanMents: I've clarified this in e9d6b9116480ad4ce1df470f493a4a93d769d7ba. Hope it makes sense.

MarkvanMents (approved):
Thanks for the PR.

content/en/docs/refguide/runtime/custom-settings/_index.md
atennapel: Thanks! Yes, it's correct.

Issue #10263, opened 2025-10-22
Deprecated Page Widget Documentation


JMack2500:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [input-reference-set-selector](https://docs.mendix.com/refguide/input-reference-set-selector/)

Document link: [input-reference-set-selector.md](https://github.com/mendix/docs/blob/development/content/en/docs/refguide/modeling/pages/input-widgets/input-reference-set-selector.md)

## My Issue/Suggestion

This widget does not exist anymore and the docs reference it as if it does. I needed to find a forum post (https://community.mendix.com/link/spaces/community/questions/144506) to learn that the ComboBox has replaced it's functionality. The "Mendix Crash Course" still references this and Maia doesn't appear to "know" this either, just referring me to a "forbidden widgets" tab. There are several similar widgets like this I think. Thank you.

JMack2500:
Including the "Drop-Down" widget too.

MarkvanMents:
Hi @JMack2500 
Thanks for raising this issue.
I have spoken to the developers and clarified the behavior in this PR: #10275 
Thanks for wanting to make the Mendix documentation better. Please let us know if you spot anything else.
Yours
Mark

PR #10275, opened 2025-10-24
Clarify Dojo widget behavior in React apps

MarkvanMents:
Dojo widget availability in React apps is not clearly documented. See #10263


MarkvanMents:
Approved over Slack.

PR #10279, opened 2025-10-27
History pane 11.4 public beta

tknippenberg:


quinntracy:
@tknippenberg It appears that this information needs to also be in the [View Menu](https://docs.mendix.com/refguide/view-menu/) doc. Can you please add the necessary details into that doc?

PR #10294, opened 2025-10-30
Clarify 'Support Tools' section in help menu

tknippenberg:
Updated the 'Support Tools' section to clarify tool window navigation.


tknippenberg:
This is already current state in Mendix 11. I expect it may also apply to 10 LTS (and possibly before), but haven't checked.

tknippenberg:
If it's unclear from the 'files changed' tab. The only thing I added is
`**Tool Windows** -> `
for 3 options.

PR #10304, opened 2025-11-03
removed ticket 261700 entry from 10.24.11 to add to 10.24.10

furkanarabaci:


MarkvanMents (approved):
Thanks @furkanarabaci 
Looks great. Approving.

PR #10313, opened 2025-11-04
Workstation: Adding a test example to explain file device syntax and configuration

mendix-nleimer:
- Added a reproducible example to explain how to use the file device and it's syntax.
- Fixing misleading message syntax of file device 
- Renaming page to title for better clarity what the page is about


katarzyna-koltun-mx:
Merging into my branch to fix broken links.

PR #10314, opened 2025-11-04
Release Notes for DatePicker rendering issue

MxKevinBeqo:
Added release notes for the issue where the native DatePicker would not render in cases where there was a visibility condition.


MarkvanMents (approved):
Thanks for the update @MxKevinBeqo 
Just moved it into numerical order of ticket number and approving and merging.

PR #10332, opened 2025-11-06
Update amazon-translate.md

strategicalliances:
Updated the Amazon Translate documentation with the Terminology update


strategicalliances:
Hi @katarzyna-koltun-mx can you have a look at this update?
The code is already updated in the Marketplace, so feel free to put it live when you can.

PR #10423, opened 2025-11-27
Add PostgreSQL 18 and MS SQL 2025 to system requirements

dahfjkg:
To be published with 11.6 release


basdebakker:
The same applies to 10.24 and 9.24, so we should update `refguide10` and `refguide9` as well.
Other than that, LGTM.

quinntracy (approved):

PR #10431, opened 2025-12-01
Right-click menu doc

quinntracy:


content/en/docs/refguide/modeling/domain-model/right-click-menu.md
MarkvanMents: Should this come here, or one level down since we are talking about right-clicking on an entity? So `refguide/modeling/domain-model/entities/right-click-menu.md` ?

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
```diff
@@ -0,0 +1,64 @@
+---
+title: "Right-Click Menu"
```
MarkvanMents: Perhaps need to explicitly say "Right-Click Menus in the Domain Model"? We may want to extend this to other parts of Studio Pro if it seems useful.

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
```diff
@@ -0,0 +1,64 @@
...
+url: /refguide/right-click-menu/
+weight: 80
+#If moving or renaming this doc file, implement a temporary redirect and let the respective team know they should update the URL in the product. See Mapping to Products for more details.
+---
+
+## Introduction
+
+The entity right-click menu allows you to manage your entity by, for example, reviewing its properties, find usages within your app, and generating overview pages based on its data. You can also right-click an entity’s attributes to manage its data.
```
MarkvanMents: I wonder if we should split the examples into a separate sentence, to make the first sentence shorter and punchier? And since we are now talking about lots of different right-click menus, we can drop "entity" and talk about something like "Right-click menus in the domain model allow you to …"
Also
```suggestion
The entity right-click menu allows you to manage your entity by, for example, reviewing its properties, finding usages within your app, and generating overview pages based on its data. You can also right-click an entity’s attributes to manage its data.
```

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
```diff
@@ -0,0 +1,64 @@
...
+## Entity Menu Overview
+
+The entity right-click menu items are described in the table below.
+
+| Menu item        | Description                     | Shortcut key |
+|------------------|---------------------------------|--------------|
+| Go to Generalization |                  | |
+Select Generalization |                   | |
+| Add | Add an attribute, association, or validation rule to your entity          | |
+| Move to  | Move the entity to the domain model of one of the following locations: <br><ul><li> Adminsitration</li><li>Atlas_Core</li><li>Atlas_Web_content</li><li>BusinessEvents</li><li>DataWidgets</li><li>FeedbackModule</li><li> NanoflowCommons</li><li>WebActions</li> | |
```
MarkvanMents: Won't this just be the domain models from all the (editable) modules in the app?

MarkvanMents (commented):
This looks like a good start. Presumably we can link to other documentation for most of these functions - or do you think there is so much missing that we will need to add detail here?

I've made a couple of comments, but let me know if you need more from me.

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
quinntracy: Agreed

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
quinntracy: Probably, but the UI lists all these options, so that's why I listed them here too

quinntracy:
@MarkvanMents To confirm, does the "Find changes" option also refer to the [Find, Find Advanced, and Find Usages](/refguide/find-and-find-advanced/) doc? I can't find any specific documentation about the exact UI, but Maia says "The "Find changes" option for an attribute allows you to search for changes or usages of different elements, documents, XPaths, changes to various elements, in your app via the "Find Advanced" option under the Edit menu in Mendix Studio Pro." When clicked, it opens to the Find Results tab.

MarkvanMents:
> @MarkvanMents To confirm, does the "Find changes" option also refer to the [Find, Find Advanced, and Find Usages](/refguide/find-and-find-advanced/) doc? I can't find any specific documentation about the exact UI, but Maia says "The "Find changes" option for an attribute allows you to search for changes or usages of different elements, documents, XPaths, changes to various elements, in your app via the "Find Advanced" option under the Edit menu in Mendix Studio Pro." When clicked, it opens to the Find Results tab.

Trying it out in an app it seems that this is like "Find Usages" but only returns a usage where this particular attribute value is set/changed (e.g. in a microflow **Change Object** action). "Find Usages", in contrast, returns everywhere that the attribute is used (e.g. displayed on a page or as a parameter to a microflow) and includes cases where the value of the attribute isn't changed.

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
```diff
@@ -0,0 +1,65 @@
...
+
+The right-click menu in the domain model allows you to manage an entity, as well as its attributes and associations. For example, you can use the menus to review properties, find usages within your app, and generating overview pages based on an entity's data.
+
+## Entity Menu Overview
+
+The entity right-click menu items are described in the table below.
+
+| Menu item        | Description                     | Shortcut key |
+|------------------|---------------------------------|--------------|
+| Go to Generalization | Go to the [generalization](/refguide/entities/#generalization) entity of the chosen entity. | |
```
nerush: Generalization can be lowercase, as in the menu item.

content/en/docs/refguide/modeling/domain-model/right-click-menu.md
```diff
@@ -0,0 +1,65 @@
...
+| Select Generalization | Choose the entity the selected entity should [inherit](/refguide/generalization-and-association/#generalization-specialization-and-inheritance) from. | |
```
nerush: Generalization can be lowercase, as in the menu item.

PR #10436, opened 2025-12-01
Agent Editor in SP +1Q, Observability available now

dlhartveld:


MarkvanMents (approved):
Thanks for the update, David.
Approving and merging.

PR #10454, opened 2025-12-05
11.6: Update information about one way navigability display

jelmersnippe:
From 11.6 onwards one way navigable associations no longer show as a dotted line, they will only have an icon showing it's direction

### DIFF
--- a/content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
+++ b/content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
@@ -82,7 +82,7 @@ One-way navigable associations are associations that allow navigation to associa
 
 You can retrieve data from all other domain model associations in both directions.
 
-One-way navigable associations are represented by an icon next to the association name showing the direction of its navigability.
+One-way navigable associations are represented by an icon next to the association name showing the direction of its navigability. In Mendix versions below 11.6.0, they are also shown with a dashed arrow.
 
 ### On Delete Behavior {#delete-behavior}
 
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/domain-model/associations/association-properties.md
```diff
@@ -82,7 +82,7 @@ One-way navigable associations are associations that allow navigation to associa
 
 You can retrieve data from all other domain model associations in both directions.
 
-One-way navigable associations are represented by a dashed arrow in Domain model editor. An icon shows the direction of its navigability.
+One-way navigable associations are represented by an icon next to the association name showing the direction of its navigability.
```
MarkvanMents: Can you let me know what the icon looks like? Is it one of our standard ones, or will I need to try to get an SVG to upload the the website?

MarkvanMents:
```suggestion
One-way navigable associations are represented by an icon next to the association name showing the direction of its navigability. In Mendix versions below 11.6.0, they are also shown with a dashed arrow.
```

jelmersnippe: image

jelmersnippe: The 'icon' is not an svg but a filled polygon, rotated in the direction of the association. This is already present in the current version, the change that will be happening is the removal of the dashed line.
The screenshot above is taken in the current (pre 11.6) version, in which you can see the direction indicators

MarkvanMents: Thanks for the update. I'll have a think about whether we should show the icon in the documentation, but it looks clear enough that it perhaps isn't needed.

MarkvanMents (approved):
Thanks for the PR and answering my questions.
Approving ready for 11.6 release.

PR #10461, opened 2025-12-08
Redeploy app after project change

birdie1971:
Customers need to redeploy their app after they link it to another/new project.


OlufunkeMoronfolu (approved):
LGTM

PR #10469, opened 2025-12-09
Add Support for Multiple Workflow Endpoints

Yiyun333:


content/en/docs/refguide/modeling/application-logic/workflows/workflow-elements/end-event.md
```diff
@@ -6,14 +6,18 @@ weight: 50
 
 ## Introduction
 
-An **End Event** in workflows defines where a workflow process ends. When the workflow execution reaches this element, the process is completed and no further actions are taken.
+An **End** event in workflows defines where a workflow process ends. When the workflow execution reaches this element, the process is completed and no further actions are taken.
 
-A workflow can have multiple end events, as shown in the example below:
+A workflow can have multiple end events. Starting from Studio Pro 11.6, workflows can also terminate through any one of these end events. This means that activities with multiple outgoing paths, such as user tasks or decisions, can direct each path to its own **End** event, as shown in the example below:
```
timvandervoornmx: This line reads if we couldn't end through these events before: `Starting from Studio Pro 11.6, workflows can also terminate through any one of these end events.`

Not really sure what to say as an alternative but I think we should address it.

PR #10473, opened 2025-12-10
[UIA-1288] Add certificate requirements section

ozgeMendix:


content/en/docs/marketplace/platform-supported-content/modules/document-generation.md
```diff
@@ -191,6 +192,12 @@ You should have installed a stable release of [Node.js](https://nodejs.org/).  M
 
 {{% alert color="info" %}}Make sure that the Mendix Runtime has the applicable permissions to run the *node* executable.{{% /alert %}}
 
+###### Certificate Requirements
+
+Both the Mendix Runtime user and the Node.js component used by the DocGen Service must trust the Certificate Authority (CA). The user account running the Mendix Runtime needs read access to the CA certificate to verify the domain name, and the Node.js installation must also trust the same CA certificate.
+
+If Node.js does not have access to the certificate CA by default, set the `NODE_EXTRA_CA_CERTS` environment variable to point to your certificate file (`.cert` or `.pem` format).
```
MxMurshed: In some cases, Node.js cannot reach the certificate CA. To grant it access, you can add the `NODE_EXTRA_CA_CERTS` environment variable in the host machine to point to the app certificate CA `.ca` or `.pem`. This will include the certificate in the Node.js process when executed to run the document generation service. You can follow [this guide](https://support.humly.com/add-additional-certificate-authority-for-node-js) for details on setting it up.

content/en/docs/marketplace/platform-supported-content/modules/document-generation.md
```diff
@@ -191,6 +192,12 @@ You should have installed a stable release of [Node.js](https://nodejs.org/).  M
 
 {{% alert color="info" %}}Make sure that the Mendix Runtime has the applicable permissions to run the *node* executable.{{% /alert %}}
 
+###### Certificate Requirements
+
+Both the Mendix Runtime user and the Node.js component used by the DocGen Service must trust the Certificate Authority (CA). The user account running the Mendix Runtime needs read access to the CA certificate to verify the domain name, and the Node.js installation must also trust the same CA certificate.
```
MxMurshed: The Mendix Runtime and Node.js that is installed on the host machine must be able to verify the domain name of incoming requests by accessing the app's certificate CA. Since both run in the context of the user who is used to run the Mendix runtime, that user should have read access to the Certificate CA `.cert` or `.pem` file.

content/en/docs/marketplace/platform-supported-content/modules/document-generation.md
```diff
@@ -48,6 +48,7 @@ The [PDF Document Generation](https://marketplace.mendix.com/link/component/2115
...
 * The access and refresh tokens used to secure requests to the cloud service are stored unencrypted in the app database. No user roles have read access to these tokens, and all communication with the cloud service is encrypted by requiring HTTPS. However, do consider this when sharing a backup of the database with other developers. We will introduce encryption at a later stage.
 * If you have the [Application Performance Monitor (APM)](/appstore/partner-solutions/apd/) or [Application Performance Diagnostics (APD)](/appstore/partner-solutions/apd/) add-on enabled in your app, or the log level of the **Services** log node set to *Trace*, the PDF Document Generation module will not be able to generate documents when used in Mendix Cloud. This limitation is only applicable for apps built in Mendix 9.24.5 and below and Mendix 10.0.0.
+* When running the Document Generation Service in an air-gapped environment, Chromium cannot access the internet, so external dependencies such as fonts, stylesheets, or images will not load. Only resources included in the Mendix app or available on the internal network will work.
```
MxMurshed: * When running the Document Generation Service in an air-gapped environment, Chromium will not have access to the internet, hence, any external dependencies such as fonts, stylesheets, or images will not be loaded, which could cause the document generation service to fail to generate or result in incomplete documents.

allard-mx: This limitation is only applicable when self-hosting the service (or during development), maybe it would be good to make that explicit, as it could otherwise be a bit confusing imo. I would prefer something like the below. We could also consider to move it to an alert within the 'Running on self-hosted for airgapped and on-premises' section.

> When self-hosting the Document Generation Service in an airgapped environment, the browser will not have access to the internet. Be aware that any external dependencies such as fonts, stylesheets, or images will not be loaded, and could cause the document generation to fail or result in incomplete documents.

What do you think?

content/en/docs/marketplace/platform-supported-content/modules/document-generation.md
allard-mx: I would prefer something like the below. I think we do not need to mention that the service uses the Node.js runtime as we have a section about configuring Node.js right above this one. I've also updated the text to combine both paragraphs, as I think they're closely related to each other: you only need to configure read access in case the `NODE_EXTRA_CA_CERTS` path was required, as in the other scenario Node.js already recognized the CA from the trust store.

> Both the Mendix and Node.js runtime need to trust the SSL/TLS certificate that is used to secure connections to the Mendix app. In case you are using a custom Certificate Authority (CA), make sure that the Node.js runtime trusts the CA. If Node.js does not recognize the CA by default, set the `NODE_EXTRA_CA_CERTS` environment variable to point to the applicable certificate file (`.cert` or `.pem` format). Be aware that the user account that is used to execute the Mendix runtime needs read access to this file.

Let me know what your thoughts are.

content/en/docs/marketplace/platform-supported-content/modules/document-generation.md
allard-mx: See the above comment; I think we can combine these two paragraphs into a single one.

ozgeMendix:
Hi @NicoletaComan, all comments have been resolved. 
Could you please review and merge the PR?

NicoletaComan (approved):

NicoletaComan:
Thank you! PR reviewed and merged.

PR #10475, opened 2025-12-11
Update view entity description and remove beta alert

AssassinLYZ:
Removed beta status alert for view entities and updated description.


quinntracy:
@AssassinLYZ Is this to be released with 11.6?

PR #10482, opened 2025-12-12
GenAI Portal: Model upgrade info

Karuna-Mendix:


content/en/docs/marketplace/genai/mendix-cloud-genai/navigate_mxgenai.md
linaaaf, line 1: @Karuna-Mendix 

'Ensure you are using Mendix Cloud GenAI Connector v5.4.0 or above to see the upgraded model version reflected in your GenAI connectors after upgrading.'

This is correct but we can just add it later. For now, it is only important, that 5.3.0 is used - because only then, the latest Cohere model is supported. Just remove it for now

PR #10500, opened 2025-12-18
Clarified that optimistic locking is in beta

jdiehl:


MarkvanMents (approved):
Thanks for this, Jonathan.
Approving and Merging.

PR #10519, opened 2025-12-22
Do not claim that autonumbers do not have gaps

dahfjkg:


MarkvanMents (approved):
Thanks for the PR.
Looks good.
Approving and merging.

PR #10536, opened 2025-12-30
Isolating cookie tracking to test GSC crawl

OlufunkeMoronfolu:


layouts/partials/hooks/body-end.html
```diff
@@ -31,6 +31,8 @@
 </script>
 
 <!-- Olu: Added analytics as defined by Marketing/Legal. Read more here: https://github.com/mendix-web/mendix-analytics-package. You need access from the Mendix Web team -->
+{{ if not .Params.disable_usercentrics }}
```
MarkvanMents, line 34: So we need this for both development and production?

MarkvanMents (approved):
Looks good. Thanks for doing this.
I've asked one question, but if the answer is "yes" then all looks good, so I'll approve it anyhow.
Hope this helps with debugging, even if it isn't _the_ solution to the problem.

layouts/partials/hooks/body-end.html
OlufunkeMoronfolu: Yes, we do. It's the same script for both environments.

PR #10440, opened 2025-12-03
[RUN-3090] Document changes for Amazon S3 SDK upgrade in migration guide

emirmx:

PR #10550, opened 2026-01-06
[RUN-3090] Update for s3 sdk changes.

emirmx:
Explain settings changes and how to perform key update.


content/en/docs/refguide/runtime/custom-settings/_index.md
```diff
@@ -197,10 +197,10 @@ For deployments to Mendix Cloud, SAP BTP, and Mendix on Kubernetes, these settin
...
+| <a id="commendixstorages3UseV2Auth" href="#commendixstorages3UseV2Auth">com.<wbr>mendix.<wbr>storage.<wbr>s3.<wbr>UseV2Auth</a> | Lets the authentication policy use `Signature Version 2` instead of the default `Signature Version 4`. Set this setting to `true` when the endpoint does not support `Signature Version 4`.<br/><br/>This setting is not supported in Mendix version 11.6.0 and above. | false |
+| <a id="commendixstorages3EncryptionKeys" href="#commendixstorages3EncryptionKeys">com.<wbr>mendix.<wbr>storage.<wbr>s3.<wbr>EncryptionKeys</a> | List of keys which can be used to encrypt and decrypt data at rest in S3. The right key to decrypt the data with is automatically selected depending on with which key it was encrypted. Each encryption key consists of a key id, the encryption algorithm and the actual key (Base64 encoded).<br/><br/>In Mendix version 11.6.0 and above, only AES encryption algorithm is supported. For migration information see [S3 Encryption Key Update](/refguide/s3-encryption-key-update/).Example: {{< figure src="/attachments/refguide/runtime/custom-settings/code_snippet_2.png" class="no-border" >}} |   |
```
MarkvanMents: There is an awful lot of information packed into this table cell. Also, the code snippet is an image which makes it bad for accessibility purposes, and also means it can't be copied easily.
Now we have the S3 Encryption document, I would suggest removing the code snippet image from here and putting it in the S3 document as a proper code block.

content/en/docs/refguide/runtime/custom-settings/s3-encryption-key-update.md
```diff
@@ -0,0 +1,50 @@
...
+    // END USER CODE
+  }
+
+  …
+}
+```
+
+This Java action can be repeatedly called from a microflow that reads objects of type `System.FileDocument`' in batches. For example, see the following microflow:
+
+{{< figure src="/attachments/refguide9/runtime/custom-settings/batch-file-update-microflow.png" class="no-border" alt="Microflow which contains a loop which retrieves a list of FileDocument objects and rewrites them to rotate the encryption key." >}}
```
MarkvanMents, line 65: In this microflow, why to we pass the list of FileDocuments to the Java action before we count them? I feel this makes the purpose less clear.

content/en/docs/refguide/runtime/custom-settings/s3-encryption-key-update.md
```diff
@@ -0,0 +1,50 @@
...
+
+Here a list of FileDocument objects is retrieved, for example 100, ordered by the `FileID` attribute, and passed to the `JA_RefreshFileContents` Java action. This is repeated until all the files are processed. We store the last processed file's `FileID` attribute in case the process gets interrupted. This microflow can be executed in a task queue.
```
MarkvanMents: I think we could make this clearer by talking about the batch?
```suggestion
Here, a list of FileDocument objects is retrieved, for example 100, ordered by the `FileID` attribute, and passed to the `JA_RefreshFileContents` Java action. This is repeated until all the batches of files are processed. The microflow logs and stores the `FileID` attribute of the last processed file in each batch in case the process gets interrupted.

This microflow can be executed in a task queue.
```

MarkvanMents (changes requested):
Hi Emir,
Thanks for the PR. I've moved it under Custom-Settings as it seems to fit better there.
You can see my changes in https://github.com/mendix/docs/pull/10550/files/84ed2b1e9517ec4f54e7f554fc695e77e8d84c5f..cf05d674becae2ebdf6ed47dd196c7123bdd34d4.
I have a couple of questions - can you have a look at them?
Thanks.

content/en/docs/refguide/runtime/custom-settings/s3-encryption-key-update.md
emirmx: If you are suggesting swapping the order, i.e. first count and terminate if zero then pass the list to the java action, then agree it would be better.

content/en/docs/refguide/runtime/custom-settings/_index.md
emirmx: I applied your suggestion to refguide9, can apply to other others if it is ok

content/en/docs/refguide/runtime/custom-settings/s3-encryption-key-update.md
MarkvanMents: Yep - it took me a few minutes to work out why we were counting them after they had been processed - I thought the Java action might be changing something which the count then revealed.
I'm sure most people would work it out quicker, but there may be others who are as slow as me 😁

content/en/docs/refguide/runtime/custom-settings/_index.md
MarkvanMents: That looks great - exactly what I was thinking of.
Please go ahead and do it for Mx10 and Mx11 if you can.
Thanks.

MarkvanMents (approved):
Thanks for the PR.
Looks good.
Can I merge this and #10440 now, or do they need to wait for something?

PR #10557, opened 2026-01-07
Update SSO BYOIDP setup instructions

johanflikweert:
Added steps for configuring optional claims in the ID token.


JaapF (approved):
Ersel confirmed the step is correct.
So yeah, let's merge this step.

Karuna-Mendix (approved):

PR #10611, opened 2026-01-19
Clarify that XPath constraints could be case-sensitive

MarkvanMents:
There are some circumstances when XPath constraints on strings will be case-sensitive, although these are rare.

Also improved the navigation by removing redundant "XPath " from every entryl.

### DIFF
--- a/content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-constraint-functions/xpath-contains.md
+++ b/content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-constraint-functions/xpath-contains.md
@@ -1,12 +1,17 @@
 ---
 title: "XPath contains"
+linktitle: "contains"
 url: /refguide/xpath-contains/
 weight: 16
 ---
 
 ## Overview
 
-The `contains()` function tests whether a string attribute contains a specific string (case-insensitive) as a sub-string.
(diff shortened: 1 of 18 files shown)


content/en/docs/refguide/modeling/xpath/xpath-constraints/xpath-constraint-functions/xpath-contains.md
```diff
@@ -1,12 +1,17 @@
 ---
 title: "XPath contains"
+linktitle: "contains"
```
dahfjkg, line 3: What is the purpose of the new `linktitle` tag? And how safe is it to use words like `true`, `false` and `not` there? Would it be safer to add `xpath-` to link titles?

MarkvanMents: It means that the navigation just has the name of the function, rather than it being cluttered with the word "XPath" repeated multiple times.
image
image

dahfjkg: Ah, that actually looks better indeed

dahfjkg:
Looks good!

PR #10618, opened 2026-01-19
Update Studio Pro prerequisites (.NET 10 Upgrade)

rodrigocmoreira:
We are upgrading Studio Pro 11.6.3 and 11.7.0 onwards to work with .NET 10. 
We also need to make this a release note.


MarkvanMents:
Can you check that this is mentioned in the 11.7 release notes when they are frozen?

quinntracy:
The note has been added to 11.7.

quinntracy (approved):

PR #10626, opened 2026-01-20
Update the LLM model used by Maia in Mendix Portal

Yiyun333:


Yiyun333:
Technical review completed

PR #10637, opened 2026-01-22
UPdated release notes for Operator 2.25.1

Nidhi251289:
As reported by PMP, they have a number of environments where the S3 endpoint is just a hostname, without a schema.

In that case, Go's `url.Parse` will return an empty string hostname; in those cases, the hostname will be in `endpointURL.Path`.

This implementation prepends a placeholder schema, so that `url.Parse` can use its built-in hostname and path validation.


katarzyna-koltun-mx:

PR #10641, opened 2026-01-22
[ MOBILE ] chore: changelog update ( geolocation bug fix mobile )

karahanharunn:


ConnorLand:
These changes are live now, right? I don't have to wait until SP 11.7? @karahanharunn

karahanharunn:
Yes @ConnorLand, They are already live. We don't need to wait.

ConnorLand (approved):

PR #10655, opened 2026-01-26
[RUN-3090] Update s3 upgrade docs

emirmx:


MarkvanMents (approved):
@emirmx - I've proofread these docs and am happy with them.
Can you have a quick look and check that it still looks OK.
I also found a couple of references to 11.6 which I changed to the Mx10/9 versions as that seemed to make sense.
Let me know if this looks OK now.
Thanks.

emirmx:
> @emirmx - I've proofread these docs and am happy with them. Can you have a quick look and check that it still looks OK. I also found a couple of references to 11.6 which I changed to the Mx10/9 versions as that seemed to make sense. Let me know if this looks OK now. Thanks.

Thanks Mark. It looks good!

PR #10722, opened 2026-02-05
Remove duplicate WSC release notes

Karuna-Mendix:


Karuna-Mendix:
The decision is made to remove the WSC release notes as they are already available at the Marketplace. See the conversation here- https://mendix.slack.com/archives/C05N89ELXRV/p1770373597558479

PR #10727, opened 2026-02-08
add Korean translation

jsp-mx:
Hi,

Add korean translation under content/ko/docs


MarkvanMents:
Hi @jsp-mx,

Thank you for your PR and wanting to make the Mendix documentation better. However, I am going to reject the PR and close it for a number of reasons:

* Our site already has a Korean translation which you can access in the bottom right of the screen
    
    image

* Our site is continually updated and we would need to have a mechanism for keeping the Korean docs up-to-date (the current real-time translation does this automatically)
* The automatic translation (while not in any way perfect) does have the option to add a glossary and recognizes code blocks and other features which it knows not to translate
* Mendix localization is focused on multiple languages (Korean, Chinese, and Japanese) so we would like to have a consistent UX for all these languages, as well as other languages which we may add later
* We already have issues with the time it takes to build and upload our site when we make changes (which can happen several times a day) and uploading four times as many pages (if we did this for the three languages mentioned above) would make this much slower and hamper our ability to align our documentation updates with releases.

Thanks again for wanting to improve the documentation. I would love to hear your ideas and any issues there are for Korean users of the current documentation. Please get in touch if you would like to discuss this.

Mark

jsp-mx:
Then can I post the Korean version within my local repo?

jsp-mx:
The reason why I do the translation is that automated translation is broken Korean and more hard to understand.

MarkvanMents:
You are welcome to do it in your local repo.

I'd be interested to know how bad the Korean is, and whether there are specific terms that are badly translated - we do have a glossary with certain terms translated by the Siemens translation team, but we haven't been able to review that in Korean so feedback on specific problems would be useful. You can use an issue here, or contact me internally.

We have investigated alternatives to our current translation mechanism, but alternatives seemed even worse 😒 (at least for Chinese which is the language we were able to review).

As I said, this is the best maintainable solution for now, but we continue to look at our pipeline and translation tools to see if there are alternatives.

PR #10738, opened 2026-02-10
Update mx-dock to use different script per env

OlufunkeMoronfolu:


MarkvanMents (approved):
Tested - looks great. I can see that it is loading different MxDocks depending on environment.
Only thing I noticed is that the search bar isn't restricted to just docs, but I guess that is because that needs the real URL to work properly.
Approving.

PR #10740, opened 2026-02-10
Added the workaround for Back up documentation

Nidhi251289:
Updated the workaround for restoring database backups.


Nidhi251289:
@katarzyna-koltun-mx if this is reviewed, please go ahead with publishing it

PR #10769, opened 2026-02-17
Clarify user time zone sources and precedence

Robinfr:
## Summary
- clarify that browsers may report a time zone identifier with UTC offset and explain the fallback behavior
- state that explicitly configured user time zones take precedence for server operations
- applies to Mendix 9, 10, and 11

### DIFF
--- a/content/en/docs/refguide/runtime/date-and-time-handling/_index.md
+++ b/content/en/docs/refguide/runtime/date-and-time-handling/_index.md
@@ -20,7 +20,7 @@ For brevity we will call these time zones user time, UTC and server time.
 
 ## Changes
 
-The server needs to know the time zone of each user. Unfortunately, the web browser cannot report this information. It only reports the current offset to UTC. This is not enough to determine the exact time zone and properly deal with daylight saving time of future and past dates. For this reason web applications often offer you the option of setting your time zone and we have now built this option into Mendix, too. You can explicitly set a time zone for each user and this time zone is then used for server operations.
+The server needs to know the time zone of each user. On startup, the browser reports both the current offset to UTC and, when available, a time zone identifier (for example, `Europe/Amsterdam`) from the underlying platform. The runtime prefers the identifier to properly handle daylight saving time for future and past dates; if no identifier is available, it falls back to the offset. You can explicitly set a time zone for each user, and that configured value takes precedence for server operations.
 
 The entity TimeZone has been added to the System module along with an association between User and TimeZone. The TimeZone entity has three attributes: a code (based on the Olson database), a description suitable for showing to the user ("(GMT-05:00) New York") and a raw offset that is using for sorting the list of time zones.
 
@@ -34,7 +34,7 @@ To make use of the new date/time handling you have to take some action after con
 
 ### Single Time Zone App
 
-If you do nothing in a single time zone app where the server is also in that time zone the situation for server operations is actually slightly worse than before. Users will not have a time zone and if that is the case the server uses the current offset from UTC sent by the web browser. This offset is not enough to determine the exact time zone and this means that daylight saving time (DST) will not be taken into account. In practice this means that dates and times in the future and past - past DST changes - are one hour off.
+If you do nothing in a single time zone app where the server is also in that time zone the situation for server operations depends on what the browser can report. Users will not have a time zone and the server will use the time zone identifier reported by the browser when available, or otherwise the current offset from UTC. When only the offset is available, daylight saving time (DST) will not be taken into account. In practice this means that dates and times in the future and past - past DST changes - are one hour off.
 
 To make sure that all users have their time zone set, you have to set the default time zone in the app settings in Studio Pro. When the application is started again all existing users will get this default time zone. Also, when new users are created they will also get this default time zone automatically.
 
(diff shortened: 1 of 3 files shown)


Robinfr:
Can be merged after writers review

content/en/docs/refguide/runtime/date-and-time-handling/_index.md
```diff
@@ -6,65 +6,68 @@ description: "Describes date and time handling for a Mendix application."
...
-3. Server: The time zone that the server is running in is only used for scheduling the time at which scheduled events run. When defining a scheduled event you can choose whether you want to use server time or UTC time. For everything else, the server time is irrelevant.
+1. User/client – The time zone where the client is running. This is used for presenting dates and times to the end-user. The time zone is now (optionally) stored with each end-user to properly deal with daylight saving time (DST). If the time zone is not set, DST is not applied.
+2. Coordinated Universal Time (UTC) – The platform stores all dates in UTC. This is a time standard that is often used in servers to provide an unambiguous date format. It does not change with DST. Every time a date or time is presented to the end-user it is localized to the time zone of the client unless specified differently.
```
MarkvanMents: UTC does not only not change with DST, it does not change depending on where you are on the globe!

```suggestion
2. Coordinated Universal Time (UTC) – The platform stores all dates in UTC. This is a time standard that is often used in servers to provide an unambiguous date format. It does not change with DST, nor with where you are in the world. Every time a date or time is presented to the end-user it is localized to the time zone of the client unless specified differently.
```

content/en/docs/refguide/runtime/date-and-time-handling/_index.md
```diff
@@ -6,65 +6,68 @@ description: "Describes date and time handling for a Mendix application."
...
 
 ## Existing App
 
-To make use of the new date/time handling you have to take some action after converting your existing app. Those actions depend on the type of the app: single time zone or multiple time zone. In a single time zone app all users are in the time zone or they are at least willing to use the same time zone. The time zone of the server is not important, so an app is still single time zone if all users are in the Netherlands but the server is in England. Multiple time zone apps have users in different time zones. Let us see what you need to do in each case.
+To make use of the date/time handling you have to take some action after converting your existing app. Those actions depend on the type of the app: single time zone or multiple time zone. In a single time zone app all end-users are in the time zone or they are at least willing to use the same time zone. The time zone of the server is not important, so an app is still single time zone if all end-users are in the Netherlands but the server is in England. Multiple time zone apps have end-users in different time zones. Let us see what you need to do in each case.
```
MarkvanMents: All this section is about converting an existing app (which I think was written prior to Mendix 8). I think we should update it a bit?

```suggestion
To make use of the date/time handling you have to take some action when configuring your app. Those actions depend on the type of the app: single time zone or multiple time zone. In a single time zone app all end-users are in the time zone or they are at least willing to use the same time zone. The time zone of the server is not important, so an app is still single time zone if all end-users are in the Netherlands but the server is in England. Multiple time zone apps have end-users in different time zones. Let us see what you need to do in each case.
```

content/en/docs/refguide/runtime/date-and-time-handling/_index.md
```diff
@@ -6,65 +6,68 @@ description: "Describes date and time handling for a Mendix application."
...
+In the `Administration` module, the `Account_NewEdit` page adds a time zone selector to set the time zone of the end-user. By default, the `MyAccount` page does not have this selector. It is the administrator's task to correctly set the time zone. You can add it yourself if you want your end-users to be able to set their time zone.
 
-Studio Pro adds a setting to the **App Settings** dialog box. On the **Runtime** tab you can specify a **Default time zone**. This time zone is used for new users, but it is also applied to all users that do not have a time zone yet when starting your application.
+Studio Pro adds a setting to the **App Settings** dialog box. On the **Runtime** tab you can specify a **Default time zone**. This time zone is used for new end-users, but it is also applied to all end-users that do not have a time zone yet when starting your application.
 
 ## Existing App
```
MarkvanMents:
```suggestion
## Configuring Your App
```

content/en/docs/refguide/runtime/date-and-time-handling/_index.md
```diff
@@ -6,65 +6,68 @@ description: "Describes date and time handling for a Mendix application."
...
 
 ### Single Time Zone App
 
-If you do nothing in a single time zone app where the server is also in that time zone the situation for server operations is actually slightly worse than before. Users will not have a time zone and if that is the case the server uses the current offset from UTC sent by the web browser. This offset is not enough to determine the exact time zone and this means that daylight saving time (DST) will not be taken into account. In practice this means that dates and times in the future and past - past DST changes - are one hour off.
+If you do nothing in a single time zone app where the server is also in that time zone the date and time used for server operations depends on what the browser can report. End-users will not have a time zone and the server will use the time zone identifier reported by the browser when available, or otherwise the current offset from UTC. When only the offset is available, DST will not be taken into account. In practice this means that dates and times during daylight savings, including past and future dates, are one hour off.
```
MarkvanMents: I think there is some old text left in here?

```suggestion
If you do nothing in a single time zone app where the server is also in that time zone the date and time used for server operations depends on what the browser can report. The server will use the time zone identifier reported by the end-user's browser when available, or otherwise the current offset from UTC. When only the offset is available, DST will not be taken into account. In practice this means that dates and times during daylight savings, including past and future dates, are one hour off.
```

MarkvanMents (changes requested):
@Robinfr 

Thanks for the PR.

I've proofread this entire document as it was talking about some previous setting which seems to have been before Mendix 8, as the Mendix 8 version of [Date and Time Handling](https://docs.mendix.com/refguide8/date-and-time-handling/) talks about "Previously operations like … all used the server time zone."

I have a couple of places where I suggest some other clarification - can you have a look at that. Thanks.

If you are happy, I can copy the Mx11 version to Mx10 and Mx9.

content/en/docs/refguide/runtime/date-and-time-handling/_index.md
```diff
@@ -6,65 +6,68 @@ description: "Describes date and time handling for a Mendix application."
 
 ## Introduction
 
-The Mendix Server operations use the time zone of the user instead of the server time zone. Previously operations like generating documents, exporting to Excel/CSV and date computations in microflows/OQL all used the server time zone. This is fine if the server and all users of your application are in the same time zone. If they are not, however, generated output can contain unexpected results.
+The Mendix Server operations use the time zone of the end-user instead of the server time zone, where possible. This ensures that operations like generating documents, exporting to Excel/CSV, and date computations in microflows/OQL do not contain unexpected results.
```
MarkvanMents:
```suggestion
The Mendix Server operations use the time zone of the end-user instead of the server time zone, where possible. This ensures that operations like generating documents, exporting to Excel/CSV, and date computations in microflows/OQL do not produce unexpected results.
```

MarkvanMents (approved):

PR #10790, opened 2026-02-23
Doc MP Security Fixes

NicoletaComan:


content/en/docs/marketplace/upload-content/_index.md
```diff
@@ -64,16 +64,21 @@ Some fields only apply to specific component types, so follow the on-screen prom
     If you are uploading a solution, the **Upload Component Source** section is not displayed.    
     If you are uploading an industry template, selecting a component source in the **Upload Component Source** section is optional.    
 
-5. Select the **Studio Pro Version** on which you built the content.    
+5. Select the purpose of this release:
```
mameletzi, line 67: Assuming users might read the document top down, I am not sure if it makes sense to mention the release purpose here. Or maybe mention it, but add a sentence to indicate that this option is shown only when uploading a new version. I know it is clearly mentioned at section 5, but I would maybe mention it here as well.

NicoletaComan: That's a very good point, thank you.

PR #10791, opened 2026-02-23
Update SBOM Generation with Mx11 example output

quinntracy:


grootjans:
Checked, looks good to me!

PR #10840, opened 2026-03-05
Beta release modernized JavaScript Action Editor

WimJongeneel:
- To be released in 11.9
- Nees R&D review


EnasAbdelrazek:
Since we introduced some design changes, should we briefly mention them or/and include screenshots from the new design?

WimJongeneel:
> Since we introduced some design changes, should we briefly mention them or/and include screenshots from the new design?

I don't think that is needed tbh. All functionality is still in the same tabs, and pretty much all of it at the same place. Most of the design updates are just that we use more modern components. 

I think we should for now just keep the old screenshot, and when we move to the opt-out phase we can replace them with new one.

EnasAbdelrazek (approved):

Yiyun333 (approved):

Karuna-Mendix:
Merging for 11.9 release

PR #10854, opened 2026-03-09
Doc optimization for GSC and mobile viewers

OlufunkeMoronfolu:
Here is what to expect with this PR:
- Fixed the long-standing mobile unresponsiveness. Now, when on small screens, users can see the hamburger menu, which opens the sidebar when clicked. Test by minimizing your screen.
- The doc content is now reordered to enable better crawling and should solve our crawling problem with Google. Test this by using either the browser's:
  - View page source
  - Inspect

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -464,6 +464,28 @@ img {
     }
   }
 
+    /* Olu - Position the mobile close button and remove extra top spacing so content aligns */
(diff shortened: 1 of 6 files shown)


MarkvanMents (approved):
Hi @OlufunkeMoronfolu 
This looks great. 
Haven't been able to fully review the code, but can't see any issues when I test it.
Let's get it out there and check that Google is happy indexing the restructured page.
Thanks for much for picking this up.
Mark

PR #10860, opened 2026-03-09
Update debugging section with breakpoint limitations

jdiehl:
Added note about breakpoint debugging limitations during application startup.

Please add this note to mx10 and 9 as well


MarkvanMents (approved):
Approving and merging.

PR #10863, opened 2026-03-10
Add a release note for the fix reversing an association’s direction.

eng-myousif:
Add a release note for the fix that resolves the issue where reversing an association’s direction did not work.


MarkvanMents (approved):
Proofread and looks good.
Approving and merging.
Thanks for the PR.

PR #10888, opened 2026-03-13
Clarify JDK installation requirements in documentation

jdiehl:
Removed incorrect duplication. Please back port to Mx10


MarkvanMents (approved):
Thanks for the PR @jdiehl 
I've backported that bullet point to Mx10 and am approving and publishing.

PR #10899, opened 2026-03-17
Committing a Rolled-Back Object

MarkvanMents:


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -105,15 +105,39 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
 
-However, a Mendix commit is not the same as a database commit. For an object of a persistable entity, the saved value is not committed to the database until the microflow and any microflows from which it is called, complete. This means that errors in a microflow can initiate a rollback. If a microflow activity errors and has **Error handling** set to **Rollback** or **Custom with rollback**, the value of the object is rolled back to the value it had at the start of the microflow. See [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for more information.
+However, a Mendix commit is not the same as a database (SQL) COMMIT. When you use a **Commit object(s)** activity, Mendix actually performs an INSERT into the database. For an object of a persistable entity, the saved value is not committed to the database until the microflow and any microflows from which it is called, complete. This means that, although the current end user can see the updated version of the object, other end users will not see the updated object until the microflows end.
```
basdebakker: Only if the object is new, does a commit action perform a database `INSERT`. Otherwise, it performs a database `UPDATE`.

basdebakker: "the saved value is not committed" it still confusing, as it is not clear what type of commit this refers to. After the commit action, the object is in the database, but the transaction has not yet been committed. The effect is that any retrieve from database in that same transaction/microflow will see the (new state of) the object, but other transactions (including, but not limited to, other users) will not see the change. When the microflow ends, the transaction is committed and the change becomes visible globally.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -24,6 +24,15 @@ There are four error handling options:
 * **Custom without rollback**
 * **Continue**
 
+In summary, the four error handling options do the following:
+
+| Option | Behavior |
+| --- | --- |
+| Rollback (default) |  All changes are reverted to the state before the microflow started, the microflow aborts, and a system error message is shown. |
+| Custom with rollback |  All changes are reverted to the state before the microflow started, an error flow is followed, and the microflow's subsequent behavior depends on whether the error handling flow ends with an error or end event. |
+| Custom without rollback |  Changes made before the error are kept, an error flow is followed, and the microflow's subsequent behavior depends on whether the error handling flow ends with an error or end event. |
```
basdebakker: It may be more accurate to say that "Changes made before the activity causing the error are kept". If you commit a list of objects and the third one fails, the changes for the first two objects will be rolled back before the error flow starts.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -24,6 +24,15 @@ There are four error handling options:
...
+| Continue |  The microflow proceeds as if no error occurred, keeping all changes, and no error message is logged or displayed to the end-user. |
```
basdebakker: Similarly, it keeps all changes from before the activity causing the error.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -171,8 +184,29 @@ Consider the following best practices for error handling:
...
 * Where possible, avoid using the **Continue** option and use **Custom without rollback** instead.
+* Do not try to commit objects which were rolled back, Mendix will no longer see them as being changed. See [How Commits Work](/refguide/committing-objects/#how-commits-work) in *Commit Object(s)* for more information about this.
 * Do not overdo it – you can specify a lot of complicated error handling combinations, but this makes it more difficult (and slower) for the Mendix Runtime to evaluate the microflow, and it also makes it more difficult to predict the exact behavior in case of an exception.
 
+## What Error Handling Gives You
+
+### Consistent Data
+
+Everything that happens in your app happens in a transaction. By default either all instructions are completed or nothing is completed. If you don't specify any error handling and the microflow you are trying to run encounters an error, it will appear that nothing has changed. That means that all the objects you created or changed will be reverted, and the end-user will get an error. This ensures that processes and data remain consistent.
```
basdebakker: We don't tend to use "instructions", maybe "activities"? But this only applies to local objects, any calls to external systems cannot be reverted by a transaction rollback.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -171,8 +184,29 @@ Consider the following best practices for error handling:
...
+
+### Isolated Changes
+
+While updating or creating your objects, you do not want other end-users to see information until your microflow has finished executing. 
+
+To ensure that every end-user or process can only see persisted data, data changed in a microflow is only available within that microflow. None of the changes made inside the microflow will be available outside the microflow, not even to the end-user that initiated the microflow. The information will only be available to all end-users of the application once the microflow has successfully completed all the activities.
+
+### Protection from Parallel Updates
+
+When an object is updated, your app will place a lock on that object until the transaction/microflow ends. While the microflow is running, nothing else can read or write that same object and anything that attempts to do so will have to wait. This lock is released automatically as soon as the microflow ends, allowing any waiting processes to continue normally.
```
basdebakker: You can read an object that was updated and therefore locked by another transaction. Locks do not block reads, only other writes. The value you will see will not include any changes by uncommitted transactions.

MarkvanMents: Tried to clarify in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Added UPDATE in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

MarkvanMents: Tried to clarify in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Updated in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Updated in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Added this information in 89a5f3336953ea3f5b8ba388e6b806e496d294c4

MarkvanMents (commented):
Thanks for all the comments @basdebakker 
I've tried to address them in 89a5f3336953ea3f5b8ba388e6b806e496d294c4 - please let me know if it needs more clarification.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
basdebakker: I don't know how much it matters in practice, but strictly speaking the text is still incorrect. Other microflows (or just client retrieves) that belong to that end-user's app will also see the old version until the updating microflow finishes. Until that time, only the updating microflow itself will see the new version.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -105,15 +105,39 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
 
-{{% alert color="warning" %}}
-Deleting an object and then committing it can have different outcomes depending on whether the object has already been committed or not. If the object has already been committed, the delete will remove the object from the database, and the subsequent commit will have no effect. If the object is new (that is, it has not been committed before), the delete will do nothing, but the commit will store the object in the database. Therefore, this sequence of actions (a delete followed by a commit) may lead to unexpected results if the object has not been committed before.
-{{% /alert %}}
+#### What Gets Committed
+
+When you work on an object in memory, Mendix records whether the object has been changed. When you perform a **Commit object(s)** activity, the current value is written to the database and Mendix marks the object as unchanged in memory. This has a couple of consequences that you might not expect:
```
lvanengelen-mx: Earlier in this document, you changed 'current value' to 'changes to current values'. Should you make that distinction here as well?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -24,6 +24,15 @@ There are four error handling options:
 * **Custom without rollback**
 * **Continue**
 
+In summary, the four error handling options do the following:
+
+| Option | Behavior |
+| --- | --- |
+| Rollback (default) |  All changes are reverted to the state before the microflow started, the microflow aborts, and a system error message is shown. |
```
lvanengelen-mx: I think a system error message is only shown if the error is not caught using error handling on a higher level and is allowed to bubble up.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -24,6 +24,15 @@ There are four error handling options:
...
+| Custom with rollback |  All changes are reverted to the state before the microflow started, an error flow is followed, and the microflow's subsequent behavior depends on whether the error handling flow ends with an error or end event. |
```
lvanengelen-mx: 'All changes ... state before the microflow started' might be a bit ambiguous here.

In the following setup (client request triggers MF1, MF1 triggers MF2, MF2 triggers MF3, which fails)
```
request -> MF1
MF1: Start -> ... database stuff ... -> call MF2 Rollback -> End
MF2: Start -> ... more database stuff ... -> call MF3 Custom with rollback
MF3: FAIL
```
all database operations in 'database stuff' and 'more database stuff' are reverted. If MF1 did 'call MF2 Custom without rollback' instead, only 'more database stuff' is reverted.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -105,15 +105,39 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
+
+* If you commit an object which is then rolled back due to an error, committing the object again will not write the latest version to the database. You can understand this as the following sequence (see [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for a more detailed discussion of how rollbacks work during error handling):
+
+    1. Your microflow starts and creates a savepoint.
+    1. You change your object – it is marked as changed.
+    1. You perform a **Commit object(s)** activity which sends the changes to the database – the object is marked as unchanged.
+    1. An error occurs, the microflow ends and data in the database is rolled back to the savepoint.
```
lvanengelen-mx: If I read this correctly, this is an error _after_ the first commit, so unrelated to the commit itself. Maybe make that a bit more explicit as in first passing I read this as the commit itself failed, which results in different behavior.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -105,15 +105,39 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
+    1. You perform a **Commit object(s)** on the object again, but the changes are not written to the database because:
+
+        * The object in memory still has your changes
```
lvanengelen-mx: Unless any other changes were made after the first commit, I do not think this is correct. As the commit succeeded, the changes were applied to the object in memory and the object does not have changes. It's just that the earlier changes are now the current values. I.e. rolling back this object would keep the same values.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -105,15 +105,39 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
+
+* Deleting an object and then committing it has different outcomes depending on whether the object has already been committed or not:
+
+    * If the object has already been committed, the delete will remove the object from the database, and the subsequent commit will have no effect.
+    * If the object is new (that is, it has not been committed before), the delete will do nothing, but the commit will store the object in the database. Therefore, this sequence of actions (a delete followed by a commit) may lead to unexpected results if the object has not been committed before.
+
+
+#### Committing Non-Persistable Entities
+
+Mendix mimics this behavior for non-persistable entities. Committing a non-persistable entity means that you cannot use a rollback object activity to go back to the previous values, although the rollback error handling in a microflow rolls back to the original values.
```
lvanengelen-mx: I do not understand this part. Rollback on NPEs and PEs should work the same. After a commit the original values are replaced with the updated values, so a rollback will revert to the updated values. But that is the same for both NPEs and PEs.

basdebakker: A transaction rollback (which is what "the rollback error handling in a microflow" refers to) will rollback PEs to their original values from before the microflow, regardless of whether they were committed (as in commit object activity, not as in transaction commit).

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
emirmx: Maybe the next bullet point covers it? Or combine with it?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
emirmx: > If MF1 did 'call MF2 Custom without rollback' instead, only 'more database stuff' is reverted.

But aren't we talking about `Custom with rollback` here?

lvanengelen-mx: Yes, I just wanted to indicate how the call MF2 error handling configuration influences to which state the database is rolled back. This, to indicate that you do not know to what level the data is rolled back in the case of nested transactions. It could be to the database state at the start of the microflow calling the failing action (i.e. the start of MF2) or a database state at another level entirely.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: I am not sure I understand @basdebakker but as far as I know, in general, nothing of the in-memory state of a mendix object is touched on a transaction commit or rollback. The two are independent.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: But doesn't it always end up in the state before this MF is called - the error propagation might then change it again - but for an instant when this particular MF errors the rollback will only go as far as the calling microflow and then work out what to do with the error there?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Updated in ff31fe7c06a52b948f2ee63f92f2661fba363e70

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Updated in ff31fe7c06a52b948f2ee63f92f2661fba363e70

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Combined bullets in ff31fe7c06a52b948f2ee63f92f2661fba363e70

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Should we change this - I'm not sure how to make it clearer. I'm trying to say that PEs and NPEs work the same, but obviously I'm not being clear.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Updated in ff31fe7c06a52b948f2ee63f92f2661fba363e70

MarkvanMents (commented):
Thanks for the comprehensive review.
Resolved some of these (I hope) - there are still two discussions outstanding where I'm not sure what best to say.
Can you have another look and see if there are still changes I need to make.
Thanks for all your work so far.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
basdebakker: What @lvanengelen-mx is saying is just my point (I think). A transaction rollback will rollback any PEs to the state they were in before the transaction started. But it will not affect NPEs. So in that sense they are treated differently.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: Some feedback after our discussion:
>  All changes are reverted to the state before the microflow started, the microflow aborts, 

'the microflow': is it the microflow in which the error handling is defined or the microflow that is called by the Call Microflow action on which error handling is defined.

> and a system error message is shown unless custom error handling is added in a microflow which called this one

I would phrase it as cause -> effect, something like: 'if the error is not handled, a system error message is shown'

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: Not sure if I fully understand, probably because I am not 100% sure about what you mean by state. In the following:
```
request -> MF1
MF1: S -> A1 -> Call MF2 Rollback -> E
MF2: S -> A2 -> Call MF3 Custom with rollback -> E
MF3: S -> FAIL
```
when the `Call MF3` fails, execution continues in its error flow (execution state) while the database resets to the last savepoint, which is all the way to the start of `MF1`.

So we should probably be careful with the word 'state' as well :-)

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: To me, 'the object in memory still has changes' here can mean two things:
- the attributes have the same value as those that you set, but the original values are still there as well. I.e. rolling back this object will undo your changes
- the attributes have the same value as those that you set and the original values are the same. I.e. rolling back this object will keep the same values as the values you set.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: Bas and I came to the conclusion that we had a misunderstanding on what we meant with PE/NPE: state in memory vs state in the database.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Updated in c920e2e1b012ff60ec7f55c03d4895bdf448594d

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Updated to simplify - customers will need to look at the examples for the full picture.
c920e2e1b012ff60ec7f55c03d4895bdf448594d

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -103,21 +103,53 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
 
-However, a Mendix commit is not the same as a database commit. For an object of a persistable entity, the saved value is not committed to the database until the microflow and any microflows from which it is called, complete. This means that errors in a microflow can initiate a rollback. If a microflow activity errors and has **Error handling** set to **Rollback** or **Custom with rollback**, the value of the object is rolled back to the value it had at the start of the microflow. See [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for more information.
+However, a Mendix commit is not the same as a database (SQL) COMMIT. When you use a **Commit object(s)** activity, Mendix actually performs an INSERT or UPDATE on the database. For an object of a persistable entity, the database COMMIT is not performed until the microflow and any microflows from which it is called, complete. This means that, although a retrieve from the database by the end-user's app will see the updated version of the object, the updated object will not be seen globally by other end-users until the microflows end.
```
MarkvanMents: I have a contradiction between this statement and [Line 210 of error-handling-in-microflows](https://github.com/mendix/docs/pull/10899/changes#diff-c6263e928b74fedf46fedf9dbdac95a8259bbd6122d7d1cd3e7e0ad7e600da67R210) (which I pulled in from the Mx8 documentation as people said it was useful). In what circumstances does the current end-user continue to see their changes?

If they have uncommitted changes, are these stored in the client and still available if they start a transaction and retrieve the object? Or are they lost if they aren't committed?

This statement implies that the changes continue to be visible to the user even outside the microflow (e.g. if you retrieve into a data container) whereas line 210 of error-handling-in-microflows says the changed data is only visible within the microflow.

I guess this one is more accurate otherwise a **Change Object** with **Commit**: _No_ and **Refresh in Client**: _Yes_ wouldn't work?

Can you confirm?

MarkvanMents (commented):
Hi @basdebakker and @lvanengelen-mx 
I've tried to clarify things after my discussion with Leroy in these commits: https://github.com/mendix/docs/pull/10899/changes/23debd16bc93f77be2831a1c26d06c541ac3cd31..6aba2b817fcc7015908d2dc90198d8eff665aaae

I also have one outstanding question where I am not sure which statement is more accurate.

I hope these changes take us towards more clarity. Please let me know what you think.

Mark

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -103,13 +103,20 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
+
+1. The state in the database where the values are shared with every other user of the app. This is the state you get when you retrieve an object which is not already in the app's memory.
+2. The state in memory where values changed by the app can be seen.
+
+Non-persistable entities behave like the in-memory version of a persistable object.
+
 ### Committing Objects
 
-When you commit an object, all changes to the current values are saved. This means that you cannot roll back to the previous values of the object using the **Rollback object** activity of a microflow.
+When you commit an object which is in memory, all changes to the database values are saved. Once it is committed, you cannot roll back to the previous values of the object using the **Rollback object** activity of a microflow.
```
lvanengelen-mx: > all changes to the database values are saved

Do you mean all changes form the in-memory object are saved to the database? Or after the in-memory changes are saved to the database, the database values and in-memory object are in sync? The latter is not necessarily true if some other changes were committed earlier by some other user.

lvanengelen-mx: Maybe something like:
> When you commit an object which is in memory, its changes are saved to the database. The object is now unchanged and a roll back will have no effect.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: That line 210 says:
> To ensure that every end-user or process can only see persisted data, data changed in a microflow is only available within that microflow. None of the changes made inside the microflow will be available outside the microflow, not even to the end-user that initiated the microflow. The information will only be available to all end-users of the application once the microflow has successfully completed all the activities.

I do not fully understand what this is meant to say. Changes made by the user in the client are (ephemerally) stored there and applied on in-memory objects when calling a microflow. So the end-user has access to the changes, because they are in the browser.

>  In what circumstances does the current end-user continue to see their changes?

As long as they do not refresh the page and the changes are not committed, these are stored in the client and passed to the runtime when necessary, where they become in-memory objects. When the microflow re-retrieves such an object from the database, this is another copy in memory with the same object id. By itself, this does not interfere with the changes in the client.

However, if the **Refresh in Client** setting is used on the object retrieved from the database, the runtime will tell the client in the response that it should update its state cache with the retrieved object for that id. This will remove the original change.

Returning the retrieved object as the microflow return value keeps the change.

To be honest, I was not 100% sure about the behavior in these 3 scenarios, so I reproduced them in a test project.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -103,13 +103,20 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
 
-Another consequence of this is that, in contrast to an explicit **Rollback object** call, errors in a microflow can initiate a rollback. If a microflow activity errors and has **Error handling** set to **Rollback** or **Custom with rollback**, the value of the object is rolled back to the value it had at the start of the microflow. See [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for more information.
+Another consequence of this distinction is that, in contrast to an explicit **Rollback object** call, which rolls back to the last Mendix commit, errors in a microflow can initiate a database rollback. If a microflow activity errors and has **Error handling** set to **Rollback** or **Custom with rollback**, the value of the object in the database is rolled back to the value it had at the last savepoint. See [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for more information.
```
lvanengelen-mx: > ... the value of the object in the database is rolled back to the value it had at the last savepoint

Explicitly mention that the in-memory changes are kept?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
 * If you commit an object which is then rolled back due to an error, committing the object again will not write the latest version to the database. You can understand this as the following sequence (see [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for a more detailed discussion of how rollbacks work during error handling):
```
lvanengelen-mx: > If you commit an object which is then rolled back due to an error

Should we say instead that not _the object_ is rolled back but the transaction is rolled back to the last savepoint?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
...
 
     1. Your microflow starts and creates a savepoint.
-    1. You change your object – it is marked as changed.
-    1. You perform a **Commit object(s)** activity which sends the changes to the database – the object is marked as unchanged.
-    1. An error occurs after the **Commit object(s)** has successfully sent changes to the database, the microflow ends and data in the database is rolled back to the savepoint.
+    1. You change your object in memory – it is marked as changed.
+    1. You perform a **Commit object(s)** activity which sends the changes to the database – the object in memory is marked as unchanged.
+    1. An error occurs after the **Commit object(s)** has successfully sent changes to the database – the microflow ends and data in the database is rolled back to the savepoint.
```
lvanengelen-mx: Should we make it more explicit that the error occurs in another action after the commit?

Our implementation of commit could be seen as `Prologue > Database Commit > Epilogue` Any error in `Epilogue` should restore the object to its original state, as that is the contract of the **Commit** action. So this scenario is really about errors in the activity _after_ the commit.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
...
     1. You perform a **Commit object(s)** on the object again, but the changes are not written to the database because:
 
         * The object in memory still has your changes but it was marked as unchanged after your previous commit
```
lvanengelen-mx: I would avoid 'changes' here, as that implies that it is still in a changed state. Maybe 'committed values' is better?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
...
         * The **Commit object(s)** activity does not see the changed marker and so does not recognize that your object in memory has changes which need to be written.
 
-    If you want to keep the changes in the latest version you will have to work around this behavior.
+    If you want to keep the changes in the version which is in memory, you will have to work around this behavior by creating a new object which contains the changes.
```
lvanengelen-mx: I doubt that creating a new object is correct in general: you probably want it to be the same (i.e. with same object id) object. What would work is update any attribute value (even to its current value). That will mark the in-memory object as changed for the database. Not sure if we want to recommend that to users though...

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
...
-
+    * If the object in memory is new (that is, it has not been committed before you delete it), the delete will do nothing. However, the subsequent commit will write the object to the database. Therefore, this sequence of actions (a delete followed by a commit) may lead to unexpected results if the object has not been committed before.
 
 #### Committing Non-Persistable Entities
 
-Mendix mimics this behavior for non-persistable entities. Committing a non-persistable entity means that you cannot use a rollback object activity to go back to the previous values, although the rollback error handling in a microflow rolls back to the original values.
+Mendix mimics this behavior for non-persistable entities. This means that:
+
+* Performing a commit on a non-persistable entity means that you cannot use a **Rollback object** activity to go back to the previous values
+* If an error occurs in a microflow, and you have error handling which does a rollback, the values roll back to the state a the previous savepoint.
```
lvanengelen-mx: > the values roll back to the state a the previous savepoint.

The *database* values roll back to the state of the previous savepoint. But as non-persistable entities are not stored in the database, this does not matter for non-persistable entities.

Database rollback and object rollback are unrelated concepts. In-memory objects do not roll back on failure.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -118,29 +125,31 @@ When you work on an object in memory, Mendix records whether the object has been
...
 
 ### Autocommit and Associated Objects {#autocommit-and-associated-objects}
 
-When an object is committed through a default **Save** button, a commit activity, or web services, it always triggers the commit events. The platform also evaluates all associated objects. To guarantee data consistency, the platform may also autocommit associated objects.
+When an object in memory is committed through a default **Save** button, a commit activity, or web services, it always triggers the commit events. The platform also evaluates all associated objects. To guarantee data consistency, the platform may also autocommit associated objects.
```
lvanengelen-mx: > it always triggers the commit event

Maybe make it explicit that this also holds when the object is not marked as changed (and therefore no actual database save will take place)

lvanengelen-mx: > The platform also evaluates all associated objects.

What does this mean?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/events/error-event.md
```diff
@@ -10,13 +10,17 @@ This event can only be used in **Microflows**.
 
 ## Introduction
 
-An error event defines where a microflow will stop and throw an error that occurred earlier. If you call a microflow, you may want to know whether any errors occurred within the microflow or not. This event throws the error again, so the caller of the microflow can catch them. When you use this event, all database actions within the current transaction will be rolled back (for more information, see [Error Handling in Microflows](/refguide/error-handling-in-microflows/)).
+An error event defines where a microflow will stop and throw a new error based on the error that occurred earlier. If you call a microflow, you may want to know whether any errors occurred within the microflow or not.
+
+When you use this event, it creates a new error with the same information as the original error. Because this is a new error, even if the error that occurred earlier was caught **without rollback**, all database actions within the current transaction will be rolled back (for more information, see [Error Handling in Microflows](/refguide/error-handling-in-microflows/)).
```
lvanengelen-mx: > all database actions within the current transaction will be rolled back

Do you mean 'since the last savepoint' (considering transaction is only on the outer level)? Or don't we make that distinction here?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -28,12 +28,18 @@ In summary, the four error handling options do the following:
 
 | Option | Behavior |
 | --- | --- |
-| Rollback (default) |  All changes are reverted to the state before the microflow started, the microflow aborts, and a system error message is shown unless custom error handling is added in a microflow which called this one. |
-| Custom with rollback |  All changes are reverted to the state before the microflow started, an error flow is followed, and the microflow's subsequent behavior depends on whether the error handling flow ends with an error or end event. |
-| Custom without rollback |  Changes made before the activity causing the error are kept, an error flow is followed, and the microflow's subsequent behavior depends on whether the error handling flow ends with an error or end event. |
-| Continue |  The microflow proceeds as if no error occurred, keeping all changes made before the activity causing the error, and no error message is logged or displayed to the end-user. |
+| Rollback (default) | Reverts all changes, aborts the microflow, and returns the error to the calling microflow. |
+| Custom with rollback | Reverts all changes and follows the error flow. |
```
lvanengelen-mx: > Reverts all changes

Should we make it explicit that this about _database_ changes and that it reverts changes to the state of the last savepoint? Using 'all' might indicate all changes in the request?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -43,62 +49,62 @@ You can also include decisions in the error handling flow based on what you want
 
 It is important to know that when a microflow is started by an end-user (for example, by clicking a button), a savepoint is always created at the very beginning of the microflow. This is the case for **Microflow 1** in all the following examples. Knowing how savepoints are created and function is key to understanding how error handling works.
 
-In the following subsections, we introduce the functionality of each error handling option using simple examples.
+In the following subsections, we introduce the functionality of each error handling option using simple examples. In the examples we have an activity which generates the error. In a real app, however, an error can be generated by any activity. Any changes which happen within the activity which fails will not be applied (for example, if a [Call REST Service](/refguide/call-rest-action/) returns an error, any data the REST call retrieved will not be kept). The error handling only applies to activities which happened before the activity which errors.
```
lvanengelen-mx: > In the examples we have an activity which generates the error. In a real app, however, an error can be generated by any activity.

I do not see the contrast between 'an activity' and 'any activity'. Maybe change 'an activity' to 'a Microflow call action' or 'a specific activity'?

lvanengelen-mx: > The error handling only applies to activities which happened before the activity which errors.

Isn't this redundant? I can't see what the alternative would be.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -43,62 +49,62 @@ You can also include decisions in the error handling flow based on what you want
...
+When **Microflow 1** starts with a button click, a savepoint is created at the very beginning of **Microflow 1**. When an error occurs in **Microflow 2**, the default error handling is to roll back all changes to the state of this savepoint (so changes made in **Create Order** and in **Create Customer** are both reverted.), **Microflow 2** terminates, then **Microflow 1** terminates, the error is logged, and a system error message is shown to the end-user.
 
 {{% alert color="warning" %}}
-Rollback does not mark any uncommitted objects as changed. See [How Commits Work](/refguide/committing-objects/#how-commits-work) in *Commit Object(s)* to see the implications of this.
+After a rollback, previously committed objects remain marked as unchanged in memory, even though their database state has been reverted. See [How Commits Work](/refguide/committing-objects/#how-commits-work) in *Commit Object(s)* to see the implications of this.
```
lvanengelen-mx, line 63: > After a rollback

Explicit '_database_ rollback'?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -43,62 +49,62 @@ You can also include decisions in the error handling flow based on what you want
...
+When **Microflow 1** starts with a button click, a savepoint is created at the very beginning of **Microflow 1**. When an error occurs in **Microflow 2**, everything rolls back to the state of this savepoint because the error handling is set to **Custom with rollback**. Hence, changes made in **Create Order** and in **Create Customer** are both reverted. A custom error is logged using a **Log message** activity. After that, the error event throws a new error (with the same information as the original error) to terminate **Microflow 2**, and **Microflow 1** will terminate because there is no custom error handing there.
 
-#### Error Handling Which Ends with an End Event
+### Error Handling - Custom without Rollback
 
-In this example, the error handling in **Microflow 2** is set to **Custom with rollback** and the error handling flow ends with an end event.
+Setting **Custom without rollback** will not stop data changes within the activity which failed from being rolled back. For example, if a [Call REST Service](/refguide/call-rest-action/) returns an error, any data the REST call retrieved will not be kept.
```
lvanengelen-mx: > will not stop data changes within the activity which failed from being rolled back.

I find this sentence difficult to understand. Maybe interpreting this as 'rolling back' assumes too much. E.g. if the error handling is configured on a **Call microflow** action, then some of its side effects might be undone, some might not. In the given example of **Call REST Service** nothing is 'rolled back' per se, it's just that the return value is discarded.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -43,62 +49,62 @@ You can also include decisions in the error handling flow based on what you want
...
 
 {{% alert color="info" %}}
-Using custom error handling without rollback and ending the error handling flow with an error event here makes it possible to create a custom log message with details about the **NewOrder** (if you used error handling with rollback, the information about the **NewOrder** would not be available when creating the custom log message). However, after logging this message, everything will still be rolled back.  
+Using custom error handling without rollback and ending the error handling flow with an error event here makes it possible to create a custom log message with details about the **NewOrder** (if you used error handling with rollback, the information about the **NewOrder** would not be available when creating the custom log message). However, after logging this message, everything will still be rolled back as the error event is not handled in **Microflow 1**.
```
lvanengelen-mx: > (if you used error handling with rollback, the information about the **NewOrder** would not be available when creating the custom log message)

I do not understand why this would be the case. The database state is rolled back to the savepoint, but the data in the **NewOrder** object should still be available.

I have asked it within my team, waiting for response.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -208,7 +214,7 @@ To ensure that every end-user or process can only see persisted data, data chang
 When an object is updated, your app will place a write lock on that object until the transaction/microflow ends. While the microflow is running, nothing else can write to that same object and anything that attempts to do so will have to wait. This lock is released automatically as soon as the microflow ends, allowing any waiting processes to continue normally.
 
 {{% alert color="info" %}}
-This doesn't prevent two end-users editing the same object. Two end-users can read the same object and change it. However, neither will see the changes that the other has made. They can both commit the changes and the change committed second will be applied.
+This doesn't prevent two end-users editing the same object. Two end-users can read the same object and change it. However, neither will see the changes that the other has made. They can both commit the changes and the change committed second will overwrite the changes committed first.
```
lvanengelen-mx, line 217: > They can both commit the changes and the change committed second will overwrite the changes committed first.

Or it might error out, depending on whether [optimistic locking](https://docs.mendix.com/refguide/optimistic-locking/) is enabled.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: We have no idea why this data would not be available. I tried within a test project, and I can read the attributes from the object. Or is it meant that retrieving the data from the database won't show the changed values? Which would be logical, as they were rolled back 😃

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/events/error-event.md
MarkvanMents: Good point - we are trying to avoid using "transaction". Would it be best to say "all database actions within the microflow?". I don't want to start talking about savepoints as we are only talking about how we exit the microflow here.

Should I say
```suggestion
When you use this event, it creates a new error with the same information as the original error. Because this is a new error, even if the error that occurred earlier was caught **without rollback**, all database actions within the current microflow will be rolled back (for more information, see [Error Handling in Microflows](/refguide/error-handling-in-microflows/)).
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Good idea:
```suggestion
| Rollback (default) | Reverts all database changes, aborts the microflow, and returns the error to the calling microflow. |
| Custom with rollback | Reverts all changes and follows the error flow. |
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Good point. How about:
```suggestion
In the following subsections, we introduce the functionality of each error handling option using simple examples. In the examples we have a microflow call which generates the error. In a real app, however, an error can be generated by any activity. No results from the activity which fails will be available (for example, if a [Call REST Service](/refguide/call-rest-action/) returns an error, any data the REST call retrieved will not be kept). The error handling only applies to activities which happened before the activity which errors.
```

MarkvanMents: I added this because someone pointed out that people could think that if something like a change object with commit fails on the commit part, people might think that the change object bit of the activity still happens. This is supposed to say that nothing that happens within the activity will still be kept.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: I don't think we should say that here. Mendix uses the term "Rollback" not "Database Rollback" and I think we should be consistent with that. Unless you want to change the UX and say "Custom with Database Rollback" for the error handling option?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: This is the same thing as I tried to say before. How about:

```suggestion
Setting **Custom without rollback** will not stop data changes within the actual activity which failed being lost. For example, if a [Call REST Service](/refguide/call-rest-action/) returns an error, any data the REST call retrieved will not be kept.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: You need to log the information while you still have the object in context. So you can't error and then log it in the calling microflow. That is what this is saying: you can do the custom error handling to log information from the object, and then rethrow the error. But it sounds like there is a confusion about what is being rolled back here. Perhaps we can make it clearer:

```suggestion
Using custom error handling and ending the error handling flow with an error event here makes it possible to create a custom log message with details about the **NewOrder**. However, after logging this message, everything will still be rolled back as the error event is not handled in **Microflow 1**.  
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
MarkvanMents: Good point - this was written before optimistic locking was introduced. Perhaps we should say:

```suggestion
This doesn't prevent two end-users editing the same object. Two end-users can read the same object and change it. However, neither will see the changes that the other has made. They can both commit the changes and the result depends on whether [optimistic locking](/refguide/optimistic-locking/) in implemented.

* Without optimistic locking, the change committed second will overwrite the changes committed first.
* WIth optimistic locking, the change committed second will throw an error.
```

MarkvanMents (commented):
Responded to error-events.md and error-handling-in-microflows.md
Please let me know if suggestions are good.
Will look at committing-objects.md in separate review.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: I suggest:

```suggestion
When you commit an object which is in memory, changes to the values are saved in the database. Once it is committed, you cannot roll back to the previous values of the object using the **Rollback object** activity of a microflow.
```

But I want to specifically talk about **Rollback object** as this is different from the rollback which occurs with an error.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents:
```suggestion
Another consequence of this distinction is that, in contrast to an explicit **Rollback object** call, which rolls back to the last Mendix commit, errors in a microflow can initiate a database rollback. If a microflow activity errors and has **Error handling** set to **Rollback** or **Custom with rollback**, the value of the object in the database is rolled back to the value it had at the last savepoint. The object in memory will, however, keep any changes to the values. See [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for more information.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Good idea:

```suggestion
* If you commit an object and the transaction is then rolled back due to an error, committing the object again will not write the latest version to the database. You can understand this as the following sequence (see [Error Handling in Microflows](/refguide/error-handling-in-microflows/) for a more detailed discussion of how rollbacks work during error handling):
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: I'll make it more explicit:

```suggestion
    1. An error occurs in an activity after the **Commit object(s)** activity has successfully sent changes to the database – the microflow ends and data in the database is rolled back to the savepoint.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: How about:

```suggestion
        * The object in memory still has the updated values but it was marked as unchanged after your previous commit
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: How about:

```suggestion
    If you want to keep the changes in the version which is in memory, you will have to work around this behavior by modifying the object again.
```

Is this vague enough? I feel like having said that the developer still has the modified version we need to suggest a solution that doesn't involve throwing the changes away and having to do the whole process (which may having involved end-user input) all over again.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: So is there no difference for Non-persistable entities between custom with rollback and custom without rollback. They will both behave as custom without rollback? Perhaps I could say that.

```suggestion
* If an error occurs in a microflow, and you have custom error handling, it will always act as a custom without rollback as persistable entities do not have a separate state in the database.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: I wonder if this, although true, is too much detail. I think "it always" is enough?

MarkvanMents: Good question - I don't actually know what it does but it has some sort of consistency check on all the associated objects. Not sure what else to say here.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
```diff
@@ -171,8 +190,33 @@ Consider the following best practices for error handling:
...
+
+{{% alert color="info" %}}
+This applies to data held within your app and its database. Calls to external systems will not be reversed if a transaction errors. You have to handle these cases specifically.
+{{% /alert %}}
+
+### Isolated Changes
+
+While updating or creating your objects, you do not want other end-users to see updated information until your microflow has finished executing. 
+
+To ensure that every end-user or process can only see persisted data, data changed in a microflow is only available within that microflow. None of the changes made inside the microflow will be available outside the microflow, not even to the end-user that initiated the microflow. The information will only be available to all end-users of the application once the microflow has successfully completed all the activities.
```
MarkvanMents: To address https://github.com/mendix/docs/pull/10899/changes#r3063102012 and try to explain how information is isolated:

```suggestion
To ensure that every end-user or process can only see persisted data, data changed and committed in a microflow is not made available to all end-users of the application until the top-level microflow has successfully completed all the activities. Before then, only the end-user running the microflow has access to the data. 
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: Thanks for the investigation. I think there is too much to add to the documentation here, but I have tried to clarify line 210 of error-handling-in-microflows.md in this suggestion: https://github.com/mendix/docs/pull/10899/changes#r3074354106

MarkvanMents (commented):
@lvanengelen-mx  
Thanks for the comprehensive review.
I've made some more suggestions in response to your review. Please let me know if you have any further points.
I've now responded to everything, I think. If nothing is too misleading, I'd like to publish what we have (as it is an improvement to what is already out there). We can always come back to this at a later date (I'm sure we will).

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: I think this is fine.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: That change looks ok to me.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: I was having doubts about my earlier observation and did an experiment. The change has to be on the attribute that needs to be committed. Changing attribute A won't force a commit on changed attribute B, i.e. the changed state is tracked _per attribute_.

So I like the change that you did but maybe we should use _attributes_ instead of object.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: > So is there no difference for Non-persistable entities between custom with rollback and custom without rollback. 

I would even say for persistable entities there is no difference between the error handling types. Only because persistable entities are used to interact with the database there are some related effects.

I think your change makes sense by itself, although I have the feeling that if it would be clear in general that database rollback is a different concept to object rollback, this remark would be superfluous.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
lvanengelen-mx: Yes, if documented, this behavior should be documented as part of event handling.

lvanengelen-mx: If we do not know what it means, do we have to keep it? I tried finding out by looking at the history, but it was already there at the start of the repo it seems.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/events/error-event.md
lvanengelen-mx: I would prefer not to limit it to the actions in the microflow here, as it may very well be that more database actions will be rolled back from the calling microflow.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: ~~an~~ a microflow call

lvanengelen-mx: > The error handling only applies to activities which happened before the activity which errors.

I interpreted this to mean that error handling only applies to activities that come in the flow before the activity with error handling. I do not see how to read it so it applies to activities with 'multiple steps', like the change activity. Can we rephrase it?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: `[optimisti locking]` -> `[optimistic locking]`

I would say 'enabled' instead of 'implemented'.

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/microflows/error-handling-in-microflows.md
lvanengelen-mx: The entry for 'Continue' now mentions:
> No error is logged or shown to the end-user.

After the latest changes, no other entry references error logging or dialogs. So, shouldn't we remove it here?

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents:
```suggestion
        * The object in memory still has the updated values but the attributes were marked as unchanged after your previous commit
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents:
```suggestion
    If you want to keep the changes in the version which is in memory, you will have to work around this behavior by changing the attributes of the object again.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents: OK - we won't discuss errors here at all.

```suggestion
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
MarkvanMents:
```suggestion
When an object in memory is committed through a default **Save** button, a commit activity, or web services, it always triggers the commit events. To guarantee data consistency, the platform may also autocommit associated objects.
```

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/events/error-event.md
MarkvanMents:
```suggestion
When you use this event, it creates a new error with the same information as the original error. Because this is a new error, even if the error that occurred earlier was caught **without rollback**, all database actions within the current microflow will be rolled back. In addition, other database changes might be rolled back depending on how the error is handled in the calling microflow. For more information, see [Error Handling in Microflows](/refguide/error-handling-in-microflows/).
```

MarkvanMents (commented):
Discussion with Leroy

content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/object-activities/committing-objects.md
```diff
@@ -103,21 +103,53 @@ When inside a [nanoflow](/refguide/nanoflows/), the object is refreshed across t
...
+
+    * If the object has already been committed, the delete will remove the object from the database, and the subsequent commit will have no effect.
+    * If the object in memory is new (that is, it has not been committed before you delete it), the delete will do nothing. However, the subsequent commit will write the object to the database. Therefore, this sequence of actions (a delete followed by a commit) may lead to unexpected results if the object has not been committed before.
+
+#### Committing Non-Persistable Entities
+
+Mendix mimics this behavior for non-persistable entities. This means that:
+
+* Performing a commit on a non-persistable entity means that you cannot use a **Rollback object** activity to go back to the previous values
+* If an error occurs in a microflow, and you have error handling which does a rollback, the values roll back to the state a the previous savepoint.
```
MarkvanMents: Rewrite to remove mention of errors.
```suggestion
Mendix mimics this behavior for non-persistable entities. This means that performing a commit on a non-persistable entity means that you cannot use a **Rollback object** activity to go back to the previous values.
```

MarkvanMents:
Agreed to publish after discussion with Leroy.
Can revisit if we get further feedback.
Thanks everyone for your input.

PR #10914, opened 2026-03-20
[UIA-1455] Add global inbox documentation

semiramaria:
The PR is ready.
The modules are not released yet. So, the links for the marketplace are not available. We should update them when we have the links


Yiyun333 (approved):

Yiyun333:
Only links still need to added before the release.

PR #10935, opened 2026-03-26
Portable App Distribution

katarzyna-koltun-mx:


content/en/docs/deployment/general/pad/pad-best-practices.md
```diff
@@ -0,0 +1,45 @@
...
+    This allows you to maintain a single entry point for configuration while still controlling the order of included files.
+
+## The `etc/Default` File
+
+The `etc/Default` file is primarily provided for convenience and serves as an example. It is not mandatory to use it as-is.
+
+The system is designed to support multiple configuration files, allowing you to pick and choose which ones you want to use. This is particularly beneficial in Cloud environments where specific configurations (for example, `StudioPro.conf` for local testing) might be replaced with custom, deployment-specific files.
+
+In a typical Cloud deployment, it is often recommended to replace files like `StudioPro.conf` (which is geared towards local testing) with your own custom configuration tailored for the Cloud environment.
+
```
GulamY, line 54: We can also add that the same setting can be specified in multiple configuration files. 

The configurations are loaded in this order

1. include file("etc/StudioPro.conf")
2. include file("etc/constants/defaults.conf")
3. include file("etc/configurations/Default.conf")
4. include file("etc/variables.conf")
5. include file("etc/constants/variables.conf")

If a the same setting is present let's say in StudioPro.conf and etc/variables.conf the latter will take precedence

GulamY (commented):
Looks good. Maybe just one clarification might be needed. Otherwise this looks comprehensive

content/en/docs/deployment/general/pad/pad-best-practices.md
katarzyna-koltun-mx: Done.

PR #10937, opened 2026-03-27
Add alert about the Windows does not have the executability bit. +x on Linux 

mruiserrmendix:


MarkvanMents (changes requested):
@mruiserrmendix 
Hi Mario,
The description is for a very small change, but there are 15 commits in this PR.
Have you branched off the wrong branch here?
We can either merge into the original branch, or do you want to create a new PR with just this alert in that we can merge independently?

mruiserrmendix:
Hi Mark,

Uhhh, I probably branched you on the wrong branch. yes.
If you are able to merge into the original branch, I would appreciate it, as it will help avoid another mistake. :) 
Thank you so much.

Mario

MarkvanMents (approved):
I've moved this to merge with the appropriate branch.

PR #10943, opened 2026-03-30
Clarify that authentication types other than basic are possible

joostverhoog:


quinntracy:
@joostverhoog I'm guessing this is just a small revision--can it be merged?

joostverhoog:
> @joostverhoog I'm guessing this is just a small revision--can it be merged?

Yes, it's simple and can be merged. It's based on feedback that it looks like we only support basic authentication. We support basically anything, but only basic authentication is built in.

PR #10954, opened 2026-03-31
[WTF] Include Layout Grid Row Extraction in Docs

LEGIO-SEXTA-FERRATA:
The feature is to be shipped with 11.9.

### DIFF
--- a/content/en/docs/refguide/modeling/pages/structure-widgets/layout-grid.md
+++ b/content/en/docs/refguide/modeling/pages/structure-widgets/layout-grid.md
@@ -179,7 +179,7 @@ A new column is added, the weight 1 is automatically set for it.
 Sometimes on a native page, data is not clearly visible when using a native layout grid. To see the data more clearly, nest the layout grid inside the scroll container.
 {{% /alert %}}
 
-### Performing Other Actions on Rows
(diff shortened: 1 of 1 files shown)


content/en/docs/refguide/modeling/pages/structure-widgets/layout-grid.md
```diff
@@ -185,6 +185,7 @@ In adding to inserting a new row, you can perform the following actions when rig
 
 * **Move up** – moves a row up in the layout grid, you can use a shortcut for it  <kbd>Ctrl</kbd> + <kbd>↑</kbd> 
 * **Move down** – moves a row down in the layout grid, you can use a shortcut for it  <kbd>Ctrl</kbd> + <kbd>↓</kbd> 
+* **Extract to separate layout grid** – extracts the right-clicked row into a new, separate layout grid
```
ancajuravle: @LEGIO-SEXTA-FERRATA Should we call it as we do in Studio Pro: **Extract to new grid**? And since in the end it wasn't added after the move actions, not sure if we should keep it also in the docs in sync or leave it as you proposed? Also don't forget to drop the 10 and 9 back-ports.

Image

ancajuravle (approved):

ConnorLand (approved):

PR #10957, opened 2026-03-31
Protect shared Claude settings

dbreseman:


MarkvanMents (approved):
@dbreseman 
Thanks for this, Dana. Looks great and very clear.
I'm guessing the GitHub Action works as it is flagging the change to CLAUDE.md 🙂
Approving.

PR #10962, opened 2026-04-01
Remove experimental status of OQL statements

basdebakker:
For Mendix 11.10.


MarkvanMents:
This PR is replaced with #10974 as OQL Statements was made GA in 11.8 but documentation was not updated then.
Closing this in favor of #10974

PR #10997, opened 2026-04-09
Forgot Password: disable anonymous users

Karuna-Mendix:
6.5.0

### DIFF
--- a/content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
+++ b/content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
@@ -142,9 +142,9 @@ You may have to configure an email alias on your SMTP server if you are using a 
 
 ### Disabling Anonymous Users
 
-Starting from version X.X.X of the module, you can disable anonymous users. Additionally, check whether `Anonymous` user role can be removed from the app. If your application does not require anonymous access, removing the role is recommended as a security best practice.
+Starting from version 6.5.0 of the module, you can disable anonymous users. Additionally, check whether `Anonymous` user role can be removed from the app. If your application does not require anonymous access, removing the role is recommended as a security best practice.
 
-To disable anonymous users, set the `EnableAnonymousUserRole` constant to *false* in the module. By default this value is set to *true*. Once the constant value is set to *false*, end users can be authenticated via the `https://<your-app-url>/forgotpassword/v1/login` URL. When accessing the URL, users are automatically redirected to a single page where they can sign in, sign up, or reset their password.
+To disable anonymous users, set the `EnableAnonymousUserRole` constant to *False* in the module. By default this value is set to *True*. Once the constant value is set to *False*, end users can be authenticated via the `https://<your-app-url>/forgotpassword/v1/login` URL. When accessing the URL, users are automatically redirected to a login page where they can sign in, sign up, or reset their password.
 
 {{% alert color="info" %}}
 You can use either a Mendix page or an endpoint to access the login page, but not both at the same time.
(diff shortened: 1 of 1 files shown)


content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
```diff
@@ -140,6 +140,30 @@ Consider a scenario, where your SMTP username is *user@example.com* and you have
 You may have to configure an email alias on your SMTP server if you are using a different **fromAddress** in your email template than the email address of your selected SMTP account. Some SMTP servers will not send emails if the **fromAddress** is not associated with the SMTP account.
 {{% /alert %}}
 
+### Disabling Anonymous Users
+
+Starting from version 6.5.0 of the module, you can disable anonymous users. Additionally, check whether `Anonymous` user role can be removed from the app. If your application does not require anonymous access, removing the role is recommended as a security best practice.
```
JaapF: "Starting from [...], you don't need anonymous userrole in your application to have your end-users reset their password. If you migrate your app from a previous version of this module, we advise you to check whether [...]".

"[...] best practice, since usage of an anonymous userrole introduces risk of access to entities and microflows without protection from the Mendix runtime security model."

content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
```diff
@@ -140,6 +140,30 @@ Consider a scenario, where your SMTP username is *user@example.com* and you have
...
+
+To disable anonymous users, set the `EnableAnonymousUserRole` constant to *False* in the module. By default this value is set to *True*. Once the constant value is set to *False*, end users can access the login flow via the `https://<your-app-url>/forgotpassword/v1/login` URL. When accessing the URL, users are automatically redirected to a login page where they can sign in, sign up, or reset their password.
```
JaapF, line 149: I think we had agreed to put the default to FALSE.
It breaks backward comptability but enforces customers to make a deliberate decision.

content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
```diff
@@ -140,6 +140,30 @@ Consider a scenario, where your SMTP username is *user@example.com* and you have
...
+
+{{% alert color="info" %}}
+You can use either a Mendix page or an endpoint to access the login page, but not both at the same time.
+{{% /alert %}}
+
+#### Translating Non-Mendix Pages of the Module
```
JaapF: We call it "non-Mendix pages" in informal conversations.
I thunk it's better to call them "html pages" and clarify that access to these pages is not guarded by the Mendix runtime security model.

content/en/docs/marketplace/platform-supported-content/modules/forgot-password.md
```diff
@@ -140,6 +140,30 @@ Consider a scenario, where your SMTP username is *user@example.com* and you have
 You may have to configure an email alias on your SMTP server if you are using a different **fromAddress** in your email template than the email address of your selected SMTP account. Some SMTP servers will not send emails if the **fromAddress** is not associated with the SMTP account.
 {{% /alert %}}
 
+### Disabling Anonymous Users
+
+Starting from version 6.5.0 of the module, you can disable anonymous users. Additionally, check whether `Anonymous` user role can be removed from the app. If your application does not require anonymous access, removing the role is recommended as a security best practice.
+
```
JaapF, line 148: I think we need a bit of an introduction somewhere. In short it would say we provide 2 options for password reset pages:
(1) using Mendix-pages which requires anonymous userrole. This is old behavior, i.e. older versions
(2) using html pages which does not require anonymous userrole. This is recommended and is the new default. These pages are provided with the module. Texts on these pages can be translated.

PR #11030, opened 2026-04-16
Update skill names and description

MarkvanMents:
Skill names were clashing with inbuilt Clause skills.

Renaming to make it easier to find. and to avoid name clashes in future.

Also clarifying whether skills work on a single document or on a whole PR.


dbreseman (approved):

MarkvanMents:
Discussed with @dbreseman - merging to development.

PR #11040, opened 2026-04-20
Add Mapping Model API release notes

paulmendix:
Add release notes for changes added to the Model API for Mapping documents.

Includes some automatic formatting updates suggested by Prettier that I have installed in VS Code (like using - over * as bullet point marker),  this should not affect the Markdown output.

Contact: Paul Rijneveld


quinntracy:
Merged into [PR #11086 ](https://github.com/mendix/docs/pull/11086) to combine 11.10 release notes.

PR #11042, opened 2026-04-20
Appcircle docs

furkanarabaci:


ConnorLand:
@furkanarabaci this isn't tied to a release, right?

furkanarabaci:
> @furkanarabaci this isn't tied to a release, right?

No, this is an independent disclaimer, doesn't need to be in a specific release.

PR #11048, opened 2026-04-20
Interrupting Notification Event Sub-Process

DmytroKost:
For 11.10


content/en/docs/refguide/modeling/application-logic/workflows/event-sub-processes.md
```diff
@@ -53,16 +53,25 @@ Event sub-processes are triggered by a [Notify workflow](/refguide/notify-workfl
...
+
+After you confirm the change:
+
+* The sub-process is re-created with the start event of the specified type together with all the event sub-process activities. The new start event can be triggered after the workflow is redeployed and becomes in progress.
+* The workflow will become incompatible if the changed event sub-process is already being executed in one of the ongoing workflow instances.
+
+The event sub-process is re-created upon type switch because in-place conversion would produce states that contradict BPMN 2.0 concepts. Per BPMN, an interrupting event sub-process cancels the parent process scope and all other active sub-processes the moment it is triggered, while a non-interrupting one runs in parallel without affecting them. These are mutually exclusive execution models: an event sub-process instance belongs to exactly one of them from the moment it starts. Changing the type in place on an already-active instance would leave it in a state that is neither valid interrupting nor valid non-interrupting behavior, violating the fundamental BPMN distinction between the two.
```
Remcovh: What do we mean by type here? I would consider type, timer, message, etc. And this, if I read correctly, is about switching interrupting behaviour.

Remcovh: Maybe we can enrich this part with an example

> Changing the type in place on an already-active instance would leave it in a state that is neither valid interrupting nor valid non-interrupting behavior, violating the fundamental BPMN distinction between the two.

Changing the type in place on an already-active instance is not possible because switching to interrupting, for example, would require retroactively canceling the parent scope and any parallel sub-processes that may have already completed or progressed.

DmytroKost: I understand the concern about retroactive patching, but the limitation is conceptual, not technical. In BPMN 2.0, interrupting and non-interrupting behaviors are mutually exclusive lifecycles, not toggleable settings. Changing the type mid-flight would create an illegal state that violates the spec, and since we **never** retroactively patch history or undo completed work, the sub-process must be re-created to ensure a valid execution model moving forward.

Yiyun333: @Remcovh "Type" is a pretty general term, but in this context it’s clear we’re talking about switching between interrupting and non-interrupting, so I think it works fine here 🙂

Yiyun333 (approved):

PR #11077, opened 2026-04-23
Add alt text skill

dbreseman:


.claude/skills/docs-alt-text/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+   - Use Read tool to view the actual image
+   - Understand what the image shows BEFORE reading context
+
+2. **STEP 2 - Read surrounding context**
+   - Read the heading, preceding/following text, list item, or numbered step
+   - Understand the image's purpose within the documentation
+   - Consider if context + image together make the image informative or decorative
+
+3. **STEP 3 - Apply W3C decision tree**
+   - **Technical docs assumption:** Images are informative unless obviously decorative
```
MarkvanMents, line 26: My assumption is always the other way around. I believe that you should understand the documentation if the images were removed.

But I know everyone else thinks the opposite way, as described here, so I won't start the argument all over again. 😁

.claude/skills/docs-alt-text/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+   - Categorize as: UI screenshot, procedural step, diagram, code example, graph, or configuration screen
+   - Only use `alt=""` if certain the image is pure decoration (borders, spacers, decorative backgrounds)
```
MarkvanMents: But I do think


```suggestion
   - Only use `alt=""` if certain the image is pure decoration (borders, spacers, decorative backgrounds, duplication of surrounding textual information)
```

MarkvanMents: Should this be in STEP 4 where the alt text is being generated. It shouldn't impact the categorization?

.claude/skills/docs-alt-text/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+   - Understand what the image shows BEFORE reading context
+
+2. **STEP 2 - Read surrounding context**
+   - Read the heading, preceding/following text, list item, or numbered step
+   - Understand the image's purpose within the documentation
+   - Consider if context + image together make the image informative or decorative
+
+3. **STEP 3 - Apply W3C decision tree**
+   - **Technical docs assumption:** Images are informative unless obviously decorative
+   - Categorize as: UI screenshot, procedural step, diagram, code example, graph, or configuration screen
```
MarkvanMents:
```suggestion
   - Categorize as: UI screenshot, procedural step, diagram, code example, graph, decorative, or configuration screen
```

.claude/skills/docs-alt-text/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+
+## W3C Guidelines Summary
+
+**Informative images:**
+- Convey meaning or information → provide descriptive alt text
+- Be "the most concise description possible"
+- Focus on what information it communicates, not what it looks like
+
+**Decorative images (rare in technical docs):**
+- Pure visual styling with no informational value → use `alt=""`
```
MarkvanMents: Not what W3C says. It actually says
[For example, the information provided by the image might already be given using adjacent text, ](https://www.w3.org/WAI/tutorials/images/decorative/) which is why I don't think they are as rare as you think.

What happens if you tell Claude this instead of "pure visual styling"?

.claude/skills/docs-alt-text/SKILL.md
dbreseman: Understood! I hear this may be an ongoing conversation, so we can definitely revisit.

dbreseman: Understood! I hear this may be an ongoing conversation, so we can definitely revisit.

.claude/skills/docs-alt-text/SKILL.md
dbreseman: Agreed—I shuffled some things around to make the flow more organized.

.claude/skills/docs-alt-text/SKILL.md
dbreseman: Updated (and simplified)

.claude/skills/docs-alt-text/SKILL.md
dbreseman: Thanks for catching that. Updated now.

.claude/skills/docs-alt-text/SKILL.md
MarkvanMents: I still feel we can miss this guidance off and let Claude decide based on the information we have provided. But I'm happy to try this in a separate experiment later.

MarkvanMents (approved):
Looks good.
I still would like to know what Claude actually does if we just give it facts and no guidance as to what is the assumption, it always wants to please so I think this will lead to everything getting an alt-text.
But I'll do my own tests later and report back. I think we should go with this for now, and work Agile.
Thanks for doing this. I think the name of the new skill is clear enough that it can't be confused with any of the other skills. I think it will be very useful.

PR #11157, opened 2026-05-07
We do support SourceDatabaseJdbcUrl on PostgreSQL

basdebakker:
A tiny fix to remove a long obsolete statement.


MarkvanMents (approved):
Looks good. Thanks Bas.

PR #11165, opened 2026-05-08
Landing page hero section new design

OlufunkeMoronfolu:


layouts/landingpage/landingpage.html
```diff
@@ -17,23 +17,24 @@
...
-                                    </div>
-                                    <img src="/images/woman-stack-of-books-laptop.svg" alt="" class="d-none d-lg-block ms-4" style="margin-top: -.5rem;">
+                           <div class="lp-hero d-flex align-items-center text-white">
+    
+                                <div class="hero-text">
+                                    <h1 class="page-title">{{ .Page.Title }}</h1>
+                                    <h5 class="lp-description">{{ .Page.Description }}</h5>
+                                </div>
+
+                                <div class="hero-image">
```
dbreseman:
```suggestion
                                <div class="hero-image d-none d-lg-block">
```

assets/scss/_front-page.scss
```diff
@@ -2,31 +2,56 @@
...
+    padding: 2rem 2.5rem;
+
+    @media (max-width: 767px) {
+        min-height: unset;
+    }
+}
+
+.hero-text  {
+    flex: 0 0 60%;
+}
```
dbreseman, line 41:
```suggestion
.hero-text  {
    flex: 0 0 100%;
    // Stretch out text once image disappears (smaller viewports)
    @include media-breakpoint-up(lg) {
        flex: 0 0 60%;
    }
}
```

content/en/docs/_index.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: "Welcome to Mendix Docs"
+title: "Mendix Docs"
 linktitle: "Docs"
 url: /
 type: landingpage
 layout: landingpage
-description: Learn how to create, deploy, and manage your app landscape. 
+description: Browse the documentation for Mendix products, read about latest releases, and make contributions for the Mendix community. Learn how to create, deploy, and manage your app landscape.
```
dbreseman: We should workshop these front matter changes with the group. I like the longer description but I wonder if we should change it more to highlight AI. I personally tend prefer keeping "Welcome to" in the title, since it feels friendly and aligns better with the other Portal pages like https://marketplace.mendix.com/link/components and https://community.mendix.com/link/home.

assets/scss/_front-page.scss
```diff
@@ -99,3 +124,19 @@ p.rn-date {
     font-weight: 400;
     margin-bottom: 0rem;
 }
+
+// Mobile hamburger bar on landing page — matches hero background
+.lp-mobile-bar {
+    background-color: #0A1324;
```
dbreseman: We should double-check this with Amit. It's very similar to https://github.com/mendix/docs/blob/4cbc0f9c6f9ac48e01decf33f0b38898598e4a85/assets/scss/_variables_project.scss#L14 (from https://designsystem.mendix.com/doc/p/Utilities/Colors-and-Backgrounds/d49d94a9-4f75-463e-bf0b-f7151d090a39), so I wonder if this hex code is correct or if we can just use $secondary.

assets/scss/_front-page.scss
```diff
@@ -2,31 +2,56 @@
...
     margin-bottom: 2rem !important;
 }
 
+.lp-hero {
+    background-color: #0A1324;
+    min-height: 400px;
+    margin-left: -2.5rem;
+    margin-right: -2.5rem;
+    padding-left: 2.5rem;
+    padding-right: 2.5rem;
```
dbreseman:
```suggestion
```

dbreseman: I suggest removing these because they conflict with line 29

assets/scss/_front-page.scss
```diff
@@ -2,31 +2,56 @@
...
+    padding: 2rem 2.5rem;
+
+    @media (max-width: 767px) {
```
dbreseman:
```suggestion
    @include media-breakpoint-down(md) {
```

dbreseman: Adjusted to use Bootstrap's mixin (and match when the image disappears)

assets/scss/_front-page.scss
```diff
@@ -2,31 +2,56 @@
...
+    max-width: none;
 }
 
 .lp-row{
     margin-bottom: 2rem !important;
 }
 
+.lp-hero {
+    background-color: #0A1324;
+    min-height: 400px;
```
dbreseman: Do you think we should push back on this spec, or do you like it as is? I think some of the Portal pages (like https://marketplace.mendix.com/link/components) have shorter hero heights, and it might be nice if it weren't quite so big so that more of the story cards fit on smaller screens by default.

dbreseman:
Hey @OlufunkeMoronfolu, this looks really good! Thanks for putting it together. I made a few comments and suggestions—the biggest one is about getting the image to disappear on smaller viewports. Do you want to share this on Slack or at a group meeting so we can see if the Docs team has feedback on the new design?

assets/scss/_front-page.scss
OlufunkeMoronfolu: There appears to be a mismatch between the color we were given and what PDS is using now. I also checked the community and marketplace backgrounds, and they’re all still referencing `#0A1324`. Might be something we want to confirm with design.

I also saw the PDS release updated the BG, so the ...24 might be obsolete. But let's confirm that before we substitute the variable.

assets/scss/_front-page.scss
OlufunkeMoronfolu: Yes, let's try that. At least make a case for showing more of the cards on screen (like we currently do), or showing the leading icons for the second cards if we stay within the 300–350px range.

MarkvanMents: 400px looks OK to me, but it could be a bit less? Not less than 330px though. Control Center has a much larger one (686), Compass 435, but I think 400px would be OK if they insist.

content/en/docs/_index.md
MarkvanMents: I think this is too much. I don't think anyone will read it. I liked the previous length, but we could rethink what it actually says - for example "Learn how to" should perhaps be left for Academy, and we might want to have something about AI?.

I like "Welcome to Mendix Docs" and Marketplace has "Explore Mendix Marketplace" so it would be consistent to have additional words.

I also don't think we need to say "Browse the documentation", there are multiple ways of accessing it. I like highlighting that you can contribute, but not sure this is the best place for it. Perhaps we could have link to the contribution instructions in the **Collaborate** section lower down instead?

assets/scss/_front-page.scss
OlufunkeMoronfolu: This is now verified, and we will be using #0A1324, moving forward.

content/en/docs/_index.md
OlufunkeMoronfolu: We've reverted to the old byline for now and will revisit this in about a month, once we have a clearer direction for the upcoming agentic enterprise changes.

assets/scss/_front-page.scss
OlufunkeMoronfolu: We have now used 350px to fit more of the story cards on the screen.

dbreseman (approved):

PR #11176, opened 2026-05-11
Remove Standalone option

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Not proceeding at this time, may reopen later.

PR #11190, opened 2026-05-12
Add documentation for the LPAD and RPAD OQL functions

passalaqua:
~Should be added in 11.11.0~

Changed to 11.12.0 in a196c91a4d95b04b39d1723f291996c8a9dbff5f


katarzyna-koltun-mx:
Awaiting 11.11 release.

passalaqua:
> Awaiting 11.11 release.

Just FYI, this still requires approval from the team and for the implementation story to be finalised.

katarzyna-koltun-mx:
Got it, will wait for the go ahead.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+LPAD ( expression , length_expression [, pad_expression ] )
+```
+
+#### expression
+
+`expression` specifies the expression of type `string` to pad.
+
+#### length_expression
+
+`length_expression` specifies, as an `integer`, the length of the resulting string.
```
pijuskri: I would propose a different wording:

> `length_expression` specifies the length of the resulting string. The expression must be of type `integer` or `long`

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+
+{{% alert color="info" %}}
+If `length_expression` is smaller than the length of `expression`, this function truncates it. This behavior is database specific.
+{{% /alert %}}
+
+#### pad_expression
+
+`pad_expression` is an optional parameter that specifies the character or string to pad with. If not specified, the space character is used.
```
pijuskri, line 1370: Let's visualize the character:

> the space character (" ") is used

pijuskri (changes requested):

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: I went with this description to be consistent with the documentation for TRIM and, imo, adding such a visualization is always a bit confusing. If I were to change here, I think it's better to change in the rest of the page as well.

pijuskri: I don't find it to be a big deal either way, so lets keep it as is. Other parts of the documentation that refers to spaces also dont show `(" ")`

pijuskri:
Approved by team for technical review

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+
+#### Syntax
+
+```sql
+LPAD ( expression , length_expression [, pad_expression ] )
+```
+
+#### expression
+
+`expression` specifies the expression of type `string` to pad.
```
MarkvanMents, line 1355: Any special behavior for `NULL`? For Trim it returns `NULL` but does this just return a whole load of `pad_expression`?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+
+#### length_expression
+
+`length_expression` specifies the length of the resulting string. The expression must be of type `integer` or `long`
+
+{{% alert color="info" %}}
+If `length_expression` is smaller than the length of `expression`, this function truncates it. This behavior is database specific.
```
MarkvanMents, line 1365: Truncates the right side, even though we are doing `LPAD`?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+
+
+```sql
+SELECT LPAD('hello', 10, 'x') AS padded FROM Sales.Order
+```
+
+| padded     |
+|:-----------|
+| xxxxxhello |
+
```
MarkvanMents, line 1393: Is it worth an example with a string `pad_expression`?

```sql
SELECT LPAD('hello', 10, 'abc') AS padded FROM Sales.Order
```

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1507,6 +1560,57 @@ SELECT ROUND((Price : 7), 2) as RoundedPrice, Price : 7 FROM Sales.Order
...
+#### expression
+
+`expression` specifies the expression of type `string` to pad.
+
+#### length_expression
+
+`length_expression` specifies the length of the resulting string. The expression must be of type `integer` or `long`
+
+{{% alert color="info" %}}
+If `length_expression` is smaller than the length of `expression`, this function truncates it.
```
MarkvanMents, line 1649: Still normal truncation - so the same as LPAD?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1507,6 +1560,57 @@ SELECT ROUND((Price : 7), 2) as RoundedPrice, Price : 7 FROM Sales.Order
 | 0.33         | 3.33333333 |
 | 1.17         | 1.17142857 |
 
+### RPAD {#rpad-function}
```
MarkvanMents, line 1621: Same comments as for LPAD

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
```diff
@@ -1329,6 +1331,57 @@ SELECT * FROM Sales.Customer WHERE LOWER(LastName) = 'doe'
...
+
+#### Examples
+
+```sql
+SELECT LPAD('hello', 10) AS padded FROM Sales.Order
+```
+
+| padded     |
+|:-----------|
+|      hello |
```
MarkvanMents: This looks wrong when built. May have to think around this (`&nbsp;` would work - but still difficult to see that it is space padded rather than right justified)

Image

)

MarkvanMents (commented):
Thanks for the PR. I have a few questions:

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: They just return `null` if either the expression or length are null. If the padding pattern is `null`, it is usually null (HSQLDB is the exception and throws an error).

And if `expression` is the empty string or the length is zero, then Oracle returns null, but other databases return the empty string.

And if the padding pattern is the empty string, then Oracle and MariaDB return null, HSQLDB uses the space character, MySQL returns the empty string, and other databases return the original string. It's kind of a mess and I suppose those are cases where things can be described as "it's database dependent", so I didn't add to this page.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: Weirdly, yes. We were just as confused when implementing. The only exception is HSQLDB, which does the right thing, but possibly by accident given how special it is.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: Probably a good idea. I can write it as part of the same query (shorter page) or as a separate example (clearer). Which one do you prefer?

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: Yep.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: I suppose we can change the query to

`SELECT '|' + LPAD('hello', 10) + '|' AS padded FROM Sales.Order`

The resulting string would be `|     hello|`. Would we render it correctly with spaces?

passalaqua: We could also just render the result as

```
·····hello
```

Or something similar and indicate `·` represents a whitespace.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Is it worth noting that this edge case behavior depends on the database and that the customer should test them on their configuration? I'm not sure that all our users would think about this, or might raise questions about it.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Thanks for the answer. We've said that it's database specific, so let's just leave it like that, rather than trying to give an actual answer.

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Good idea. Let's do that. using the `·` character so it is clear we didn't mean `.`

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: Let's go for the clearer one.
(I need to review this page and split it up anyhow if we want it to be translatable - another few lines won't make any difference)

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
MarkvanMents: OK - should we add the note about it being database specific to encourage the user to test it?

MarkvanMents (commented):
Thanks for your answers. I've had a quick look. 
Let me know if you have any questions, or would like me to draft any changes.
Mark

content/en/docs/refguide/modeling/domain-model/oql/oql-expression-syntax.md
passalaqua: `RPAD` is consistent. It's only `LPAD` that behaves differently in `HSQLDB` and other dbs.

passalaqua:
For the record, both this and https://github.com/mendix/docs/pull/11250/ affect the same file and will need to be merged for 11.12.

MarkvanMents (approved):
Thanks for the PR. Ready for publication with 11.12 release.

PR #11220, opened 2026-05-19
Update web-extensibility-api.md

SofiBili:
Changes for the 11.11


MarkvanMents (approved):
Approving so it is ready to be published when 11.11 is released.

PR #11237, opened 2026-05-22
Update Hugo, Docsy, and Node packages

OlufunkeMoronfolu:
## Summary
The following dependency updates were made:
- Hugo Extended from `^0.126.1` to `^0.156.0`
- Docsy from `0.10.0` to `0.15.0`
- Node.js CI version (.travis.yml): `20.9` to `24.15`

### CSS fixes (Docsy 0.15 compatibility)

- Overrides Docsy 0.15's new rule that makes all top-level TOC entries bold by default.
- Overrides Docsy 0.15's new hover background colour on TOC links so hover states remain as is.
- Fixes font-weight variable reference: `$font-weight-body-text` → `$td-font-weight-body-text` (renamed in Docsy 0.15).

### Config fix
- Updated [module] replacements in `hugo.toml` to reflect the correct module paths for Font Awesome and Bootstrap under the new Hugo/Docsy versions.

### Layout fixes (Docsy 0.15 compatibility)
- Added `layouts/docs/single.html` to route docs single pages to `_default/content.html` via `.Render "content"`, instead of Docsy's new `_td-content.html` which renders the page description field as visible body text. This preserves the existing behaviour where description is used for metadata only (search snippets and link previews).

### Other fixes
- Duplicate metadata causing build  errors
- Fixed feedback partial to now use the full site params context (`.Site.Params.ui.feedback`) rather than a scoped context (`.`). In Docsy 0.15, the partial is invoked differently, so `{{ .yes }}` no longer resolves.

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -88,7 +88,7 @@
     max-width: 90%;  // NK - fixed width of all content to 90%
 
     p, li, td {
-        font-weight: $font-weight-body-text;
+        font-weight: $td-font-weight-body-text;
     }
 
     // NK - custom additions for margin styling in lists
@@ -339,6 +339,11 @@ img {
         // Hugo's ToC is a mouthful, this can be used to style the top level h2 entries.
         > ul > li > ul > li > a {}
 
+        // Override Docsy 0.15+ which makes top-level (h2) items bold by default
+        > ul > li > :not(ul) {
+            font-weight: $font-weight-normal;
+        }
+
         a {
             // NK - styling for ToC
             color: $gray-800;
@@ -346,9 +351,11 @@ img {
             padding-bottom: .25rem;
             padding-top: .25rem;
 
-            &:hover {
+            &:hover,
+            &:focus {
                 color: $link-hover-color;
                 text-decoration: underline;
+                background-color: transparent; // Override Docsy 0.15+ hover background
             }
         }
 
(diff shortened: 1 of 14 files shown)


assets/scss/_styles_project.scss
```diff
@@ -339,16 +339,23 @@ img {
         // Hugo's ToC is a mouthful, this can be used to style the top level h2 entries.
         > ul > li > ul > li > a {}
 
+        // Override Docsy 0.15+ which makes top-level (h2) items bold by default
+        > ul > li > :not(ul) {
+            font-weight: $font-weight-normal;
+        }
```
MarkvanMents: Took me time to work this one out - perhaps we should mention that this is related to the ToC, not the actual h2 entry?

MarkvanMents (approved):
@OlufunkeMoronfolu 
Looks great. Have made one comment but it is only very minor and only comes from my own confusion looking at just the code changes.
I've run it (once I upgraded Node) and the site looks great.

I think that before we release it we should make a final GitHub "Release version" with the old Hugo/Docsy. Then create a new (major?) release version with the new updates so it is easy to see which version people are using and we can advise people to upgrade clearly.

Did Docsy ever implement their own dynamic ToC? If so, we should investigate whether it works as well as our one - but that can come later.

I'm happy for this to go - but let's wait until after 11.11 goes out as everyone will need time to do a new `npm install`.

Also think about a post in Slack/Teams so that dev's are aware if they are doing local builds.

Let me know if you want to discuss.

PR #11240, opened 2026-05-22
Fix: hidden feedback widget and unresponsive button click

OlufunkeMoronfolu:
In addition to fixing the hidden feedback widget on index pages, I also found the following problems:
- Clicking the buttons has no effect
- There was a console error `Uncaught TypeError: Cannot read properties of null (reading 'href')` in the `feedback.html` file.

All errors have now been fixed and should now work properly.


MarkvanMents (approved):
Thanks @OlufunkeMoronfolu - looks good.

PR #11244, opened 2026-05-26
Remove Beta flag for Optimistic Locking

akkrishnakumar:
RUN-5172

### DIFF
--- a/content/en/docs/refguide/runtime/optimistic-locking.md
+++ b/content/en/docs/refguide/runtime/optimistic-locking.md
@@ -8,7 +8,8 @@ beta: true
 ## Introduction
 
 {{% alert color="info" %}}
-This feature is available in Beta starting with Mendix version 11.5.0.
+This feature is Generally Available starting with Mendix version 11.11.0.
+This feature was available as Beta since Mendix version 11.5.0.
 {{% /alert %}}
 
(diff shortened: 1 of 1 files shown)


akkrishnakumar:
Waiting for review from Runtime Core team

emirmx:
> Waiting for review from Runtime Core team

Looks good

content/en/docs/refguide/runtime/optimistic-locking.md
```diff
@@ -7,10 +7,6 @@ beta: true
 
 ## Introduction
```
MarkvanMents, line 8: @akkrishnakumar 

We still need to say that this was introduced in 11.5.0, for customers who are using earlier 11.x versions.
When was it made GA? I can't see anything in the release notes. We need to add something there since 11.5 says it is in public beta.

akkrishnakumar: I understand. 

Which means in 12, we can remove this ?


I found this in the release notes of 11.5

> Optimistic Locking
We made optimistic locking publicly available as a beta feature. This allows apps to safeguard against concurrent modifications. It can be enabled in the Runtime tab of the App Settings dialog box. Enabling this feature gives each entity an extra system attribute (MxObjectVersion of type Long). Its behavior will be automatically handled by the system. For more information, see [Optimistic Locking](https://docs.mendix.com/refguide/optimistic-locking/).  



I'll add the details

MarkvanMents: Yes, in Mx12 we only have to say what the status is in 12, so it will be GA from 12.0.0 so we don't have to say anything.

MarkvanMents (approved):
Thanks for the PR, Akhil.
Looks good - I've also removed the "BETA" lozenge which was displayed in the navigation pane.
Approving and merging, since 11.11 is already released.

PR #11247, opened 2026-05-26
Add auto-assign-pr action

dbreseman:
* Auto-assigns PRs to Technical Writers team members when they create the PRs
* Only assigns if no one is already assigned; does nothing otherwise
* Only applies to Technical Writers team members; does nothing if someone else creates the PR


MarkvanMents (approved):
Haven't tested it but looks good.
Perhaps mention it at the group meeting and then implement it.
We can always remove it if it gets annoying.
Thanks for picking this up.

PR #11258, opened 2026-05-27
Add Vale

dbreseman:


.github/workflows/vale.yml
```diff
@@ -0,0 +1,35 @@
...
+      - 'content/en/docs/**/*.md'
+
+# Allow reading repo contents and posting PR comments
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  vale:
+    # Only run on Mendix repo, not forks, and not on draft PRs
```
MarkvanMents, line 16: Does this mean it won't run on developer-submitted PRs?
Would it be better to run it on all PRs that target `development`?
Perhaps we can extend it to that if it is successful?

.gitignore
```diff
@@ -9,6 +9,7 @@ hugo.exe
 /.claude/audit.log
 /.claude/settings.local.json
 /.mcp.json
+/.vale/styles/Microsoft/
```
MarkvanMents, line 12: Is this something that the Vale package creates itself?

.vale/styles/Mendix/Acronyms.yml
```diff
@@ -0,0 +1,74 @@
...
+  - GUI
+  - HTML
+  - HTTP
+  - HTTPS
+  - IDE
+  - JAR
+  - JSON
+  - JSX
+  - LESS
+  - LLDB
```
MarkvanMents, line 43: Should we have MDA and MPR here? The acronyms don't actually make much sense (even if you can remember what they are). Are there any other Mendix acronyms which don't need explaining?

.vale/styles/Mendix/Acronyms.yml
```diff
@@ -0,0 +1,74 @@
...
+  - list
+  - table
+  - blockquote
+# Ensures that the existence of 'first' implies the existence of 'second'.
+# Known issue: Doesn't recognize hyperlinks for acronym definitions
+# Modified from Microsoft.Acronyms to accept sentence-case definitions (e.g., "personal access token (PAT)" not just "Personal Access Token (PAT)")
+first: '\b([A-Z]{3,5})\b'
+second: '(?:\b[A-Za-z][a-z]+ )+\(([A-Z]{3,5})\)'
+# ... with the exception of these:
+exceptions:
```
MarkvanMents, line 18: I presume this also covers `jar` and `zip` which I tend to see in lowercase.

.vale/styles/Mendix/AmericanSpelling.yml
```diff
@@ -0,0 +1,47 @@
...
+  '\bdialogue([ds])?\b': 'dialog$1'
+
+  # Double-L patterns
+  '\bcancell(ed|ing)\b': 'cancel$1'
+  '\btravell(ed|ing|er)\b': 'travel$1'
+  '\bmodell(ed|ing)\b': 'model$1'
+
+  # Common individual words
+  '\bgrey\b': 'gray'
+  '\bwhilst\b': 'while'
```
MarkvanMents, line 39: 😒

.vale/styles/Mendix/CompoundWords.yml
```diff
@@ -0,0 +1,21 @@
...
+  - list
+  - table
+  - blockquote
+action:
+  name: replace
+swap:
+  'data-set': 'dataset'
+  'data set': 'dataset'
+  'end point': 'endpoint'
+  'check box': 'checkbox'
```
MarkvanMents, line 20: 'end user': 'end-user'?

MarkvanMents (approved):
Looks great. Can't wait to try it out locally.
I've made a couple of suggestions - but we can always add/change things later as we get used to the tool
Thanks for doing this.
Mark.

.gitignore
dbreseman: Yes! It's the Microsoft package from Vale: https://github.com/vale-cli/packages#available-styles

.vale/styles/Mendix/Acronyms.yml
dbreseman: I'll add those. I'm sure we will come up with more acronyms to add as we go along, so this will be a growing list.

.vale/styles/Mendix/Acronyms.yml
dbreseman: Yes, those won't be flagged—the regex only matches capitalized acronyms, so anything lowercase will just slip through.

.github/workflows/vale.yml
dbreseman: Good point, and you're right—this will only include internal PRs rather than all PRs that target development. I think let's plan to expand it in a week if we're happy with how it's going?

PR #11264, opened 2026-05-29
Clarify that file uploads during sync are not incremental

jdiehl:
Adds a sentence to the Upload Phase section of the Offline Synchronization page clarifying that — unlike downloads — file uploads are not incremental. The full file content is uploaded whenever the containing object has been modified in the local database, regardless of whether the file content itself changed.

This was flagged by a customer who noticed slow synchronization when attributes of an entity inheriting from System.Image were modified. Applied to refguide (SP11), refguide10 (SP10), and refguide9 (SP9).


MarkvanMents (approved):
@jdiehl 
I've moved the "unlike the..." comment to the download phase as that comes second and it is a bit strange to refer to something the customer hasn't read yet.
Also broken the text into paragraphs to make it a bit more readable, but haven't changed anything.
Approving - please let me know if you are OK with the changes and I'll publish.
Thanks for the PR.

MarkvanMents:
No response from @jdiehl - haven't changed meaning so merging anyhow.

PR #11270, opened 2026-05-29
Update multi-domain support: OIDC SSO

Karuna-Mendix:


JaapF (commented):
Looks clear and solid to me!

PR #11285, opened 2026-06-02
Add GenAI glossary page

dbreseman:


content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+A workflow pattern where specific steps are handled through agentic decision-making instead of fixed logic. The workflow keeps the overall structure and control; agentic components handle steps that require reasoning, interpretation, or adaptive action.
+
+## Embeddings {#embeddings}
+
+Mathematical representations of text as numeric vectors, where conceptually similar text strings result in similar vectors. This enables semantic search based on meaning rather than exact character matches. Embeddings are typically stored in a [vector database](#vector-database) and are a core component of [RAG](#retrieval-augmented-generation).
+
+## Generative AI (GenAI) {#generative-ai}
+
+A category of artificial intelligence technology that can generate new content such as text, images, code, or other outputs based on learned patterns from training data. Generative AI can be used to accelerate business processes, provide user-friendly interactions, and enhance apps with features such as chatbots, content generation, text analysis, and language translation.
```
linaaaf: This is all about text generation, vision capabilities (image analysis) and image generation could be alternatives (if we want)

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Hallucination {#hallucination}
+
+When an LLM generates incorrect, nonsensical, or fabricated information that is not grounded in its training data or the provided context. For example, an LLM might confidently state that a fictional person won an award that never existed, or cite nonexistent research papers. To reduce hallucinations, techniques like [RAG](#retrieval-augmented-generation) ground responses in verified data, and patterns like [human in the loop](#human-in-the-loop) add validation steps for critical decisions.
+
+## Human in the Loop {#human-in-the-loop}
+
+A design pattern where an agent pauses at defined points to request human input—such as approval, correction, or a decision—before continuing. The human actively participates in the agent’s execution rather than passively observing.
```
linaaaf: very AI-y dashes :D instead it could be '(e-g- approval, (...))'

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+In Mendix, this is implemented through user access approval settings on [tools](#tool-calling), which control when tools get executed and whether they are visible to the user. For example, a tool that deletes records might require explicit user confirmation before execution. The ConversationalUI module supports human-in-the-loop interactions out of the box. For implementation details, see [Human in the Loop](/agents/genai-for-mx/conversational-ui/#human-in-the-loop).
+
+## Knowledge Base {#knowledge-base}
+
+A storage system for discrete pieces of information that can be retrieved and used to augment prompts. Combining a knowledge base with [embeddings](#embeddings) (vector representations of the data) creates a [vector database](#vector-database) that supports semantic similarity searches. Knowledge bases are essential for [RAG](#retrieval-augmented-generation) patterns and semantic search.
```
linaaaf: This throws me a bit off: 

Yes, a knowledge base is not always a vector base. But when using the knowledge base to store embeddings vectors it is a vector base 

How about something like 

'When storing knowledge in the form of embedding vectors inside of a vector database, (...)'

Instead of 'Combining a knowledge base with (...) creates a vector database (...).'

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Retrieval Augmented Generation (RAG) {#retrieval-augmented-generation}
+
+A pattern that combines [embeddings](#embeddings)-based knowledge retrieval with text generation. RAG retrieves relevant information from a [knowledge base](#knowledge-base), augments the prompt with that information, and generates a response grounded in the retrieved data. This allows models to answer questions about private or domain-specific data. For example, when a user asks "What is our refund policy?", RAG retrieves relevant policy documents from the knowledge base, adds them to the prompt, and generates an accurate answer based on the actual policies.
```
linaaaf: Just knowledge retrieval (instead of embeddings-based (...)) as not every knowledge base is necessarily a vector base

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## System Prompt {#system-prompt}
+
+Instructions that define the model's behavior, role, and constraints. System prompts establish how the LLM should respond and what personality or expertise it should adopt. These are typically set by developers and stay the same across interactions. For example, "You are a helpful customer service assistant for an e-commerce platform. Always be polite, concise, and follow company policies when answering questions."
+
+## Token {#token}
+
+The basic unit of text that LLMs process. Models break down text into tokens, which can be words, parts of words, or individual characters depending on the model's tokenization method. For example, "agentic workflow" might be broken into tokens like `["agent", "ic", " work", "flow"]`.
```
linaaaf: This is for text input only, image input would also be tokenized but differently.

I think if your replace 'unit of text' with 'unit of data' and then frame the next part as an example for how it works for text input it is clear

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Tool Calling {#tool-calling}
+
+Also known as tool use or function calling, a capability that allows LLMs to call specific functions or tools to gather additional information or perform actions. The model returns a structured request indicating which function to call and with what parameters, which the app then executes and returns results to the model. Tool calling enables agent patterns like ReAct (Reasoning + Acting), where agents iteratively reason about what to do and then call tools to take action. For more information, see [Tool Calling](/agents/function-calling/).
```
linaaaf: The difference between tool and function is not super clear to me but every function is a tool but not every tool a function

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Trace {#trace}
+
+A trace is a structured record of execution across a system or workflow. In the context of an agent, it captures the sequence of inputs, reasoning steps, tool calls, intermediate actions, and outputs that led to a result.
```
linaaaf: the trace also captures token consumption (the cost), the duration of all and each step and whether the execution was successful

most important would be the token consumption part :)

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+title: "Glossary"
+url: /agents/glossary/
+linktitle: "Glossary"
+weight: 5
+description: "Quick reference for key agentic and generative AI terms used in Mendix documentation."
+---
+
+## Introduction
+
+This glossary provides brief definitions of key agentic and generative AI terms used in the Mendix documentation. To explore working implementations of these concepts, see the [GenAI Showcase App](https://marketplace.mendix.com/link/component/220475), which demonstrates RAG, tool calling, MCP integration, and more.
```
linaaaf: Since the term RAG is explained below only, use Retrieval-Augmented-Generation instead and link to that section? For tool calling also

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Agent {#agent}
+
+An AI system that combines an LLM with a reasoning loop and can be extended with memory, tools, skills, and the ability to call and execute functions. It can plan, evaluate results, and iterate toward a goal.
```
linaaaf: Technically we do not yet support memory and skills, at least this is not what we call it, especially skills is pretty connected to Claude Code (secretly this is also knowledge tools). Let's talk about this shortly

content/en/docs/genai/concepts/glossary.md
```diff
@@ -0,0 +1,101 @@
...
+
+In Mendix, you can build agents using Agents Kit, a collection of starter apps, connectors, and modules that support implementations from simple text generation to complex multi-step agentic workflows. For more information, see [GenAI Agents](/agents/agents/).
+
+## Agent Orchestration {#agent-orchestration}
+
+The coordination layer—often built within an app or workflow—that manages how agents work together: which agent takes responsibility, what context is shared, and how handoffs, retries, failures, and recovery are handled.
```
linaaaf, line 21: Remove AI-y dashes

content/en/docs/genai/concepts/glossary.md
dbreseman: I made pretty minimal edits here. Does this work?

content/en/docs/genai/concepts/glossary.md
dbreseman: I actually think this is clearest way to write this one—can't let AI claim all the em dashes from the humans!

content/en/docs/genai/concepts/glossary.md
dbreseman: Should I simplify to just "can be extended with tools"?

content/en/docs/genai/concepts/glossary.md
linaaaf: I would prefer 'related to' instead of 'known as' and then it works. The other edits are helpful 👍

PR #11293, opened 2026-06-03
Change Boundary Event Type

DmytroKost:
For 11.12


content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -90,6 +107,12 @@ After you confirm the change:
 
 Boundary events are re-created upon interrupting behavior change because in-place conversion can result in invalid states. An interrupting boundary event must abort its parent activity when triggered, meaning an activity cannot have more than one active interrupting boundary event. Converting an already-triggered non-interrupting boundary event to interrupting in place violates this rule: the parent activity remains in progress, resulting in an interrupting boundary event whose parent is never aborted. Conversely, converting an already-triggered interrupting boundary event to non-interrupting in place leaves it active on an already-aborted parent activity, contradicting the rule that a non-interrupting boundary event must not abort its parent.
 
+#### Implications of Changing the Boundary Event Type
```
Remcovh, line 112: Took me a bit to understand this section. What do you think about the following? We can even wrap it in an alert for more emphasis. 

**Changing a non-interrupting Boundary Event**

Due to technical limitations, changing an ongoing **non-interrupting** boundary event creates a **partially resolvable conflict** of type [Current Activity Moved out of Path](/refguide/workflow-versioning/#current-activity-moved-out-of-path) on running workflow instances, which requires manual intervention. This does not apply to interrupting boundary events.

For resolution steps, see [Workaround for Non-resolvable and Partially Resolvable Conflicts](/refguide/workflow-versioning/#workaround-for-non-resolvable-and-partially-resolvable-conflicts).

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
```diff
@@ -75,6 +75,23 @@ To configure the properties of a boundary event, double-click the event to open
...
 When a boundary event is added to an ongoing activity, any workflow instances currently executing that activity will schedule the new boundary event accordingly. The only exception occurs when an ongoing boundary event is removed from the ongoing activity, the workflow is redeployed, and then the removal is reverted. In this case, the re-added boundary event will not be scheduled.
 
+### Changing Boundary Event Type
+
+You can change the type of an existing boundary event via the context menu. To do so:
+
+1. Right-click the boundary event to open its context menu.
+2. Click **Change event**.
+3. Select one of the available options:
+
```
Remcovh, line 85: I think we can leave out the image and options. So just 

You can change the type of an existing boundary event via the context menu, without having to delete and re-add it. To do so:

1. Right-click the boundary event to open its context menu.
2. Click Change event.
3. Select the desired boundary event type.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
DmytroKost: > changing an ongoing non-interrupting boundary event

may imply any change to its configuration (like first execution type). I think it makes sense to mention "type" like I did and give an example.

> which requires manual intervention

This is true for every conflict. I think the problem is not manual intervention but rather the limitation that one cannot continue the workflow.

> For resolution steps, see

That linked section focuses on a workaround. The resolution steps are mentioned in the first reference to the conflict Current Activity Moved out of Path.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
DmytroKost: I rephrased the into to the section like you suggested.

Regarding the image, if we can show part of UI we are referring to I see no reasons not to. In this case its a single graphic so it doesn't even clutter the documentation.

If you think we should remove this picture then we could just as well remove the rest of them from the page. right?

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
Remcovh: > may imply any change to its configuration (like first execution type). I think it makes sense to mention "type" like I did and give an example.

Make sense, let's keep type

> This is true for every conflict. I think the problem is not manual intervention but rather the limitation that one cannot continue the workflow.

Can we put the emphasis on "one cannot continue the workflow"?

> That linked section focuses on a workaround. The resolution steps are mentioned in the first reference to the conflict Current Activity Moved out of Path.

So it would become something like this?

> Changing a non-interrupting Boundary Event Type
> 
> Due to technical limitations, changing an ongoing non-interrupting boundary event type (for example, from Timer to Notification) creates a partially resolvable [Current Activity Moved out of Path](/refguide/workflow-versioning/#current-activity-moved-out-of-path) conflict on running workflow instances. Affected workflow instances cannot be continued. This does not apply to interrupting boundary events.
> 
> For more information on how to handle such conflicts, see [Workaround for Non-resolvable and Partially Resolvable Conflicts](/refguide/workflow-versioning/#workaround-for-non-resolvable-and-partially-resolvable-conflicts).

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
Remcovh: Good point, lets keep the picture.

content/en/docs/refguide/modeling/application-logic/workflows/boundary-events.md
DmytroKost: Applied the suggestions

OlufunkeMoronfolu (approved):

PR #11329, opened 2026-06-09
Add 11.6 MTS support exception info box

jdiehl:
Adds an info box under the MTS section to clarify that Mendix 11.6 is an exception to standard MTS support policy and will be supported until 3 months after the 11.18 LTS release.


MarkvanMents (approved):
Looks good. I've just rewritten it a little bit, but no changes to meaning.
Approving and merging.

PR #11337, opened 2026-06-10
Fixed CVEs in new releases

Nidhi251289:
Fixed the CVEs in the new Operator and PCLM releases


katarzyna-koltun-mx (approved):

PR #11350, opened 2026-06-12
Update macOS requirements

GulamY:
Specify that Studio Pro for macOS is now only arm64


quinntracy:
@GulamY To confirm, are Intel-based Macs still supported in beta mode for Mendix 11 versions below 11.12?

GulamY:
We're planning to make 10.24 and 11.6 patch releases Apple Silicone only as well. The non lts/mts will not receive this update and will remain Intel (x64) based.

quinntracy (approved):

PR #11359, opened 2026-06-16
Revise license configuration in documentation

mruiserrmendix:
Updated license configuration syntax and added environment variable references.

### DIFF
--- a/content/en/docs/deployment/general/subscription-management/licensing-apps.md
+++ b/content/en/docs/deployment/general/subscription-management/licensing-apps.md
@@ -169,7 +169,7 @@ For more instructions on how to do this, see [Linux Deployment](/developerportal
 
 ### Portable App Distribution{#portableappdistribution}
 
-To activate the license on a Mendix app using Portable App Distribution on either Docker, Cloud Foundry, or Unix-like servers, follow these steps:
+To activate the license on a Mendix app using Portable App Distribution on Docker, Cloud Foundry, or Unix-like servers, follow these steps:
 
 1. Open the `$ConfigName.conf` in `etc/configurations`.
 2.  Add the `LicenseID` and `LicenseKey` to your runtime configuration:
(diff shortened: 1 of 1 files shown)


content/en/docs/deployment/general/subscription-management/licensing-apps.md
```diff
@@ -176,10 +176,12 @@ To activate the license on a Mendix app using Portable App Distribution on eithe
...
+runtime {
+ license {
+   id = 
+    key = 
+  }
 }
 ```
 
-The values for these properties can also be passed by using environment variables for your deployment type, or by creating a separate config file.
+The values for these properties can also be passed by using the environment variables `RUNTIME_LICENSE_ID` and `RUNTIME_LICENSE_KEY` for your deployment type, or by creating a separate config file.
```
MarkvanMents: What do you mean by environment variables for your deployment type?
We are just talking about PAD deployment here, so surely the customer just has one license ID and license key for PAD, which is what they should be putting here?

MarkvanMents (commented):
@mruiserrmendix 
Hi Mario,
One quick question below which wasn't clear to me.

content/en/docs/deployment/general/subscription-management/licensing-apps.md
mruiserrmendix: Yes, we are just talking about PAD deployment.
My meaning of deployment type is Linux, Docker, K8S, CF

MarkvanMents: @mruiserrmendix so a customer could have licenses for multiple PAD deployments, for Linux, Docker, K8s etc. and needs to make sure they supply the correct license details.

MarkvanMents: Updated as discussed in Slack.

MarkvanMents (commented):
Thanks for the PR @mruiserrmendix - approving and merging.

MarkvanMents (approved):

PR #11360, opened 2026-06-16
Add release notes for version 3.8.0

MXLino:
Added release notes for version 3.8.0, detailing new features, improvements, and bug fixes for Workstation Management, Workstation Client, and Workstation Connector.

### DIFF
--- a/content/en/docs/releasenotes/workstation/_index.md
+++ b/content/en/docs/releasenotes/workstation/_index.md
@@ -23,7 +23,7 @@ These release notes cover changes made to the [Mendix Workstation](/mendix-works
 * Printer Device Support – We added full support for setting up and managing new printer device types directly within the Management UI. This expands the range of peripherals you can integrate with your Workstations.
 * Manual Bulk Configuration Push – A new **Sync Clients** button has been added to the station overview page, allowing you to manually sync client configurations for workstations that are out of sync. This provides a quick way to ensure all clients have the latest settings.
 * Workspace Environment Types – Workspaces can now be assigned an environment type (Test, Acceptance, Production) when created, and this can be changed in the workspace settings. This helps you better organize and manage your different development and deployment stages.
-* Serial Device Detection by pnpId – We added setup UI support for detecting serial devices using their pnpId. This provides another robust method for reliably identifying and connecting to serial devices.
+* Serial Device Detection by pnpId – We added support for detecting serial devices using their pnpId. This provides another robust method for reliably identifying and connecting to serial devices.
 
 #### Improvements
 
--- a/content/en/docs/releasenotes/workstation/_index.md
+++ b/content/en/docs/releasenotes/workstation/_index.md
@@ -34,6 +34,9 @@ These release notes cover changes made to the [Mendix Workstation](/mendix-works
 #### New Features
 
 * Printer Device Support – The Workstation Client now fully supports the new printer device type. This means your Workstation Clients can seamlessly interact with configured printers.
+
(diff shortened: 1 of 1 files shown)


content/en/docs/releasenotes/workstation/_index.md
```diff
@@ -10,6 +10,45 @@ cascade:
...
+#### Improvements
+
+*   **Streamlined Station Detail Page:** The station detail page UI has been improved by making options like "detect card reader" and "dev mode edit" less prominent. This declutters the interface and highlights the most frequently used settings.
+
+
+### Workstation Client
+
+#### New features
+
+*   **Printer Device Support:** The Workstation Client now fully supports the new printer device type. This means your Workstation Clients can seamlessly interact with configured printers.
```
MXLino: Here, we need to add a note that currently only the following is supported: 
Text in raw format and label printers

content/en/docs/releasenotes/workstation/_index.md
```diff
@@ -10,36 +10,73 @@ cascade:
...
+
+#### Fixes
+
+* Studio Pro Login Blocked – Fixed an issue where the client blocked sign-in to Studio Pro when six or more devices were configured, due to an overlap in port usage. Studio Pro users can now sign in without issues, regardless of the number of configured devices.
+
+### Workstation Connector
+
+#### Improvements
+
+* Improved Get/Create Device Action – The `Get/Create Device` action now returns an empty object instead of throwing an error when no matching device is found. This makes your custom integrations more robust and prevents unexpected crashes.
```
MarkvanMents, line 51: Is this a fix rather than an Improvement? Sounds like if it crashed that was something that needed to be fixed.

content/en/docs/releasenotes/workstation/_index.md
```diff
@@ -10,36 +10,73 @@ cascade:
...
+* Printer Device Support – We added full support for setting up and managing new printer device types directly within the Management UI. This expands the range of peripherals you can integrate with your Workstations.
+* Manual Bulk Configuration Push – A new **Sync Clients** button has been added to the station overview page, allowing you to manually sync client configurations for workstations that are out of sync. This provides a quick way to ensure all clients have the latest settings.
+* Workspace Environment Types – Workspaces can now be assigned an environment type (Test, Acceptance, Production) when created, and this can be changed in the workspace settings. This helps you better organize and manage your different development and deployment stages.
+* Serial Device Detection by pnpId – We added setup UI support for detecting serial devices using their pnpId. This provides another robust method for reliably identifying and connecting to serial devices.
```
MarkvanMents: I'm not sure I understand this. I'm not sure we need a reference to the UI here. Could we just say

```suggestion
* Serial Device Detection by pnpId – We added support for detecting serial devices using their pnpId. This provides another robust method for reliably identifying and connecting to serial devices.
```

MarkvanMents (commented):
Thanks for the PR @MXLino 
Looks good.
I've just applied some of our style manual and have a couple of very minor questions.

content/en/docs/releasenotes/workstation/_index.md
MXLino: Previously, you had to perform the empty check before calling this action to prevent a meaningless error message from appearing if no device existed. Therefore, the original behavior was known and expected, but from a developer's perspective, it wasn't very user-friendly, hence the improvement.

MarkvanMents: Thanks for the explanation. I'll leave it as an improvement.

MarkvanMents (approved):
Thanks for the response to my questions.
Approving and publishing.

PR #11377, opened 2026-06-18
Clarify System.User access rules

hnejra:


MarkvanMents (commented):
Hi Arjen
I've moved the not in the info box into the warning box as I think it is best to have all the information there.
I have also tried to clarify the structure to make the points stand out, and made the sentences a bit shorter. I don't think I've changed the meaning.

Please review the Mx11 version and, once we are happy, I can copy the changes to Mx10 and Mx9 as well.

Thanks for the PR - I think the more clarity we can have about how System.User works, the better.

MarkvanMents (approved):
Thanks for the PR
Approving and Merging.

PR #11397, opened 2026-06-23
Upload SP 11.12.0 LTS RNs

ConnorLand:
This release may have LTS dependencies that need to be added into this PR. I'll put in as many as I know and then check it with Maria.


content/en/docs/releasenotes/studio-pro/11/11.12.md
```diff
@@ -0,0 +1,164 @@
+---
+title: "11.12"
+url: /releasenotes/studio-pro/11.12/
+description: "The release notes for Mendix Studio Pro 11.12 (including all patches) with details on new features, bug fixes, and known issues."
+weight: 89
```
MarkvanMents: Weight should be 88.
Can you make sure that changing this is part of the process?

MariaShaposhnikova:
Hi @ConnorLand , what do you mean by LTS dependencies?

content/en/docs/releasenotes/studio-pro/11/11.12.md
MariaShaposhnikova: It also needs an lts badge, please add it to the metadata.

ConnorLand:
> Hi @ConnorLand , what do you mean by LTS dependencies?

I mean any doc changes to do with LTS status that Mark/you/Hans deV are handling. Within this PR, the only docs I have changing are the RNs and the latest releases.

content/en/docs/releasenotes/studio-pro/11/11.12.md
ConnorLand: I changed the weight and added the label. It is already a part of the process. I'll make sure to square away the metadata earlier in the RN production flow as per the process.

PR #11404, opened 2026-06-24
[RUN-5208] Update offline docs for java update tool

emirmx:
This is for Studio Pro 11.13 release.


atennapel:
Looks good from the technical side.

emirmx:
@quinntracy We released a new version of the tool and updated the docs accordingly with the new commit.

quinntracy (approved):

PR #11426, opened 2026-06-26
Update backups API documentation for optional comment

sagaranand015:
Clarified that the comment attribute is optional and specified that an empty JSON object is needed for a successful response.


sagaranand015 (commented):
Looks Good.

PR #11444, opened 2026-06-30
SAML limitation: need for restart with multiple nodes

Karuna-Mendix:
https://mendix.atlassian.net/browse/TW-2379


Karuna-Mendix:
Phani approved this PR via Slack.

PR #11483, opened 2026-07-03
Restoring the pipeline api yaml

katarzyna-koltun-mx:


katarzyna-koltun-mx:
I have mistakenly pushed this fix directly into production, backporting it into dev.

PR #11493, opened 2026-07-07
SCIM Module: update "Attribute Mapping" section

Karuna-Mendix:


Karuna-Mendix:
Phani approved via slack.

PR #11510, opened 2026-07-09
Add T Error Known Issue

ConnorLand:
From 10.21 and above, and 11.1 and above. Both major version RNs may have slightly different text, but their Zendesk Problem # is the same so it's clear it's the same core KI.


content/en/docs/releasenotes/studio-pro/10/10.21.md
```diff
@@ -27,6 +28,7 @@ weight: 79
     * Fixed in [10.24.4](/releasenotes/studio-pro/10.24/#fix-debugger-runtime-only) and [11.1.0](/releasenotes/studio-pro/11.1/#fix-debugger-runtime-only).
 * In the React client, the debugger does not reconnect when restarting an app.
     * Fixed in [10.24.4](/releasenotes/studio-pro/10.24/#fix-debugger-reconnect) and [11.1.0](/releasenotes/studio-pro/11.1/#fix-debugger-reconnect).
+* When opening, editing, or creating a new microflow or workflow in Studio Pro, this error occurs: `t is not a function. Please Close and reopen this UI component. If the issue persists, contact Mendix Support.` The microflow or workflow is not created, and the editor does not open, preventing further development activities.
```
ConnorLand: Recieving text from an SME on this

PR #11539, opened 2026-07-15
Mention constants and enums in docs for OQL

dahfjkg:
To be included in 11.13 release. Requires team review.


pijuskri:
Approved by team

MarkvanMents (approved):
I assume that the constant name and enumeration value are checked at design time so that you get an error and can't compile the app if they are entered incorrectly? Don't think we have to mention that - the customer will find out.
Approving ready for publication with 11.13

PR #11545, opened 2026-07-15
Upload 11.6.8 SP RNs

ConnorLand:
Ordered the tickets and implemented the updated fix language for the t error fix


content/en/docs/releasenotes/studio-pro/11/11.6.md
```diff
@@ -8,6 +8,39 @@ mts: true
...
+- We fixed an issue in React client mode where navigating between pages with the same layout left widgets unresponsive. (Ticket 276710)
+- We fixed an issue where Studio Pro crashed when pasting content from the clipboard while another application was also accessing the clipboard. (Ticket 278333)
+- On SQL Server, when a unique constraint or foreign key constraint is dropped, we now use `IF EXISTS`. This avoids an automatic rollback of the synchronization transaction in case the constraint does not exist. (Ticket 279545)
+- We fixed an issue where runtime fails to finish executing a queued task started by a user if the user is deleted during task execution. (Ticket 279891)
+- We fixed an issue where the native `Switch` widget did not render correctly on iOS. (Ticket 280839)
+- <a id="fix-t-is-not-a-function"></a>We fixed the `t is not a function` crash in microflow and workflow editors, introduced by the recent Microsoft update for WebView. If you installed an older version of the Microsoft Edge WebView as a workaround and you have no other affected Studio Pro installations, we recommend uninstalling it to prevent future compatibility issues. (Ticket 282591, 282696)
```
ConnorLand: With 11.6.8, we now have a fix for this KI in this minor version. Currently all the links to the fix point to the fix in 11.12, but I should probably change all the fix links in 11.6 to the fix in 11.6.8, right @MariaShaposhnikova ?

Helps customers that want to stay on 11.6 do so.

lmeijvogel: You could also add the page editor to this list, but I think we also didn't mention it in the 11.12 RNs.

lmeijvogel: In that case:

 <a id="fix-t-is-not-a-function"></a>We fixed the `t is not a function` crash in microflow, workflow, and page editors, introduced by the recent Microsoft update for WebView. If you installed an older version of the Microsoft Edge WebView as a workaround and you have no other affected Studio Pro installations, we recommend uninstalling it to prevent future compatibility issues. (Ticket 282591, 282696)

ConnorLand: I did so.

PR #11550, opened 2026-07-16
Unreleased feature "Branching off from module should retain link to origin module" documentation part removed

MarkTr0:


MariaShaposhnikova:
Closed as the feature was released.

PR #11602, opened 2026-07-23
Allow setting and showing version for all modules from mx.exe

MarkTr0:
Important: This is feature for 11.14, not 11.13


quinntracy:
@MarkTr0 Has this update been backported to 11.12.3? I see a release note that has this same information. If so, I would like to create a separate PR with the information and make a clear distinction that it is for 11.12.3 and above (as well as 11.14 and above when this PR is released).

MarkTr0:
> @MarkTr0 Has this update been backported to 11.12.3? I see a release note that has this same information. If so, I would like to create a separate PR with the information and make a clear distinction that it is for 11.12.3 and above (as well as 11.14 and above when this PR is released).

We have not backported anything related to this feature yet. Could you please share where it is already in release notes?

quinntracy (approved):

PR #11615, opened 2026-07-27
Clarify Mendix on Kubernetes subscription requirement after Operator license removal

MXClyde:
## Summary
- The Operator License alert on the Mendix on Kubernetes page implied a subscription was only needed for technical support, which caused real customer confusion after the standalone Operator license key was removed in Operator 2.23.0 (a customer asked support whether Mendix on Kubernetes still needed to be purchased at all).
- Rewords the alert and the following paragraph to make clear that removing the license key was a technical change only, and a valid Mendix on Kubernetes subscription is still required to use the Operator, regardless of Operator version.
- Also fixes a confusingly-ordered sentence in the Runtime License section that described the Runtime/Operator license relationship starting from the wrong subject.

## Test plan
- [x] Rendered the page locally to confirm formatting/alert shortcodes are unaffected
- [x] Verified no other page content references the old wording

### DIFF
--- a/content/en/docs/deployment/private-cloud/_index.md
+++ b/content/en/docs/deployment/private-cloud/_index.md
@@ -105,10 +105,10 @@ This percentage can be adjusted by providing a custom value in Custom JVM Option
 ### Operator License
 
 {{% alert color="info" %}}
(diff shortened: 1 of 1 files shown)


katarzyna-koltun-mx (approved):
@MXClyde approved, should I merge directly?

MXClyde:
> @MXClyde approved, should I merge directly?

Yes please!

PR #11642, opened 2026-07-30
MendixOnAzure: Add documentation for pre-provisioned Grafana alerts

walkline:
Mendix on Azure will provision a set of Grafana alert rules covering applications, nodes, and the PostgreSQL server. Document them ahead of the release.

Add a reference page describing each rule, its severity, condition, and pending period. The page frames the rules as examples that customers own and are expected to adjust, because they are provisioned once at cluster creation and are not changed afterwards.

Email notifications need an SMTP relay that Azure Managed Grafana does not provide, so include the az grafana update configuration and the parameters that most often need attention. Also explain the No Data state, which is normal when no app is deployed.

Link to the new page from the monitoring page.

### DIFF
--- a/content/en/docs/deployment/mx-azure/mx-azure-alerts.md
+++ b/content/en/docs/deployment/mx-azure/mx-azure-alerts.md
@@ -7,9 +7,9 @@ description: "Describes the example alert rules that are provisioned with Mendix
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


katarzyna-koltun-mx (approved):

katarzyna-koltun-mx:
Reviewed and waiting for the go ahead (probably close to Aug 20).

PR #11670, opened 2026-08-04
Module ID as selected element

ester-nl:
for 11.14 and maybe 11.12.3 patch as well (tbd)


quinntracy (approved):

ConnorLand:
@quinntracy I resolved conflicts so all versions are present. Please check later to ensure it was correct.

Issue #10730, opened 2026-02-09
use-the-plotly-images-rest-service-endpoint


edwinvanelk:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [use-the-plotly-images-rest-service-endpoint](https://docs.mendix.com/appstore/widgets/charts-plotly-images-rest/)

Document link: [charts-plotly-images-rest.md](https://github.com/mendix/docs/blob/development/content/en/docs/marketplace/platform-supported-content/widgets/charts/charts-plotly-images-rest.md)

## My Issue/Suggestion

Does the Plotly Images REST Service Endpoint still exist? My sample app does not work anymore and the endpoints are not reachable anymore.

Can you update the documentation?

ConnorLand:
Hello,

We had wide-ranging issues throughout docs when referring to plotly.js versions. I have recently worked with a developer to modernize this aspect, as well as modernize Charts docs in general. The old Plotly how to is deleted, and in its place we have clearer reference information and smoother tutorials: https://github.com/mendix/docs/pull/11681

Issue #7649, opened 2024-04-24
any-chart - update plotly library


lucadesimoni:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [any-chart](https://docs.mendix.com/appstore/modules/any-chart/)

Document link: [any-chart.md](https://github.com/mendix/docs/blob/development/content/en/docs/appstore/modules/any-chart.md)

## My Issue/Suggestion

Please update Any Chart to a more recent version of plotly library to supply additional chart types and functionalities

ConnorLand:
Hello,

We had wide-ranging issues throughout docs when referring to plotly.js versions. I have recently worked with a developer to modernize this aspect, as well as modernize Charts docs in general. The old Plotly how to is deleted, and in its place we have clearer reference information and smoother tutorials: https://github.com/mendix/docs/pull/11681

Issue #5311, opened 2022-11-15
any-chart


Olerichi:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [any-chart](https://docs.mendix.com/appstore/modules/any-chart/)

Document link: [any-chart.md](https://github.com/mendix/docs/blob/development/content/en/docs/appstore/modules/any-chart.md)

## My Issue/Suggestion

Updated version to plotly.js version 1.51.2???  -> this is what is stated in the Marketplace Module

ConnorLand:
Hello,

We had wide-ranging issues throughout docs when referring to plotly.js versions. I have recently worked with a developer to modernize this aspect, as well as modernize Charts docs in general. The old Plotly how to is deleted, and in its place we have clearer reference information and smoother tutorials: https://github.com/mendix/docs/pull/11681

PR #11681, opened 2026-08-06
[WC-3458]: update chart docs

rahmanunver:
## What

- Corrects the plotly.js version in the Charts docs: the reference guides said **1.47.4**, but Charts has shipped plotly.js 3.x since v6.0.0. Fixed across the Mendix 9, 10, and 11 guides.
- Removes the Plotly Images REST endpoint how-to. The `api.plot.ly/v2/images` service no longer exists, so the page can't be followed.

## Why

The version claim predates two plotly major releases and was reported by a customer. Charts 6.0.0 runs on Mendix 9.6+, so the wrong version was live on every supported guide.

The docs now name the major version and point to the Marketplace listing for the exact one, so this won't drift again. The retired page's URLs are kept as aliases, so old links redirect rather than 404.


ConnorLand:
Thanks @rahmanunver ! This PR links up to existing GH issues, so I'll edit/merge then tell the issue people about the solution:

- https://github.com/mendix/docs/issues/5311
- https://github.com/mendix/docs/issues/7649
- https://github.com/mendix/docs/issues/10730

ConnorLand:
I'll respond here when it's merged: https://mendix.slack.com/archives/C63U54B6K/p1784552823134959

ConnorLand (approved):

ConnorLand (approved):

PR #11723, opened 2026-08-12
NIST artifact signing

katarzyna-koltun-mx:


content/en/docs/private-platform/pmp-registry-sign.md
```diff
@@ -0,0 +1,115 @@
...
+```text
+cosign sign-blob --key "$COSIGN_PRIVATE_KEY" --bundle $FILE.sig https://cdn.mendix.com/$FILE
+```
+
+The `$FILE.sig` file is distributed together with any downloadable files and can be used to verify the signature by using the following command:
+
+```text
+cosign verify-blob --key "$COSIGN_PUBLIC_KEY" --bundle $FILE.sig https://cdn.mendix.com/$FILE
+```
+
```
katarzyna-koltun-mx, line 101: In the original draft, there was the following line here: "These CDN files may be transferred to central registry in the future. In that case, user need oras tool to download them."

I removed it - if this is actually implemented, we'll need to update the doc anyway.

PR #11753, opened 2026-08-18
Clarify Manageable Roles

MarkvanMents:
There is confusion about the way that manageable roles work, and there is also a major confusing change in the UX.

This PR aims to clarify how manageable roles work.


content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
-End-users of your application only see the user roles and not the module roles.
-
-To access user roles, open **App Security** > **User roles** tab:
-
-{{< figure src="/attachments/refguide/modeling/security/app-security/user-roles/user-roles-example.png" class="no-border" >}}
+End-users of your application only see the user roles and not the underlying module roles.
 
 {{% alert color="warning" %}}
-The effects of changes to user roles are not immediate. This means that your app might show outdated pages or incorrect data. For more information, please refer documentation about [persistent sessions](/refguide/clustered-mendix-runtime/#sessions-are-always-persistent).
+The effects of changes to user roles are not immediately applied to end-users who are signed in to the app. End-users need to sign out and sign in again to pick up their new roles. This means that your app can show the wrong pages or data. For more information, refer to the documentation on [persistent sessions](/refguide/clustered-mendix-runtime/#sessions-are-always-persistent).
```
atennapel: > End-users need to sign out and sign in again to pick up their new roles.
Or wait 30s (i.e. the `SessionValidationTimeout`). After 30s the session will be revalidated by fetching it from the database, which will have the updated roles.

knorrie: I think we should be cautious in the wording. There are all kinds of places where the logged in users roles are cached and used to make decisions about for example what kind of interface elements to show or not in the web client.

MarkvanMents: I also think we should be cautious. How about we say something like:
```suggestion
The effects of changes to user roles are not immediately applied to end-users who are signed in to the app. This means that your app can show the wrong pages or data. Although the user roles are reapplied when their session is revalidated, Mendix advises that end-users sign out and sign in again to pick up their new user roles. This ensures that they have the correct user roles. For more information, refer to the documentation on [persistent sessions](/refguide/clustered-mendix-runtime/#sessions-are-always-persistent).
```

@atennapel and @knorrie let me know if you are happy with this and I can commit the change and merge the PR.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
 
 ## Introduction
 
-A user role aggregates a number of access rights on data, forms, and microflows. An end-user of the application is assigned one or more user roles by an administrator, and gets all access rights that these user roles represent.
+A user role aggregates a number of access rights to app documents such as data, pages, and microflows. Each end-user of the application is assigned one or more user roles by an end-user with user management rights, and gets all the access rights that these user roles represent.
 
-Every user role has one or more [module roles](/refguide/module-security/#module-role), which means that users with that user role have all the access rights that are defined for those module roles. A typical user role has the **System.User** module role and at least one other module role.
+Every user role has one or more [module roles](/refguide/module-security/#module-role). End-users with that user role are granted all the access rights defined for those module roles. A typical user role has the **System.User** module role and at least one other module role.
```
MariaShaposhnikova:
```suggestion
Every user role has one or more [module roles](/refguide/module-security/#module-role), which means that a user role is granted all the access rights defined for this module role. A typical user role has the **System.User** module role and at least one other module role.
```

MariaShaposhnikova: I had to reread the sentence several times to understand it and tried to make it clearer.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
 
-The purpose of the distinction between user roles and module roles is to make a module self-contained (independent from the app in which it is defined or used), so that it can be reused in different apps and/or published to the Marketplace.
+The separation of user roles and module roles means that modules can be self-contained (independent of the app in which it is defined or used). This allows it to be reused in different apps or published to the Marketplace.
```
MariaShaposhnikova:
```suggestion
The difference between user roles and module roles is that modules can be self-contained (independent of the app in which it is defined or used). This allows it to be reused in different apps or published to the Marketplace.
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
-| Documentation  | In this property you can document additional information about the user role. This information is shown to end-users who can create or edit user accounts in the application. |
-| Module roles   | A list of module roles of which the access rights are accumulated in the user role. An end-user that is assigned a user role gets all access rights of the module roles of that user role. |
-| Check security | This specifies whether the consistency of security settings is checked for this user role. You can choose to not check security for a user role. For example, user roles that are used only for web service users do not need to be checked because they never sign in to the client. For more information on the security check, see [App Security](/refguide/app-security/). |
+| Property | Description |
+| --- | --- |
+| Name | The name of the user role. This name is shown to end-users who can create or edit user accounts in the application. Blank apps usually start with the `User` and `Administrator` user roles. |
```
MariaShaposhnikova:
```suggestion
| Name | The name of the user role. This name is shown to end-users who can create or edit user accounts in the application. Blank apps usually start with the **User** and **Administrator** user roles. |
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
+| Module roles | A list of module roles whose access rights are added to this user role. An end-user that is assigned this user role gets all access rights of the module roles of that user role. |
+| Check security | Specifies whether Studio Pro checks during development that security settings for this user role are consistent.<br/>You can choose to not check security for a user role. For example, user roles that are used only for web service end-users do not need to be checked because they never sign in to the client. For more information on the security check, see the [Check Security](/refguide/app-security/#check-security) section of *App Security*. |
 
 ### User Management Properties {#user-management}
 
-A user role can be allowed to manage users with a number of other user roles (including itself), called manageable roles. This means that end-users who have this user role, can create, view, edit and delete users with at most the manageable user roles.
+A user role can be allowed to manage end-users depending on their user roles. These roles are called manageable roles (or grantable roles).
```
MariaShaposhnikova:
```suggestion
A user role can be allowed to manage end-users. User roles that can be managed by another user role are called manageable roles (or grantable roles).
```

MariaShaposhnikova: It gets confusing whose user role we mean here, so I removed it.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
 
-| Value | Description |
-| --- | --- |
-| All | End-users with this user role can manage all users and grant all user roles. Usually this option should only be configured for an administrator. |
-| Selected | End-users with this user role can manage users that have at most the selected user roles, and can grant only the selected user roles. Select **(No user roles)** to only manage users without a user role (for example, newly created users). If nothing is selected, end-users with this user role cannot manage users at all.  |
+End-users who have this user role can create, view, edit, and delete end-users with the selected roles, provided they do not also have a user role which is not selected.
```
MariaShaposhnikova:
```suggestion
End-users who have manageable user role can create, view, edit, and delete end-users with the selected roles, provided they do not also have a user role which is not selected.
```

MariaShaposhnikova: Not sure I understand what "which is not selected" means?

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
+
+Take for example an app with three defined user roles: `User`, `Administrator`, and `SubAdministrator`. Assume that you are configuring the `SubAdministrator` user role. The `SubAdministrator` user role has only the `User` user role as a manageable role.
+
+{{< figure src="/attachments/refguide/modeling/security/app-security/user-roles/manageable-roles.png" alt="User management section with only the User role selected as a manageable role" >}}
+
+The following table shows which end-users an end-user with the `SubAdministrator` user role can manage:
```
MariaShaposhnikova:
```suggestion
The following table shows which end-users the `SubAdministrator` user role can manage:
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
+| End-User Name | Has User Roles | SubAdministrator can manage |
+| --- | --- | --- |
+| User1 | `User` | Yes¹ |
+| User2 | `SubAdministrator` | No |
+| User3 | `User` and `SubAdministrator` | No² |
+
+¹ The SubAdministrator can only grant or remove the `User` user role.
+² Although `User` is a manageable role, User3 also has the `SubAdministrator` user role which is not a manageable role for a SubAdministrator.
+
+The **(No user roles)** manageable role allows this user role to manage end-users without a user role (for example, newly created end-users).
```
MariaShaposhnikova:
```suggestion
The **(No user roles)** role is a manageable role and allows another user role to manage end-users without a user role (for example, newly created end-users).
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
```diff
@@ -10,56 +10,92 @@ aliases:
...
+
+The **Select / deselect all** checkbox lets you select all the roles as manageable roles, or deselect them all.
+
+{{% alert color="warning" %}}
+If the **Select / deselect all** box is checked (that is, all roles are manageable by this user role) then adding a new user role to the app will mean that it is automatically selected as a manageable role. If any of the roles are not selected, then added user roles will not be automatically selected as manageable roles.
```
MariaShaposhnikova, line 82:
```suggestion
If the **Select / deselect all** box is checked (that is, all roles are manageable by this user role) then adding a new user role to the app will mean that it is automatically selected as a manageable role. If none of the roles are selected, then added user roles will not be automatically selected as manageable roles.
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
atennapel: Yes looks good to me.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents: That isn't what I meant.

What I mean is that if I am an end-user (Mark) and I have this user role, I am granted access rights from all the module roles which are part of this user role.
A user role is not granted anything - it is just a collection of module roles. It is only when it is manifested as an end-user that it does this.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents: Again, that is not what I meant. This is an explanation of why there are user roles and module roles and we don't just put everything into user roles.

What I am saying is that having both user roles and module roles _means that_ modules can be self contained. The self-containedness is not a difference between module roles and user roles.

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents: It is confusing, but it is necessary in this description. What I'm saying is that this user role (UserRole1) can manage end-users but only if they have a manageable role (e.g. UserRole2) - we lose information if we take out "depending on their user roles". How about:

```suggestion
A user role can be allowed to manage end-users depending on the user roles that the end-user has been granted. The end-user's user roles are called manageable roles (or grantable roles).
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents: No - _this_ user role is not the manageable role, it is the selected roles which are manageable.


```suggestion
End-users who have this user role can create, view, edit, and delete end-users with the selected manageable roles, provided they do not also have a user role which is not selected.
```

MarkvanMents: Yep - this is a mess which is why the next lines give an example 😁

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents:
```suggestion
The **(No user roles)** role is a manageable role which allows the current user role to manage end-users without a user role (for example, newly created end-users).
```

content/en/docs/refguide/modeling/security/app-security/user-roles.md
MarkvanMents: No - that changes the meaning. If _any_ of the roles are not selected then automatic selection is turned off. It only gets automatically selected if **all** the roles are selected.

_A terrible UX decision in Mx11_ 🙄

MarkvanMents (commented):
@MariaShaposhnikova 
I'm afraid I have rejected some of your suggestions.
I accepted a couple and have proposed alternatives for a couple.
However, some of your suggestions change the meaning, so I'm going to have to reject them and rethink how to clarify what I mean.

PR #11783, opened 2026-08-20
Document AKS internal CIDR ranges to avoid on Mendix on Azure

MXClyde:
## Summary
- Document the AKS internal CIDR ranges (`10.0.0.0/16` service CIDR, `10.244.0.0/16` pod CIDR) that every Mendix on Azure cluster reserves internally, so customers don't pick a VNet/on-prem range that overlaps them when planning peering or on-prem connectivity.
- Updates the *AKS Node CIDR IP Range* portal setting description (also used as the in-portal tooltip) and adds a warning callout with the full peering/on-prem guidance and the Private Endpoints workaround for unavoidable overlap.

Resolves [MXFORAZURE-549](https://mendix.atlassian.net/browse/MXFORAZURE-549).

## Test plan
- [ ] Table renders correctly and the new alert box follows existing Hugo shortcode conventions on this page
- [ ] Links/formatting check in docs preview


MXClyde:
@katarzyna-koltun-mx This is a small, self-contained doc addition (1 file, +12/-1) - can be merged immediately.

PR #11798, opened 2026-08-24
Document Teamcenter role sharing with embedded Mendix app

jdiehl:
Adds a short section to Mendix Inside Teamcenter explaining that the active Teamcenter role is shared between Active Workspace and the embedded Mendix app, and that switching roles reloads both.


brambrink (approved):
Looks good 👍

NicoletaComan:
Thank you, @jdiehl and @brambrink.
I'd just like to clarify one thing: does role sharing work automatically, OOB, or does it require any sort of setup? Are there any further configuration steps that we should describe or maybe refer to?

PR #11824, opened 2026-08-27
Advise customer on separating security and business entities

MarkvanMents:


content/en/docs/howto/security/best-practices-security.md
```diff
@@ -18,19 +18,19 @@ This document describes the common aspects to consider when deploying an applica
 
 ## Implementing Access Rules{#access-rules}
 
-The Mendix architecture includes the Mendix Client, which can compose its own queries (generated by widgets) and send them to the Mendix Runtime. If the Mendix Runtime were to execute these queries without touching them, the Mendix Client would be able to request data that the user should not be allowed to see.
+The Mendix architecture includes the Mendix Client, which can compose its own queries (generated by widgets) and send them to the Mendix Runtime. If the Mendix Runtime were to execute these queries without touching them, the Mendix Client would be able to request data that the user is not be allowed to see.
```
knorrie: Now it says "is not be allowed to see"

MarkvanMents: Thanks - fixed.

content/en/docs/refguide/modeling/security/app-security/administrator.md
```diff
@@ -72,15 +72,15 @@ The user role assigned to the Administrator. For more information, see [User Rol
 Default: *Administrator*
 
 {{% alert color="info" %}}
-The administrator is always created and has the System.Administrator role by default. The System.Administrator role allows users of your application to be managed. 
+The Administrator user role is always created and has the System.Administrator module role by default. The Administrator user role can also manage users of your application as it has all [manageable roles](/refguide/user-roles/#user-management). 
 
-For Free Apps, the user that created the application automatically has this role by default as well so you can use it to manage your users in that environment.
+For Free Apps, the user that created the application automatically also defaults to having the Administrator role so they can use it to manage the users in that environment.
```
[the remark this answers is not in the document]
knorrie: No, 'Free Apps' is the name of the product offering where you run your app in Mendix Cloud without a paid license.

content/en/docs/refguide/modeling/security/app-security/administrator.md
```diff
@@ -72,15 +72,15 @@ The user role assigned to the Administrator. For more information, see [User Rol
...
 
-This role may be helpful in case you have exceeded your user license restriction in which case you can use any user that has this System.Administrator role to sign in to manage your users.
+This role may be helpful if you exceed your user license restriction, as you can sign in as any end-user that has this Administrator user role to manage your end-users.
```
knorrie, line 79: The System.Administrator module role has entity access rights to remove System.Session objects.
However, there is no UI for this (in the System module). There is UI pages for this in the Administration module, which you cannot reach unless you also include the Administration.Administrator module role.

Technically the System.Administrator module role has nothing to do with user management, because that module role does not grant entity access on System.User. Instead it is the Manageable Roles configuration that does that. However, there is no UI for that (in the System module), so in the end it is the combination of having Manageable Roles AND having Administration.Administrator module role that causes you to be able to manage users (and only via Administration.Account usage).

However, the Administration module contains the UI for removing logged in user System.Session objects. So, probably the only way to end up without consistency check errors in Studio Pro is to have all three... Manageable Roles checked, System.Administrator module role and Administration.Administrator module role. These three go together and you have either all of them or none.

knorrie: Technically [...] has nothing to do > actually, it has, because having a user role that includes this module role means there is an exception that you can still log in while there are already (license) max sessions active.

Still, all of this only works ('by accident') if you have all 3 together.

knorrie: "only via Administration.Account usage" > Because even while having Manageable Roles configured gives you technically the option to create System.User objects, it is still not possible to commit those to the database, because it has both a) validation rule that says password attribute cannot be emtpy and b) sending a change for the password attribute from the client api is rejected.

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+While the data that is viewable and editable to each role is application-specific, the following best practices are key:
 
-* Attributes determined by the system (like the status of an order) should never be writable
-* If an anonymous user is allowed to create objects, constrain these objects to the owner (an anonymous user is actually a **System.User** object created on the fly)
-* Do not set a default rule for read-and-write access – this forces you to think about each attribute that is added to an entity
-* Security constraints should be formed as entity access rules
-* Constraints on widgets in pages should not be used as a measure of security, but can filter out irrelevant data for the context of the page
-* Keep your attributes editable within data views, because if an access rule prohibits write access, your client will display it as non-editable – this way you are aware of the (correct) working of an access rule
+* Do not make attributes determined by the system (like the status of an order) writable
+* If an anonymous user is allowed to create objects, constrain these objects to the owner (an anonymous user is a **System.User** object created on the fly)
```
knorrie: Link to a place in the docs where it shows how to do xpath id = currentuser?

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+* Do not set a default rule for read-and-write access—this forces you to think about each attribute that is added to an entity
```
knorrie: This is not a 'rule you set'. This is about the radio button "Default rights for new members" (o)None ( ) Read ( ) Read, Write, where None must be selected. This is also what is taught in class room training, security workshops and CTF stuff, to never use this.

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+* Implement security constraints as entity access rules
+* Do not use constraints on widgets in pages as a security measure, but filter out irrelevant data for the context of the page
+* Keep your attributes editable within data views, because if an access rule prohibits write access, your client will display it as non-editable—this way you are aware of the (correct) working of an access rule
```
knorrie, line 34: Here we need to (can also be todo for after this PR ;-]) add a thing about page variables! They are a really great help to make things more secure: If you have an attribute that you want to closely inspect changes on, then you can make the attribute read only in the access rights, and use a page variable (edit box) on your page for that value, and have it call a microflow. In the microflow, you then get the object (with unchanged attribute value) and separately, e.g. as a string value the new proposed value. Now you can validate the input and compare it to the previous value and decide if the change is allowed, change the actual object and commit it.

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+## Configuring User Roles and Access
 
-Examples are the `ws-doc` or `rest-doc` endpoints that enumerate all the published web and REST services of the application. An attacker could use this information to discover possible areas to exploit.
+Which users and roles are defined within an application changes, depending on the function of the app. However, there are some key guidelines to keep in mind when validating the user security:
 
-You can take the following preventative measures:
+* Disable anonymous access if it has no function within the application
+    * Some applications have anonymous access enabled, solely to serve a custom login form—this can be replaced by modifying the default *login.html* within your theme (which will also help the user experience with an improved loading time)
+* Make roles managing other user roles as strict as possible (configured via **User management** within the user role options)
+* Only allow the role of the app's administrator user (default **MxAdmin**) to create the actual administrative accounts (or configure Single Sign On (SSO))
```
knorrie: Best practice: Use this Administrator role only for local account management and/or SSO config, do not give it further access rights on your business process data. (This happens a *lot* where app developers mindlessly check everything readwrite for their business app data, even if there is no UI for it at all.)

Use a separate user role to organize 'god mode' access to business data if needed (fixing things 'in the backend').

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+To link the customer information to an end-user, create a 1-1 association, `Customer_Account`, between `MyModule.Customer` and `Administration.Account`.
 
-{{< figure src="/attachments/howto/security/best-practices-security/default-access-restrictions.png" alt="Mendix Cloud Access Restrictions Overview" class="no-border" >}}
+## Configuring User Roles and Access
 
-Examples are the `ws-doc` or `rest-doc` endpoints that enumerate all the published web and REST services of the application. An attacker could use this information to discover possible areas to exploit.
+Which users and roles are defined within an application changes, depending on the function of the app. However, there are some key guidelines to keep in mind when validating the user security:
 
-You can take the following preventative measures:
+* Disable anonymous access if it has no function within the application
```
knorrie: Link to other place where more best practices for Anonymous are mentioned?
https://docs.mendix.com/refguide/anonymous-users/

By the way, in there it should be mentioned more explicitly to best practice recommend to have a dedicated user role (preferably named "Anonymous") for it, and in modules have explicit Anonymous module roles, and not mixed with existing other module roles (for internal logged in user)

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -205,41 +176,47 @@ By default, Mendix forces a strong password policy. The same password policy tha
...
 
-The user name of the administrator can be changed in 's **App Security** settings on the **Administrator** tab.
+The user name of the administrator can be changed in Studio Pro using the app's **App Security** settings on the **Administrator** tab.
 
-When deployed to Mendix Cloud, the information about the administrator user name and role is taken into account when using the **Change admin password** button on the environment. After changing the settings in and redeploying the application, a successful admin password change will trigger the creation of a user in the app with the new name and role.
+When deployed to Mendix Cloud, the information about the administrator user name and role is taken into account when using the **Change admin password** button on the environment. After changing the settings in Studio Pro and redeploying the application, a successful admin password change will trigger the creation of a user in the app with the new name and role.
 
 {{% alert color="info" %}}
 At this point, the application does not automatically remove the user with the previous user name. Removing the old **MxAdmin** account has to be done manually.
```
knorrie: It's only the old "MxAdmin account" if the previous name was MxAdmin, not if you change it from Banana to Apple later.

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
```diff
@@ -112,16 +116,16 @@ The **Path to user...** button adds an XPath constraint so the access rule is on
...
-The XPath constraint can be constructed easily using the **Path to user...** button by selecting the **Customer** entity in the **Order** entity access rule. The created rule will look like this:
+The required XPath constraint can be constructed using the **Path to user...** button on the `Order` entity by navigating to the `Administration.Account` entity via the `Order_Customer` and `Customer_Account` associations in the entity navigation tree. The created rule will look like this:
 
 ```xpath
-[Module.Order_Customer = '[%CurrentUser%]']
+[Module.Order_Customer/Module.Customer/Module.Customer_Account='[%CurrentUser%]']
 ```
 
-Because of this XPath constraint, access defined in the access rule is only applied to orders for which the customer is the current end-user.
+Applying this XPath constraint means the current end-user can only see orders for the customer associated with the current end-user.
```
knorrie: Maybe?! change/improve wording because ambiguity: 'customer' can be used both as 'customer person' and 'customer company'. In the example here you constrain to the individual person.

A common real life case is of course that you can see orders placed by your customer company (also by other employees at your customer company). But we want to have a simple example here.

Another way to solve the confusion is to explicitly mention that the example we use is of the style of a webshop with individual visitors, not a business thing.

content/en/docs/refguide/modeling/security/app-security/administrator.md
knorrie: either all of them or none> So, this also means that the scenario we just described as example for Manageable Roles in the documentation, with Administrator, SubAdministrator and User user roles where SubA has mgb role on User is simply not possible to use it like that in practice. I mean, (without rolling all your own logic) this requires giving SubAdminstrator the Administration.Administrator module role...

To summarize, this means that right now, the only user role that might have any Manageable Roles configured should be the real user management Administrator user role, which has all these 3 complementary things. No other user role must have any manageable roles configured, because if they do an app developer has to enable more things to get it working, at all, and by doing so they will introduce unwanted side effects, because the only way in which you can get it working requires giving broader permissions to the role than you would want.

Well, there we are. Maybe this is why customers are upset currently, when they discover that more users/accounts/etc info is visible or editable than they intended to.

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: Done e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: Done e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: Done e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: Done e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: Changed e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/refguide/modeling/domain-model/entities/access-rules.md
MarkvanMents: Tried to clarify e6dcf89143b2fe9233484e69365f359e72601623

content/en/docs/refguide/modeling/security/app-security/administrator.md
MarkvanMents: This is too much to pick up now. Need to add it to backlog to look at in more detail.

MarkvanMents (commented):
Thanks for the review @knorrie 
I've responded to most of the comments in e6dcf89143b2fe9233484e69365f359e72601623.

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
 
-* Attributes determined by the system (like the status of an order) should never be writable
-* If an anonymous user is allowed to create objects, constrain these objects to the owner (an anonymous user is actually a **System.User** object created on the fly)
-* Do not set a default rule for read-and-write access – this forces you to think about each attribute that is added to an entity
-* Security constraints should be formed as entity access rules
-* Constraints on widgets in pages should not be used as a measure of security, but can filter out irrelevant data for the context of the page
-* Keep your attributes editable within data views, because if an access rule prohibits write access, your client will display it as non-editable – this way you are aware of the (correct) working of an access rule
+* Do not make attributes determined by the system (like the status of an order) writable
+* If an anonymous user is allowed to create objects, constrain these objects to the owner (an anonymous user is a **System.User** object created on the fly)—you can do this using [Path to User](/refguide/access-rules/#path-to-user) in your entity *Access Rules*
```
rvhofwegen, line 30: Best practice for object creation by anonymous users is to use non-persistable entities, so that they are only able to enter new data in memory. While storing the data is done using microflows which validate the entered data before mapping to persistable entities and send it to the database. Hence Xpath constraints to owner isn't possible. 
And yes, sometimes you need to use persistable entities and this could be an good solution. But I do not think thats a common scenario and good to use as example for in the general documentation.

content/en/docs/howto/security/best-practices-security.md
```diff
@@ -12,100 +12,93 @@ aliases:
...
+* Set **Default rights for new members** to **None** so you have to think about who has access to each attribute that is added to an entity
+* Implement security constraints as entity access rules
```
rvhofwegen, line 32: I would add the term XPath to this.
Something like Implement Xpath constraints to entity access rules to constrain the accessible data

content/en/docs/howto/security/best-practices-security.md
rvhofwegen: Indeed a good option. Often its a collection of attributes and thus an non-persistable obejct would be more suitable

content/en/docs/quickstarts/native-mobile-app.md
```diff
@@ -58,7 +58,7 @@ In the pop-up window that appears, enter a **name** for your app. Leave all othe
 Just as in part one, we will need to create the entity to store employee data. In the **domain model** for the Native Mobile Module, add a **new entity**.
 
 1. Name the entity Employee.
-1. The entity should be a generalization of the Account entity.
+1. The entity should be a specialization of the Account entity.
```
rvhofwegen, line 61: In general this is a bad practice. However as this is native, I'm not 100% sure if a split between User/account and process data collides with native limitations.

content/en/docs/refguide/modeling/security/app-security/administrator.md
```diff
@@ -72,15 +72,15 @@ The user role assigned to the Administrator. For more information, see [User Rol
 Default: *Administrator*
 
 {{% alert color="info" %}}
-The administrator is always created and has the System.Administrator role by default. The System.Administrator role allows users of your application to be managed. 
+The Administrator user role is always created and has the System.Administrator module role by default. The Administrator user role can also manage users of your application as it has all [manageable roles](/refguide/user-roles/#user-management).
```
rvhofwegen, line 75: I think in the original text " The administrator" did refer to MxAdmin as a default user. 
The administrator role is always present in a starter app, although it can be managed like any other role. hence it can be deleted as well. And Mendix will not add it back automatically.

content/en/docs/refguide/modeling/security/app-security/administrator.md
rvhofwegen: "the user that created the application automatically also defaults to having the Administrator role"

What user is referred to and what does "created the application" entile?
Mx platform user != application user

When I deploy a free app:
- I cannot login using MxAdmin (or default pw is used again?)
- There is no other account available
- I have a standard microflow which I use in all my free apps, where I do a catch or create for a custom admin user. So that I can login as my own created alternative MxAdmin

content/en/docs/refguide/modeling/security/app-security/administrator.md
knorrie: We also discussed this offline, and in the end I think where we are going to end up is that, IF you want to have local users and use multiple 'layers' of these manageable roles, you will have to adjust the Administration module to match that.

content/en/docs/howto/security/best-practices-security.md
MarkvanMents: So change this advice to
```suggestion
* Mendix recommends that you do not allow anonymous users to create or write to persistable objects. In general, anonymous users should always use non-persistable entities. Although you can constrain persistable objects to an anonymous owner (an anonymous user is a **System.User** object created on the fly) using [Path to User](/refguide/access-rules/#path-to-user) in your entity *Access Rules*, this is not advised.
```

content/en/docs/howto/security/best-practices-security.md
MarkvanMents:
```suggestion
* Implement XPath constraints as entity access rules to limit the data access.
```

content/en/docs/quickstarts/native-mobile-app.md
MarkvanMents: Thanks Rene.
I was going to try to change this example, but I don't want to block releasing the other changes so I have taken this document and will work on it with the Native team in a separate PR.
It is only still here in this PR because someone mixed up generalization and specialization and I wanted to correct that immediately.

content/en/docs/refguide/modeling/security/app-security/administrator.md
MarkvanMents: Should I mention that MxAdmin is created too?

```suggestion
The Administrator user role is always present in a starter app and has the System.Administrator module role by default. An administrator user (MxAdmin) is also created in the app. The Administrator user role can also manage users of your application as it has all [manageable roles](/refguide/user-roles/#user-management).  The Administrator role can be changed or deleted like any other user role once your app has been created.
```

MarkvanMents (commented):
Thanks for your comments, @rvhofwegen. I have made some suggestions to address them, let me know if the suggestions improve things and answer you concerns.

Mark.

content/en/docs/refguide/modeling/security/app-security/administrator.md
rvhofwegen: Rewrote it to match the header "Default: *Administrator*

The default application administrator (MxAdmin) is always created and has the Administrator user role by default. The Administrator userrole has by default the System.Administrator module role. By default the Administrator user role is allowed to manage all users of your application as it has all [manageable roles](/refguide/user-roles/#user-management). 
The Administrator role can be changed or deleted like any other user role once your app has been created.

content/en/docs/refguide/modeling/security/app-security/administrator.md
rvhofwegen: Agree with your rant 

"To summarize, this means that right now, the only user role that might have any Manageable Roles configured should be the real user management Administrator user role, which has all these 3 complementary things."

"IF you want to have local users and use multiple 'layers' of these manageable roles, you will have to adjust the Administration module to match that."

These two sentences nail exactly how it works and how it should be treated. And what isn't done propely in many Mx Apps

content/en/docs/howto/security/best-practices-security.md
rvhofwegen: Rewrote it starting from the positive side on how to do it correct and split the not advised scenario

* Mendix recommends that in case anonymous require to create and write on objects, this should be handled using non-persistable objects. And mapping entered data to persistable objects and database after validation using microflows. In order to block anonymous users to create, write and read directly in/from the database.
* If the covered scenario requires writing to objects of persistable entities, use [Path to User](/refguide/access-rules/#path-to-user) in your entity *Access Rules*. This should be well thought through before implementing.

MarkvanMents:
@rvhofwegen and @knorrie 
Are you OK with my proposed changes?
I'd like to publish this soon, we can always add and change things if we have more thoughts.
Can you confirm by c.o.b. tomorrow (15/9)? If I don't hear from you, I'll publish anyhow and we can pick up changes later.
I think better to get this out there now, rather than wait for it to be perfect?
Mark

rvhofwegen:
> @rvhofwegen and @knorrie Are you OK with my proposed changes? I'd like to publish this soon, we can always add and change things if we have more thoughts. Can you confirm by c.o.b. tomorrow (15/9)? If I don't hear from you, I'll publish anyhow and we can pick up changes later. I think better to get this out there now, rather than wait for it to be perfect? Mark

I saw 3 open discussions and suggestion from my side. before publishing these need to be addressed

MarkvanMents:
> > @rvhofwegen and @knorrie Are you OK with my proposed changes? I'd like to publish this soon, we can always add and change things if we have more thoughts. Can you confirm by c.o.b. tomorrow (15/9)? If I don't hear from you, I'll publish anyhow and we can pick up changes later. I think better to get this out there now, rather than wait for it to be perfect? Mark
> 
> I saw 3 open discussions and suggestion from my side. before publishing these need to be addressed

Hi Rene,
I've made suggestions for text changes to address your concerns. Are these all good - in which case I can commit them, or do they miss your concerns?

I don't think there are any outstanding concerns where I haven't suggested a change.

content/en/docs/refguide/modeling/security/app-security/administrator.md
knorrie: When starting an app (eg with an empty database locally) a user with the configured administrator user role and configured (in security tab in studio pro) is created. Anyone can create and choose any user role they want indeed and change the default admin user name to anything else (which we recommend doing in docs!!).

So, no 'MxAdmin' should not be mentioned here, or  maybe as 'e.g. MxAdmin by default if you did not change it' yet.

MarkvanMents: OK, I'll leave it as per my suggestion.
Is my suggestion good, or shall I wait for Rene to come back.

rvhofwegen: Im fine with it

Issue #10226, opened 2025-10-15
pluggable-widgets-api


bartheijs:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [pluggable-widgets-api](https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets/)

Document link: [_index.md](https://github.com/mendix/docs/blob/development/content/en/docs/apidocs-mxsdk/apidocs/studio-pro-11/pluggable-widgets/_index.md)

## My Issue/Suggestion

Paragraph 4.2.2.

Now: "A URL of a help page can be provided through the helpUrl property after the **description** tag:"
Should be: "A URL of a help page can be provided through the helpUrl property after the **studioProCategory** tag:"

If placed after description tag and before studioProCategory tag, I get an error in Mendix Studio Pro (10.24.3) that it can't read the xml.

ConnorLand:
Hello,

Thank you for reaching out. This was indeed wrong, and is now fixed. Also, as we resolved this issue we took the opportunity to further improve that doc. https://github.com/mendix/docs/pull/11832 

Thank you for contributing to the Mendix docs,

Connor

PR #11832, opened 2026-08-28
Change Pluggable Widgets API tag description

ConnorLand:
Small change in 10 and 11 docs, solves GitHub issue 10226:

https://github.com/mendix/docs/issues/10226

Will merge after tech verification.


weirdwater:
This is indeed more accurate. Perhaps we can also add emphasis to the fact that the order of elements is important at the start of [the section](https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets-10/#widget-description).

We can also add the recommendation to use an XML validator/linter when editing the file, as we publish a schema that it can be validated against for in-editor feedback. Either at the top of this the page, or under [Widget Definition](https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets-10/#widget-definition) For example:

> A schema for the widget definition file is made available through the [`mendix` package](https://www.npmjs.com/package/mendix) which is installed by the Pluggable Widgets Tools. Editor plugins and IDE's can use the schema to provide in-line feedback while authoring the definition file.

Issue #10749, opened 2026-02-12
build-javascript-actions-for-native-mobile


vinicius-ambrosio-truechange:
Please use the form below, leaving the prefilled data to help us. Thank you.

Page link: [build-javascript-actions-for-native-mobile](https://docs.mendix.com/howto10/extensibility/create-native-javascript-action/)

Document link: [create-native-javascript-action.md](https://github.com/mendix/docs/blob/development/content/en/docs/howto10/extensibility/build-javascript-actions/create-native-javascript-action.md)

## My Issue/Suggestion
Is there any frequent review on this docs?

The text instructs the reader to install a package developed 7 years ago !!

```
Install the dependency with the command npm install react-native-nfc-manager@1.2.2.
```
Even if it still works, it may have a lot of improvments or security issues implemented that we will no use when installing a ancient version!
This happens throughout the docs!

ConnorLand:
I've confirmed that we no longer rely on these dependencies, and our current platform-supported resources remove the need to teach users about adding/declaring them. So I'll be removing this doc section altogether. Thanks for highlighting this small command detail @vinicius-ambrosio-truechange , it's going to lead to some good improvements.

I'll make a PR for the deletion, merge it, then close this issue, 

Connor

ConnorLand:
@vinicius-ambrosio-truechange ,

Thank you for bringing this to our attention. These dependencies used to be key to completing the document, but as Mendix has evolved its dependencies and templates these sections have become less relevant. I've deleted them now and pointed users curious about dependencies to our widget declaring dependencies doc: https://docs.mendix.com/apidocs-mxsdk/apidocs/pluggable-widgets-native-dependencies/

Have a great day, and thanks for contributing to the Mendix documentation,

Connor

PR #11833, opened 2026-08-28
Update npm dependency messages

ConnorLand:
Will solve https://github.com/mendix/docs/issues/10749, but we need to determine if the devs wanted to specify only that version for a reason. If there is no reason, we'll rewrite to have the command install the LTS version of npm.


ConnorLand:
Confirmed with Jonathan Diehl over Slack that we don't want to teach users to install dependencies in this way anymore at all, so I deleted all these sections at his advisement in this PR: https://github.com/mendix/docs/commit/382130fee70e1015aeeffcd87a2702d879c3ccbb

This also means the GH issue is solved, which I will address and close: https://github.com/mendix/docs/issues/10749

PR #11839, opened 2026-08-31
Add guidance on feature rollouts and Mendix version upgrades to OTA updates doc

jdiehl:
## Summary
Adds two new sections to the "Updating Native Apps" (OTA) page, explaining
two release strategies not previously documented: rolling out features that
require a full app store release, and upgrading the Mendix version safely.
Applied to refguide (Studio Pro 11), refguide10, and refguide9.

## Changes
- New section "Rolling Out Features That Span an App Store Release":
  explains how to use OTA to prepare a feature while keeping it hidden
  behind a constant (via a nanoflow decision or conditional visibility),
  then enable it with a small, low-risk app store release. Keeps old and
  new app versions compatible while the release is under Apple review.
- New section "Upgrading the Mendix Version": recommends keeping version
  upgrades separate from feature work, keeping model changes minimal, and
  explains that a compatible model lets the old and new native client both
  work against the upgraded runtime, so the new client can ship via a
  normal app store release without OTA.
- Reordered the page so both new sections sit after "When to Use OTA
  Updates" and before "Enabling and Building an App with Mendix OTA
  Updates Enabled," keeping strategy guidance grouped with the existing
  use-case guidance.
- Backported identically to refguide10 and refguide9.

### DIFF
--- a/content/en/docs/refguide/mobile/distributing-mobile-apps/overtheair-updates.md
+++ b/content/en/docs/refguide/mobile/distributing-mobile-apps/overtheair-updates.md
@@ -115,3 +115,26 @@ To release a new version OTA, follow these steps:
 1. Deploy your new app to the cloud
 
 On the next restart of the application, the new OTA update will be downloaded and used. 
+
+## Rolling Out Features That Span an App Store Release {#rolling-out-features}
+
+Updating a native app can take several days because Apple must review each app store release. During this time, both the old and the new version of the app must keep working.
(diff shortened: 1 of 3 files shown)


ConnorLand:
Thanks for this PR @jdiehl !

ConnorLand:
Thanks for this excellent draft @jdiehl . I will take a look when I get back from vacation and merge this.

PR #11851, opened 2026-09-02
Add documentation for the new Best Practice Recommender security rules

joerih:
Make this a new page next to the already existing page about performance recommendations, and update the latter a bit to adapt it to this new structure.

### DIFF
--- a/content/en/docs/refguide/mendix-ai-assistance/best-practice-recommender/security-best-practices.md
+++ b/content/en/docs/refguide/mendix-ai-assistance/best-practice-recommender/security-best-practices.md
@@ -1,0 +1,135 @@
+---
+title: "Security Recommendations"
+url: /refguide/security-best-practices/
+description: "Describes the security best practices from Best Practice Recommender, including the best practices for anonymous users."
+#If moving or renaming this doc file, implement a temporary redirect and let the respective team know they should update the URL in the product. See Mapping to Products for more details.
+#The anchors <mxs001-mxsnnn> below are all mapped, so they should not be removed or changed.
+---
+
+## Introduction
+
+This document outlines the security best practices provided by [Best Practice Recommender](/refguide/best-practice-recommender/) in Studio Pro.
+
+## Anonymous User Best Practices {#anonymous-users}
+
+This section outlines security issues and Mendix best practices for [anonymous users](/refguide/anonymous-users/). Anonymous users can access an app without signing in, which means that every access right held by the anonymous user role is available to anyone who can reach the URL of the app.
+
+Best Practice Recommender checks the best practices in this section when both of the following conditions are met:
+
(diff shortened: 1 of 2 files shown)


joerih:
This is the documentation of a new feature that will be released in version 11.15.0.

content/en/docs/refguide/mendix-ai-assistance/best-practice-recommender/security-best-practices.md
```diff
@@ -0,0 +1,141 @@
...
+
+Without an XPath constraint, the access rule returns every object of the entity to every anonymous session. Data that was only ever meant to be visible to the visitor who submitted it, such as problem reports or form submissions, then becomes readable by all unauthenticated visitors. This is the most common way in which anonymous access leaks the data of unrelated users.
+
+#### Steps to Fix
+
+To fix the issue, add an XPath constraint to the access rule, so that it only returns the objects that the current anonymous session is allowed to see, for example by constraining on the owner of the object. If the objects cannot be narrowed down to the current session, remove the read access instead.
+
+{{% alert color="info" %}}
+This best practice is not reported for an access rule that also grants access to other user roles, because an XPath constraint applies to every module role of the access rule. Such an access rule is reported under [Do Not Share Module Roles Between Anonymous and Other User Roles](#mxs008) instead.
+{{% /alert %}}
```
MarkvanMents, line 135: Why do we recommend this - we say earlier on that putting an XPath constraint for an anonymous user is pointless as they will don't keep their user ID across sessions? But we now say it is a good idea.

knorrie: I agree, I'm not sure today what to say on this, today. If an anonymous session does not have any identity, then it makes no sense doing something so that they later can find back their own persisted data... Indeed.

joerih: Yes, good question, this isn't explained very well. I've discussed it briefly with Jonathan, and the main purpose of this rule for constraining read access is to make sure the user at least _thought_ about limiting access (because you can still write an XPath filter that does not look at the anonymous user id, but that restricts access in some other way).

MarkvanMents: I've tried to explain the contradiction by adding something in ec38d82c7166d0c82ab1875454cf69e3dc33bb3b

MarkvanMents:
@joerih and @knorrie - are you both happy with the changes I've made?

content/en/docs/refguide/mendix-ai-assistance/best-practice-recommender/security-best-practices.md
joerih: Yes, that looks good to me, thanks.

PR #11875, opened 2026-09-07
Update release notes for Private PDF Document Generation to v1.0.5

samuelreichert:


MarkvanMents (approved):
Thanks for the PR.
I'll approve and merge immediately as the release date is in the past.
@samuelreichert - if you could include release notes as part of your definition of done, then customers would be better informed, and we would have more time to discuss any questions we might have. Thanks for your help in this.

PR #11877, opened 2026-09-07
Update message definitions docs for single-definition documents

aimiliatheologimendix:
In Studio Pro 11.15, message definitions are split into separate documents, so each document now contains a single message definition.

- Update the menu path to Add other > Message definition
- Remove the list-based Managing Message Definitions section
- Move Generate mapping(s)... under the General tab
- Remove the Name field, as the document name is now the definition name
- Rename Filter to Search and match the Example Value column header
- Update both screenshots to the new UI
- Fix "a message definitions" to singular in Import Mapping Action


content/en/docs/refguide/modeling/application-logic/microflows-and-nanoflows/activities/integration-activities/import-mapping-action.md
```diff
@@ -46,11 +46,11 @@ The [import mapping](/refguide/import-mappings/) defines how to transform the XM
 
 ### Input Content Type
 
-If the import mapping is based on a [message definitions](/refguide/message-definitions/), it can import both XML and JSON. Select whether the input object contains XML or JSON.
+If the import mapping is based on a [message definition](/refguide/message-definitions/), it can import both XML and JSON. Select whether the input object contains XML or JSON.
```
HarishShankerAS, line 49: Since there is no collection document, do we need to remove the plural format in the ref guide link? message-definitions to message-definition

quinntracy: How will the UI appear in Studio Pro? When a user is adding one to their app (following right-click > **Add other**), does it appear as "Message defintion" or "Message defintions" in the menu? We suggest matching the UI. However, keep in mind that if we change the URL to /refguide/message-definitions/, this will need to be updated if the link is hard coded in Studio Pro.

aimiliatheologimendix: nope

aimiliatheologimendix: The new UI will show singular but since I know that changing URLs is a painful, my suggestion is to keep it as it is!

quinntracy (approved):

PR #11884, opened 2026-09-08
Update string-concatenation.md with a new example

MariaShaposhnikova:


content/en/docs/refguide/modeling/application-logic/string-concatenation.md
```diff
@@ -12,7 +12,7 @@ aliases:
 
 Sometimes, you might need to concatenate some text together in your app. Common cases are when you want to assemble the full name of a person based on the first and last name.
 
-Whenever you want to paste multiple strings attributes together, you want to make sure that the text looks good in all cases. For example, you do not want "firstname + middlename + lastname" to be printed as "John null Doe" because the middle name is empty.
+Whenever you want to paste multiple strings attributes together, you want to make sure that the text looks good in all cases. For example, you do not want "firstname + ' ' + middlename + ' ' + lastname" to be printed as "John  Doe" with two spaces in-between the first and last name because the middle name is empty.
```
MarkvanMents: Looks good. There is a proofreading change you could make as well.

```suggestion
Whenever you want to paste multiple string attributes together, you want to make sure that the text looks good in all cases. For example, you do not want "firstname + ' ' + middlename + ' ' + lastname" to be printed as "John  Doe" with two spaces in-between the first and last name because the middle name is empty.
```

PR #11891, opened 2026-09-10
FDS: SSO redirect changes

Karuna-Mendix:


content/en/docs/marketplace/platform-supported-content/modules/fds-gateway-login-connector.md
```diff
@@ -89,41 +89,38 @@ For more information, see [Constants](/refguide/constants/).
...
-
 <body></body>
-
 </html>
 ```
 
-Update the `originURI` cookie value in `index.html` and use `/sso-login.html` instead of `/login.html` as shown in the code below:
+Update the `<script>` tag in `index.html`:
+
+This removes the existing `if` condition and sets `originURI` to `sso-login.html`, preventing the Gateway from taking over `login.html` and performing the SSO flow.
```
JaapF: This is cryptic to me. 
- How could the gateway over take-over a login.html.
- Gateway IS supposed to execute the SSO flow - that's the whole idea of the module - delegate execution of the SSO-flow.

So ....?

Karuna-Mendix: @JaapF , please check this commit: https://github.com/mendix/docs/pull/11891/commits/cb9e43ae41023e904ad2f8d5a05b4eb174cbcd50

PR #11893, opened 2026-09-10
Prevent build-and-deploy workflow from running in forks

MarkvanMents:
## Issue

The `build-and-deploy.yml` workflow uses `push` and `pull_request` triggers scoped to the `development` and `production` branches. This caused the workflow to run in two unintended scenarios:

- **In forks**: any fork that has a `development` or `production` branch triggers the workflow on pushes to those branches, consuming the fork's Actions minutes and failing when deployment secrets are absent.
- **On fork sync**: when a fork is synced to the main repo and opens a PR, the `pull_request` trigger fires the workflow in the main repo against the fork's code.

## Resolution

Added `if: github.event.repository.fork == false` to the `build` job. Because the `deploy` job depends on `build` via `needs: build`, it is automatically skipped whenever `build` is skipped. The workflow now only runs in the main repository, for both push and pull request events.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -24,6 +24,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-22.04
+    if: github.event.repository.fork == false
     outputs:
       target: ${{ steps.env.outputs.target }}
     concurrency:


MarkvanMents:
In [lint-action](https://github.com/mendix/docs/blob/development/.github/workflows/lint-action.yml) we actually use
```
  lint-pr:
    # Check if the event is not triggered by a fork
    if: github.repository_owner == 'mendix'
    # The type of runner that the job will run on
```
But the suggested change here (`github.event.repository.fork == false`) is more generic.
If it works, should we also use this in the other actions?

.github/workflows/build-and-deploy.yml
```diff
@@ -24,6 +24,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-22.04
+    if: github.event.repository.fork == false
```
OlufunkeMoronfolu: I'd suggest we go with `if: github.repository == 'mendix/docs'` than `if: github.event.repository.fork == false.`

Mainly because the fork check reads from the event payload, which isn't present on every event type and in some cases, the job runs anyway. Meanwhile, `github.repository` is always available in the context, and is consistent with how we guard our other workflows.

The trade-off is that it hardcodes the repo name, so a rename or transfer means updating the workflow, which seems unlikely, in contrast to other scenarios.

MarkvanMents: Thanks Olu, I've updated in 04bc611b0b0cb53f19f884b6f2ded6c64d839702

MarkvanMents:
Approved by Olu in person. Merging,

PR #11895, opened 2026-09-10
Open tables in a pop-up when viewed in expanded mode

Karuna-Mendix:
https://mendix.atlassian.net/browse/TW-2751
Key points to consider:
1. Tables should open in a pop-up window when expanded.
2. The background page should remain visible but inactive.
3. The pop-up window can be closed via the “collapse table” option/button. 

Before:
table-photo
After: 

https://github.com/user-attachments/assets/940aba54-47bd-4766-a4a1-40e3a957910b

### DIFF
--- a/assets/js/expandable-table.js
+++ b/assets/js/expandable-table.js
@@ -1,0 +1,46 @@
+(function () {
+  'use strict';
+
+  function init() {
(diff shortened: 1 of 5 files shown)


Karuna-Mendix:
Need to apply the changes to all docs if this PR gets approved.

PR #11925, opened 2026-09-16
Updated Webhooks section

katarzyna-koltun-mx:


katarzyna-koltun-mx:
Merged into https://github.com/mendix/docs/pull/11947

